PR #85, opened 2022-07-19
[WIP] Support for conditional loop

xadupre:
Introduces syntax for onnx loops through function `conditional_range`. It is a proposal. It seems simple enough.

```python
@script()
def loop_range_cond(A: FLOAT["N"]) -> FLOAT["N"]:
    T = A
    cond = op.Constant(value=make_tensor('true', TensorProto.BOOL, [1], [1]))

    for i in conditional_range(10, cond):

        T = T + A * op.Cast(i, to=TensorProto.FLOAT)
        cond &= op.ReduceSum(T) > 0
    return T
```

Symbol `&=` is key. `=` creates a different variable, not the one `condition_range` receives. `&=` modifies the constant if this one is a mutable object, then `condition_range` evaluate a different condition at every iteration. This trick is not needing for the converter but is needed for the eager mode.

It works if cond is mutable. When cond is immutable, it may not work.

```python
cond = True
print(id(cond), cond)
cond &= False
print(id(cond), cond)
```
```
>>> 140735022353256 True
>>> 140735022353288 False
```
It works with a numpy array:
```python
import numpy as np
cond = np.array(True)
print(id(cond), cond)
cond &= False
print(id(cond), cond)
```
```
>>> 1416572020432 True
>>> 1416572020432 False
```


justinchuby:
JAX' [while_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html#jax.lax.while_loop) and [fori_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html) look like possibilities for expressing loop structures. Do you have any thoughts on how this design relate to or different than their apis?

gramalingam:
> JAX' [while_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html#jax.lax.while_loop) and [fori_loop](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html) look like possibilities for expressing loop structures. Do you have any thoughts on how this design relate to or different than their apis?

This PR was more about the python syntax for expressing ONNX loops, not about modifying ONNX's loop ops. Specifically, there is no demand for changing or updating ONNX loop ops. ONNXScript is concerned only with syntactic-sugar for expressing such loops.

In any case: I think some of these older PRs are not active and should likely be closed.

justinchuby:
@gramalingam yeah, I was talking more about the designs of the apis like `jax.lax.while_loop`. Since this PR also creates `conditional_range` to express a range loop

PR #144, opened 2022-09-29
Support graph-attributes for ops like Scan

gramalingam:
Higher-order operators in ONNX (like Loop and Scan) use graph-valued attributes.

This PR is the first step in supporting the use of such operators. (As of now, it does not handle references to outer-scope variables.)


onnxscript/irbuilder.py
```diff
@@ -6,14 +6,16 @@
...
 
 import onnx
-from onnx import helper
+from onnx import ValueInfoProto, helper
 from onnx.defs import onnx_opset_version
 
 import onnxscript
 from onnxscript import type_annotation as ta
 from onnxscript import values
+from onnxscript.onnx_types import ONNXType
```
justinchuby, line 18: nit: Use onnx_types.ONNXType

onnxscript/irbuilder.py
```diff
@@ -231,24 +219,46 @@ def append_function(self, opf):
...
+                    input.type.CopyFrom(io_types.to_type_proto())
+            for output in graph.output:
+                if not output.HasField("type"):
+                    output.type.CopyFrom(io_types.to_type_proto())
+        if input_types is not None:
+            for input, type in zip(graph.input, input_types):
+                input.type.CopyFrom(type.to_type_proto())
+        if output_types is not None:
+            for output, type in zip(graph.output, output_types):
+                output.type.CopyFrom(type.to_type_proto())
```
justinchuby, line 261: nice: we now have coverage warnings

onnxscript/irbuilder.py
```diff
@@ -284,20 +294,17 @@ def to_proto(f):
...
-                use
-            io_types: defines a default value for missing input and
-                output type, this is only used if *enforce_typed* is
-                True
+            use_default_type: if True, the function uses a default type
+                for inputs and outputs that do not have a type
 
-        Returns:
-            an instance of :class:`onnx.GraphProto`
+        Return:
```
justinchuby, line 306:
```suggestion
        Returns:
```

justinchuby: Return is an alias of Returns

onnxscript/irbuilder.py
```diff
@@ -306,11 +313,23 @@ def to_graph_proto(self, enforce_typed=False, io_types=None):
...
         return graph, sub_functions
 
+    def to_graph_proto(self, use_default_type: bool = True):
+        """
+        Converts the content of this class into a `onnx.GraphProto`.
+
+        :param use_default_type: if True, the function uses a default type
+            for inputs and outputs that do not have a type
+
+        :return: an instance of :class:`onnx.GraphProto`
```
justinchuby, line 328: Prefer google style doc strings

onnxscript/main.py
```diff
@@ -104,6 +104,54 @@ def transform(f):
...
+    Returns:
+        A decorator that returns its input function, but attaches a graph_proto
+        attribute representing the input function. The translation is not
+        done at this time, but previously when the outer-level function
+        was translated to an OnnxFunction. The decorator just looks up
+        and retrieves the GraphProto representation previously generated.
+
+    Example:
+
+    ::
```
justinchuby, line 123: Possible to use backticks for code blocks like markdown? I think sphinx supports it.

onnxscript/irbuilder.py
gramalingam: Done

onnxscript/main.py
gramalingam: I am not sure. I just used the format from example above (which I assume works). Happy to leave it to the experts to fix (or let me know what exactly is required.)

onnxscript/irbuilder.py
gramalingam: Let me leave it as-is for now. I think the current file-organization need to be rationalized and improved.

onnxscript/converter.py
```diff
@@ -1321,7 +1310,17 @@ def translate_block(self, stmts, name, live_defs, parent_stmt=None):
                     self.current_fn, ovar, typeinfo, debuginfo.DebugInfo(info_stmt, self)
                 )
         graph = self.exit_scope()
-        return graph.to_graph_proto()
+        return graph.to_graph_and_functions()
+
+    def translate_nested_function_def(self, fn: ast.FunctionDef):
+        """
+        Translate a nested function definition.
+        """
```
justinchuby, line 1318:
```suggestion
        """Translate a nested function definition."""
```

nit: style preference. https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

onnxscript/irbuilder.py
```diff
@@ -73,41 +75,26 @@ def __repr__(self):
     def typed_str(self):
         return f"{self.name} : {str(self.typeinfo)}"
 
-    def to_value_info(self, enforce_typed=False, default_type=None):
-        """Converts the content of this class into :class:`onnx.ValueInfoProto`.
+    def to_value_info(self, use_default_type: bool = True):
+        """
+        Converts the content of this class into :class:`onnx.ValueInfoProto`.
```
justinchuby, line 80:
```suggestion
        """Converts the content of this class into :class:`onnx.ValueInfoProto`.
```

onnxscript/irbuilder.py
```diff
@@ -284,20 +294,17 @@ def to_proto(f):
             graph, opset_imports=opset_imports, functions=functions, **kwargs
         )
 
-    def to_graph_proto(self, enforce_typed=False, io_types=None):
-        """Converts the content of this class into a `onnx.GraphProto`.
+    def to_graph_and_functions(self, use_default_type: bool = True):
+        """
+        Converts the content of this class into a `onnx.GraphProto` and
```
justinchuby, line 299:
```suggestion
        """Converts the content of this class into a `onnx.GraphProto` and
```

onnxscript/values.py
```diff
@@ -104,7 +104,17 @@ def get_schema(self):
     def has_schema(self):
         return self.opschema is not None
 
+    def adapt_kwargs(self, **kwargs):
+        """
+        Replaces function-valued attribute-values by their GraphProto representation.
```
justinchuby, line 109:
```suggestion
        """Replaces function-valued attribute-values by their GraphProto representation.
```

justinchuby (commented):
nit

onnxscript/test/models/graph_attr.py
```diff
@@ -0,0 +1,34 @@
...
+def sum_to(X: INT64):
+    '''Test use of a nested-function as a graph-attribute, using the Loop operator.'''
+    @graph(parent=sum_to)
+    def LoopBody(i: INT64, cond: BOOL, sum_in: INT64):
+        cond_out = op.Identity(cond)
+        sum_out = sum_in + i
+        scan_out = op.Identity(sum_out)
+        return cond_out, sum_out, scan_out
+    zero = op.Constant(value_int=0)
+    _, result = op.Loop (X, None, zero, body=LoopBody)
```
xadupre, line 33: How do you support variables defined in `sum_to` and used in `LoopBody` without them being an explicit input? Or maybe you don't.

gramalingam: Good question. That is not supported yet (as mentioned in the PR description above). I plan to add it next (probably as a separate PR).

onnxscript/test/models/graph_attr.py
```diff
@@ -0,0 +1,34 @@
...
+# --------------------------------------------------------------------------
+
+from onnxscript import graph, script
+from onnxscript.onnx_opset import opset15 as op
+from onnxscript.onnx_types import BOOL, INT64
+
+
+@script()
+def cumulative_sum(X: INT64['N']):
+    '''Test use of a nested-function as a graph-attribute, using the Scan operator.'''
```
justinchuby, line 13:
```suggestion
    """Test use of a nested-function as a graph-attribute, using the Scan operator."""
```

onnxscript/test/models/graph_attr.py
```diff
@@ -0,0 +1,34 @@
...
+        sum_out = sum_in + next
+        scan_out = op.Identity(sum_out)
+        return sum_out, scan_out
+    zero = op.Constant(value_int=0)
+    _, result = op.Scan (zero, X, body=Sum, num_scan_inputs=1)
+    return result
+
+@script()
+def sum_to(X: INT64):
+    '''Test use of a nested-function as a graph-attribute, using the Loop operator.'''
```
justinchuby, line 25:
```suggestion
    """Test use of a nested-function as a graph-attribute, using the Loop operator."""
```

onnxscript/test/models/graph_attr.py
justinchuby: Even though test models are not formatted, if code format is not part of what is tested, I would run black on these files.

onnxscript/test/models/graph_attr.py
```diff
@@ -0,0 +1,34 @@
...
+    @graph(parent=sum_to)
```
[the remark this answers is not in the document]
gramalingam: Yes and no. I hope we can eliminate this requirement eventually, but it is used for now. It is a long explanation. This is because of a combination of features: We'd like the translation of this nested-function to Graph happen once (when sum_to is translated by converter when `@script()(sum_to))` happens, and not when `@graph(LoopBody)` executes (during eager-mode). So, when `@graph` executes in eager-mode we want to lookup and reuse the previously produced GraphProto, which is stored in the IR created for `sum_to`. This is the reason we need access to the parent `sum_to`. We know this parent at conversion-time when `@script` is evaluated on `sum_to`, but it is not easy to get this information when `@graph` is executed during eager-mode. At least, I don't know how to get this information. May be `inspect` will let us get this information? However, the converter (during `@script` time) does have this information and can automatically add this to the AST and re-compile it. But that's a lot of work and not clear we want to do that yet (as we once discussed).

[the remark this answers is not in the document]
gramalingam: Ok, I replaced it by looking it up in the frame.

onnxscript/main.py
justinchuby: Tested; markdown doesn't work yet. reST is the way to go I think

xadupre:
It looks ok to me. Maybe some tests about wrong code and error messages could useful (missing @graph, nested functions...)

onnxscript/main.py
xadupre: You can use extension .. exec_code:: to execute the example and dispaly the output in the documentation. The documentation fails if the example is no longer valid.

gramalingam: I think I tried something like exec_code in the documentation (tutorial) and ran into the issue that our scripts don't work in that mode since we use inspect to get the source. Our documentation file examples are tested using our own test-case-driver (as you know). So, I will leave this as is for now. I am happy to update things if and when we have a answer to that.

gramalingam:
> It looks ok to me. Maybe some tests about wrong code and error messages could useful (missing @graph, nested functions...)

Added a testcase

onnxscript/test/converter_test.py
```diff
@@ -556,6 +556,14 @@ def test_graph_attr_loop(self):
         expected_output = np.array([0, 1, 3, 6, 10, 15], dtype=np.int64)
         self.check_run(sum_to, inputs, expected_output)
 
+    def test_graph_attr_loop_error(self):
+        from onnxscript.test.models.graph_attr import sum_to_error
+
+        input = np.array(6, dtype=np.int64)
+        with self.assertRaises(ValueError) as exc:
```
justinchuby, line 563: Consider assertRaisesRegex if trying to match messages? https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

PR #145, opened 2022-09-30
Add flake8-pyi linter

justinchuby:
https://pypi.org/project/flake8-pyi/ catches errors in pyi stubs. We don't have pyi's yet but it's good to have the check in place.


AllenTiTaiWang (approved):
Good to know more about mighty flake8 family.

gramalingam (approved):

PR #146, opened 2022-09-30
Limit action permissions

justinchuby:
Minimally needed permissions for actions.

Ref: https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#permission-on-pull-requests

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -11,18 +11,8 @@ jobs:
     name: Lint Python
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


AllenTiTaiWang (approved):
I think I was blocked by only using `pull-requests: read` last time, but never tried with this. I wonder if this affects any actions (alerts/warnings/errors). Maybe after merging we should see how it goes in the next PR?

justinchuby:
I don't think they should since the states are determined by the github actions system and actions do not have to write any states. We will see how it goes.

PR #147, opened 2022-10-03
Configure pydocstyle

justinchuby:
Configure pydocstyle to enable flake8 test for docstring style.

### DIFF
--- a/onnxscript/test/functions/onnx_script_test_case.py
+++ b/onnxscript/test/functions/onnx_script_test_case.py
@@ -9,7 +9,7 @@ import dataclasses
 import numbers
 import unittest
 import warnings
-from typing import Any, List, Optional, Union
+from typing import Any, Collection, Optional, Union
 
 import numpy as np
 import onnx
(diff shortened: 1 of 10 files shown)


onnxscript/test/functions/onnx_script_test_case.py
```diff
@@ -198,26 +198,31 @@ def run_eager_test(
...
+            function: the function to be tested.
+            rtol: relative tolerance. Defaults to None.
+            atol: absolute tolerance. Defaults to None.
+            skip_eager_test: not to run eager test if Ture.
+            skip_test_names: to skip these tests.
+            attrs: default attributes of the function node.
 
         """
         if skip_test_names is None:
-            skip_test_names = []
```
justinchuby, line 220: Since we only check membership (`... in skip_test_names`), I changed it to a set

onnxscript/test/functions/onnx_script_test_case.py
```diff
@@ -198,26 +198,31 @@ def run_eager_test(
...
-        Arguments:
-            function (onnxscript.OnnxFunction): the function to be tested.
-            skip_eager_test (bool): not to run eager test if Ture.
-            skip_test_names (List[str]): to skip these tests.
-            attrs (Any): default attributes of the function node.
+        Args:
+            function: the function to be tested.
+            rtol: relative tolerance. Defaults to None.
+            atol: absolute tolerance. Defaults to None.
+            skip_eager_test: not to run eager test if Ture.
```
gramalingam, line 217: "Ture" => "True"

onnxscript/test/functions/onnx_script_test_case.py
```diff
@@ -198,26 +198,31 @@ def run_eager_test(
...
+            skip_test_names: to skip these tests.
+            attrs: default attributes of the function node.
 
         """
         if skip_test_names is None:
-            skip_test_names = []
+            skip_test_names = set()
```
gramalingam, line 223: I guess `{}` would work as well? Not a big deal either way.

gramalingam (approved):
LGTM, modulo minor pre-existing typo.

onnxscript/test/functions/onnx_script_test_case.py
justinchuby: {} gives us a dict. I got trapped a few times 😅

onnxscript/test/functions/onnx_script_test_case.py
justinchuby: Done

onnxscript/test/functions/onnx_script_test_case.py
justinchuby: If there is stuff inside the brackets, e.g. `{1}`, then it is a set. There was a modernization PR I created that did some of this work.

gramalingam:
Please feel free to merge it. I would have. but not sure if you planned to make any other changes.

PR #150, opened 2022-10-05
Move build documentation to a separate CI job

justinchuby:
Because we only need to build the doc once.

Also

- Add windows machines to test and build doc
- cache pip packages for faster runs

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -11,10 +11,10 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest]
         include:
           # Python310-onnx:
           - python-version: "3.10"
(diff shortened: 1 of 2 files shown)


gramalingam:
Sounds fine to me. But, in principle, even doc generation could depend on the platform, right? If a user wants to run it on windows, for example, it could fail there even if succeeds on linux, right?

justinchuby:
Good point, I added windows machines

justinchuby:
The matrix is missing one config for py37. not sure why yet

gramalingam:
I am fine either way. @xadupre, any concerns? Thanks!

xadupre:
Support for python 3.7 ends in 8 months. Seems ok to drop it.

justinchuby:
@gramalingam I think I fixed it.

AllenTiTaiWang (approved):

.github/workflows/main.yaml
```diff
@@ -11,37 +11,41 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest, windows-latest]
```
AllenTiTaiWang, line 17: I wonder if tests are necessary to be run on two platforms.

justinchuby: I think so because we want to ensure portability among these two platforms.

justinchuby:
Merging if no more comments; revertible if needed

PR #155, opened 2022-10-13
Support for outer-scope variables in nested function (GraphProto)

gramalingam:
Add support for GraphProto attributes that reference outer-scope variables.

* In eager-mode, when invoking ORT for an op with GraphProto attributes, we lookup the values of outer-scope variables referenced by the graph, and add them as extra graph-inputs to pass them on to ORT.
* At script-time, check that the outer-scope variables are not redefined in between the function-definition site, and the place where the function is used as a GraphProto attribute. Such redefinitions are mapped to distinct variables in ONNX due to its SSA requirement.
* Update the analyses to account for variables used within nested GraphProto (function).
*


onnxscript/eager_mode_evaluator.py
```diff
@@ -92,9 +92,10 @@ def ort_to_os_value(v):
     raise TypeError(f"Unexpected ORT value type {type(v)}.")
 
 
-def call_ort(schema, *args, **kwargs):
+def call_ort(schema, args, kwargs, implicit_args={}):
```
xadupre, line 95: implicit_args=None (empty dictionary is not recommended for default value)

onnxscript/converter.py
```diff
@@ -406,6 +401,20 @@ def translate_attr(self, attr_name, expr):
...
+            if isinstance(val, irbuilder.Function):
+                # Check that outer-scope variables referenced by function have same value
+                # at function-definition site and use-as-attribute site, to avoid errors.
+                for pyvar, previous in val.outer_scope_variables:
+                    current = self.lookup(pyvar, debuginfo.DebugInfo(expr, self))
+                    if current.value != previous.value:
+                        self.fail(
+                            expr,
+                            f"Outer scope variable {pyvar} referenced by function "
+                            f"{expr.id} modified."
```
xadupre, line 413: ```{expr.id!r}``

onnxscript/converter.py
```diff
@@ -406,6 +401,20 @@ def translate_attr(self, attr_name, expr):
             val = self.lookup(expr.id, debuginfo.DebugInfo(expr, self))
             if isinstance(val, values.AttrRef):
                 return self.ir_builder.attr_ref(attr_name, val.value, val.typeinfo)
+            if isinstance(val, irbuilder.Function):
+                # Check that outer-scope variables referenced by function have same value
+                # at function-definition site and use-as-attribute site, to avoid errors.
+                for pyvar, previous in val.outer_scope_variables:
+                    current = self.lookup(pyvar, debuginfo.DebugInfo(expr, self))
+                    if current.value != previous.value:
```
xadupre, line 409: Is there a unit test to check this exception?

gramalingam: Yes, see `test_outer_scope_redefinition`.

onnxscript/eager_mode_evaluator.py
gramalingam: Done

onnxscript/values.py
```diff
@@ -104,23 +106,38 @@ def get_schema(self):
...
-                        "attribute. Did you forget to decorate it with @graph?"
-                    )
-        return kwargs
+            if isinstance(v, OnnxClosure):
+                kwargs[k] = v.function_ir.to_graph_proto()
+                for pyvar, onnxvar in v.function_ir.outer_scope_variables:
+                    closure[onnxvar.value] = v.frame.f_locals[pyvar]
+            elif callable(v):
+                raise ValueError(
+                    f"Error: function-valued attribute {v.__name__} has no graph_proto"
```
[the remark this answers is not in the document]
gramalingam: Done

onnxscript/converter.py
gramalingam: Done

gramalingam:
I am merging this in, to facilitate other changes. If there is any concern, please let me know, I can fix them. Thanks

PR #156, opened 2022-10-20
Fix documentation version

xadupre:
Signed-off-by: xadupre <xadupre@microsoft.com>

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -19,24 +19,29 @@ jobs:
         include:
           - name: python310-onnx
             python-version: "3.10"
-            onnx_standard: "1"
-            test_examples: "1"
+            onnx_standard: true
+            onnx_exp: false
+            test_examples: true
           - name: python39-onnx
             python-version: "3.9"
-            onnx_standard: "1"
-            test_examples: "0"
+            onnx_standard: true
+            onnx_exp: false
+            test_examples: false
           - name: python39-expe
             python-version: "3.9"
-            onnx_standard: "0"
-            test_examples: "0"
+            onnx_standard: false
+            onnx_exp: true
+            test_examples: false
           - name: python38-expe
             python-version: "3.8"
-            onnx_standard: "0"
-            test_examples: "0"
+            onnx_standard: false
+            onnx_exp: true
+            test_examples: false
           - name: python37-expe
             python-version: "3.7"
-            onnx_standard: "0"
-            test_examples: "0"
+            onnx_standard: false
+            onnx_exp: true
+            test_examples: false
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
@@ -52,12 +57,12 @@ jobs:
           python -m pip install -r requirements-dev.txt
 
       - name: Install standard onnx
-        if: eq(${{ matrix.onnx_standard }}, "1")
+        if: ${{ matrix.onnx_standard }}
         run: |
           python -m pip install onnx onnxruntime
 
       - name: Install experimental onnx
-        if: eq(${{ matrix.onnx_standard }}, "0")
+        if: ${{ matrix.onnx_exp }}
         run: |
           python -m pip install -r requirements-dev-experiment.txt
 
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yaml
```diff
@@ -74,6 +82,22 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
 
+      - name: Build documentation
+        if: and(${{ matrix.test_examples }}, eq('${{ matrix.os }}', 'ubuntu-latest'))
```
xadupre, line 86: It does not work. It is still executed on windows. Nothing I tried is working.

requirements-dev-experiment.txt
```diff
@@ -0,0 +1,4 @@
+--pre -f https://onnxruntimepackages.blob.core.windows.net/$web/onnx-function-experiment.html
```
gramalingam, line 1: We were told to use
```
   https://onnxruntimepackages.z14.web.core.windows.net/onnx-function-experiment.html onnx-function-experiment
```
instead (some change in the last few days). Similarly for below.

onnxscript/__init__.py
```diff
@@ -21,6 +21,7 @@
     # It's onnxscript and onnx-script now. That way, we can use __package__ here.
     __version__ = importlib_metadata.version("onnx-script")
 except importlib_metadata.PackageNotFoundError:
+    # This should not happen.
     __version__ = None
```
justinchuby, line 25: It happened to me when the package was installed in editable mode. Was there a better way handling it?

.github/workflows/main.yaml
justinchuby: Curious if the build_docs job was not enough to cover different cases?

gramalingam: @justinchuby : the background is that build_docs was failing (see my PR: https://github.com/microsoft/onnx-script/pull/155 ) ... if you have any suggestions, I would appreciate it. Xavier was trying to figure out the same, but I am not sure what his intentions are with these specific changes. I do know there is another issue fetching experimental packages which is interfering with CI currently (fixed in my PR, but not merged in), which is discussed below, not sure if that's a source of Xavier's issues.

justinchuby: Thanks! Taking a look

justinchuby: Since the CI did not fail before and does now, I wonder if there are changes in the dependencies we take (sphinx themes etc.) that caused this?

gramalingam: I wonder if it relates to [this line](https://github.com/microsoft/onnx-script/blob/3c241a092dbe8f0756f7034540585d293a390141/docs/conf.py#L49) ... the corresponding line in ONNX looks like  [this](https://github.com/onnx/onnx/blob/604af9cb28f63a6b9924237dcb91530649233db9/docs/docsgen/source/conf.py#L45)

requirements-dev-experiment.txt
gramalingam: I merged in a PR to fix the URLs. That will, at least, help isolate other sources of failure, and let us focus on the documentation issue, thanks!

.github/workflows/main.yaml
xadupre: The second line is better. The first line may work or not depending on the current directory. My point about the first comment is not related to the documentation itself but the fact I tried to avoid generating the documentation on windows to make the CI faster.

requirements-dev-experiment.txt
xadupre: I was surprised that the packages were installed for every build and then uninstalled. It is best not to install them if they are not needed.

xadupre:
I wanted to exclude the documentation generation on windows to make CI shorter.

.github/workflows/main.yaml
justinchuby: Regarding doc gen, I tried to not generate documentation at all in `test` jobs and separated them to the `build_docs` job below. Does that fit the need, or are there special things we are doing in the `test` jobs that make us want to generate the docs after tests are run?

gramalingam:
Thanks Xavier, for investigating this. I guess I can close this now.

PR #160, opened 2022-10-26
Evaluator and Mixed Evaluator

gramalingam:
* Introduce base-class evaluator (to support multiple evaluators)
* Add mixed-mode evaluator that allows in-python execution of some ops, relying on ORT for rest. Meant to be used for higher-order ops like SequenceMap


gramalingam:
@xadupre : could you take a look? It requires a review by someone, thanks!

onnxscript/evaluator.py
```diff
@@ -15,7 +18,62 @@
     InvalidGraph,
 )
 
-from onnxscript import irbuilder, tensor, utils
+from onnxscript import autocast, irbuilder, onnx_opset, tensor, utils, values
```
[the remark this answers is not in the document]
xadupre: The warning about import cycle should be investigated.

onnxscript/evaluator.py
```diff
@@ -15,7 +18,62 @@
...
+    """Base class for evaluation of ONNX ops.
+
+    The execution of onnxscript functions in eager-mode is dispatched to an Evaluator
+    instance (or, more precisely, to the eval method of the Evaluator instance).
+    The evaluator is expected to transform the input/output/attribute representation
+    supported by onnxscript to those expected by a particular backend.
+    """
+
+    def eval(self, schema, inputs, attributes):
+        closure = self.adapt_attributes(schema, attributes)
```
xadupre, line 34: I wonder if you should have a function adapt_outputs as well.

onnxscript/evaluator.py
```diff
@@ -144,3 +204,99 @@ def call_ort(schema, args, kwargs, implicit_args=None):
     # Map ORT output values to the onnxscript representation-type.
     cast_result = [ort_to_os_value(x) for x in result]
     return cast_result[0] if len(cast_result) == 1 else cast_result
+
+
+def id(schema):
```
xadupre, line 209: You should use another name. `id` is a built-in function.

onnxscript/evaluator.py
```diff
@@ -144,3 +204,99 @@ def call_ort(schema, args, kwargs, implicit_args=None):
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._python_ops = {}
+
+    def use_graph_attribute(self, schema):
+        return id(schema) not in self._python_ops
+
+    def _eval(self, schema, inputs, attributes, closure):
+        if id(schema) in self._python_ops:
```
xadupre, line 237: You call `id(schema)` twice. You should store it in a variable.

onnxscript/evaluator.py
```diff
@@ -118,6 +176,8 @@ def call_ort(schema, args, kwargs, implicit_args=None):
         opset_imports=[opset_id],
         ir_version=irbuilder.select_ir_version(schema.since_version, domain=schema.domain),
     )
+    model = onnx.shape_inference.infer_shapes(model)
```
xadupre, line 179: `model` is changed here and then serialized in `_cache_` with `SerializeToString`. The information about shape inference makes the model bigger and longer to serialize. Would it be faster to keep the original `model`? I assume this line was changed to do more thorough verifications.

onnxscript/test/eval_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+            @graph()
+            def square(y: FLOAT["N"]) -> FLOAT["N"]:  # noqa: F821
+                return op.Mul(y, y)
+
+            seq2 = op.SequenceMap(seq1, body=square)
+            return seq2
+
+        x = np.array([0.0, 1.0], dtype=np.float32)
+        output = seq_map[evaluator.ort_mixed_evaluator](x)
```
xadupre, line 24: It would be better to add another unit test to check a dummy evaluator even if this one is calling ORT underneath.

onnxscript/evaluator.py
```diff
@@ -4,6 +4,9 @@
 # --------------------------------------------------------------------------
 
 import pprint
+from abc import ABC, abstractmethod
+from contextlib import contextmanager
```
justinchuby, line 8: Prefer importing modules only?

onnxscript/evaluator.py
```diff
@@ -15,7 +18,62 @@
...
+class Evaluator(ABC):
+    """Base class for evaluation of ONNX ops.
+
+    The execution of onnxscript functions in eager-mode is dispatched to an Evaluator
+    instance (or, more precisely, to the eval method of the Evaluator instance).
+    The evaluator is expected to transform the input/output/attribute representation
+    supported by onnxscript to those expected by a particular backend.
+    """
+
+    def eval(self, schema, inputs, attributes):
```
justinchuby, line 33: Add types maybe?

onnxscript/evaluator.py
```diff
@@ -144,3 +204,99 @@ def call_ort(schema, args, kwargs, implicit_args=None):
...
+
+    def get_input(iter_num):
+        return [get_input_of(input_index, iter_num) for input_index in range(len(inputs))]
+
+    return [fun(*(get_input(i))) for i in range(len(inputs[0]))]
+
+
+# Used to control the default evaluator instance. A simple approach for now.
+
+default_ = ort_evaluator
```
justinchuby, line 275: This looks like it can be private: `_default_evaluator`

onnxscript/evaluator.py
gramalingam: done

onnxscript/evaluator.py
gramalingam: done

onnxscript/evaluator.py
gramalingam: done

onnxscript/evaluator.py
gramalingam: Well, actually onnxruntime session creation was failing for some test-case (because the output-type was not specified in the model). I am using the inference to get output-types here. (We could potentially avoid this in the future by calling onnx to do node-level inference, but currently that has limited functionality and Jacky is working on an ONNX PR to make it complete.)

onnxscript/test/eval_test.py
gramalingam: I don't understand. Happy to add a test, but what do you want to test and how?

onnxscript/evaluator.py
gramalingam: Done

onnxscript/evaluator.py
```diff
@@ -144,3 +204,99 @@ def call_ort(schema, args, kwargs, implicit_args=None):
...
+
+
+def default():
```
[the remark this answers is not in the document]
gramalingam: Done

onnxscript/evaluator.py
gramalingam: Done

onnxscript/test/eval_test.py
```diff
@@ -0,0 +1,30 @@
+import unittest
```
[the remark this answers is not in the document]
gramalingam: Done

onnxscript/values.py
```diff
@@ -122,9 +121,9 @@ def adapt_kwargs(self, kwargs):
         return kwargs, closure
 
     def __call__(self, *args, **kwargs):
-        kwargs, closure = self.adapt_kwargs(kwargs)
-        args = autocast.dynamic_cast_inputs(self.opschema, *args)
-        return self.evaluator(self.opschema, args, kwargs, closure)
+        from onnxscript import evaluator
```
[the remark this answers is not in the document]
gramalingam: I think so. I will leave this for now until various cyclic dependencies are sorted out and cleaned up.

onnxscript/values.py
```diff
@@ -165,6 +166,22 @@ def name(self):
...
+    def __getitem__(self, instance):
+        """Returns a lambda to evaluate function using given evaluator instance.
+
+        Usage:
+           script_fun(X) executes the function using the default evaluator instance.
+           script_fun[instance](X) executes the function using the given evaluator instance.
+        """
+
+        def fun(*args, **kwargs):
+            from onnxscript import evaluator
```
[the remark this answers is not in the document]
gramalingam: As above.

onnxscript/evaluator.py
gramalingam: Good point. Agree this is very important (formally pinning down the representation-types used by the different evaluators is essential, I think). But some of these is in flux, I'd prefer to do it a bit later.

onnxscript/values.py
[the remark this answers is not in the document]
justinchuby: Got it. Maybe include a TODO or FIXME then?

onnxscript/evaluator.py
```diff
@@ -0,0 +1,310 @@
...
+        opset_imports=[opset_id],
+        ir_version=irbuilder.select_ir_version(schema.since_version, domain=schema.domain),
+    )
+    model = onnx.shape_inference.infer_shapes(model)
+    # onnx.checker.check_model(model)
+    try:
+        sess = _cache_(model, ["CPUExecutionProvider"])
+    except (Fail, InvalidGraph, InvalidArgument) as e:
+        raise RuntimeError(
+            f"Unable to create onnxruntime InferenceSession "
```
xadupre, line 193: You should recall the schema.name in the error message. I guess that's what user would know, not the ONNX graph.

onnxscript/evaluator.py
xadupre: I would probably move the shape_inference in function `_cache_` then, and call it only if onnxruntime fails to load the model. The original model could still be used as a key for the cache.

onnxscript/evaluator.py
```diff
@@ -0,0 +1,310 @@
...
+
+    # Map ORT output values to the onnxscript representation-type.
+    return [ort_to_os_value(x) for x in result]
+
+
+def schema_id(schema):
+    return schema.name, schema.domain, schema.since_version
+
+
+class ORTEvaluator(Evaluator):
```
xadupre, line 220: If the plan is to have multiple evaluator, I would split this file into two: one for the base class, the other one with everything specific to onnxruntime.

onnxscript/test/evaluator_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+            @graph()
+            def square(y: FLOAT["N"]) -> FLOAT["N"]:  # noqa: F821
+                return op.Mul(y, y)
+
+            seq2 = op.SequenceMap(seq1, body=square)
+            return seq2
+
+        x = np.array([0.0, 1.0], dtype=np.float32)
+        output = seq_map[evaluator.ort_mixed_evaluator](x)
```
xadupre, line 24: You should add a unit test with a dummy evaluator to make sure the API works with another one.

onnxscript/evaluator.py
gramalingam: That's how I started, but I ended up simplifying things to avoid cyclic dependencies. I am going to leave it like this (at least until all cyclic dependencies are resolved for now).

onnxscript/test/evaluator_test.py
gramalingam: Done

onnxscript/evaluator.py
gramalingam: Done, thanks! (This is all pre-existing code, not part of current PR ... but I agree with your comment.)

onnxscript/evaluator.py
gramalingam: I think the caching can be improved and may be made more efficient. But let's do that separately, it's not the focus here. We need to select the right key for caching, but am not convinced this shape-inference will make a difference. I think the types do have to be part of the caching key, so it shouldn't make a difference.

PR #162, opened 2022-11-04
Strongly-typed static code generation for IntelliSense support

### LINKED ISSUE #83: [Usability] Generate pyi files for all ops
Having pyi files helps both users and type checkers / editors to understand what ops are available and what to expect. We can also add the documentation from the proto files to the generated pyi.

We can refer to https://github.com/microsoft/onnxruntime/blob/main/orttraining/orttraining/eager/opgen/opgen/generator.py and https://github.com/microsoft/onnxruntime/blob/main/orttraining/orttraining/eager/opgen/opgen/onnxops.py

### LINKED ISSUE #172: TensorType is not a type
It looks like TensorType is not a type. When I did

```python
def LeakyRelu(input, negative_slope: FLOAT | float = 0.01, inplace: BOOL | bool = False):
    ...
```

I got

```
TypeError: unsupported operand type(s) for |: 'TensorType' and 'type'
```

abock:
This implements static code generation for everything via `onnx.defs.get_all_schemas_with_history` while preserving the existing dynamic functionality (support for calling any arbitrary schema `(domain, version, opname)`).

**All ops are now supported (469 across core opsets 1-18, plus all the ML domains).**

Note that while we originally discussed this as being a `.pyi` typing overlay, the extra effort to go to 100% functional static code was miniscule, and the fully generated code is much nicer for debugging/stepping.

Using the `Hardmax` example, I have tested this PR for both ONNX model exporting and eager mode evaluation and observed no differences in behavior. No changes to existing ONNX Script code are needed.

_Note that the `onnx_opset/_impl` directory is meant to hide the op classes and modules themselves from IntelliSense._

## Notable Changes

1. To make any of the strong typing useful, the first commit in this PR changes the various `TensorType`s from _instances_ of `TensorType` to distinct subclasses of `TensorType`.

   Shape annotation subscripts still work by extending the class with `__class_getitem__` along with a _default instance_ stored on the class.

   Prior to this change, the _instances_ of `TensorType` would resolve as `Any` through Pylance/VS Code, because they cannot be statically reasoned about.

   No breaking changes observed to existing ONNX Script code.

2. Introduce a very simple Python AST (pygen.py) for generating the code needed to support the ops. Using Python's built-in `ast` for construction is far too tedious, especially when manipulating typings in the tree. Further, I expect to leverage this code generator in other scenarios involving ONNX and ONNX Script. For now I would like to leave it in the tree, but my intention is to spin it out.

3. Introduce the code generator for ONNX ops built upon (2), organizing each `(domain, version)` as its own Python module and class therein.

   Each version of an opset for a given domain subclasses the previous version to ensure the appropriate union across all opsets (`<=version`) are statically available. The first version of any opset subclases the original (dynamic) `onnxscript.values.Opset`.

4. Introduce the `opgen` module as the entrypoint for code generation. Generated code is run through `black` and `isort` for consistent formatting. I am inclined to want to store the resulting code in the repository so we can see diffs over time, but I have no real strong preference. The generated code is included in this PR for now for the sake of code review.

Closes #83, fixes #172


onnxscript/onnx_opset/_impl/opset1.py
```diff
@@ -0,0 +1,3072 @@
...
+        :param transB: Whether B should be transposed
+        """
+
+        schema = get_schema("Gemm", 1, "")
+        op = Op(self, "Gemm", schema)
+        return op(
+            A, B, C, alpha=alpha, beta=beta, broadcast=broadcast, transA=transA, transB=transB
+        )
+
+    def Or(self, A: BOOL, B: BOOL, axis: int = None, broadcast: int = 0) -> BOOL:
```
xadupre, line 94: `axis: int = None`, should it be `axis: Optional[int] = None`?

gramalingam:
I haven't looked at the details yet. But for variadic arguments, one key point is to distinguish the homogeneous from the heterogeneous types. In the first case, all the variadic inputs are expected to have the same type, and in the second, this is not the case.

I wonder if we can use python's `TypeVar` to make this distinction. Eg, could we use
```py
   def MyOp ( *varargs: Union[type1, type2])
```
versus
```py
   T = TypeVar("T", type1, type2)
   def MyOp ( *varargs: T)
```

I am not sure about Python's semantics for these, need to check.

setup.py
```diff
@@ -5,22 +5,66 @@
...
+        # it will be overridden with correct code as part of the generation
+        # below.
+        makedirs(base_path)
+        with open(base_path.joinpath("__init__.py"), "w"):
+            pass
+
+        from onnxscript.onnx_opset_builder import OpsetsBuilder
+        builder = OpsetsBuilder()
+        paths = builder.write(base_path)
+        subprocess.run(["black", "--quiet", *paths])
```
xadupre, line 54: If you use black, it may be better not to add the opset files to the repo. It may create unrelated changes to PR if a different version of black is used.

onnxscript/onnx_opset/_impl/opset_ai_onnx_ml2.py
```diff
@@ -0,0 +1,82 @@
...
+        default_string: str = "_Unused",
+        keys_floats: Sequence[float] = None,
+        keys_int64s: Sequence[int] = None,
+        keys_strings: Sequence[str] = None,
+        values_floats: Sequence[float] = None,
+        values_int64s: Sequence[int] = None,
+        values_strings: Sequence[str] = None,
+    ) -> Union[FLOAT, INT64, STRING]:
+        """
+        Maps each element in the input tensor to another value.<br>
```
xadupre, line 39: I wonder how this one renders since the indentation is weird and it includes html tags.

onnxscript/onnx_opset/_impl/opset5.py
```diff
@@ -0,0 +1,93 @@
...
+        First input is the data tensor, second input is a shape tensor which specifies the output shape. It outputs the reshaped tensor.
+        At most one dimension of the new shape can be -1. In this case, the value is
+        inferred from the size of the tensor and the remaining dimensions. A dimension
+        could also be 0, in which case the actual dimension value is unchanged (i.e. taken
+        from the input tensor). Shape (second input) could be an empty shape, which means converting to a scalar.
+        The input tensor's shape and the output tensor's shape are required to have the same number of elements.
+
+        :param data: An input tensor.
+        :param shape: Specified shape for output.
+        """
```
xadupre, line 89: No `:return:`?

onnxscript/onnx_opset/_impl/opset2.py
```diff
@@ -0,0 +1,170 @@
...
+        strides: Sequence[int] = None,
+    ) -> Union[DOUBLE, FLOAT, FLOAT16]:
+        """
+        LpPool consumes an input tensor X and applies Lp pooling across
+         the tensor according to kernel sizes, stride sizes, and pad lengths.
+         Lp pooling consisting of computing the Lp norm on all values of a subset
+         of the input tensor according to the kernel size and downsampling the
+         data into the output tensor Y for further processing.
+
+        :param X: Input data tensor from the previous operator; dimensions for image case are (N x C x H x W), where N is the batch size, C is the number of channels, and H and W are the height and the width of the data. For non image case, the dimensions are in the form of (N x C x D1 x D2 ... Dn), where N is the batch size.
```
xadupre, line 57: Should we split long lines into two?

onnxscript/onnx_opset/_impl/opset15.py
```diff
@@ -0,0 +1,313 @@
...
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+from typing import Tuple, Union
+
+from onnx.defs import get_schema
+
+from ...onnx_types import (
```
xadupre, line 13: Relative or absolute import?

onnxscript/onnx_opset/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+from ._impl.opset16 import Opset16 as __Opset16
+from ._impl.opset17 import Opset17 as __Opset17
+from ._impl.opset_ai_onnx_ml1 import Opset_ai_onnx_ml1 as __Opset_ai_onnx_ml1
+from ._impl.opset_ai_onnx_ml2 import Opset_ai_onnx_ml2 as __Opset_ai_onnx_ml2
+from ._impl.opset_ai_onnx_ml3 import Opset_ai_onnx_ml3 as __Opset_ai_onnx_ml3
+from ._impl.opset_ai_onnx_preview_training1 import (
+    Opset_ai_onnx_preview_training1 as __Opset_ai_onnx_preview_training1,
+)
+
+all_opsets: list[Opset] = []
```
xadupre, line 37: List?

onnxscript/onnx_opset/_impl/opset15.py
```diff
@@ -0,0 +1,313 @@
...
+        schema = get_schema("Bernoulli", 15, "")
+        op = Op(self, "Bernoulli", schema)
+        return op(input, dtype=dtype, seed=seed)
+
+    def BatchNormalization(
+        self,
+        X: Union[BFLOAT16, DOUBLE, FLOAT, FLOAT16],
+        scale: Union[BFLOAT16, DOUBLE, FLOAT, FLOAT16],
+        B: Union[BFLOAT16, DOUBLE, FLOAT, FLOAT16],
+        input_mean: Union[BFLOAT16, DOUBLE, FLOAT, FLOAT16],
```
gramalingam, line 207: Can we capture typing constraints in ONNX that require several inputs to be of the same type using python's `TypeVar`? (Not necessarily in this PR, it could be a refinement later too.)

gramalingam:
Is the plan to use qualified names to handle the conflicts around `Optional` and `Sequence`?

onnxscript/onnx_opset_builder.py
```diff
@@ -0,0 +1,375 @@
...
+def parse_input_output_type(onnx_type: str) -> TypeRef:
+    try:
+        id = ""
+        stack: list[TypeRef] = []
+        for c in onnx_type:
+            if c == "(":
+                if id == "tensor":
+                    type = TensorTypeRef()
+                elif id == "seq":
+                    type = SequenceTypeRef()
```
gramalingam, line 290: One question we need to resolve is whether we should use distinct type (annotation) for attributes and values. My instinct is to differentiate them, but maybe we might be able to get by without doing it for Sequences. It may also tie in with what warnings IDEs give, relating to ONNX Scripts restrictions (though, we may want to relax them eventually: eg., supporting the use of `[X, Y, Z]` for a sequence of tensors.)

onnxscript/onnx_opset/_impl/opset1.py
abock: Yes, I need to add Optional and variadic support

onnxscript/onnx_opset/_impl/opset15.py
abock: I think we can refine this in a subsequent PR. The problem, AFAIK, is that Python does not support function/member level "generics". So we could generate these union types through `TypeVar`, but there will be many, and how do we name them so that they're useful in the IntelliSense context? If I am wrong, please let me know - I would love to specify generics on the functions!

abock:
Regarding using `TypeVar` for generics, unfortunately it seems useless from an IntelliSense UX perspective, which defeats the entire purpose of this PR.

Screen Shot 2022-11-07 at 5 28 51 PM

onnxscript/_pygen.py
```diff
@@ -0,0 +1,507 @@
...
+            self.return_type.write(writer)
+        writer.write(":\n")
+
+        def get_doc_string():
+            doc = ""
+            if self.doc:
+                doc += self.doc + "\n\n"
+            for param in self.params:
+                if param.doc:
+                    doc += f":param {param.name}: {param.doc}\n"
```
justinchuby, line 244: Maybe create docstring in google style for consistency? https://google.github.io/styleguide/pyguide.html#doc-function-raises

justinchuby (commented):
Just a question for my learning: I wonder wondering where indentation is handled? Was it in the visitor?

justinchuby:
> Regarding using `TypeVar` for generics, unfortunately it seems useless from an IntelliSense UX perspective, which defeats the entire purpose of this PR.

Would it make sense to define a few sets of common type collections as TypeVar, so we get both?

gramalingam:
> Regarding using `TypeVar` for generics, unfortunately it seems useless from an IntelliSense UX perspective, which defeats the entire purpose of this PR.
> 
> Screen Shot 2022-11-07 at 5 28 51 PM

That's unfortunate! Yes, it is hard to reconcile the use of different tools, which may not all support all the features of python.

onnxscript/_pygen.py
```diff
@@ -0,0 +1,507 @@
...
+
+class ImportAdjuster(Visitor):
+    """
+    Adjusts typings in the codegen AST to aliases and generates
+    associated imports at the top of the module.
+    """
+
+    def __init__(self):
+        self.current_module: Optional[ModuleDef] = None
+        self.imports: dict[str, set[str]] = {}
```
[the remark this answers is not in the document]
abock: I am pretty sure ONNX Script requires 3.10 or newer. Not using the `typing` definitions was intentional. I was also under the impression that starting with 3.0 importing from `__future__` is not necessary?

onnxscript/onnx_opset/__init__.py
abock: Resolving this to [keep discussion below](https://github.com/microsoft/onnx-script/pull/162#discussion_r1016849642).

onnxscript/_pygen.py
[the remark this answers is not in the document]
justinchuby: Are we concerned about the barrier of entry if 3.10 is the lowest version supported? As far as I know many users are still on 3.7-3.9.

image

Many futures are removed but `from __future__ import annotations` is a new thing [introduced in 3.7](https://peps.python.org/pep-0585/#implementation)

gramalingam:
Re. the point about `Split` having an attribute and input with the same name. That is not in the same version: in opset 11 it was an attribute, and in opset 13, it was changed to an input. Why is this an issue or concern?

justinchuby:
@gramalingam looks like it's [opset1](https://natke.github.io/onnx/operators/onnx__Split.html#split-1):

> Attributes
> axis: Which axis to split on
> split: length of each output
> 
> Inputs
> Between 1 and 2 inputs.
> input (heterogeneous) - T: The tensor to split
> split (optional, heterogeneous) - T: Optional list of output lengths (see also arg ‘split’)
>

setup.py
```diff
@@ -5,22 +5,67 @@
 
 # -*- coding: utf-8 -*-
 
-import os
-
 import setuptools
+import subprocess
+
+from distutils.cmd import Command
```
justinchuby, line 11: Looks like distutils is deprecated https://docs.python.org/3/library/distutils.html

gramalingam:
> @gramalingam looks like it's [opset1](https://natke.github.io/onnx/operators/onnx__Split.html#split-1):
> 
> > Attributes
> > axis: Which axis to split on
> > split: length of each output
> > Inputs
> > Between 1 and 2 inputs.
> > input (heterogeneous) - T: The tensor to split
> > split (optional, heterogeneous) - T: Optional list of output lengths (see also arg ‘split’)

Oh, thanks, I didn't go back all the way :) ... I thought I would fix the issue in the op definition if it was in the latest (opset 18) by renaming the input. I guess the code-generator could make the input-name unique in such cases by adding some prefix/suffix like "input_". The attribute name is significant, but the input name is not.

abock:
Codegen is now at 100% spec coverage and all tests are passing.

However, I'm not entirely sure what to do about these failing CI steps:

- `codecov/patch` and `codecov/project`: I assume this is now scanning the generated code? Can we exclude froms scanning?

- `Lint / Lint Python`: it's funny that now the ONNX Script CI is catching errors in ONNX docs, because the docs are now code! However, this shouldn't fail ONNX Script (again, exclude from scanning?)
   - It seems all this does is look for misspellings? Can we label the step to indicate that?
   - `lintrunner -m origin/main` is happy locally... this step should run as part of `lintrunner`?

onnxscript/onnx_opset/_impl/opset_ai_onnx_ml2.py
abock: They're now run through the same function as ONNX's docsgen.

onnxscript/onnx_opset/_impl/opset5.py
abock: TODO. I'd like to discuss docstrings in more detail first.

justinchuby:
1. codecov is non-blocking. It is just using algo to suggest a target coverage. If desired `codecov.yml` can be configured: https://docs.codecov.com/docs/ignoring-paths.
2. Lint Python is non-blocking and yes it's just checking for misspellings :). It's mainly useful for PR reviews to catch typos.
3. Version: I made a PR to add the `annotations` future import to all files, so we can use the new syntax and maintain compatibility with older pythons. I think this will position us the best in terms of modern python support and allow the most users as most people are still on python 3.7 iirc (or 3.6? even though it's no longer supported https://w3techs.com/technologies/details/pl-python/3) @gramalingam @xadupre for comments?

abock:
@justinchuby does codecov only apply to the current PR? In other words if this is merged as-is does codecov start failing for subsequent PRs or is a new baseline established?

As for the Python version we discussed this in the morning meeting and decided to defer to you. I also added `from __future__ import annotations` to this PR, so we at least have consistency. My preference is to use the builtins vs `typing`s when applicable.

justinchuby:
There will be a new baseline established. The rest sounds perfect. Thanks!

justinchuby (approved):

justinchuby:
> Can we label the step to indicate that?

I will change the workflow name

gramalingam:
@abock : question: I thought your plan was not to include the generated code in the repos? If so, I guess those files need to be removed?

In this case, it would be useful to add some documentation, preferably in readme.md, informing users what they need to do to generate and use it for intellisense etc.

abock:
@gramalingam actually I'm not sure we settled on it. I originally didn't care much one way or another. I don't typically love generated code in the tree, but I can also see some value in it.

If we don't include it in the repo, it needs to be invoked during build/install, so I could use some guidance there.

To run the generator, from the top of the repo:

```bash
python -m opgen
```

abock:
@gramalingam I've added another commit to the PR that demonstrates what this would like like over time. This commit reflects updates to generated code from updating ONNX from 1.12.0 to 1.13.0rc1. You can see some of the bug fixes with regards to type changes, documentation fixes to older opsets, and the introduction of opset18.

onnxscript/onnx_opset/__init__.py
```diff
@@ -0,0 +1,191 @@
...
+    (
+        "ai.onnx.ml",
+        3,
+    ): opset_ai_onnx_ml3,
+    (
+        "ai.onnx.preview.training",
+        1,
+    ): opset_ai_onnx_preview_training1,
+}
+default_opset: Opset14 = all_opsets[
```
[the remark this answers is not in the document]
abock: mypy, you are the worst!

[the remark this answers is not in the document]
justinchuby: is default set to be Opset14? Can we have a "latest" if that is the case?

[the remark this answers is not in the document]
justinchuby: Using `typing.Cast` can assure mypy we know what we are doing.

justinchuby:
Personally I like the idea of checking these in first before we make it a build step, so that we can leverage all the goodness before we figure out the build.

gramalingam:
> @gramalingam actually I'm not sure we settled on it. I originally didn't care much one way or another. I don't typically love generated code in the tree, but I can also see some value in it.

Sounds good. I think we can merge it in its current form, and revisit this if we need to.

gramalingam (approved):

onnxscript/onnx_types.py
```diff
@@ -53,6 +55,13 @@ def __getitem__(self, shape: Optional[ShapeType]):
             shape = (None,)
         return TensorType(self.dtype, shape)
 
+    def __class_getitem__(cls, shape: Optional[ShapeType]):
+        if cls.default_instance is None:
+            raise TypeError(f"{cls} does not specify a default_instance.")
+        # pylint erroneously flags with unsubscriptable-object if
+        # using subscript notation (cls.default_instance[shape]):
+        return cls.default_instance.__getitem__(shape)
```
justinchuby, line 63: Does this return a type?

justinchuby: @abock looks like we still need https://github.com/microsoft/onnx-script/issues/219

PR #167, opened 2022-11-09
Add test case for attributes

gramalingam:
Add test case for various types of attributes (for both converter as well as eager-mode),


onnxscript/test/converter_test.py
```diff
@@ -539,7 +546,7 @@ def check_run(self, onnxfn, inputs, expected_output):
         np.testing.assert_equal(output, expected_output)
 
         # Test running model in eager mode
-        output = onnxfn(*inputs)
+        output = onnxfn._usercall(*inputs)
```
justinchuby, line 549: Wondering if there will be a public method for this?

gramalingam: I am not sure. There is a technical question: there are two conventions for argument/result types. Whether we can rationalize this and automate this or not is to be seen. I would like to hide it if possible, but am not sure we will always be able to.

gramalingam: Ok, eliminated/rewrote it.

justinchuby: Awesome thanks:) just asking questions for my education as well. I was wondering what _usercall does. Is the new version doing the same thing?

gramalingam: So, the main difference is: eager-mode execution of onnxscript functions use our own Tensor type, which is a simple wrapper around numpy arrays, used, for example, to define our own overloaded ops. However, at the top-level (user-facing), it is convenient to remove the wrapper and return numpy arrays. That's main difference between `_usercall` and `_libcall`. (I think there is some cleaning up to do with regards to these methods, which I will do as a separate PR, as it is unrelated to current PR.)

justinchuby: Thanks for explaining!

gramalingam: Any other concerns? The PR still needs an approval to be merged :)

justinchuby (approved):
LGTM!

onnxscript/test/converter_test.py
```diff
@@ -587,6 +597,11 @@ def inner():
                 Temp = op.Abs(X)
                 return op.DummyOp(body=inner)
 
+    def test_attr(self):
+        from onnxscript.test.functions import attr_test
```
justinchuby, line 601: Sorry one more thing:

justinchuby (approved):

onnxscript/test/converter_test.py
gramalingam: I prefer to leave it as is. It is a test-suite, but we make it a separate file to keep "onnxscript code" separate (will make it easier to turn off linter/style-checker like tools from being applied to these test-cases). Loading it when needed makes it easier to run test-cases incrementally for debugging etc.

PR #169, opened 2022-11-10
Add INT type in args parser

xiaowuhu:
Signed-off-by: xiaowuhu <xiaowuhu@microsoft.com>

add INT type into args parser, otherwise, in the below sample,  the param N cannot be passed into function.

```python
from onnxscript import script
from onnxscript.onnx_opset import opset15 as op
import numpy as np

@script()
def sumprod(x, N):
    sum = op.Identity(x)
    prod = op.Identity(x)
    for i in range(N):
        sum = sum + x
        prod = prod * x
    return sum, prod


x = np.array([2])
N = 3
a,b = sumprod(x, N)
print(a,b)
```


onnxscript/values.py
```diff
@@ -199,6 +199,8 @@ def _usercall(self, *args, **kwargs):
                 new_args.append(tensor.Tensor(a))
             elif isinstance(a, bool):
                 new_args.append(tensor.Tensor(np.array(a)))
+            elif isinstance(a, int):
+                new_args.append(a)
```
xadupre, line 203: why not `elif isinstance(a, (bool, int, float)):

gramalingam:
I am okay with this change for now. However, I think we need to think through the contract (types of values expected/supported in various scenarios), document it, and enforce it, in the longer term. (Eg., technically, the input is expected to be a tensor above, and `N` should be `np.array(3)`, and the above extension is just a convenience. It's fine as long as it doesn't create any other complications anywhere else. Otherwise, some of these conversions end up masking issues elsewhere in the system.)

onnxscript/values.py
xiaowuhu: if use this:
```
 elif isinstance(a, (bool, int, float)):
```

the N in below code will be Tensor(array(3)), so the range(N) will throw exception.

```
for i in range(N):
```

xadupre: According to the code I wrote a while ago at this line: https://github.com/microsoft/onnx-script/blob/main/onnxscript/test/models/signal_dft.py#L364, `range(int(N))` should work if `N` is a tensor. I might be wrong but the change you made is not consistent with the line just above, so I fear another bug arises later.

xiaowuhu: Yes, range(int(N)) works. But from the user perspective, if the 'user' is us, I feel OK. if the 'user' is external user, they may feel weird, because from the context of code, the N is already INT type.
I'm not talking my change is correct or not, I care about user experience, because we have a task from Aaron that we need to give spec about how to let user write script in reasonable way. :)

justinchuby: Is it possible to allow a scalar tensor behave like a Python scalar like numpy does?

gramalingam: > Is it possible to allow a scalar tensor behave like a Python scalar like numpy does?

Yes, since `range` works for a numpy array, maybe we can do the same. Maybe we just need to add an `__int__` method to our Tensor class? Must be something like that?

justinchuby: Looks like it's `__index__` called:

```
>>> for i in range(np.array([1,2])): print("loop", i)
...
TypeError: only integer scalar arrays can be converted to a scalar index
```

justinchuby: #181

PR #171, opened 2022-11-10
Remove _libcall usage

gramalingam:
Remove the explicit usage of _libcall. It's not needed.

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -8,6 +8,7 @@ import logging
 import sys
 import typing
 from enum import IntEnum
+from typing import NoReturn
 
 import numpy
 import onnx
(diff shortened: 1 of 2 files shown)


onnxscript/converter.py
```diff
@@ -844,21 +844,16 @@ def translate_opset_expr(self, node) -> values.Opset:
 
     def translate_callee_expr(self, node) -> values.Op:
         """Return an Op"""
```
justinchuby, line 846: Worth fixing this check?

justinchuby (approved):
Probably not the best person to review but lgtm

onnxscript/converter.py
gramalingam: Done (I hope). I added the `NoReturn` annotation to the `fail` methods.

justinchuby: Thanks!

justinchuby: Looks like CodeQL isn't smart enough. that's ok

PR #174, opened 2022-11-11
Playing with opinfo

justinchuby:


playground/test_func.py
```diff
@@ -0,0 +1,97 @@
...
+def LeakyRelu(input, negative_slope: float = 0.01, inplace: bool = False):
+    zero = op.CastLike(0, input)
+    negative_slope = op.CastLike(negative_slope, input)
+    return op.Where(input < zero, negative_slope * input, input)
+
+
+@script()
+def FuncWithStr(input, negative_slope: float = 0.01, mode: str = "foo"):
+    if mode == "foo":
+        return input
```
gramalingam, line 45: FYI: 
(a) `return` inside control-flow constructs is not supported.
(b) identity assignment is not supported (though probably easy to add)
Suggest using
```py
   if mode == "foo":
      result = op.Identity(input)
   else:
      result = ...
   return result
```

justinchuby: Perfect, thanks!

PR #181, opened 2022-11-11
Implement `__index__` and `__float__` on Tensor

justinchuby:
Implement `__index__` and `__float__` to support `range(<Tensor>)` and `float(<Tensor>)`. Also added types.


gramalingam (approved):
A test-case calling `range` on a Tensor object would be helpful.

justinchuby:
@gramalingam I added tests

PR #184, opened 2022-11-11
Add support for returning input as output

gramalingam:
Add support so that graph-inputs can be returned as graph-outputs (which requires introducing an Identity op in ONNX).


justinchuby (approved):
LGTM🚀

PR #185, opened 2022-11-12
Move common test utils to common directory

justinchuby:
* __->__ #185

Move testutils and checker to the common directory to make their functions clear and promote reuse. So we have test files structured in a way resembling the source

Also parameterized `eager_test`


gramalingam:
Looks fine to me. IMO, though, there are other restructuring that would be more helpful, but that's a separate issue, I guess. (Eg., the subfolders `functions` and `models` are misleadingly named, not sure why. But rationalizing it is more than just renaming.)

gramalingam (approved):

PR #195, opened 2022-11-15
Handle copy within loop body

gramalingam:
A continuation of the recent PR to handle copy-assignments. This handles such assignments inside a loop body. (Forgot to merge this change with the previous PR).

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -1232,6 +1232,8 @@ class Converter:
         )
         for pv in loop_state_vars:
             ov = self.py_var_to_onnx_var(pv, debuginfo.DebugInfo(loop_stmt, self))
+            if ov not in self.current_fn.assigned_names:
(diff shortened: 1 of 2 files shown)


onnxscript/converter.py
```diff
@@ -1231,6 +1231,8 @@ def translate_loop_stmt(self, loop_stmt: Union[ast.For, ast.While]):
         )
         for pv in loop_state_vars:
             ov = self.py_var_to_onnx_var(pv, debuginfo.DebugInfo(loop_stmt, self))
+            if ov not in self.current_fn.assigned_names:
```
justinchuby, line 1234: Possible to add a comment on the situation this is handling? I think that would help future readers a lot

justinchuby (approved):

onnxscript/converter.py
gramalingam: Good point. Added some comments.

PR #199, opened 2022-11-18
Cleanup _libcall and _usercall distinction

gramalingam:
Cleanup _libcall and _usercall distinction


onnxscript/values.py
```diff
@@ -183,77 +183,73 @@ def fun(*args, **kwargs):
 
         return fun
 
+    def wrap(self, inputs):
```
justinchuby, line 186: I would consider a more descriptive name?

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@ def fun(*args, **kwargs):
...
+        Args:
+            inputs: a list/tuple of inputs to an ONNX function
+
+        Returns:
+            a pair (wrapped_inputs, flag) where flag indicates whether any numpy array
+            was wrapped into a Tensor.
+        """
+        has_array = False
+
+        def do_wrap(input):
```
justinchuby, line 206: Looks like do_wrap do not need to clse over local variables? If that's the case I'd prefer module level functions which are easier to test and maintain

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@ def fun(*args, **kwargs):
...
+            if isinstance(input, np.ndarray):
+                nonlocal has_array
+                has_array = True
+                return tensor.Tensor(input)
```
justinchuby, line 210: Possible to return has_array as the second return var?

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@ def fun(*args, **kwargs):
...
+            return output.value
+        elif output is None:
+            return None
+        elif isinstance(output, list):
+            return [self.unwrap(elt) for elt in output]
+        elif isinstance(output, tuple):
+            return tuple([self.unwrap(elt) for elt in output])
+        elif isinstance(output, np.ndarray):
+            return output
+        raise TypeError(f"Unexpected type {type(output)}.")
```
justinchuby, line 246: List the expected ones, or annotate input?

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@ def fun(*args, **kwargs):
...
-                if isinstance(r, tensor.Tensor):
-                    unwrapped.append(r)
-                else:
-                    raise TypeError(
-                        f"Unexpected output type {type(r)} for an output {i} "
-                        f"in function {self.function!r}."
-                    )
-            return tuple(unwrapped)
-        raise TypeError(f"Unexpected output type {type(res)} in function {self.function!r}.")
+        new_args, has_array = self.wrap(args)
```
justinchuby, line 250: Just for my learning, way is input with nparray treated differently?

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@
...
+            elif isinstance(input, tensor.Tensor):
+                return input
+            elif isinstance(input, (bool, int, float)):
+                return tensor.Tensor(np.array(input))
+            elif input is None:
+                return None
+            elif isinstance(input, list):
+                return [do_wrap for elt in input]
+            elif isinstance(input, tuple):
+                return tuple([do_wrap(elt) for elt in input])
```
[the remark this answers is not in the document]
justinchuby: We should probably disable this rule in these lines as a generator can have a bigger overhead for this type of uses.

onnxscript/values.py
gramalingam: First, eager-mode execution uses Tensor(...) as a wrapper around numpy arrays to define overloaded operators like +. So, if the user input is a nparray, it is automatically wrapped used a Tensor object.

Second, for user convenience, we unwrap such a Tensor wrapper when returning values to the user at the top-level.

justinchuby: I see thanks! I think it'd be very helpful to include this in a comment too

onnxscript/values.py
gramalingam: Yes, I considered that. It might be better to do that. (I ended up with this choice in case we want to define derived classes where we want to alter this behavior by overriding this method. But I don't see any immediate use for such overriding.)

onnxscript/values.py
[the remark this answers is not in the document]
gramalingam: Will it? I don't fully understand what the compiler does in the two cases. Does list-comprehension work-in-place without using a generator? Why can't `tuple(e for x in l)` produce similar code to `[e for x in l]`?

onnxscript/values.py
```diff
@@ -183,77 +183,73 @@
...
+                has_array = True
+                return tensor.Tensor(input)
+            elif isinstance(input, tensor.Tensor):
+                return input
+            elif isinstance(input, (bool, int, float)):
+                return tensor.Tensor(np.array(input))
+            elif input is None:
+                return None
+            elif isinstance(input, list):
+                return [do_wrap for elt in input]
```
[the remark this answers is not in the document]
gramalingam: Oops, renaming copy-paste eliminated the `(elt)` part after `do_wrap` !

onnxscript/values.py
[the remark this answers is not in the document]
justinchuby: `tuple(e for x in l)`:

a generator is first created by `e for x in l`, then `tuple` consumes it to initialize the tuple. 

`tuple([e for x in l])`:

a list is created by the comprehension, then used by `tuple`.

Last time I tested the generator version was slower. This may have changed, and the generator version is admittedly cleaner.

onnxscript/values.py
```diff
@@ -142,6 +142,70 @@ class OnnxClosure:
     function: Any
 
 
+def adapt_to_eager_mode(inputs):
```
justinchuby, line 145: Should we hide this function? Same for the other

```suggestion
def _adapt_to_eager_mode(inputs: Union[np.ndarray, tensor.Tensor, bool, Number, list[Any], tuple[Any, ...]):
```

onnxscript/values.py
```diff
@@ -142,6 +142,70 @@ class OnnxClosure:
...
+        elif input is None:
+            return None
+        elif isinstance(input, list):
+            return [adapt(elt) for elt in input]
+        elif isinstance(input, tuple):
+            return tuple(adapt(elt) for elt in input)
+        raise TypeError(f"Unexpected input type {type(input)}.")
+
+    result = adapt(inputs)
+    return result, has_array
```
justinchuby, line 183: optional but I think some tests will be helpful

justinchuby (approved):

gramalingam:
@justinchuby do you know why the lint check is failing? It's complaining about "cls.atol = 1e-7  # type: ignore[attr-defined]", but I didn't touch this line at all ... I guess it must already have been there

onnxscript/values.py
gramalingam: For the type, the ideal would be to use a recursive type definition (https://github.com/python/mypy/issues/731) ... do you think we can use it, or is it too experimental still?

justinchuby:
I fixed it in the previous PR. Does rebasing help? The issue was because lintrunner previously ignored some mypy errors that it shouldn’t and fixed it in the package

onnxscript/values.py
justinchuby: I see! I didn't consider the recursion. lgtm!

onnxscript/values.py
justinchuby: From the comments it's one by default for mypy 0.990. since we are using 0.991 it may be worth a try

gramalingam:
I ended up leaving the type-definitions commented out: it is a bit non-trivial to get all typing annotations to work (as other existing code doesn't have necessary type annotations), and I will need to get other things setup to tackle something like this.

PR #200, opened 2022-11-21
Enable delayed type annotation evaluation according to PEP 563

justinchuby:
- Add `from __future__ import annotations` to all files needed
- Use the new style type annotation https://peps.python.org/pep-0604/
- Fix mypy errors

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,7 +2,8 @@
 # Please move this into pyproject.toml, when it does.
 
 [flake8]
-max-line-length = 120  # 120 is a hard limit. Black keeps lines generally at 88.
+# 120 is a hard limit. Black keeps lines generally at 88.
+max-line-length = 120
 
 # W503 talks about operator formatting which is too opinionated.
 # E203 is need to support black formatting.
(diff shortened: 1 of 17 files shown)


.flake8
```diff
@@ -2,15 +2,16 @@
 # Please move this into pyproject.toml, when it does.
 
 [flake8]
-max-line-length = 120  # 120 is a hard limit. Black keeps lines generally at 88.
+# 120 is a hard limit. Black keeps lines generally at 88.
```
justinchuby, line 5: Fix syntax

.flake8
```diff
@@ -2,15 +2,16 @@
...
+max-line-length = 120
 
 # W503 talks about operator formatting which is too opinionated.
 # E203 is need to support black formatting.
 # https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html
 # Dxx ignores are for pydocstyle. It should mirror the [tool.pydocstyle] section in
 # pyproject.toml. Mirrored here for flake8-docstring support. See pyproject.toml for
 # reasons for disabling.
-extend-ignore = E203, W503, D1, D202, D205, D041, D415
```
justinchuby, line 13: typo

gramalingam:
Out of curiosity: is there a compelling reason to replace `Optional[T]` by `T | None` ? It seems to me that the first is a more readable abbreviation. Replacing `Union(X, Y)` by `X | Y` makes sense.

gramalingam (approved):

justinchuby:
Done

PR #208, opened 2022-11-23
feat(atenlib): create tests with OpInfo

justinchuby:
* #235
* __->__ #208
* #223
* #212


onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,227 @@
...
+    torch.int16,
+    torch.int32,
+    torch.int64,
+    # Floating types
+    torch.float16,
+    torch.float32,
+    torch.float64,
+)
+
+# Convenience tuples for creating dtype lists when skipping or xfailing tests
```
fatcat-z, line 32: Typo: xfailing -> failing.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def xfail(
+    op_name: str,
+    variant_name: str = "",
+    *,
+    dtypes: Optional[Collection[torch.dtype]] = None,
+    reason: Optional[str] = None,
+):
+    """Expects a OpInfo test to fail.
```
fatcat-z, line 72: Typo: a->an

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def skip(
+    op_name: str,
+    variant_name: str = "",
+    *,
+    dtypes: Optional[Collection[torch.dtype]] = None,
+    reason: Optional[str] = None,
+):
+    """Skips a test case in OpInfo.
```
fatcat-z, line 98: What's the difference between "a test case" and "an OpInfo test"?

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Done

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: They are the same. Updated.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,227 @@
...
+        dtypes=dtypes,
+        reason=reason,
+    )
+
+
+def add_decorate_info(
+    all_opinfos: Sequence[opinfo_core.OpInfo],
+    test_class_name: str,
+    base_test_name: str,
+    skip_or_xfails: Iterable[DecorateMeta],
```
fatcat-z, line 121: Could this be optional, or the default value could be None?

justinchuby: skip_or_xfails is always required. Empty list if nothing to skip. Thoughts?

fatcat-z: Either an empty list or None works for me.

gramalingam:
Typo in folder name "function_libs"

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Going for empty list

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4,3 +4,42 @@
...
+    del name
+
+    def decorator(func):
+        return func
+
+    return decorator
+
+
+# TODO: put this in nn
+@atenop("aten::relu6")
```
gramalingam, line 25: supporting decorators like this is tricky. Even though currently onnxscript does not produce any error messages when applied to this, I think the safer thing to do is to report an error message since we don't know the effect of the decorator in general. I wonder what you intend to use this for (in case there's a requirement that onnxscript could support).

justinchuby: I intend to use the decorator to register the function to a dictionary-type container. The decorator should be a no-op from onnxscript's perspective.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -151,9 +156,26 @@ def aten_cross_entropy_loss(
...
+    input_scale: Annotated[float, Is[lambda x: x == 1.0]] = 1.0,
 ) -> TensorType:
-    raise NotImplementedError()
+    # del scale
+    # del input_scale
+    return op.Elu(self, alpha=alpha)
+
+
+def aten_elu__int(
+    self: IntType,
```
justinchuby, line 170: @fatcat-z @xiaowuhu thoughts on reconciling data type that is not supported by the onnx native op?

fatcat-z: Even these are aten ops, I still suggest the function signature is ONNX compatible only. We don't need to support any Torch data type which is unsupported by ONNX in these function signatures.
We need to transfer those data types to ONNX compatible ones (Numpy/Python) before we call these aten_op() functions. If necessary, we might provide a helper lib which helps to convert data from PyTorch data format to ONNX data format. 
Such lib could be a new Python package or lives in ONNX Script package.

justinchuby: Sounds like a good idea to me

justinchuby: @titaiwangms is offering to help think about this aspect.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,253 @@
...
+@dataclasses.dataclass
+class DecorateMeta:
+    """A dataclass for storing information about a test case to skip or xfail.
+
+    Adapted from functorch: functorch/test/common_utils.py
+    """
+
+    op_name: str
+    variant_name: str
+    decorator: Callable
```
justinchuby, line 68: TODO

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,253 @@
...
+        reason=reason,
+    )
+
+
+def add_decorate_info(
+    all_opinfos: Sequence[opinfo_core.OpInfo],
+    test_class_name: str,
+    base_test_name: str,
+    skip_or_xfails: Iterable[DecorateMeta],
+):
```
justinchuby, line 130: Add return type

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4105,10 +4107,10 @@ def aten_select_scatter(self: TensorType, src: TensorType, dim: int, index: int)
     raise NotImplementedError()
 
 
-def aten_selu(self: TensorType) -> TensorType:
+def aten_selu(self: TFloat) -> TensorType:
     # selu(Tensor self) -> Tensor
 
-    raise NotImplementedError()
+    return op.Selu(self)
```
fatcat-z, line 4113: Why do we put the logic like this within ONNX Script? As we discussed before, the logic relative to PyTorch specifically should be left in PyTorch exporter side, right?

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,253 @@
...
+        "TestOutputConsistency",
+        "test_output_match",
+        skip_or_xfails=EXPECTED_SKIPS_OR_FAILS,
+    )
+    def test_output_match(self, device: str, dtype: torch.dtype, op):
+        """Base test method for testing each opset, used by instantiate_device_type_tests."""
+        # device is provided by instantiate_device_type_tests, but we only want to run in cpu.
+        assert device == "cpu"
+
+        samples = op.sample_inputs(
```
fatcat-z, line 214: Why isn't there "shape" information for a sample input?

requirements-dev.txt
```diff
@@ -12,6 +12,8 @@ sphinx-gallery
 pydata_sphinx_theme
 
 # ATen lib
+typing_extensions
```
fatcat-z, line 15: Could you re-sort them in alphabetical order?

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -0,0 +1,253 @@
...
+                torch_output = op(*inputs, **cpu_sample.kwargs)
+                try:
+                    function_output = scripted_function(*input_numpy, **cpu_sample.kwargs)
+                except onnxruntime.capi.onnxruntime_pybind11_state.NotImplemented:
+                    self.skipTest(
+                        f"ONNX Runtime doesn't support running {op.name} with dtype {dtype}",
+                    )
+
+                # Use torch testing to ensure dtypes and shapes match
+                torch.testing.assert_close(
```
fatcat-z, line 241: Do we need to consider the case of multiply outputs?

How about this:

assert [torch.allclose(o, torch.tensor(o_ort)) for o, o_ort in zip(torch_output, function_output)]

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: More of (2) should be in the function lib, so only the glue logic (not expressed by functions) lives in the exporter. We should aim for minimal op logic on the exporter (but allow full control at the same time so changes can be made independently)

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: The shape is not explicitly used in this test because onnxscript can handle it in its evaluator. Any considerations?

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Good catch. I am inclined to keep it as is for now and expand to multi input when needed

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: About "minimal op logic on the exporter", may I know the reason why we are doing this? What's the difference between calling these troch_aten ops and onnx scripts ops from PyTorch exporter?

fatcat-z (approved):
LGTM, thanks!

requirements-dev.txt
justinchuby: Done

justinchuby:
Accidentally merged this into #223. I think that's ok and I will close this.

PR #214, opened 2022-11-29
opgen: run lintrunner instead of black/isort

abock:
The SPACES linter has some terrible O(N^?) perf when explicitly passing paths that comprise tens of thousands of lines of code, so skip it here. The generator doesn't write lines with trailing spaces anyway.

### DIFF
--- a/opgen/__main__.py
+++ b/opgen/__main__.py
@@ -30,8 +30,7 @@ with open(opsets_path.joinpath("__init__.py"), "w", encoding="utf-8"):
 
 builder = OpsetsBuilder(".".join(module_base_names), MIN_REQUIRED_ONNX_OPSET_VERSION)
 paths = builder.write(repo_root)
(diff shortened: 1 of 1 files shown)


justinchuby:
Thanks for pointing out the issue with SPACES! I will take a closer look.

Regarding this change, I am curious of your thoughts on bringing in lintrunner as a build dependency, presuming we want to have this code gen process as a build step? I think it would be a more robust process relying on only isort and black, given that lintrunner is not working well on Windows yet and it also takes longer to run.

If we are going to not check in the generated code, lintrunner will ignore those files so nothing will be done.

If we are going to keep the generated code checked in, I think we can simply exclude them from lintrunner so that (1) the build is more robust, with less dependencies (2) the build is faster.

I do see the value in running mypy etc. on the generated code to make sure the types are right. I think this can be done by a manual run in lintrunner there.

What do you think?

PR #216, opened 2022-11-29
Stub out a test harness for pygen

abock:


opgen/pygen-test.py
```diff
@@ -0,0 +1,62 @@
+# --------------------------------------------------------------------------
```
justinchuby, line 1: nit: file name `pygen_test.py`

justinchuby (approved):
thanks!

PR #223, opened 2022-11-30
feat(atenlib): generate all aten signatures

justinchuby:
* (to be filled)

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/linalg.py
+++ b/onnxscript/function_libs/torch_aten/ops/linalg.py
@@ -1,0 +1,246 @@
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""torch.ops.aten operators under the `linalg` module.
+
+- No inplace operators.
+- All functions should not have the script() decorator. This is because
+    we want to delay the compilation of the function.
+"""
+from __future__ import annotations
+
+from typing import Optional, Sequence
+
+from onnxscript import TensorType
+
+
+def aten_linalg_cholesky(self: TensorType, *, upper: bool = False) -> TensorType:
+    raise NotImplementedError()
+
+
+def aten_linalg_cholesky_ex(
+    self: TensorType, *, upper: bool = False, check_errors: bool = False
+) -> tuple[TensorType, TensorType]:
+    raise NotImplementedError()
+
+
+def aten_linalg_cond(self: TensorType, p: Optional[float] = None) -> TensorType:
+    raise NotImplementedError()
+
+
+def aten_linalg_cross(self: TensorType, other: TensorType, *, dim: int = -1) -> TensorType:
+    raise NotImplementedError()
+
+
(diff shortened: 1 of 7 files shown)


onnxscript/function_libs/torch_aten/ops/linalg.py
```diff
@@ -0,0 +1,246 @@
...
+    raise NotImplementedError()
+
+
+def aten_linalg_matrix_rank(
+    self: TensorType, tol: float, hermitian: bool = False
+) -> TensorType:
+    raise NotImplementedError()
+
+
+def aten_linalg_multi_dot(tensors: TensorType[...]) -> TensorType:
```
gramalingam, line 162: `TensorType[...]` does not likely do what you intended here. Is this supposed to be a variadic operator?

justinchuby: I was hoping to write “a tensor of any type and any rank”. What would be the right way to do it?

onnxscript/function_libs/torch_aten/ops/linalg.py
```diff
@@ -0,0 +1,246 @@
...
+def aten_linalg_householder_product(input: TensorType, tau: TensorType) -> TensorType:
+    raise NotImplementedError()
+
+
+def aten_linalg_inv(A: TensorType) -> TensorType:
+    raise NotImplementedError()
+
+
+def aten_linalg_inv_ex(
+    A: TensorType, *, check_errors: bool = False
```
gramalingam, line 71: Currently, onnxscript doesn't handle things like keyword-only-arguments. Should probably decide what to support (specifically, in terms of what's required, what's not, distinguishing ONNX value-parameters vs. attributes, etc.)

justinchuby: I think it’s fine to not support them, presuming the order of the attributes doesn’t matter and can be supplied by name?

gramalingam: Yes, that's the convention being followed currently: specify actual inputs positionally and attribute-values as name=value pairs.

fatcat-z:
Is this a one-time work? Or we will keep the generator script and re-run it once PyTorch/ONNX has an update?

justinchuby:
Error cases

1. tensor list
 
```python
def aten_unflatten_dense_tensors(flat: TensorType, tensors: TensorType[...]) -> TensorType:
    # unflatten_dense_tensors(Tensor flat, Tensor[] tensors) -> Tensor[]

    raise NotImplementedError()
```

fatcat-z:
> The generator is designed to be one time to get us started. Thoughts?

I'd like to keep the code for the generator in the repo, to see if we can update it to regenerate some functions caused by ONNX improvement.

onnxscript/function_libs/torch_aten/ops/linalg.py
justinchuby: This actually should be a list of tensors. I will fix

justinchuby:
@fatcat-z PTAL

fatcat-z (approved):
Thanks.

PR #226, opened 2022-12-01
Cleanup use of default opset in converter

gramalingam:
A script function should implicitly or explicitly indicate which ONNX opset version it uses.

This PR adds a search in the converter to identify any implicit use of an ONNX opset (indicated by a call to `someopset.OpName(...)`) first. The identified opset is used, for example, in translation of overloaded operators.

For a function that contains no such call (for example, one that uses only overloaded operators such as + and *), the script parameter should specify the default_opset to use. 

Cleanup all examples where a default needs to be specified.

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -183,7 +183,7 @@ class Converter:
     def default_opset(self):
         if self.default_opset_ is None:
             raise RuntimeError(
-                "Default_opset must be specified in script for functions "
+                "default_opset must be specified in script for functions "
                 "that do not contain any use of an ONNX opset."
             )
         return self.default_opset_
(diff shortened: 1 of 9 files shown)


onnxscript/converter.py
```diff
@@ -182,14 +182,10 @@ def __init__(
     @property
     def default_opset(self):
         if self.default_opset_ is None:
-            if self.current_fn is None:
-                raise RuntimeError("Unable to return a default opset, None was detected yet.")
-            warn(
-                f"No default opset was defined or detected in function "
-                f"{self.current_fn.name!r}, the converter uses opset 15."
+            raise RuntimeError(
+                "Default_opset must be specified in script for functions "
```
abock, line 186: `default_opset` (lowercase `D`)?

abock (approved):

PR #228, opened 2022-12-02
Fix TensorType to return new TensorTypes for shapes instead of TensorType instances

### LINKED ISSUE #219: bug: tensor types are not recognized as types
Seen in https://github.com/microsoft/onnx-script/pull/223

1. "INT64" expects no type arguments, but 1 given  [MYPY]

	```python
	def aten_ones(size: INT64[...]) -> TensorType:
	```

2. Unexpected "..."  [MYPY]

	```python
	def aten_row_stack(tensors: TensorType[...]) -> TensorType:
	```

3. ~`INT64` does not accept a default value~ #229

	image

abock:
This ensures that `TensorType` is never actually instantiated and instead new `TensorType` _subclasses_ are created (and cached) for each shape.

This is a stepping stone to a fully generic TensorType where the dtype and shape are modeled into the type itself, but it appears that won't be easily realized until we can depend on Python 3.11.

Importantly, this adds unit tests to ensure if we can move to a proper generic that behavior will remain the same.

Fixes #219

### DIFF
--- a/onnxscript/test/onnx_types_test.py
+++ b/onnxscript/test/onnx_types_test.py
@@ -6,6 +6,7 @@
 # mypy: disable-error-code=misc
 
 """Unit tests for the onnx_types module."""
+from __future__ import annotations
 
 import unittest
 
@@ -24,13 +25,13 @@ class TestOnnxTypes(unittest.TestCase):
             FLOAT[...]()
 
     @parameterized.expand(tensor_type_registry.items())
-    def test_type_properties(self, dtype: DType, tensor_type: TensorType):
+    def test_type_properties(self, dtype: DType, tensor_type: type[TensorType]):
         self.assertEqual(tensor_type.dtype, dtype)
         self.assertIsNone(tensor_type.shape)
-        self.assertEqual(tensor_type[...].shape, ...)
-        self.assertEqual(tensor_type[...].dtype, dtype)
-        self.assertEqual(tensor_type[1, 2, 3].shape, (1, 2, 3))
-        self.assertEqual(tensor_type[1, 2, 3].dtype, dtype)
+        self.assertEqual(tensor_type[...].shape, ...)  # type: ignore[index]
+        self.assertEqual(tensor_type[...].dtype, dtype)  # type: ignore[index]
+        self.assertEqual(tensor_type[1, 2, 3].shape, (1, 2, 3))  # type: ignore[index]
+        self.assertEqual(tensor_type[1, 2, 3].dtype, dtype)  # type: ignore[index]
 
     @parameterized.expand([(dtype,) for dtype in tensor_type_registry])
     def test_dtype_bound_to_subclass(self, dtype: DType):


justinchuby (approved):
LGTM🚀

onnxscript/onnx_types.py
```diff
@@ -2,20 +2,17 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+
 from __future__ import annotations
 
-from typing import Optional, Tuple, Union
+from abc import ABC
```
justinchuby, line 8: I would import abc and use abc.ABC

onnxscript/onnx_types.py
```diff
@@ -36,131 +33,144 @@ def check_shape(shape):
         check_dim(shape)
 
 
-class TensorType:
-    """ONNX Script representation of a tensor type."""
+tensor_type_registry: dict[DType, "TensorType"] = {}
+tensor_type_shape_cache: dict[DType, "TensorType"] = {}
```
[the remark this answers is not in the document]
abock: Good catch. I had moved these out of `TensorType` originally since I didn't want them accessible on the subclasses. In that context quoting is necessary.

onnxscript/onnx_types.py
abock: Any particular reason?

onnxscript/onnx_types.py
```diff
@@ -2,20 +2,17 @@
...
 import onnx.helper
 
-# Representations of ONNX types in ONNX Script.
-# Currently restricted to tensor types.
-# Example type annotations in ONNX Script.
-#    x : FLOAT (a scalar-tensor of rank 0)
-#    x : FLOAT[...] (a tensor of unknown rank)
-#    x : FLOAT['M', 'N'] (a tensor of rank 2 of unknown dimensions, with symbolic names)
-#    x : FLOAT[128, 1024] (a tensor of rank 2 of known dimensions)
+DType = int
```
gramalingam, line 14: nit: a comment clarifying that this is ONNX's TensorProto::DataType enum may be worth it.

onnxscript/onnx_types.py
```diff
@@ -36,131 +33,144 @@ def check_shape(shape):
...
+                )
+            tensor_type_registry[dtype] = cls
+        else:
             check_shape(shape)
 
     def __getitem__(self, shape: Optional[ShapeType]):
-        if self.shape is not None:
+        raise NotImplementedError("should not be reached")
+
+    def __class_getitem__(cls, shape: Optional[ShapeType]) -> TensorType:
```
gramalingam, line 87: A Python question: does Python overload `TensorType` to also represent its meta-type?

onnxscript/onnx_types.py
gramalingam: We can eventually use our own symbolic encoding of the element-type, I guess

abock: Actually I think we can alias that directly.

onnxscript/onnx_types.py
gramalingam: I may not be using the right terminology. This returns an instance of `type` class, not an instance of `TensorType` class. Wondering how that is represented in Python? I assume mypy doesn't complain, so trying to understand this

gramalingam: Shouldn't the return type be `type` or `Type[TensorType]` or something like that? `Type[TensorType]` doesn't seem quite right, since it returns a derived class, but I am just trying to clarify my question.

gramalingam (approved):
LGTM, thanks. One thing I am not entirely clear is how well other tools, like mypy, will be when dealing with onnxscript code with type annotations, but that's a different question, may be worth talking about in our meeting. E.g., would using something like `Annotated[TensorType, Dtype(dtype), Shape(shape)]` work better with other type-checkers and tools?

onnxscript/onnx_types.py
abock: Oh, got it. Yes, good catch. It should be annotated through `Type[]`. Basically new types are created (and cached for reuse) on demand as shapes are specified through the type.

abock:
As for tooling, I am very unimpressed with mypy in particular. There is an actual bug in mypy that is causing it to not see the `__class_getitem__` and then complain that the type is not generic (it is not) or subscriptable (it is).

Python 3.11 provides support for variadic generics and syntax sugar to promote Python literals to `Literal[T]`. We could make it fully generic today via `TypeVarTuple` and `Unpack` through `typing_extensions`, but the usage syntax is gross.

Today:

```python
def F(tensor: FLOAT[Literal[128], Literal["M"], Literal[1024]]): ...

def G(tensor: FLOAT[Literal[Ellipsis]]): ...
```

With 3.11:
```python
def F(tensor: FLOAT[128, "M", 1024]): ...

def G(tensor: FLOAT[...]): ...
```

abock:
Hold on merging this. Trying to figure out what stopped working on 3.7 and 3.8.

onnxscript/onnx_types.py
justinchuby: Mostly because of this https://github.com/microsoft/onnx-script/wiki/Coding-style#avoid-relative-imports-import-only-modules, as suggested by https://google.github.io/styleguide/pyguide.html#23-packages.

`ABC` is usually unique enough, but I think maintaining a single rule of importing only modules can save us from deciding what to import as modules and what not to.

PR #231, opened 2022-12-05
Patch for tensor types

justinchuby:
Now mypy can correctly check types:

image

It still doesn't like that that they are not generics, but I think we can disable `type-arg` and `name-defined` for now.

```
  Error (MYPY) type-arg
    "FLOAT" expects no type arguments, but 1 given

        28  |        self.assertSameGraph(static_shape, static_shape_txt)
        29  |
        30  |        @script()
    >>> 31  |        def symbolic_shape(A: FLOAT["N"], B: FLOAT["N"]) -> FLOAT["N"]:  # noqa: F821
        32  |            C = op.Add(A, B)
        33  |            return C
        34  |

  Error (MYPY) name-defined
    Name "N" is not defined

        28  |        self.assertSameGraph(static_shape, static_shape_txt)
        29  |
        30  |        @script()
    >>> 31  |        def symbolic_shape(A: FLOAT["N"], B: FLOAT["N"]) -> FLOAT["N"]:  # noqa: F821
        32  |            C = op.Add(A, B)
        33  |            return C
        34  |
```

### DIFF
--- a/onnxscript/irbuilder.py
+++ b/onnxscript/irbuilder.py
@@ -62,7 +62,7 @@ class TensorType(Type):
 
 
 class Var:
-    def __init__(self, varname, typeinfo, info) -> None:
+    def __init__(self, varname, typeinfo: ONNXType, info) -> None:
         if not isinstance(varname, str):
             raise ValueError(f"varname must be a string not {type(varname)!r}.")
         self.name = varname
--- a/onnxscript/onnx_types.py
+++ b/onnxscript/onnx_types.py
@@ -5,7 +5,6 @@
 
 from __future__ import annotations
 
-from abc import ABC
 from typing import ClassVar, Optional, Tuple, Union
 
 import onnx
@@ -16,7 +15,7 @@ DType = onnx.TensorProto.DataType
 DimType = Union[int, str, type(None)]
 
 
-def check_dim(dim):
+def _check_dim(dim):
     if not isinstance(dim, (int, str, type(None))):
         raise TypeError(f"Invalid dimension {dim}")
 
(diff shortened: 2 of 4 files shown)


justinchuby:
I don't think this works because FLOAT is now an instance of TensorType, not a subclass.

justinchuby:
Will wait for

- https://github.com/python/mypy/issues/11501

justinchuby:
Alternatively, we will use this and run mypy with python3.11. So mypy is happy and we can pass in literals as types even with python <3.11.

Will need to fix dtype and shape.

```python
from typing import Any, Generic, TypeVar, Union, Tuple, Literal
import typing
from typing_extensions import TypeVarTuple, Unpack
DimType = Union[int, str, type(None)]
ShapeType = Union[Tuple[DimType, ...], DimType, type(Ellipsis)]
TDType = TypeVar("TDType")
TShape = TypeVarTuple("TShape")
class _DTYPE_FLOAT:
    dtype = 100
class TensorType(Generic[TDType, Unpack[TShape]]):
    def __class_getitem__(cls, *items: Any):
        wrapped = []
        for item in items[0]:
            if type(item) in {int, str, type(...), type(None)}:
                wrapped.append(Literal[item])
            else:
                wrapped.append(item)
        return super().__class_getitem__(
            tuple(wrapped)
        )
    def dtype(self) -> TDType:
        return typing.get_args(self.__orig_class__)[0]
    def shape(self) -> TShape:
        return typing.get_args(self.__orig_class__)[1:]
class FLOAT(TensorType[_DTYPE_FLOAT, Unpack[TShape]]):
    pass
print(TensorType[int, 1, 2])
print(TensorType[int, 1, 2]().dtype())
print(TensorType[int, 1, 2]().shape())
M = TypeVar("M", bound=int)
print(FLOAT[1, 2, ..., "N", M])
```

PR #232, opened 2022-12-05
IR cleanup (part 1)

gramalingam:
Cleanup the IR and IR builder: add type annotations, rename.


onnxscript/irbuilder.py
```diff
@@ -344,35 +353,25 @@ def to_graph_proto(self, use_default_type: bool = True):
...
 
-            Default values for attributes are introduced in onnx==1.13.0.
-            If an earlier version of onnx is installed, it ignores the default
-            values of the function arguments.
+        Note: Default values for attributes are an experimental feature in ONNX.
+        Conversion ignores default values for attributes if the ONNX version installed
+        doesn't support it.
         """
         opsets = self.get_opset_import()
         if domain != "":
```
justinchuby, line 377: domian can be a type other string right

justinchuby (approved):

onnxscript/irbuilder.py
gramalingam: I don't understand the question. The type is string. (But I don't know why we have this parameter, I think we should eventually eliminate it. I will look into it.)

gramalingam: Oh, you are referring to the next line, I see now. Anyway, let me see if I can delete this parameter.

onnxscript/irbuilder.py
```diff
@@ -328,17 +329,17 @@ def to_graph_and_functions(
         Returns:
             a pair of a :class:`onnx.GraphProto` and list of :class:`onnx.FunctionProto`
         """
-        sub_functions = {}
+        called_functions: Dict[str, onnx.FunctionProto] = {}
```
justinchuby, line 332:
```suggestion
        called_functions: dict[str, onnx.FunctionProto] = {}
```

nit

onnxscript/type_annotation.py
```diff
@@ -62,9 +69,41 @@ def is_attr(pytype: type):
...
 def is_valid(typeinfo):
     return is_attr(typeinfo) or is_tensor(typeinfo)
+
+
+def validate(typeinfo):
+    if not is_valid(typeinfo):
+        raise ValueError(f"Unsupported type annotation {typeinfo}")
+
+
+def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[type]:
```
justinchuby, line 91:
```suggestion
def get_return_types(typeinfo: type | Sequence[type]) -> Sequence[type]:
```

nit

gramalingam: Currently importing only `typing` in this file (due to 3.8 and 3.7 version differences for get_origin/get_args).

PR #234, opened 2022-12-05
fix: remove default_opset annotation

justinchuby:
Previous Opset14 was incorrect.

### DIFF
--- a/onnxscript/onnx_opset/__init__.py
+++ b/onnxscript/onnx_opset/__init__.py
@@ -9,6 +9,9 @@
 # mypy: disable-error-code=override
 # pylint: disable=W0221,W0222,W0237,W0246,R0901
 # --------------------------------------------------------------------------
+from __future__ import annotations
+
+from typing import Any
 
 from onnx.defs import onnx_opset_version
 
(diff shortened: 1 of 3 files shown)


onnxscript/onnx_opset/__init__.py
```diff
@@ -183,7 +183,7 @@
         1,
     ): opset_ai_onnx_preview_training1,
 }
-default_opset: Opset14 = all_opsets[
+default_opset: Opset18 = all_opsets[
```
gramalingam, line 186: Do we need this type-annotation? I suppose intellisense depends on this. On the other hand, this seems incorrect, since the type really depends on the installed ONNX version.

Maybe we don't even need this variable?

justinchuby: Let me try that

gramalingam (approved):

justinchuby:
Removing the annotation is no good because intellisense will stop knowing what default_opset is

gramalingam:
> Removing the annotation is no good because intellisense will stop knowing what default_opset is

Yes, that's what I mentioned earlier! The real answer may be that users don't ever use it directly.

justinchuby:
Ahh ok sorry I misread. That makes sense to me. I will abort the pr for now

PR #237, opened 2022-12-07
Minor fixes to type annotation checks

gramalingam:
Minor fixes to the type annotation checks:

* use typing_extensions for get_origin, get_args
* use warning error message for invalid type annotation (for now; revisit design later; an exception terminating script seems too drastic; but simple warning message is harder to test, need a better mechanism.)

### DIFF
--- a/onnxscript/type_annotation.py
+++ b/onnxscript/type_annotation.py
@@ -4,35 +4,99 @@
 # --------------------------------------------------------------------------
 from __future__ import annotations
 
-from typing import List
+import typing
 
 import onnx
 
-pytype_to_attrtype_map = {
+from onnxscript.onnx_types import TensorType
+
+_pytype_to_attrtype_map = {
     float: onnx.AttributeProto.FLOAT,
     int: onnx.AttributeProto.INT,
     str: onnx.AttributeProto.STRING,
-    List[  # pylint: disable=unhashable-member # TODO: Need change
-        int
-    ]: onnx.AttributeProto.INTS,
 }
 
+_listtype_to_attrtype_map = {
+    float: onnx.AttributeProto.FLOATS,
+    int: onnx.AttributeProto.INTS,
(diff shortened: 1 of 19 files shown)


onnxscript/type_annotation.py
```diff
@@ -25,60 +27,35 @@
...
-    elif hasattr(t, "__args__"):
-        return t.__args__  # type: ignore[no-any-return]
-    else:
-        raise ValueError(f"Unsupported type annotation {t}")
-
-
 def _get_element_type(t: type) -> type:
     """Returns the element type for a list or sequence type."""
-    return _get_args(t)[0]
+    return get_args(t)[0]
```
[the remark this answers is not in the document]
justinchuby: We should remove warn_return_any = true from pyproject.toml. It doesn't look like it's doing much good

onnxscript/type_annotation.py
```diff
@@ -25,60 +27,31 @@
...
-    """Returns the element type for a list or sequence type."""
-    return _get_args(t)[0]
-
-
 def pytype_to_attrtype(pytype: type) -> typing.Optional[onnx.AttributeProto.AttributeType]:
     if pytype in _pytype_to_attrtype_map:
         return _pytype_to_attrtype_map[pytype]
-    if _get_origin(pytype) in _list_constructors:
-        elt_type = _get_element_type(pytype)
+    if get_origin(pytype) in _list_constructors:
```
justinchuby, line 33: Can `_list_constructors` be a frozenset?

gramalingam: Not sure. It was a plain set originally, but there was warning messages about the elements being not hashable or something like that. I changed it to a list, since it shouldn't matter much for a 3-element collection. +1 for the naming convention, will do.

justinchuby: I see. thanks for the context

Issue #242, opened 2022-12-08
onnxscript produced graph fails onnx graph checker


justinchuby:
```
Traceback (most recent call last):
  File "/home/justinchu/dev/onnx-script/onnxscript/poc/os_graph_builder.py", line 145, in <module>
    onnx_model = gb.make_model(model_name)
  File "/home/justinchu/dev/onnx-script/onnxscript/poc/os_graph_builder.py", line 77, in make_model
    checker.check_model(self.onnx_model)
  File "/home/justinchu/anaconda3/envs/onnx/lib/python3.10/site-packages/onnx/checker.py", line 106, in check_model
    C.check_model(protobuf_string)
onnx.onnx_cpp2py_export.checker.ValidationError: Graph must be in single static assignment (SSA) form, however 'other' has been used as output names multiple times.
```

From script

```python
@script()
def aten_add(self, other, alpha: float = 1) -> TensorType:
    # add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
    if alpha != 1:
        other = op.Mul(other, alpha)  # type: ignore[arg-type]
    return op.Add(self, other)
```

cc @gramalingam

PR #243, opened 2022-12-08
poc: Use the functionlib

justinchuby:
image

The graph is incorrect since alpha's default value (`1.0`) was not translated into the proto (`0` below). The graph is also overly complex and require further simplification. Should we simplify this in an optimization pass?

Aten Sub function:

image

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -63,8 +63,10 @@ def aten_adaptive_max_pool1d(
 def aten_add(self, other, alpha: float = 1) -> TensorType:
     # add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
     if alpha != 1:
-        other = op.Mul(other, alpha)  # type: ignore[arg-type]
-    return op.Add(self, other)
+        other_scaled = op.Mul(other, alpha)  # type: ignore[arg-type]
+    else:
+        other_scaled = other
+    return op.Add(self, other_scaled)
 
 
 def aten_addbmm(
@@ -4353,9 +4355,11 @@ def aten_sub(self, other, alpha: float = 1) -> TensorType:
     # sub.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
 
     if alpha != 1:
-        other = op.Mul(other, alpha)  # type: ignore[arg-type]
+        other_scaled = op.Mul(other, alpha)  # type: ignore[arg-type]
+    else:
+        other_scaled = other
 
-    return op.Sub(self, other)
+    return op.Sub(self, other_scaled)
 
 
 def aten_subtract(self: TensorType, other: TensorType, alpha: float = 1) -> TensorType:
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -63,8 +63,10 @@ def aten_adaptive_max_pool1d(
 def aten_add(self, other, alpha: float = 1) -> TensorType:
     # add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
     if alpha != 1:
-        other = op.Mul(other, alpha)  # type: ignore[arg-type]
-    return op.Add(self, other)
+        other_scaled = op.Mul(other, alpha)  # type: ignore[arg-type]
+    else:
+        other_scaled = other
```
fatcat-z, line 68: Such logic is needless. This will cause a different behavior than the PyTorch aten op.
In PyTorch torch.add() method, if the alpha is 0, then the returned value is just self.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4353,9 +4355,11 @@ def aten_sub(self, other, alpha: float = 1) -> TensorType:
     # sub.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
 
     if alpha != 1:
-        other = op.Mul(other, alpha)  # type: ignore[arg-type]
+        other_scaled = op.Mul(other, alpha)  # type: ignore[arg-type]
+    else:
+        other_scaled = other
```
fatcat-z, line 4360: Same comment as above.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: More importantly, I wonder how much we want alpha to appear here, as in most cases it’s just 1, but the generated function will always have a conditional inserted. Will that hurt performance? Do we need a mechanism to specialize the functions in the optimization phase? Can we leave this to optimization, or should we dispatch to a specialized version already during conversion?

gramalingam: Conditionals that depend only on attribute-values can always be optimized away. We will need such optimizations in onnxruntime, and we do plan to add it in onnxruntime / onnx (extending the existing constant-folding optimization). In short, we need it, but it would be better if we don't duplicate this effort.

fatcat-z: > onnx complained that the previous logic needs to be SSA. I think this is a gap be need to bridge #242 or find a better solution for

Per my understanding, to resolve SSA issue, we just need to keep the line 66 as the latest one. But the logic of line 67-68 doesn't have a relationship to SSA issue, right?

gramalingam: I don't understand the SSA issue. While ONNX does not allow multiple assignments to the same variable, the onnx script converter is supposed to take care of this and rename variables as needed. If there is a problem, there must be a bug in the converter. Eg., see https://github.com/microsoft/onnx-script/blob/main/onnxscript/test/if_test.py ...

justinchuby: Awesome. so it sounds like we can keep the functions as complete as possible and leave the simplification to optimizers. As long as the optimization doesn't only exist in onnxruntime.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Resolved to focus on the discussion above.

onnxscript/function_libs/torch_aten/ops/core.py
gramalingam: https://github.com/microsoft/onnx-script/pull/244 should fix the SSA issue, hopefully.

PR #244, opened 2022-12-09
Add input parameter names to used vars

gramalingam:
The unique-name generation mechanism does not take input parameters into account. Fix this bug.

Fixes https://github.com/microsoft/onnx-script/issues/242


justinchuby (approved):
Thanks a lot!

PR #245, opened 2022-12-09
fix: annotate script()

justinchuby:
* (to be filled)

To allow mypy to analyze typing for annotated functions. Otherwise it complains that "Untyped decorator makes function "ones_like" untyped  [misc]"

### DIFF
--- a/onnxscript/main.py
+++ b/onnxscript/main.py
@@ -9,6 +9,7 @@ import ast
 import inspect
 import sys
 import textwrap
+from typing import Any, Callable, Optional
 
 import onnx.helper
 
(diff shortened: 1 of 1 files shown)


onnxscript/main.py
```diff
@@ -52,7 +53,9 @@ def script_check(f: ast.FunctionDef, opset, global_names, source, default_opset=
     return convert.top_level_stmt(f)
 
 
-def script(opset=None, default_opset=None, **kwargs):
+def script(
+    opset: Optional[values.Opset] = None, default_opset: Optional[Any] = None, **kwargs: Any
```
gramalingam, line 57: I think `Optional[values.Opset]` for default_opset too

PR #247, opened 2022-12-09
feat(atenlib): implement aten functions 1/n

justinchuby:
* __->__ #247

Implemented ops

- lt
- gt
- round
- clamp_min
- clamp_max
- clamp
- repeat
- ones_like

Test:
- Create the ability to skip particular sub tests
- Create a helper function to transform torch inputs into numpy for onnxscript to run on


onnxscript/function_libs/torch_aten/ops/common.py
```diff
@@ -0,0 +1,13 @@
...
+
+import onnx.helper
+
+from onnxscript.onnx_opset import opset18 as op
+
+
+def ones_like(x, onnx_dtype: int):
+    shape = op.Shape(x)
+    return op.ConstantOfShape(
+        shape, value=onnx.helper.make_tensor("one", onnx_dtype, [1], [1])
```
gramalingam, line 12: One issue here is that ONNX attributes should be compile-time (that is, script-time) constants, and, hence, cannot depend on a parameter like onnx_dtype. Is the idea that output type should be the same as input type `x`? Or, can it be a different dtype?

Anyway, I would suggest using either `CastLike` or `Cast`, depending on what exactly you want.

gramalingam: The only thing we can do with an attribute-parameter in ONNX is to forward it as the attribute-value of another call.

gramalingam: Here's one option:

```py
def ones_like(x, dtype:int):
    shape = op.Shape(x)
    one = op.Constant(value_float=1.0)
    one_dtype = op.Cast(one, to=dtype)
    return op.Expand(one_dtype, shape)
```
I wrote it out in full to focus on the ONNX representation. In onnxscript, we can abbreviate it a little bit, as below:
```py
def ones_like(x, dtype:int):
    shape = op.Shape(x)
    one_dtype = op.Cast(1, to=dtype)  # Forwarding attribute-parameter as attribute-value is ok.
    return op.Expand(one_dtype, shape)
```

gramalingam: The constraints come from various ONNX ops.

gramalingam: The other point I would like to understand is about the intended usage. In an exporter, for example: you could choose to
* add multiple ONNX ops (a subgraph) to the graph being constructed (representing a single pytorch op)
* add a single call to an ONNX function (op),

They are slightly different (for example, the first approach is more general). The first can be done using an approach we have discussed previously, something like below:
```py
def specialized_ones_like(dtype: int):
   # This is outside the script, so okay:
   one_dtype = onnx.helper.make_tensor("one", dtype, [1], [1])
   @script()
   def f(x):
      shape = op.Shape(x)
      return op.ConstantOfShape(shape, value=one_dtype)
   return f.to_function_proto().node # Depending on intended usage
```

justinchuby: Thanks for the example! I love how clean it is. I am probably going to think more about the specialized version before using it requires the onnx function to be defined in a function scope to be able to close over dtype.

Question: `f.to_function_proto().node` looks cool which I didn't know about. What is the `node` we get here?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3916,10 +3942,19 @@ def aten_renorm(self: TensorType, p: float, dim: int, maxnorm: float) -> TensorT
     raise NotImplementedError()
 
 
-def aten_repeat(self: TensorType, repeats: INT64) -> TensorType:
+def aten_repeat(self, repeats: INT64["M"]):
     # repeat(Tensor self, SymInt[] repeats) -> Tensor
 
-    raise NotImplementedError()
+    # FIXME(justinchuby): When repeats.shape == [0]
```
justinchuby, line 3948: This is an example where we need shape dependent logic.

justinchuby:
👀 something is wrong on Windows

image

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -747,16 +749,31 @@ def aten_clamp(
...
 
-    raise NotImplementedError()
+    max_ = op.CastLike(max_, self)
+    return op.Clip(self, None, max_)
+
+
+def aten_clamp_max_tensor(self, max_):
+    # clamp_max(Tensor self, Scalar max) -> Tensor
+
+    return op.Min(self, max_)
```
fatcat-z, line 762: op.Max()?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -747,16 +749,31 @@ def aten_clamp(
     raise NotImplementedError()
 
 
-def aten_clamp_max(self: TensorType, max: float) -> TensorType:
+def aten_clamp_max_scalar(self, max_):
```
fatcat-z, line 752: Is it possible to validate the shape of max_ to know it is a scalar or tensor instead of having such information in the function name?

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -250,7 +377,7 @@ def test_output_match(self, device: str, dtype: torch.dtype, op):
                 # Use torch testing to ensure dtypes and shapes match
                 torch.testing.assert_close(
                     torch.tensor(function_output),
-                    torch_output,
+                    output_torch,
```
fatcat-z, line 380: Nit: I like torch_output more.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -186,14 +200,123 @@ def wrapped(fn):
...
+        compiled = onnxscript.script()(function)
+        compiled.to_function_proto()
+
+
+def _convert_tensor_to_numpy(input: Any) -> Any:
+    if isinstance(input, torch.Tensor):
+        return input.detach().cpu().numpy()
+    if isinstance(input, (tuple, list)):
+        if len(input) == 0:
+            return np.array((), dtype=np.int64)
```
fatcat-z, line 281: Why is not the default value float32?

onnxscript/function_libs/torch_aten/ops/common.py
justinchuby: Done

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
justinchuby: I found that it’s usually an index tuple so it needs to be int. If there are more subtle cases we can update the logic to handle those

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: I haven't thought of a good way. Any suggestions?

justinchuby: Ah, for validation, we can specify it in the signature / via some data structure. We can discuss today.

justinchuby: They still need to be two functions though.

fatcat-z (approved):
LGTM, thanks!

PR #252, opened 2022-12-13
feat(atenlib): ops 2/n

justinchuby:
* #256
* #255
* __->__ #252


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4461,7 +4459,13 @@ def aten_symeig(
 def aten_t(self: TensorType) -> TensorType:
     # t(Tensor(a) self) -> Tensor(a)
 
-    raise NotImplementedError()
+    # TODO(justinchuby): Make rank a function
+    rank = op.Shape(op.Shape(self))  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
justinchuby: Good call! I didn’t know there’s Size. Will update

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -407,12 +407,18 @@ def aten_leaky_relu_backward(
...
 
 
-def aten_linear(
-    input: TensorType, weight: TensorType, bias: Optional[TensorType] = None
-) -> TensorType:
+def aten_linear(input, weight, bias=None) -> TensorType:
     # linear(Tensor input, Tensor weight, Tensor? bias=None) -> Tensor
 
-    raise NotImplementedError()
+    # FIXME(justinchuby): Enable the test
```
fatcat-z, line 413: In current PyTorch exporter, if the rank of input is 2 and the node of bias is not None, it will return a result of addmm(). Code is [here](https://github.com/pytorch/pytorch/blob/dcb73aa291816d0ee2d2e21b6b8a966da97945ad/torch/onnx/symbolic_opset9.py#L6139).

Why do we remove that logic here?

fatcat-z: For the optimization purpose, then I think we should add some comments here to mark this point so that we won't forget it when we start the optimization work.

Even these functions were named with an aten_ prefix, they are still ONNXScript functions which are essentially same to those op.Functions. I think we don't need to avoid calling each other within this aten functions lib. Do you have any examples/ideas to describe the cons?

justinchuby: Since then have ATen signatures, we have seen many cases where additional params need to be provided to make the call correct. Conceptually, the ATen functions interface with ATen ops (external facing), and so calling them internally also overloads the responsibility, which I hope to maintain a clear conceptual model of.

This also helps with the representation of the exported graph. When users see an ATen function, they will know there is a corresponding op in their model.

onnxscript/function_libs/torch_aten/ops/core.py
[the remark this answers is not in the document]
justinchuby: Done

fatcat-z (approved):
LGTM, thanks!

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: Added a comment

PR #255, opened 2022-12-14
feat(atenlib): ops 3/n

justinchuby:
* #256
* __->__ #255
* #252

A bunch of matmul and math ops


onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -187,6 +192,32 @@ def wrapped(fn):
...
+        "acosh",
+        dtypes=BOOL_TYPES + INT_TYPES,
+        reason="Acosh is not defined on bool or int tensors",
+    ),
+    xfail(
+        "addmm",
+        dtypes=[torch.uint8, torch.int8, torch.int16],
+        reason="MatMul is not defined on int16/int8/uint8 tensors",
+    ),
+    xfail(
```
fatcat-z, line 210: What's the different of 2 "addmm" here?

justinchuby: There is a “variant” param which represents a different test. OpInfos are identified by name and variant. The variant is usually empty. When there is a variant OpInfo we need to specify it so that they match.

fatcat-z (approved):
LGTM, thanks!

PR #259, opened 2022-12-15
prototype: atenlib typing and overload registration

justinchuby:
TODO: Extract typing information to a metadata data structure

Reference: #257

### DIFF
--- a/onnxscript/function_libs/torch_aten/registration.py
+++ b/onnxscript/function_libs/torch_aten/registration.py
@@ -1,0 +1,70 @@
+"""Registry for aten functions."""
+
+from __future__ import annotations
+
+from typing import Any, Optional
+
+
+class OverloadedFunction:
+    """Overloaded function."""
+
+    def __init__(self, name: str):
+        self.name = name
+        self.default: Optional[Any] = None
+        self.overloads: list[Any] = []
+
+
+class Registry:
+    """Registry for aten functions."""
+
+    def __init__(self):
+        self._registry: dict[str, OverloadedFunction] = {}
+
+    def register(self, name: str, overload: bool = False):
+        """Register a function."""
+
+        def wrapper(func):
+            if overload:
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_aten/registration.py
```diff
@@ -0,0 +1,70 @@
...
+class OverloadedFunction:
+    """Overloaded function."""
+
+    def __init__(self, name: str):
+        self.name = name
+        self.default: Optional[Any] = None
+        self.overloads: list[Any] = []
+
+
+class Registry:
```
justinchuby, line 17: Code gen should then read the registry to generate individual symbolic functions that dispatches to onnx functions & adds glue nodes. So users don't need to interact with this registry, which tends to be opaque.

onnxscript/function_libs/torch_aten/registration.py
```diff
@@ -0,0 +1,70 @@
...
+    """Registry for aten functions."""
+
+    def __init__(self):
+        self._registry: dict[str, OverloadedFunction] = {}
+
+    def register(self, name: str, overload: bool = False):
+        """Register a function."""
+
+        def wrapper(func):
```
justinchuby, line 26: TODO: Parse and validate the function signature.

onnxscript/function_libs/torch_aten/registration.py
```diff
@@ -0,0 +1,70 @@
...
+        self.overloads: list[Any] = []
+
+
+class Registry:
+    """Registry for aten functions."""
+
+    def __init__(self):
+        self._registry: dict[str, OverloadedFunction] = {}
+
+    def register(self, name: str, overload: bool = False):
```
justinchuby, line 23:
```suggestion
    def register(self, name: str | tuple[str], overload: bool = False):
```

PR #260, opened 2022-12-16
feat(atenlib): atenlib function registry

justinchuby:
* __->__ #260

Registry for all atenlib functions, which supports overloading. This change implements the `torch_op` decorator for registering all functions in the ATen lib so they are discoverable. It is not designed to be used by users, but rather as an info container so code gen knows which ops are implemented.

`torch_op` also compiles the functions into OnnxFunction.


fatcat-z (approved):
LGTM, thanks!

PR #261, opened 2022-12-16
feat(atenlib): use typevars for functions

justinchuby:
* __->__ #261

This change implements

```python
TFloat = TypeVar("TFloat", bound=_FloatType)
TInt = TypeVar("TInt", bound=_IntType)
TReal = TypeVar("TReal", bound=_RealType)
```

and uses them to annotate the functions. This information will be used to generate symbolic functions which adds glue nodes and dispatches to the ONNX functions.

In order for type checking to work, the generated ops need to use `TypeVar`s as well.

- Also fix mypy ignore directives

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -2,11 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
-# mypy: disable-error-code=misc
-# mypy: disable-error-code=arg-type
-# mypy: disable-error-code=type-arg
-# mypy: disable-error-code=valid-type
-# mypy: disable-error-code=assignment
+# mypy: disable-error-code="misc,arg-type,type-arg,valid-type,assignment,return-value"
 """torch.ops.aten operators under the `core` module.
 
 - No inplace operators.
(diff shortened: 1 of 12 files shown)


justinchuby:
@abock I think onnxscript is not complaining about the typing annotations now so we are good (hopefully?).

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -19,26 +15,27 @@
...
-def aten_abs(self):
+def aten_abs(self: TReal) -> TReal:
     # abs(Tensor self) -> Tensor
 
     return op.Abs(self)
 
 
 @torch_op("aten::acos")
-def aten_acos(self):
+def aten_acos(self: TFloat) -> TFloat:
```
fatcat-z, line 31: In current exporter implementation, actually even the input is tensor(bool) it still can work.
If we limit the input to TFloat like this, will it break export?

justinchuby: The function does not take bools because onnx Acos does not. The current plan is to use the generated symbolic functions to handle dtype conversion.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -416,7 +413,7 @@ def aten_leaky_relu_backward(
 
 
 @torch_op("aten::linear")
-def aten_linear(input, weight, bias=None) -> TensorType:
+def aten_linear(input: TFloat, weight: TFloat, bias: Optional[TFloat] = None) -> TensorType:
```
fatcat-z, line 416: The result type of both MatMul and Add belongs to TFloat, should we change the TensorType to TFloat?

fatcat-z (approved):
Approved

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: Good catch! Done.

PR #264, opened 2022-12-25
add empty/empty_like/log_softmax/cat function

xiaowuhu:
Signed-off-by: xiaowuhu <xiaowuhu@microsoft.com>
add ops: empty, empty_like, log_softmax, cat


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3408,10 +3413,19 @@
     raise NotImplementedError()
 
 
-def aten_new_empty(self: TensorType, size: INT64) -> TensorType:
+@torch_op("aten::new_empty")
+def aten_new_empty(self, size: INT64, dtype: int = -1) -> TensorType:
```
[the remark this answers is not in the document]
fatcat-z: Please update the signature here like other implemented ones.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3408,10 +3413,19 @@ def aten_negative(self: TensorType) -> TensorType:
...
+    """new_empty.
+
+    Note: dtype is an onnx enum. Users should convert torch dtype to onnx dtype
+    before calling this function.
+    """
     # new_empty(Tensor self, SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor
 
-    raise NotImplementedError()
+    vals = np.empty(shape=size).flatten()
```
fatcat-z, line 3425: All of operations within this function will be transferred to ONNX finally. I think it could not handle numpy functions. 

How about op.Flatten()?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3408,10 +3413,19 @@ def aten_negative(self: TensorType) -> TensorType:
...
+    empty_tensor = op.Constant(value=make_tensor("new_empty", TensorProto.FLOAT, size, vals))
```
fatcat-z, line 3426: The parameter vals needs to be a value instead of an ONNX data structure here, so probably we need to use [] instead.

justinchuby:
Thanks for the PR, just a few comments for your consideration to help you get started before I am back:

1. I would consider starting with canonical ops. Other ops should be deprioritized https://github.com/microsoft/onnx-script/issues/258 because we should be able to convert models to the canonical representation. We can fill in other ops as needed. 
2. Consider avoiding np and onnx make tensor. Most if not all operations can be done without them. 
3. Styling wise, (1) avoid abbreviations as much as possible (e.g.) `vals` and (2) maintain consistency on the return value. For now I suggest `result` until/unless a better name is agreed upon.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1509,11 +1511,28 @@ def aten_embedding_sparse_backward(
...
+"""
+@torch_op("aten::empty_like")
+def aten_empty_like(self, dtype: int = 7):
     # empty_like(Tensor self, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None, MemoryFormat? memory_format=None) -> Tensor
 
-    raise NotImplementedError()
+    # using RandomUniform value to simulate np.empty()
 
+    input_shape = op.Shape(self)
+    result = op.RandomUniform(dtype, high=1e30, low=-1e30, shape=input_shape)
```
fatcat-z, line 1533: Shape is an attribute so we could not use a result from op.Shape() to set it. Probably op.RandomUniformLike() is more suitable.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1509,11 +1511,28 @@ def aten_embedding_sparse_backward(
...
 
-def aten_empty_like(self: TensorType, memory_format: Optional[str] = None) -> TensorType:
+"""
+@torch_op("aten::empty")
+def aten_empty(size: INT64, dtype : int = 1) -> TensorType:
+    # empty(SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None, MemoryFormat? memory_format=None) -> Tensor
+
+    # using RandomUniform value to simulate np.empty()
+
+    result = op.RandomUniform(dtype, high=1e30, low=-1e30, shape=size)
```
fatcat-z, line 1521: size is an integer value, but the attribute of shape requires a list. So, we could not use it directly.

onnxscript/autocast.py
```diff
@@ -18,7 +18,11 @@ def cast_inputs(get_type_info, cast, opschema, *args):
     This is used by the converter in a static-mode, as well as by the eager-mode
     execution in a dynamic-mode.
     """
-    if opschema is not None:
+    # original code:
+    # if opschema is not None:
```
fatcat-z, line 22: Suggest removing line 21 and 22 since code history will record this and we don't need to add comments for it.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3502,10 +3514,13 @@ def aten_ones_like(self, dtype: int = -1):
     # ones_like(Tensor self, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None, MemoryFormat? memory_format=None) -> Tensor
 
     shape = op.Shape(self)
+    print(shape)
```
fatcat-z, line 3517: Let's remove such debug code here.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3502,10 +3514,13 @@ def aten_ones_like(self, dtype: int = -1):
...
+    print(type(shape))
```
fatcat-z, line 3518: Let's remove such debug code here.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3502,10 +3514,13 @@ def aten_ones_like(self, dtype: int = -1):
...
     if dtype == -1:
         one = op.CastLike(1, self)  # type: ignore[arg-type]
     else:
         one = op.Cast(1, to=dtype)  # type: ignore[arg-type]
+    print(shape)
```
fatcat-z, line 3523: Let's remove such debug code here.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1509,10 +1509,22 @@ def aten_embedding_sparse_backward(
     raise NotImplementedError()
 
 
-def aten_empty_like(self: TensorType, memory_format: Optional[str] = None) -> TensorType:
+@torch_op("aten::empty")
+def aten_empty(size, dtype: int = 1):
+    # empty(SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None, MemoryFormat? memory_format=None) -> Tensor
+
+    # using RandomUniform value to simulate np.empty()
+    result = op.RandomUniform(dtype, 1e10, -1e10, 0.5, size)
```
justinchuby, line 1517: Is RandomUniform needed, or can it just be a bunch of zeros?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4553,8 +4568,17 @@ def aten_tile(self: TensorType, dims: Sequence[int]) -> TensorType:
...
     # to_dense(Tensor self, ScalarType? dtype=None) -> Tensor
+    # list_values = [0] * self.dims[0] * self.dims[1]
+    # index_length = len(self.indices)
+    # for i in range(index_length):
+    #     p = self.indices[i]
+    #     v = self.values.float_data[i]
+    #     list_values[p] = v  # this doesn't work
+    # data = op.Constant(value=list_values)
+    # shape = op.Constant(value=self.dims)
+    # return op.Reshape(data, shape)
```
justinchuby, line 4581: We probably won't need this function ever. I would just remove the whole thing

onnxscript/values.py
```diff
@@ -207,7 +207,12 @@ def adapt(input: ExtendedModeValue) -> EagerModeValue:
         elif input is None:
             return None
         elif isinstance(input, list):
-            return [adapt(elt) for elt in input]
+            # orginal code:
+            # return [adapt(elt) for elt in input]
```
fatcat-z, line 211: IIUC, the purpose of this function is to make sure each element is a Tensor. If a user tends to keep the original input data type, this function is not supposed to be called then.

Once this function is call, then it has to make sure each element is a Tensor, not a value.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Since the function just need to produce _any_ tensor

gramalingam: I agree. zeros may be better than random. Even that looks a bit unsatisfactory, since presumably the returned tensor is subsequently updated imperatively to fill in the values. A better exporter might be able to eliminate the use of such ops with a better analysis of the context where the result of this op is used. But as a fallback, a tensor with zeros may be ok.

onnxscript/values.py
gramalingam: Agree. Can you clarify what is the use-case that is triggering this? ONNX inputs are supposed to: tensors, or lists of tensors, etc. ONNX doesn't support separate scalar types, scalar numbers are encoded as tensors of rank 0. What is the input list [2, 3] supposed to be in the ONNX encoding?

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: use Zeros() instead

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: use Zeros_like instead.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -390,6 +400,11 @@ def _logcumsumexp_input_wrangler(
         matcher=lambda sample: "scale_factor" in sample.kwargs,
         reason="fixme: the scale_factor tests",
     ),
+    skip(
+        "log_softmax",
+        matcher=lambda sample: isinstance(sample.args[0], int),
+        reason="input rank cannot be 0",
```
justinchuby, line 406: I would use an if condition to handle the rank=0 case so that we support the op fully

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -197,6 +197,13 @@ def _logcumsumexp_input_wrangler(
     return args, kwargs
 
 
+def _log_softmax_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    kwargs["dim"] = args.pop()
```
justinchuby, line 203: Why is this needed? Do we need dim as a positional argument then instead?

onnxscript/function_libs/torch_aten/ops/special.py
```diff
@@ -205,10 +206,12 @@ def aten_special_log_ndtr(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_special_log_softmax(self: TensorType, dim: int, dtype: int = -1) -> TensorType:
+@torch_op("aten::log_softmax")
+def aten_special_log_softmax(self: TTensor, dim: int, dtype: int = FLOAT.dtype) -> TTensor:
```
justinchuby, line 210: Just to confirm, does onnx logsoftmax support all tensor types?

xiaowuhu: good suggestion. changed to TFloatOrBFloat16

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: op.LogSoftmax doesn't support rank=0 case, the input test case is array(n),  converted to tensor(array(n)), not tensor(array([n])), so the op.LogSoftmax() will be failed in runtime.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: dim is input, but in op it is a attribute.

justinchuby (approved):

onnxscript/autocast.py
```diff
@@ -18,7 +18,7 @@ def cast_inputs(get_type_info, cast, opschema, *args):
     This is used by the converter in a static-mode, as well as by the eager-mode
     execution in a dynamic-mode.
     """
-    if opschema is not None:
+    if opschema is not None and len(opschema.inputs) > 0:
```
justinchuby, line 21: From discussion: revert this change

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -326,7 +337,9 @@ def _topk_input_wrangler(
 EXPECTED_SKIPS_OR_FAILS = (
     xfail("amax", reason="ONNX Runtime 1.13 does not support ReduceMax-18"),
     xfail("amin", reason="ONNX Runtime 1.13 does not support ReduceMin-18"),
-    skip("clamp", reason="enable when onnxscript supports optional inputs"),
+    skip("clamp", reason="Enable when onnxscript supports optional inputs"),
+    skip("empty", reason="Using zeros to simulate empty"),
```
justinchuby, line 341: Use xfail?

PR #267, opened 2023-01-04
Fix type annotation issue of Sequence and fix a list issue in eager mode.

fatcat-z:
1. While calling pytepe_to_attrtype() function, the annotation of typing.Sequence will be identified to collections.abc.Sequence by get_origin(). This will impact the further execution, so adding collections.abc.Sequence into the _LIST_CONSTRUCTORS.
2. While calling a function in eager mode, a list type of input should not be transformed to be a Tensor list. Update the code logic accordingly.

### DIFF
--- a/onnxscript/type_annotation.py
+++ b/onnxscript/type_annotation.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------
 from __future__ import annotations
 
+import collections
 import inspect
 import typing
 
@@ -24,7 +25,7 @@ _LISTTYPE_TO_ATTRTYPE_MAP = {
     str: onnx.AttributeProto.STRINGS,
 }
 
-_LIST_CONSTRUCTORS = [list, typing.List, typing.Sequence]
+_LIST_CONSTRUCTORS = [list, typing.List, typing.Sequence, collections.abc.Sequence]
 
 
 def pytype_to_attrtype(pytype: type) -> typing.Optional[onnx.AttributeProto.AttributeType]:
--- a/onnxscript/values.py
+++ b/onnxscript/values.py
@@ -207,7 +207,7 @@ def _adapt_to_eager_mode(inputs: ExtendedModeValue) -> EagerModeValue:
         elif input is None:
             return None
         elif isinstance(input, list):
-            return [adapt(elt) for elt in input]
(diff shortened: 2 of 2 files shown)


xiaowuhu (approved):

onnxscript/type_annotation.py
```diff
@@ -24,7 +25,7 @@
     str: onnx.AttributeProto.STRINGS,
 }
 
-_LIST_CONSTRUCTORS = [list, typing.List, typing.Sequence]
+_LIST_CONSTRUCTORS = [list, typing.List, typing.Sequence, collections.abc.Sequence]
```
justinchuby, line 28: Could you update this to a frozenset for faster lookup?

onnxscript/values.py
```diff
@@ -207,7 +207,7 @@ def adapt(input: ExtendedModeValue) -> EagerModeValue:
         elif input is None:
             return None
         elif isinstance(input, list):
-            return [adapt(elt) for elt in input]
+            return input
```
justinchuby, line 210: I am still hesitant about this. What happens for example when we call Concat with a list of numpy arrays?

fatcat-z: > ~I am still hesitant about this. What happens for example when we call ConcatSequence with a list of numpy arrays?~
> 
> nvm. Looks like it works. does this look good to you @gramalingam ?

I just found out the issue. Actually, this change is not necessary, we should reverse it back.

Below is the code which calls this function:

> def __call__(self, *args, **kwargs):
        """Implements an eager-mode execution of an onnxscript function."""
        new_args, has_array = _adapt_to_eager_mode(args)
        result = self.function(*new_args, **kwargs)

Ideally, only inputs will be passed into _adapt_to_eager_mode for this adaption, those attributes are not supposed to be passed into this method because attributes should be in kwargs.

The real problem is somehow, both inputs and attributes were saved into args hence they were passed into _adapt_to_eager_mode() together. We should remove those attributes from args so that they won't be passed into _adapt_to_eager_mode() method.

gramalingam: I agree with @fatcat-z : the issue may be higher up, in distinguishing inputs from attributes. Fixing that may be the right fix.

PR #274, opened 2023-01-05
feat(atenlib): index_select; trace_only ops

justinchuby:
Implement `aten_index_select`; enable tests for transpose.

This change also added the **`trace_only`** argument to mark functions as trace only to work around functions that cannot be compiled. The logic can then be tested, but the argument should be removed later when onnxscript extends support for the syntax.

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -4687,7 +4687,20 @@ def aten_transpose(self, dim0: int, dim1: int):
 
     # FIXME(justinchuby): onnxscript raises Unsupported expression type
     # Script the function when this is fixed
-    return op.Transpose(self, [dim0, dim1])
+    self_rank = op.Size(op.Shape(self))
+
+    if self_rank == 0:
+        result = self
+    else:
+        # Python code, change when onnxscript supports this
+        self_rank_val = self_rank.value  # type: ignore[attr-defined]
+        dims = list(range(self_rank_val))
+        dims[dim0], dims[dim1] = dims[dim1], dims[dim0]
+        # Python code ends
+
+        result = op.Transpose(self, perm=dims)
+
+    return result
 
 
 def aten_triangular_solve(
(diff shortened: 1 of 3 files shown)


justinchuby:
cc @bowenbao

BowenBao:
> We need to fine a way to call the function that have the attribute dim before the input index.

Sounds like a restriction posed from onnx script function. Is there something fundamentally blocking from mixing the order of inputs and attributes? cc @gramalingam

gramalingam:
> > We need to fine a way to call the function that have the attribute dim before the input index.
> 
> Sounds like a restriction posed from onnx script function. Is there something fundamentally blocking from mixing the order of inputs and attributes? cc @gramalingam

No, I don't think there is anything fundamentally blocking it. But we need to add the logic to onnxscript translator to process the parameters accordingly, so it is a missing feature.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2174,10 +2176,19 @@ def aten_index_reduce(
     raise NotImplementedError()
 
 
-def aten_index_select(self: TensorType, dim: int, index: TensorType) -> TensorType:
+@torch_op("aten::index_select")
+def aten_index_select(self: TTensor, dim: int, index: TInt) -> TTensor:
     # index_select(Tensor self, int dim, Tensor index) -> Tensor
 
-    raise NotImplementedError()
+    if op.Size(op.Shape(index)) == 0:
```
gramalingam, line 2183: Otherwise, is `index` expected to be 1-dimensional? Or, can it have any other rank?

justinchuby: Yes, it should be 1d I think.

justinchuby: should I do `op.Reshape(index, [-1])` instead?

gramalingam: Yes, if we expect it to be 0d or 1d, I think reshape, as you mention, is a better option.

justinchuby: Done. thanks!

fatcat-z (approved):
Approved.

PR #275, opened 2023-01-05
fix: forward fix #267

justinchuby:
Revert changes to values; updated list to a frozenset for faster lookup.


fatcat-z (approved):
LGTM, thanks!

PR #285, opened 2023-01-06
feat(atenlib): arange with overloads

justinchuby:
arange ops

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -240,10 +240,37 @@ def aten_any(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_arange(end: float) -> TensorType:
+@torch_op("aten::arange")
+def aten_arange(end: TReal, dtype: int = -1) -> TReal:
     # arange(Scalar end, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor
 
-    raise NotImplementedError()
+    if dtype != -1:
+        end = op.Cast(end, to=dtype)
+
+    return op.Range(0, end, 1)
+
+
+@torch_op("aten::arange", overload=True)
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -240,10 +240,37 @@ def aten_any(self: TensorType) -> TensorType:
...
 
-    raise NotImplementedError()
+    if dtype != -1:
+        end = op.Cast(end, to=dtype)
+
+    return op.Range(0, end, 1)
+
+
+@torch_op("aten::arange", overload=True)
+def aten_arange_start(start: TReal, end: TReal, dtype: int = -1) -> TReal:
```
fatcat-z, line 254: Since TReal has limited dtypes, what if the given dtype doesn't exist in the scope of TReal?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -240,10 +240,37 @@ def aten_any(self: TensorType) -> TensorType:
...
+    # arange.start(Scalar start, Scalar end, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor
+
+    if dtype != -1:
```
fatcat-z, line 257: Since TReal has limited dtypes, what if the given dtype doesn't exist in the scope of TReal?

Should we set a default value it to dtype in case the given value of dtype is not in the scope of TReal?

justinchuby: Good point. I am considering if I should cast after Range

fatcat-z: > Good point. I am considering if I should cast after Range

If we pass start/end/step with an unexpected dtype into op.Range, will it work successfully?

justinchuby: Updated, it will work if dtype is specified, otherwise they need to be casted before entering the function.

fatcat-z (approved):
LGTM, thanks!

PR #286, opened 2023-01-06
feat(atenlib): where, full

justinchuby:
Implement where, full, new_full, full_like


onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -169,6 +169,7 @@ def wrapped(fn):
     "exp": core_ops.aten_exp,
     "exp2": core_ops.aten_exp2,
     "fmod": core_ops.aten_fmod,
+    "full_like": core_ops.aten_full_like,
```
fatcat-z, line 172: Why don't we add aten::full here?

justinchuby: we don't have an OpInfo for full so... I think it is innocuous enough to not include it for now?

justinchuby: I added a todo inline to note.

xiaowuhu (approved):

PR #289, opened 2023-01-06
Bump numpy to 1.22.0

abock:


justinchuby (approved):

abock:
Looks like this bump may leave behind Python 3.7 support... let's hold on this.

requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 autopep8
 click
-numpy==1.21.5
+numpy==1.22.0
```
jcwchen, line 3:
```suggestion
numpy==1.22.0; python_version >= "3.8"
numpy==1.21.5; python_version < "3.8"
```
nit: The fastest way is removing Python 3.7 (EOL is 27 Jun 2023). Or, perhaps we can just make Python 3.7 still use numpy==1.21.5 for now.

abock:
Per discussion this morning we're going to drop support for 3.7 as it EOLs in June anyway, so I will update this PR to do that as well.

ZainRizvi:
Oh man, this had the very unfortunate side effect of breaking Onnx builds on PyTorch. (It uses this script to install Onnx, and still supports python 3.7)

https://github.com/pytorch/pytorch/actions/runs/3878248792/jobs/6614365795

justinchuby:
Should we pin a commit in PyTorch for now?

PR #290, opened 2023-01-06
feat(atenlib): expand registry to allow testing all supported dtypes

### LINKED ISSUE #417: Test functions on all supported dtypes
- [x] Construct OpSchema from functions
    #626 
- [x] Create logic that will match OpSchema
    #700

justinchuby:
Fix #417


justinchuby:
Use OpSchema instead

PR #292, opened 2023-01-08
fix(atenlib): Cast SymInt to INT64 to fix Windows CI

justinchuby:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2178,7 +2179,7 @@ def aten_index_reduce(
 
 # FIXME(#277): Script when attributes can come before inputs
 @torch_op("aten::index_select", trace_only=True)
-def aten_index_select(self: TTensor, dim: int, index: TInt) -> TTensor:
+def aten_index_select(self: TTensor, dim: int, index: IntType) -> TTensor:
```
fatcat-z, line 2182: Why can't we keep TInt here? Looks like they are the same to me.

If we are using Txxx in most of places, I'd like to use the same style.

justinchuby: Since mypy requires a TypeVar to appear at least twice in the signature of a generic function, we are forced to use a plain type here.

fatcat-z (approved):
Approved, thanks!

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Actually, we can probably get away with TInt here. Let me try again and replace as much with TInt.

abock (approved):

PR #293, opened 2023-01-09
Add permute and pow ops and fix attribute value issue.

### LINKED ISSUE #287: Inputs and attributes could not be separated by _adapt_to_eager_mode method.
Ideally when an aten_xxx() function was called in eager mode, we will process all positional inputs so the data types are expected, but will leave kwargs alone.

But currently both inputs and attributes will be passed into _adapt_to_eager_mode() method as args which caused we process those attributes unexpectedly. This will cause make_attribute() failed because the dtype of attribute is wrong.

fatcat-z:
1. Add 2 new ops: permute and pow.
2. Fix (partially) #287: convert the values of attributes from tensor back to numpy value when we call Op.

### DIFF
--- a/onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
@@ -254,7 +254,7 @@ OPINFO_FUNCTION_MAPPING: dict[
     "expand": core_ops.aten_expand,
     "erf": core_ops.aten_erf,
     "fmod": core_ops.aten_fmod,
-    # "full": (core_ops.aten_full, _full_input_wrangler),
+    "full": (core_ops.aten_full, _full_input_wrangler),
     "full_like": core_ops.aten_full_like,
     "gt": core_ops.aten_gt,
     "index_select": core_ops.aten_index_select,
--- a/onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
@@ -408,6 +408,16 @@ SKIP_SUBTESTS: tuple[DecorateMeta, ...] = (
         matcher=lambda sample: "scale_factor" in sample.kwargs,
         reason="fixme: the scale_factor tests",
     ),
+    skip(
+        "permute",
+        matcher=lambda sample: len(list(filter(lambda v : v < 0, sample.args[0]))) > 0,
+        reason="Negative value in perm is not supported",
+    ),
+    skip(
+        "permute",
+        matcher=lambda sample: len(sample.args[0]) == 0,
+        reason="Empty perm is not supported",
+    ),
 )
 
 duplicate_opinfo(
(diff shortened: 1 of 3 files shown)


gramalingam:
Hi, a quick question: if I understand right, the code change converts attribute values that are tensors into numpy arrays.

First, I don't think that would fix issue 287 (completely). That's not a problem, just want to clarify that the PR description is a bit misleading. I believe we will still need to do something extra to resolve 287 fully. Or, did I misunderstand?

Maybe this change fixes the issue for a particular example? Can you point me to the example where this is used? I don't see it in this PR's changes. Thanks.

fatcat-z:
> Maybe this change fixes the issue for a particular example? Can you point me to the example where this is used? I don't

In aten_permute() method, the value of dims will be changed to tensors which are not expected by onnx.helper.make_attribute() function.

In eager mode, when we call an aten_xxx() method, all of positional inputs of it will be changed to tensors by _adapt_to_eager_mode() method. But inside the function, we still need to keep some of inputs to be a Numpy value (in aten_permute, it is 'dims') so that we can pass it into ONNX op as an attribute.

This will only happen in eager mode because _adapt_to_eager_mode will only be called in such mode. Probably I need to update the #287 for a clarification?

gramalingam:
> In eager mode, when we call an aten_xxx() method, all of positional inputs of it will be changed to tensors by _adapt_to_eager_mode() method. But inside the function, we still need to keep some of inputs to be a Numpy value (in aten_permute, it is 'dims') so that we can pass it into ONNX op as an attribute.

Ok. My main question is whether this positional input of the aten_xxx() method is supposed to be an attribute (of aten_xxx) or an input (of aten_xxx)?

I understand the first case: the problem happens if attributes are passed in positionally instead of as a keyword-argument. We can fix this.

The second case, however, cannot be supported in ONNX. Specifically, in ONNX, there is no way to convert an input (which is not statically known) into an attribute. I just want to clarify this ... we should NOT be writing such functions.

gramalingam:
In the first case (where `dims` is an attribute of `aten_permute`), the problem is the existing assumption/restriction in onnxscript that inputs are passed in positionally, while attributes are passed in as keyword-arguments. In this case, I can see that the proposed change will work as a temporary fix, but the right long term solution will be to generalize the way onnxscript handles parameters to separate out inputs from attributes. I do plan to do this generalization, and the current proposed fix can be used as a temporary fix until this happens.

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -464,6 +464,12 @@ def _topk_input_wrangler(
         matcher=lambda sample: len(sample.args[0]) == 0,
         reason="Empty perm is not supported",
     ),
+    skip(
+        "slice",
+        # kwargs {dim, start, end, step} is empty, we cannot give the default value
+        matcher=lambda sample: len(sample.kwargs) == 0,
```
justinchuby, line 470: Could you help me understand why this is needed better?

fatcat-z: This was not introduced by this PR, it should be checked-in by another PR #304

PR #299, opened 2023-01-10
feat(atenlib): onnxscript API on torchscript

### LINKED ISSUE #273: Implement torchscript as a graph backend
To use onnxscript APIs for constructing graphs in pytorch, we would like to create a version that uses torchscript as the graph IR. To do this, we probably need

1. Implement an `Evaluator` that calls `g.op` to add ops and functions as nodes
2. Implement a `Tensor` that wraps around the `torch.Value`s

titaiwangms:
Fixes #273 

An evaluator using onnxscript API to build graph using torchscript (we will have our own graph to replace torchscript.)

works with https://github.com/pytorch/pytorch/pull/91930

1. Op object triggers graph_eval instead of eval to begin op insertion for the graph (Need to figure out how to keep both functionality if they need to stay on the same API.)
2. TorchscriptEvaluator serves as a manager of input/output graph. In this demo, both of them are torchscript, so this class seems simple (should be more complicated depending on ONNX graph builder API and format requirements).
3. tensor.py is not modified because onnxscript does not really use the graph value. That is, it will only have to be compatible with graph builder type only in the future implementation.


justinchuby:
Very nice! Do functions need special handling?

justinchuby:
cc @xiaowuhu @fatcat-z

titaiwangms:
> Very nice! Do functions need special handling?

In this DEMO, not really. Special handling depends on the future graph builder API. I guess the most effort would be on how to make a fake tensor from a given torch fake tensor (extract information from torch graph and create node in new graph). onnxscript API itself is more like an interface here, and not necessarily check or adapt the value format.

onnxscript/values.py
```diff
@@ -136,7 +153,9 @@ def adapt_kwargs(self, kwargs):
     def __call__(self, *args, **kwargs):
         from onnxscript import evaluator  # pylint: disable=import-outside-toplevel
 
-        return evaluator.eval(self.opschema, args, kwargs)
+        # Hack:
+        # TODO(titaiwang): how to keep both graph_eval and eval if they share the same API?
```
BowenBao, line 157: Related to https://github.com/microsoft/onnx-script/issues/297. But for now I guess you can context manage in exporter to change default evaluator via onnx-script api.

onnxscript/evaluator.py
```diff
@@ -12,6 +12,8 @@
 
 import numpy as np
 import onnx
+import torch
```
xadupre, line 15: It means onnxscript takes a dependency on torch. Maybe that would be worth adding this in a separate folder.

justinchuby: In the future this code will live in torch so I think we are good? In reverse, torch will take dependency on onnxscript.

justinchuby: breaking it out to make it clear

BowenBao: > In the future this code will live in torch

Agree it'd be clearer for all such code to live in a dedicated directory.

justinchuby: Done

justinchuby:
Replaced by https://github.com/microsoft/onnx-script/pull/336 to allow collaboration in the main repo. Closing this for now if you don't mind @titaiwangms

PR #304, opened 2023-01-10
add ops: slice

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4394,7 +4395,22 @@ def aten_slice(
 ) -> TensorType:
     # slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)
 
-    raise NotImplementedError()
+    return op.Slice(self, start, end, dim, step)  # type: ignore[arg-type]
```
justinchuby, line 4398:
```suggestion
    return op.Slice(self, start, end, dim, step)
```

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -387,6 +388,11 @@ def _logcumsumexp_input_wrangler(
         matcher=lambda sample: "scale_factor" in sample.kwargs,
         reason="fixme: the scale_factor tests",
     ),
+    skip(
+        "slice",
+        matcher=lambda sample: isinstance(sample.args[3], int),
```
justinchuby, line 393: I would reshape the inputs.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4409,16 +4409,37 @@ def aten_sinh(self: TFloat) -> TFloat:
...
     dim: int = 0,
     start: Optional[INT64] = None,
     end: Optional[INT64] = None,
     step: INT64 = 1,
-) -> TensorType:
+) -> TTensor:
     # slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)
 
-    raise NotImplementedError()
+    is_scalar = op.Size(op.Shape(start)) == 0
```
fatcat-z, line 4422: Since start and end are Optional, we should leverage OptionalHasElement/OptionalGetElement to get their values.

xiaowuhu: add TODO to implementation later due to HasElement() cannot work now.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4409,16 +4409,38 @@ def aten_sinh(self: TFloat) -> TFloat:
...
     # slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)
 
-    raise NotImplementedError()
+    # TODO: using OptionalHasElement() to check start/end value
+    is_scalar = op.Size(op.Shape(start)) == 0
+    if is_scalar:
+        start = op.Cast(start, to=INT64.dtype)
+        start = op.Unsqueeze(start, op.Constant(value_ints=[0]))
+
+    is_scalar = op.Size(op.Shape(end)) == 0
```
justinchuby, line 4428: I would use end_is_scalar to avoid redefining this value.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4409,16 +4409,38 @@ def aten_sinh(self: TFloat) -> TFloat:
...
     end: Optional[INT64] = None,
     step: INT64 = 1,
-) -> TensorType:
+) -> TTensor:
     # slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)
 
-    raise NotImplementedError()
+    # TODO: using OptionalHasElement() to check start/end value
+    is_scalar = op.Size(op.Shape(start)) == 0
+    if is_scalar:
```
justinchuby, line 4424: start_is_scalar

onnxscript/test/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -419,6 +420,12 @@ def _topk_input_wrangler(
         matcher=lambda sample: "scale_factor" in sample.kwargs,
         reason="fixme: the scale_factor tests",
     ),
+    skip(
+        "slice",
+        # kwargs {dim, start, end, step} is empty, we cannot give the default value
+        matcher=lambda sample: len(sample.kwargs) == 0,
+        reason="start and end must be 1-D array",
```
justinchuby, line 427: Adding to Reason: … onnx runtime 1.13 does not support optional input?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4409,16 +4409,38 @@ def aten_sinh(self: TFloat) -> TFloat:
...
+        start = op.Cast(start, to=INT64.dtype)
```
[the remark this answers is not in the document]
xiaowuhu: 1. torch test case doesn't have non-scalar values as input, so we need to have additional test case.
2. it is trade off: when the function need input is INT64, but the caller give INT32. who is responsible to cast INT32 to INT64? I'm not sure.

xiaowuhu:
> Actually, we should use Reshape and avoid conditioning here (expensive) as Rama suggested, since the inputs start, end, etc have always a single element, and we don’t need to squeeze them back: [#274 (comment)](https://github.com/microsoft/onnx-script/pull/274#discussion_r1062967679)

yes. I did want to use Reshape, but at that time I forgot how to set [-1] as op.Constant()  :(

justinchuby (approved):
LGTM🚀

PR #306, opened 2023-01-11
Add a new op embedding.

fatcat-z:
Add a new op embeding and its test.

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -1506,16 +1506,15 @@ def aten_einsum(
     raise NotImplementedError()
 
 
+@torch_op("aten::embedding")
 def aten_embedding(
-    weight: TensorType,
-    indices: TensorType,
-    padding_idx: int = -1,
-    scale_grad_by_freq: bool = False,
-    sparse: bool = False,
-) -> TensorType:
+    weight: TTensor,
+    indices: TTensor,
+    **kwargs,
+) -> TTensor:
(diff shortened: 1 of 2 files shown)


xiaowuhu (approved):

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1511,16 +1511,15 @@ def aten_einsum(
...
 def aten_embedding(
-    weight: TensorType,
-    indices: TensorType,
-    padding_idx: int = -1,
-    scale_grad_by_freq: bool = False,
-    sparse: bool = False,
-) -> TensorType:
+    weight: TTensor,
+    indices: TTensor,
+    **kwargs,# pylint: disable=unused-argument
```
justinchuby, line 1518:
```suggestion
    **_,
```

Don’t know if ** works in onnxscript, but if it does:

justinchuby: Do we need to change behavior based on padding_idx?

justinchuby:
@gramalingam so variadic keyword arguments work?

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: According to current implementation, it will only impact Training. I think ONNX Script won't know if current export is under Training mode, so the warning about possible wrong training result should be thrown by exporter, not onnx-script.

Thoughts?

justinchuby: I think we should also support any training behaviors when possible.

PR #307, opened 2023-01-11
Fix onnxfs2_test regression for onnx 1.13

### LINKED ISSUE #249: Fix tests for onnx 1.13
ONNX 1.13 produced some errors in tests: https://github.com/microsoft/onnx-script/actions/runs/3678076190/jobs/6220897226

```
E           onnx.onnx_cpp2py_export.checker.ValidationError: Field 'type' of 'value_info' is required but missing.
```

### LINKED ISSUE #298: Field 'type' of 'value_info' is required but missing.
onnx checker in ONNX main raises

```
onnxscript/test/common/onnx_script_test_case.py:145: in run_converter_test
    onnx.checker.check_model(model)
../../anaconda3/envs/onnx/lib/python3.10/site-packages/onnx/checker.py:119: in check_model
    C.check_model(protobuf_string, full_check)
E   onnx.onnx_cpp2py_export.checker.ValidationError: Field 'type' of 'value_info' is required but missing.

The above exception was the direct cause of the following exception:
onnxscript/test/functions/onnxfns2_test.py:97: in test_onnxfns_reduce_log_sum_exp
    self.run_onnx_test(
onnxscript/test/common/onnx_script_test_case.py:251: in run_onnx_test
    self.run_converter_test(param, case.model)
onnxscript/test/common/onnx_script_test_case.py:154: in run_converter_test
    raise AssertionError("Verification of model failed.") from e
E   AssertionError: Verification of model failed.
```

for 

```python
@script()
def ReduceLogSumExp(data, axes: List[int], keepdims: int):
    t_axes = op.Constant(value_ints=axes)
    return op.Log(op.ReduceSum(op.Exp(data), t_axes, keepdims=keepdims))
```

BowenBao:
Root cause is new "Reduce ops" spec promotes `axes` from attributes to inputs, resulting in onnx test cases having more inputs than expected by these onnx-script tests. This PR updates these onnx-script test functions to match the signature of the updated onnx test cases.

However, since onnx test cases are not versioned, I feel tests like this that seemingly want to represent ~~older opset~~ onnx operator with function should not depend on it.

Fixes #249 
Fixes #298

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,6 @@ steps:
     python -m pip install -r requirements-dev.txt
   displayName: 'Install dependencies'
 
-# TODO(#249): Fix tests for onnx 1.13
 - script: |
       if [ '$(onnx.standard)' == '1' ]
         then
--- a/onnxscript/test/functions/onnxfns2_test.py
+++ b/onnxscript/test/functions/onnxfns2_test.py
@@ -18,12 +18,10 @@ class TestOnnxFns(onnx_script_test_case.OnnxScriptTestCase):
         reason="onnxruntime does not support that scenario.",
     )
     def test_onnxfns_reduce_sum_square(self):
-        default_axes = None
         default_keepdims = 1
 
         self.run_onnx_test(
             onnxfns2.ReduceSumSquare,
-            axes=default_axes,
             keepdims=default_keepdims,
             # default attributes are not supported yet.
             skip_test_names=[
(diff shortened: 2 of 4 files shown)


justinchuby:
Thanks a lot for fixing the ci! cc @gramalingam  for comments around test usage in the PR description

gramalingam:
Thanks, seems worth looking into a bit more. At the least, the definition of the Reduce functions does target opset18 (where the attribute was promoted to input), so onnxfns2.py should be importing opset18 instead of opset15. But I believe it should support the case where the axes is not specified or axes=None.

Adding @liqunfu .

PR #313, opened 2023-01-13
add ops: ge, le, maxmium, minmium, softmax

xiaowuhu:


justinchuby (approved):
Lgtm!

justinchuby:
Wait: ge and le should be greater/less than or equal to. We are using greater/less here. I merged prematurely @xiaowuhu

PR #314, opened 2023-01-13
feat(atenlib): add op(max)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
     """max(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, op.Constant(value_int=[-1]))
+
+    output = 1
+
+    if dim_or_other is None and keepdim is None:
```
fatcat-z, line 3270: Should we use OptionalHasElement to replace those "is None" statements?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+        result = _aten_max_with_no_dim(self)
+    elif keepdim is None:
+        if isinstance(dim_or_other, int):
+            result, indices = _aten_max_with_dim(self, dim_or_other)
+            output = 2
+        else:  # dim_or_other is tensor
```
fatcat-z, line 3276: If dim_or_other is a Tensor, then the mapped aten op is maximum. I think we don't need to consider dim_or_other is tensor here.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+        result, indices = _aten_max_with_dim_Keepdim(self, dim_or_other, keepdim)
+        output = 2
+
+    if self_rank == 0:
+        result = op.Squeeze(result)
+
+    if output == 2:
+        if self_rank == 0:
+            indices = op.Squeeze(indices)
+        return result, indices
```
fatcat-z, line 3288: For the case of 'dim_or_other is None and keepdim is None', indices will not be constructed.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, op.Constant(value_int=[-1]))
+
+    output = 1
+
+    if dim_or_other is None and keepdim is None:
+        result = _aten_max_with_no_dim(self)
+    elif keepdim is None:
+        if isinstance(dim_or_other, int):
```
titaiwangms, line 3273: if dim_or_other could be int, it should be updated in annotation

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+        output = 2
+
+    if self_rank == 0:
+        result = op.Squeeze(result)
+
+    if output == 2:
+        if self_rank == 0:
+            indices = op.Squeeze(indices)
+        return result, indices
+    else:
```
titaiwangms, line 3289:
```suggestion

```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+        return result
```
titaiwangms, line 3290:
```suggestion
    return result
```

titaiwangms: save one else

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,65 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+@torch_op("aten::max", trace_only=True)
+def aten_max(self: TReal, dim_or_other: TReal = None, keepdim: BOOL = None) -> TReal:
     """max(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, op.Constant(value_int=[-1]))
+
+    output = 1
```
titaiwangms, line 3268: If we are really recording output number, I suggest we put output=2 here, and we can save one line in the following if/else condition ...

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: but pytorch have test case for this scenario.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: yes. so in this case, the output==1, we will return result only.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: I followed your suggestion. But I think 'else' can be more readable. :)

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: done.

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: I think that test case was designed for torch.max() function. But according to the dim_or_other, it will be mapped to different aten op.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: refine the logic in another way

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Then how to pass the test case? I think it is an overload.

fatcat-z (approved):
LGTM, thanks.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3257,10 +3257,58 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
+
+
+@torch_op("aten::max", overload=True)
+def _aten_max_with_other(self: TReal, other: TReal) -> TReal:
+    result = op.Max(self, other)
+    return result
+
+
+@torch_op("aten::max", overload=True)
+# def _aten_max_with_dim(self: TReal, dim: int, keepdim: bool) -> tuple[TReal, TInt]:
```
justinchuby, line 3306: nit: Dead code

PR #316, opened 2023-01-13
Extend type annotation support

### LINKED ISSUE #207: Support Annotated types
We would like to annotate the types with `typing.Annotated` (https://docs.python.org/3/library/typing.html#typing.Annotated) to include run time value checking for the ATen lib. Annotate should only be used for attributes.

```python
@onnxscript.script()
@atenop("aten::elu")
def Elu(
    self,
    alpha: float = 1.0,
    scale: Annotated[float, Is[lambda x: x == 1.0]] = 1.0,   # Here, just need to take the float type out
    input_scale: Annotated[float, Is[lambda x: x == 1.0]] = 1.0,
):
    # del scale
    # del input_scale
    return op.Elu(self, alpha=alpha)
```

### LINKED ISSUE #280: Support for annotating Optional inputs
We would like to make this work:

```python
@torch_op("aten::clamp")
def aten_clamp(
    self: TReal, min_: Optional[TReal] = None, max_: Optional[TReal] = None
) -> TReal:
    # clamp(Tensor self, Scalar? min=None, Scalar? max=None) -> Tensor

    if op.OptionalHasElement(min_):
        min_ = op.OptionalGetElement(min_)
        min_clamp = op.CastLike(min_, self)
    else:
        # TODO: get the correct min number for the dtype
        min_clamp = op.Constant(value_float=float("-inf"))

    if op.OptionalHasElement(max_):
        max_ = op.OptionalGetElement(max_)
        max_clamp = op.CastLike(max_, self)
    else:
        max_clamp = op.Constant(value_float=float("inf"))

    # Enforce the lower and upper bounds
    clamped = op.Max(op.Min(self, max_clamp), min_clamp)
    return clamped
```

onnxruntime complains that

E           onnxruntime.capi.onnxruntime_pybind11_state.InvalidGraph: [ONNXRuntimeError] : 10 : INVALID_GRAPH : This is an invalid model. In Node, ("", OptionalHasElement, "", -1) : () -> ("output0",) , Error Node () has input size 0 not in range [min=1, max=1].

## Reference

- #76

gramalingam:
Extend the type-annotations recognized by the converter to support

* Union[...]
* Optional[...]
* TypeVar with bounds
* Annotated[...]

TODO: Existing usage uses TypeVar with bounds, like `TInt = TypeVar("TInt", bound=IntType)`. May be useful to also support `T = TypeVar("TInt", INT32, FLOAT32)`.

Fixes #207 fixes #280


onnxscript/type_annotation.py
```diff
@@ -28,31 +28,91 @@
 _LIST_CONSTRUCTORS = frozenset([list, typing.List, typing.Sequence, collections.abc.Sequence])
 
 
-def pytype_to_attrtype(pytype: type) -> typing.Optional[onnx.AttributeProto.AttributeType]:
+def remove_annotation(typeinfo):
+    """Remove Annotated wrapper if present, otherwise return typeinfo as is."""
+    if hasattr(typing, "Annotated"):
```
justinchuby, line 33: Do we want to handle Annotated from typing extensions as well?

onnxscript/type_annotation.py
```diff
@@ -28,31 +28,91 @@
...
+        # Present in Python 3.9+
+        if get_origin(typeinfo) is typing.Annotated:
+            return get_args(typeinfo)[0]
+    return typeinfo
+
+
+def is_primitive_attr_type(typeinfo) -> bool:
```
justinchuby, line 40: nit: annotations for params

onnxscript/type_annotation.py
```diff
@@ -28,31 +28,91 @@
...
     if pytype in _PYTYPE_TO_ATTRTYPE_MAP:
         return _PYTYPE_TO_ATTRTYPE_MAP[pytype]
-    if get_origin(pytype) in _LIST_CONSTRUCTORS:
-        args = get_args(pytype)
-        elt_type = args[0]
+    type_constructor = get_origin(pytype)
+    # Remove Optional wrapper if present, which is represented as an Union[..., None]
+    if type_constructor is typing.Union:
+        # Filter out None, since typing.Optional[X] evaluates to Union[X, None]
+        args = [x for x in get_args(pytype) if x is not type(None)]  # noqa: E721
```
justinchuby, line 52: Just making sure, is type(None) None?

onnxscript/type_annotation.py
```diff
@@ -28,31 +28,91 @@
...
+            return True
+        elif (len(args) == 1) and args_value_check[0] is False:
+            # Handle the case of optional attribute: eg. Optional[int]
+            # Note that we do not allow Union[int, float] for attributes.
+            return False
+        else:
+            raise ValueError(f"Unsupported type annotation {typeinfo}")
+    # Handle TypeVars:
+    if isinstance(typeinfo, typing.TypeVar):
+        if hasattr(typeinfo, "__bound__"):
```
justinchuby, line 101: A test for this may be helpful? I have seen cases where they type becomes string from inspection signature - I might have done something wrong but just to make sure.

onnxscript/type_annotation.py
```diff
@@ -28,31 +28,91 @@
...
+        args_value_check = [is_value_type(x) for x in args]
+        if all(args_value_check):
+            # Handles cases like Optional[INT32] as well as Union[FLOAT16, FLOAT, DOUBLE]
+            return True
+        elif (len(args) == 1) and args_value_check[0] is False:
+            # Handle the case of optional attribute: eg. Optional[int]
+            # Note that we do not allow Union[int, float] for attributes.
+            return False
+        else:
+            raise ValueError(f"Unsupported type annotation {typeinfo}")
```
justinchuby, line 98:
```suggestion
            raise ValueError(f"Unsupported type annotation '{typeinfo}'")
```

Quote user inputs for clarity? Here and elsewhere

onnxscript/type_annotation.py
gramalingam: No. I believe the easiest way to specify the type of None is `type(None)`.

onnxscript/type_annotation.py
gramalingam: I did try it out. All examples such as [TInt](https://github.com/microsoft/onnx-script/blob/9439e109d1246cd5d75a76ce3f9f07ff58d3bd31/onnxscript/function_libs/torch_aten/typing.py#L59) from torch_aten\typing.py go through this, and I think you already have enough testcases of these?

Not sure I understood your point about string. Are you talking about delayed-evaluation using strings?

onnxscript/type_annotation.py
gramalingam: Would be nice. But seems non-trivial (to say anything other than `Any`).

justinchuby: I think Any can be better than not annotated.

onnxscript/type_annotation.py
justinchuby: I see. Just thought `Union[x, None]` will contain None but I must be wrong

onnxscript/type_annotation.py
```diff
@@ -30,31 +30,91 @@
...
+            raise ValueError(f"Unsupported type annotation '{typeinfo}'")
+    # Handle TypeVars:
+    if isinstance(typeinfo, typing.TypeVar):
+        if hasattr(typeinfo, "__bound__"):
+            bound = typeinfo.__bound__
+            return is_value_type(bound)
+    raise ValueError(f"Unsupported type annotation {typeinfo}")
+
+
+def is_attr_type(pytype: type):
```
justinchuby, line 109: Hide all helper functions with prefix `_`?

onnxscript/type_annotation.py
```diff
@@ -30,31 +30,91 @@
...
-    if get_origin(pytype) in _LIST_CONSTRUCTORS:
-        args = get_args(pytype)
-        elt_type = args[0]
+    type_constructor = get_origin(pytype)
+    # Remove Optional wrapper if present, which is represented as an Union[..., None]
+    if type_constructor is typing.Union:
+        # Filter out None, since typing.Optional[X] evaluates to Union[X, None]
+        args = [x for x in get_args(pytype) if x is not type(None)]  # noqa: E721
+        if len(args) == 1:
+            return pytype_to_attrtype(args[0])
```
justinchuby, line 56: Is it true that we still use None when there are multiple types? Can the info be stored somewhere?

justinchuby (approved):

onnxscript/type_annotation.py
justinchuby: I would just include a few unit tests for showing all of the possible usage so that it's well documented

the string thing could be me not using it the right way. I will consult with you later :)

onnxscript/type_annotation.py
```diff
@@ -13,24 +13,36 @@
...
+# * TypeVars with above bounds
+# * Above types with annotation attached
+TypeAnnotationValue = typing.Any
+
+# Map from python type to corresponding ONNX AttributeProto type
 _PYTYPE_TO_ATTRTYPE_MAP = {
     float: onnx.AttributeProto.FLOAT,
     int: onnx.AttributeProto.INT,
     str: onnx.AttributeProto.STRING,
-    bool: onnx.AttributeProto.INT,
```
justinchuby, line 20: 👀

gramalingam: (Weird UI ... my comment disappeared.) I had to add bool as many aten examples use it, and I was running into issues there.

onnxscript/type_annotation.py
gramalingam: Oh, I think it's my comment that is misleading ... I should say `Union[X, type(None)]` in the comment ... I think the typing support implicitly treats (promotes) `None` as `type(None)` but they are distinct values.

Issue #305, opened 2023-01-11
Supply attributes by position and inputs by keyword


justinchuby:
The ability to supply attributes as positional arguments and inputs and keyword arguments

Eager and the translator both need to change.

PR #323, opened 2023-01-16
feat(atenlib): add ops(gelu, celu)

xiaowuhu:
add GELU,CELU op


onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -198,10 +198,11 @@ def aten_binary_cross_entropy_backward(
     raise NotImplementedError()
 
 
-def aten_celu(self, alpha: float = 1.0):
+@torch_op("aten::celu")
+def aten_celu(self: TFloat, alpha: float = 1.0) -> TFloat:
     # celu(Tensor self, Scalar alpha=1.0) -> Tensor
 
-    raise NotImplementedError()
+    return op.Celu(self, alpha=alpha)
```
fatcat-z, line 205: The input/output of Celu is limited to float32, so we might need a dtype to cast the value back.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -198,10 +198,15 @@ def aten_binary_cross_entropy_backward(
     raise NotImplementedError()
 
 
-def aten_celu(self, alpha: float = 1.0):
+@torch_op("aten::celu")
+def aten_celu(self: TFloat, alpha: float = 1.0, dtype: int = -1) -> TFloat:
```
justinchuby, line 202:
```suggestion
def aten_celu(self: RealType, alpha: float = 1.0, dtype: int = -1) -> FLOAT:
```

since casting is done in the function, we can support any inputs. 

Is dtype part of the aten signature? Could you point me to the doc?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -323,10 +328,32 @@ def aten_fractional_max_pool3d_backward(
...
+@torch_op("aten::gelu")
+def aten_gelu(self: TReal, approximate: str = "none") -> TReal:
     # gelu(Tensor self, *, str approximate='none') -> Tensor
 
-    raise NotImplementedError()
+    self = op.Cast(self, to=FLOAT.dtype)
+
+    if approximate == "tanh":
+        # GELU(x) = 0.5 * x * {1 + Tanh[\sqrt(2/pi) * (x + 0.044715 * x^3)]}
+        inner1 = op.Div(2.0, 3.141592653589793)
```
justinchuby, line 339: naming: inner1: consider a more meaningful name.

onnxscript/values.py
```diff
@@ -213,6 +213,8 @@ def adapt(input: ExtendedModeValue) -> EagerModeValue:
             return input
         elif isinstance(input, (bool, int, float)):
             return tensor.Tensor(np.array(input))
+        elif isinstance(input, str):
```
justinchuby, line 216: This happens because approximate was passed in positionally. We should create a wrangler instead until #305 is fixed

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: We don't really know what the input dtype is. I think this can be handled by the caller.

onnxscript/function_libs/torch_aten/ops/nn.py
fatcat-z: > since casting is done in the function, we can support any inputs.
> 
> Is dtype part of the aten signature? Could you point me to the doc?

No, dtype is not a part of the aten signature and it was added just for knowing the input data type so we can cast it back after calling celu which only supports float32.

If we can ask the caller to make sure the input data type, then we don't need to add dtype here and avoid those casting work.

onnxscript/values.py
xiaowuhu: Is there any "str" type signature that I can reference?

justinchuby: strs can be attributes last time I checked

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -332,12 +332,10 @@ def aten_gelu(self: TReal, approximate: str = "none") -> TReal:
 
     if approximate == "tanh":
         # GELU(x) = 0.5 * x * {1 + Tanh[\sqrt(2/pi) * (x + 0.044715 * x^3)]}
-        inner1 = op.Div(2.0, 3.141592653589793)
-        inner1 = op.Sqrt(inner1)
         self_cube = op.Pow(self, 3)
```
justinchuby, line 335: nit: cubed

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -198,10 +198,11 @@ def aten_binary_cross_entropy_backward(
     raise NotImplementedError()
 
 
-def aten_celu(self, alpha: float = 1.0):
+@torch_op("aten::celu")
+def aten_celu(self: TFloat, alpha: float = 1.0) -> TFloat:
```
justinchuby, line 202:
```suggestion
def aten_celu(self: FLOAT, alpha: float = 1.0) -> FLOAT:
```

because only one type is supported. TFloat can be one of a few types.

onnxscript/values.py
justinchuby: Now that I think about it again, does having an if condition on a str make it an input? So this change is desired. @gramalingam

justinchuby (approved):
with wrangler and reverting the edit in values.py

PR #324, opened 2023-01-17
feat(atenlib): add ops (ArgMax, ArgMin, Detach)

xiaowuhu:
add ops: argmax, argmin, but use trace_only = True mode, due to OptionalHasElement(dim) cannot work now.


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -331,20 +331,38 @@ def aten_arctanh(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_argmax(
-    self: TensorType, dim: Optional[int] = None, keepdim: bool = False
-) -> TensorType:
+@torch_op("aten::argmax", trace_only=True)
+def aten_argmax(self: TReal, dim: int = None, keepdim: bool = False) -> TReal:
```
justinchuby, line 335:
```suggestion
def aten_argmax(self: TReal, dim: Optional[int] = None, keepdim: bool = False) -> TReal:
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1383,10 +1401,11 @@ def aten_det(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::detach")
 def aten_detach(self: TensorType) -> TensorType:
     # detach(Tensor(a) self) -> Tensor(a)
 
-    raise NotImplementedError()
+    return op.Identity(self)
```
justinchuby, line 1408:
```suggestion
    return self
```

This should be good

xiaowuhu: if return self only, no op.xxx() function called inside this function, the test cannot be passed.

justinchuby: I see. SG

justinchuby (approved):

PR #326, opened 2023-01-17
ci: do not run lints on drafts to make reviewing easier

justinchuby:
image

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -6,6 +6,7 @@ on:
       - main
       - 'gh/**/base' # ghstack base branches
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
 
 jobs:
   optional-lint:
@@ -35,6 +36,8 @@ jobs:
   enforce-style:
     name: Enforce style
     runs-on: ubuntu-latest
+    # Do not run on drafts to make reviewing easier
+    if: github.event.pull_request.draft == false
     steps:
       - uses: actions/checkout@v3
       - name: Setup Python


gramalingam (approved):
Very helpful, thanks!

.github/workflows/lint.yaml
```diff
@@ -35,6 +36,8 @@ jobs:
   enforce-style:
     name: Enforce style
     runs-on: ubuntu-latest
+    # Do not run on drafts to make reviewing easier
+    if: github.event.pull_request.draft == false
```
justinchuby, line 40: Does this work in commits?

justinchuby: Let's just merge it and try.

titaiwangms (approved):

.github/workflows/lint.yaml
justinchuby: Good: it works

gramalingam:
A question: does this distinguish between CI failures that affect the browser view of the code from those that don't? I think that the important thing to disable is the first category, not necessarily the second.

gramalingam:
Is there a difference between what is checked and reported by the messages logged in the browser view of code and what is checked by lintrunner etc. ?

justinchuby:
They should be the same

justinchuby:
Oh! I _could_ just skip the web reporting part of it in draft mode and retain the lintrunner run; is that what you would suggest?

gramalingam:
Yes, I think that could be useful. It is not invasive to the reviewer, but can remind the PR creator if they have anything to fix.

justinchuby:
SG!

PR #328, opened 2023-01-17
fix(ci): do not submit sarif on drafts to make reviewing easier

justinchuby:
Do not submit sarif on drafts to make reviewing easier, but always run lintrunner.

Follow up of #326


gramalingam (approved):
Thanks!

PR #335, opened 2023-01-18
feat: generalized `Evaluator` and `Tensor` interfaces

justinchuby:
Update the evaluator and tensor's interfaces for graph capturing evaluators.

- Add `eval_function` to Evaluator and calls it from functions
- Add `rank` to `Tensor`
- Make all helper functions in evaluator private

### DIFF
--- a/onnxscript/evaluator.py
+++ b/onnxscript/evaluator.py
@@ -12,6 +12,7 @@ from typing import Any, Optional
 
 import numpy as np
 import onnx
+import onnx.helper
 
 from onnxscript import autocast, irbuilder, onnx_opset, tensor, utils, values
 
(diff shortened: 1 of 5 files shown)


titaiwangms (approved):

justinchuby:
@gramalingam so this part fails with the api change in the evaluator:

```python
    # This is a bit fragile. We want to get the ONNXFunction object representing
    # the outer-scope ONNXScript function from the execution stack. The caller of
    # @graph is the original script function (cumulative_sum in the above example),
    # and the caller of that function is the wrapper function/method in the
    # corresponding OnnxFunction object.
    # Currently, there is no support for eager-mode execution of nested functions,
    # so we don't need to handle doubly nested functions (e.g., a function defined
    # inside Sum in the above example).

    function_frame = sys._getframe(1)  # pylint: disable=protected-access
    wrapper_frame = sys._getframe(2)  # pylint: disable=protected-access
    onnx_function = wrapper_frame.f_locals["self"]
    nested_functions = onnx_function.function_ir.nested_functions
```

justinchuby:
I fixed it by updating the frame it looks for.

justinchuby:
I am going to merge this to process with the graph capturing evaluator. LMK if changes need to be made and I will send a patch

PR #336, opened 2023-01-19
feat(atenlib): onnxscript API on torchscript

### LINKED ISSUE #273: Implement torchscript as a graph backend

justinchuby:
An evaluator using onnxscript API to build graph using torchscript (we will have our own graph to replace torchscript.)

works with https://github.com/pytorch/pytorch/pull/92717

1. Op object triggers graph_eval instead of eval to begin op insertion for the graph (Need to figure out how to keep both functionality if they need to stay on the same API.)
2. TorchscriptEvaluator serves as a manager of input/output graph. In this demo, both of them are torchscript, so this class seems simple (should be more complicated depending on ONNX graph builder API and format requirements).
3. tensor.py is not modified because onnxscript does not really use the graph value. That is, it will only have to be compatible with graph builder type only in the future implementation.

Fixes #273

Co-authored-by: Ti-Tai Wang <titaiwang@microsoft.com>


onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+]
+
+
+ValidArgumentType = Union["TorchScriptTensor", str, int, float, bool]
+
+# TODO(justinchuby): Build a context manager to handle source information.
+
+
+class TorchScriptTensor(onnxscript_tensor.Tensor):
+    """A onnxscript tensor that wraps a torchscript Value."""
```
wschin, line 35: Please add document (probably a design doc) describing how it should be used with other components (feel free to call me for details). The name is confusing; torchscript value can be non-tensor objects, so maybe we can call it TorchScriptValue.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+        onnx_func: The onnx function.
+        args: The positional arguments.
+        kwargs: The keyword arguments.
+
+    Returns:
+        A tuple of (onnx_inputs, onnx_attributes).
+    """
+    function_ir = onnx_func.function_ir
+    # FIXME(justinchuby): There can be attributes in the args too.
+    # The first len(func_ir.inputs) arguments are onnx inputs
```
wschin, line 101: Is this always true? If not, please add enough information so that future DRI can easily fix it.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+    def symbolic_value(self) -> torch.Value:
+        """The symbolic Value in torch.Graph."""
+        return self._value
+
+
+@beartype
+def _split_args_kwargs_to_input_attr(
+    onnx_func: onnxscript.OnnxFunction, args, kwargs
+) -> tuple[list[ValidArgumentType], dict[str, ValidArgumentType]]:
+    """Split the function args and kwargs to onnx inputs and attributes.
```
wschin, line 89: What does "the function" mean? `aten::add.default`, `aten:add`, ONNX Add, or ONNXScript Add Function?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+
+    Args:
+        onnx_func: The onnx function.
+        args: The positional arguments.
```
wschin, line 93: `The positional arguments` doesn't seem carry much information. Is it the `args` passed into `aten::add.default`?

onnxscript/tensor.py
```diff
@@ -14,6 +14,7 @@
 from onnxscript import onnx_opset
 
 
+# TODO(titaiwang): FakeTensor acceptable
```
wschin, line 17: What does this mean?

onnxscript/values.py
```diff
@@ -294,8 +294,8 @@ def __call__(self, *args, **kwargs):
         # FIXME(after #225): Move import to the top of the file.
         from onnxscript import evaluator  # pylint: disable=import-outside-toplevel
 
-        new_args, has_array = _adapt_to_eager_mode(args)
-        result = evaluator.default().eval_function(self, *new_args, **kwargs)
+        # new_args, has_array = _adapt_to_eager_mode(args)
```
wschin, line 297: Remove this line maybe?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+
+    assert len(node_ouputs) == n_outputs
+    # Add all attributes
+    for key, value in sorted(attributes.items()):
+        _add_attribute_to_torchscrpt_node(node, key, value)
+
+    return node_ouputs
+
+
+class TorchScriptGraph:
```
wschin, line 269: Please rename `TorchScriptGraph` to `TorchScriptWrapper`. I thought it `torch._C.Graph` until seeing this definition.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+        return self._graph
+
+    @beartype
+    def eval_function(
+        self,
+        function: onnxscript.OnnxFunction,
+        args: Sequence[ValidArgumentType],
+        kwargs: Sequence[ValidArgumentType],
+    ):
+        # args/kwargs are TorchScriptTensor/python built-in based
```
wschin, line 210: Does it mean when I see
```
z = torch.ops.aten.add.default(x, y, 5)
```

I can call
```
etor = TorchScriptEvaluator(...)
z = etor.eval_function(torch.ops.aten.add.default, x, y)
```
? I am just wondering what do "args" and "kwargs" mean here, and how this is used inside converter.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @beartype
+    def _add_torchscript_op(
+        self,
+        name: str,
+        onnx_inputs,
+        onnx_attributes,
+        n_outputs: int,
+    ) -> TorchScriptTensor | tuple[TorchScriptTensor, ...]:
+
```
wschin, line 359: In terms of documentation, we don't distinguish between "internal" and "external" functions, so they need to be documented equally well.

```
# Add a TorchScript node (type: torch._C.Node) into the underlying TorchScript graph (type: torch._C.Graph). For example, to add an `aten::add`, you can call this function with
  name = "aten::add"
  x = TorchScriptTensor(...)
  y = TorchScriptTensor(...)
  onnx_inputs = [x, y]
  onnx_attributes = [5]
``

File: onnxscript/function_libs/torch_aten/graph_building.py (hunk 1 above)
Comment (justinchuby) [reply onnxscript/function_libs/torch_aten/graph_building.py:35]: This is an onnxscript tensor. A little confusing indeed. We will see if we could name it better

File: onnxscript/function_libs/torch_aten/graph_building.py (hunk 7 above)
Comment (justinchuby) [reply onnxscript/function_libs/torch_aten/graph_building.py:269]: This is an onnxscript graph. We will consider a better name

File: onnxscript/values.py (hunk 6 above)
Comment (gramalingam) [reply onnxscript/values.py:297]: Where is the "adapt" happening now? I don't see it in `eval_function`. It has to happen somewhere (for the normal evaluators).

Comment (justinchuby) [reply onnxscript/values.py:297]: I moved it to the Evaluator in https://github.com/microsoft/onnx-script/pull/352

File: onnxscript/function_libs/torch_aten/graph_building.py (hunk 1 above)
Comment (gramalingam) [reply onnxscript/function_libs/torch_aten/graph_building.py:35]: It may also be worth having a document/diagram illustrating the system and the flow of data through it. A key point is to pin down the representation-types used for different-types of values in different parts of the system (as much of the work is simply transforming between equivalent-representations used by different parts of the system).

Eg., I wonder if this could just be a wrapper around numpy arrays (which could serve as common intermediary format). Just to avoid too many layers of wrappers and inter-dependences.

File: onnxscript/function_libs/torch_aten/graph_building.py [hunk 10]
```diff
@@ -0,0 +1,446 @@
...
+from onnxscript import tensor as onnxscript_tensor
+
+__all__ = [
+    "TorchScriptTensor",
+    "TorchScriptGraph",
+    "TorchScriptEvaluator",
+]
+
+
+ValidArgumentType = Union["TorchScriptTensor", str, int, float, bool]
```
gramalingam, line 29: How about lists of tensors? Or, optionals?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+        ).onnx_type()
+
+    def symbolic_value(self) -> torch.Value:
+        """The symbolic Value in torch.Graph."""
+        return self._value
+
+
+@beartype
+def _split_args_kwargs_to_input_attr(
+    onnx_func: onnxscript.OnnxFunction, args, kwargs
```
gramalingam, line 87: What are the types of args/kwargs?

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: That’s a great point. Thinking about this again I wonder if it makes sense to only adapt numpy arrays (but not Python values) to Tensor in eager execution. So values flowing through systems in onnxscript all the way into Evaluator are either Tensor or Python primitives. 

This subclass of Tensor in particular needs to be symbolic and cannot have just a numpy array but needs to carry a reference to the node in the underlying graph (torch.Value) I think

justinchuby: (Working on the graph)

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Good catch

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+                inputs=(),
+                attributes=dict(value=torch.tensor(constant, dtype=torch.float)),
+            )[0]
+
+        raise TypeError(
+            f"Constant input `{constant}` of type '{type(constant)}' is not supported"
+        )
+
+    @beartype
+    def _add_torchscript_op(
```
gramalingam, line 352: nit: `_add_torchscript_op_call` may help distinguish an op from an invocation of an op.

gramalingam: Similarly for other methods below

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+    if isinstance(value, float):
+        return node.f_(key, value)
+    if isinstance(value, int):
+        return node.i_(key, value)
+    if isinstance(value, str):
+        return node.s_(key, value)
+    if isinstance(value, torch.Tensor):
+        return node.t_(key, value)
+    if isinstance(value, Sequence):
+        if isinstance(value, float):
```
gramalingam, line 235: Doesn't look right. May be check either value[0] or all values?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+            return node.fs_(key, list(value))
+        if isinstance(value, int):
```
gramalingam, line 237: Same as above

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: SG!

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Good catch thanks

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @beartype
+    def eval_function(
+        self,
+        function: onnxscript.OnnxFunction,
+        args: Sequence[ValidArgumentType],
+        kwargs: Sequence[ValidArgumentType],
+    ):
+        # args/kwargs are TorchScriptTensor/python built-in based
+        inputs, attributes = _split_args_kwargs_to_input_attr(function, args, kwargs)
```
gramalingam, line 211: Maybe it would be useful to do this in onnxscript *before* it calls the evaluator? Isn't this generic behavior we want shared by all evaluators?

onnxscript/function_libs/torch_aten/graph_building.py
gramalingam: Sorry, I misunderstood. So, this is a placeholder, not a real tensor (is this what's referred to as FakeTensor elsewhere?)

It seems a bit fragile (to make Tensor.value None). Is it possible to get by without making this inherit from onnxscript Tensor?

gramalingam: That is rely more on duck-typing, than actual inheritance-subtyping

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Good point. I was going to propose the same but thought it would be a longer discussion. Does the current split logic look good to you?

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: I considered making a `TensorProtocol` for this, but then the functions will then be annotated as `TensorProtocol[TReal]` etc., which is not exactly clean.

Alternatively we can rename `Tensor` to `ConcreteTensor`, then make `Tensor` the protocol, but then we need a tensor constructor e.g. `def tensor(nparray) -> ConcreteTensor` like pytorch does.

What do you think?

justinchuby: > Sorry, I misunderstood. So, this is a placeholder, not a real tensor (is this what's referred to as FakeTensor elsewhere?)

That's accurate, although it _can_ carry a value when available too

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Updated doc: Split the args and kwargs supplied to `onnx_func`...

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Updated: The positional arguments supplied to `onnx_func`.

onnxscript/values.py
justinchuby: Done

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: In spirit yes; in reality some wrapping needs to be done:

When x, y are symbolic tensors of type `TorchScriptTensor`:

```python
evaluator = TorchScriptEvaluator(graph)
with onnxscript.evaluator.default_as(evaluator):
    z = atenlib.aten_add(x, y, 5)
```

When x, y are np arrays:

```python
# This is using ORT by default
z = atenlib.aten_add(x, y, 5)
```

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Removed.

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: updated

```python
def _split_args_kwargs_to_input_attr(
    onnx_func: onnxscript.OnnxFunction, args: Sequence[ValidArgumentType], kwargs: dict[str, ValidArgumentType]
) -> tuple[Sequence[ValidInputType], dict[str, ValidArgumentType]]:
```

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: @gramalingam I also do want to inherit all the magic methods `__add__` etc. so we don't lose ops when uses use python symbols in a trace only function

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+    # Construct a dictionary of attributes with their names specified in the function
+    # definition
+    onnx_attributes = {}
+
+    # (1) Some/All attributes are supplied as positional arguments
+    attr_name_to_protos = collections.OrderedDict(
+        (attr.name, attr) for attr in function_ir.attr_protos
+    )
+
+    assert len(function_ir.attr_protos) >= len(attributes_in_args)
```
gramalingam, line 114: Don't understand what this is checking. In function, attributes are split into those with and without default values. In the args, they are split into args and kwargs. So, what is this checking?

gramalingam: Similar questions about the logic below (unless function_ir was changed recently in a way I am unaware of).

justinchuby: so some attributes are in args and some are in kwargs. `len(function_ir.attr_protos)` is the number of total attributes. We are asserting that the number of total attributes is more than or equal to the number of attributes in args.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,446 @@
...
+
+    Returns:
+        A tuple of (onnx_inputs, onnx_attributes).
+    """
+    function_ir = onnx_func.function_ir
+    # FIXME(justinchuby): There can be attributes in the args too.
+    # The first len(func_ir.inputs) arguments are onnx inputs
+    onnx_inputs = args[: len(function_ir.inputs)]
+    # The rest is onnx attributes
+    attributes_in_args = args[len(function_ir.inputs) :]
```
justinchuby, line 104: What about when # of inputs > # args, aka. all args are inputs and some kwargs are also inputs?

onnxscript/function_libs/torch_aten/graph_building.py
gramalingam: AFAIK, `len(function_ir.attrs) + len(function_ir.attr_protos)` is the number of total attributes

justinchuby: hmm. how do we get the type for attrs? I thought attr_protos contains everything

justinchuby:
schema: schema.outputs

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -51,10 +51,11 @@ def aten_acosh(self: TFloat) -> TFloat:
 
 
 @torch_op("aten::add")
-def aten_add(self: TReal, other: TReal, alpha: float = 1) -> TReal:
+def aten_add(self: TReal, other: TReal, alpha: float) -> TReal:
```
justinchuby, line 54:
```suggestion
def aten_add(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -51,10 +51,11 @@ def aten_acosh(self: TFloat) -> TFloat:
...
 
 @torch_op("aten::add")
-def aten_add(self: TReal, other: TReal, alpha: float = 1) -> TReal:
+def aten_add(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
     # add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor
-    if alpha != 1:
-        other = op.Mul(other, alpha)
+    one = op.Constant(value_float=1.0)
+    alpha = op.CastLike(alpha, one)
+    other = op.Mul(other, alpha)
```
justinchuby, line 58:
```suggestion
    alpha = op.CastLike(alpha, other)
    other = op.Mul(other, alpha)
```

justinchuby:
```
  _ TestTorchScriptTracingEvaluator.test_traced_graph_on_single_node_is_same_as_compiled_graph _
  [gw1] linux -- Python 3.10.9 /home/runner/work/onnx-script/onnx-script/.tox/test-onnx-preview/bin/python
  onnxscript/function_libs/torch_aten/graph_building_test.py:46: in test_traced_graph_on_single_node_is_same_as_compiled_graph
      x = self.onnxscript_graph.add_input("x", torch.randn(2, 3, 4))
  <@beartype(onnxscript.function_libs.torch_aten.graph_building.TorchScriptGraph.add_input) at 0x7f3d9c0dc700>:52: in add_input
      ???
  onnxscript/function_libs/torch_aten/graph_building.py:282: in add_input
      torch_value = self._graph.addInput(input_name)
  E   TypeError: addInput(): incompatible function arguments. The following argument types are supported:
  E       1. (self: torch._C.Graph) -> torch::jit::Value
  E   
  E   Invoked with: graph():
  E     return ()
  E   , 'x'
```

justinchuby:
```
_______________________ TestFxToOnnxWithOnnxRuntime.test_gpt2_tiny ________________________
  File "/home/justinchu/dev/pytorch/test/onnx/test_fx_to_onnx_with_onnxruntime.py", line 123, in test_gpt2_tiny
    onnx_model = fx_onnx.export_without_kwargs(
  File "/home/justinchu/dev/pytorch/torch/onnx/_internal/fx/export.py", line 614, in export_without_kwargs
    return _export(
  File "/home/justinchu/dev/pytorch/torch/onnx/_internal/fx/export.py", line 526, in _export
    onnx_model = onnxscript_graph.to_model_proto(ts_initializers, opset_version)
  File "<@beartype(onnxscript.function_libs.torch_aten.graph_building.TorchScriptGraph.to_model_proto) at 0x7f3e2cca9670>", line 55, in to_model_proto
  File "/home/justinchu/dev/onnx-script/onnxscript/function_libs/torch_aten/graph_building.py", line 442, in to_model_proto
    onnx_model = onnx.shape_inference.infer_shapes(onnx_model, True, True)
  File "/home/justinchu/anaconda3/envs/pytorch/lib/python3.9/site-packages/onnx/shape_inference.py", line 40, in infer_shapes
    inferred_model_str = C.infer_shapes(
onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] Shape inference error(s): (op_type:aten_rsub, node name: aten_rsub_22): [ShapeInferenceError] (op_type:Sub, node name: n3): B has inconsistent type tensor(int32)
(op_type:aten_getitem, node name: aten_getitem_37): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_getitem, node name: aten_getitem_39): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_getitem, node name: aten_getitem_41): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_expand, node name: aten_expand_57): [TypeInferenceError] Input 0 type is missing.
(op_type:aten_where, node name: aten_where_87): [ShapeInferenceError] (op_type:Where, node name: n0): condition typestr: B, has unsupported type: tensor(float)
(op_type:aten_getitem, node name: aten_getitem_150): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_getitem, node name: aten_getitem_152): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_getitem, node name: aten_getitem_154): [ShapeInferenceError] (op_type:Gather, node name: n2): data typestr: T, has unsupported type: seq(tensor(float))
(op_type:aten_expand, node name: aten_expand_170): [TypeInferenceError] Input 0 type is missing.
(op_type:aten_where, node name: aten_where_200): [ShapeInferenceError] (op_type:Where, node name: n0): condition typestr: B, has unsupported type: tensor(float)
```

onnxscript/function_libs/torch_aten/graph_building_test.py
```diff
@@ -0,0 +1,68 @@
...
+        self.opset_version = 17
+        self.onnxscript_graph = graph_building.TorchScriptGraph()
+        self.tracer = graph_building.TorchScriptTracingEvaluator(self.onnxscript_graph)
+
+    def to_model_proto(self):
+        # TODO(titaiwang): initializer API
+        return self.onnxscript_graph.to_model_proto(
+            initializers={}, opset_version=self.opset_version
+        )
+
```
justinchuby, line 27:
```suggestion

```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -17,6 +17,7 @@
 from onnxscript.function_libs.torch_aten.registration import torch_op
 from onnxscript.function_libs.torch_aten.tensor_typing import (
     IntType,
+    RealType,
```
justinchuby, line 20:
```suggestion
```

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -0,0 +1,485 @@
...
+    @beartype
+    def eval_function(
+        self,
+        function: onnxscript.OnnxFunction,
+        args: Sequence[ValidArgumentType],
+        kwargs: Dict[str, ValidArgumentType],
+    ):
+        # args/kwargs are TorchScriptTensor/python built-in based
+        param_schemas = param_manipulation.extract_param_schema_from_function(function)
+        inputs, attributes = param_manipulation.separate_input_attributes_from_arguments(
```
justinchuby, line 208: @gramalingam used in here

titaiwangms (approved):

PR #340, opened 2023-01-19
feat(atenlib): implementation of conv2d

justinchuby:
Replacement of #333.


justinchuby:
TODO: Expand the input tensor to 4d

PR #348, opened 2023-01-20
feat(atenlib): add ops(sum)

xiaowuhu:
add op: sum with trace_only = True, due to OptionalHasElement() cannot work now


justinchuby (approved):
thanks!

PR #350, opened 2023-01-20
fix(atenlib): better trace only ops

justinchuby:
Update the trace only ops to still compile if (rank) conditions, so the symbolic evaluator can work without shape information.


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -728,18 +744,14 @@ def aten_cartesian_prod(tensors: Sequence[TensorType]) -> TensorType:
...
-    a = op.SequenceEmpty()
-    for i in range(num_of_input):
-        if op.Size(tensors[i]) == 0 and dim != 0:
-            # Skip empty tensors when concatenating along non-zero dimension
-            continue
-        a = op.SequenceInsert(a, tensors[i])
-    return op.ConcatFromSequence(a, axis=dim)
+    # NOTE: Having empty tensors when concatenating along non-zero dimension
+    # is not supported.
+    # TODO: Filter these tensors out before calling ConcatFromSequence.
```
justinchuby, line 753: ... with Sequence ops

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3522,25 +3541,42 @@ def aten_native_group_norm_backward(
 @torch_op("aten::native_layer_norm", trace_only=True)
 def aten_native_layer_norm(
     input: TReal,
-    normalized_shape: INT64,
+    normalized_shape: Sequence[int],
```
justinchuby, line 3544: Why can we don't use this directly? otherwise

```suggestion
    normalized_shape: Sized,
```

justinchuby:
```
  _ ERROR collecting onnxscript/test/function_libs/torch_aten/ops_correctness_test.py _
  onnxscript/test/function_libs/torch_aten/ops_correctness_test.py:19: in <module>
      from onnxscript.function_libs.torch_aten.ops import core as core_ops
  onnxscript/function_libs/torch_aten/ops/core.py:3562: in <module>
      def aten_native_layer_norm_onnx(
  onnxscript/function_libs/torch_aten/registration.py:74: in wrapper
      processed_func = onnxscript.script()(func)
  onnxscript/main.py:107: in transform
      result = script_check(ast, opset, env, src, default_opset=default_opset)
  onnxscript/main.py:53: in script_check
      return convert.top_level_stmt(f)
  onnxscript/converter.py:1352: in top_level_stmt
      fn_ir = self.translate_function_def(stmt)
  onnxscript/converter.py:1326: in translate_function_def
      type_annotation = self.eval_constant_expr(fn.returns)
  onnxscript/converter.py:410: in eval_constant_expr
      return eval(cpl, self.globals, locals)  # noqa: DUO104
  <ast>:8: in <module>
      ???
```

only in python3.8

gramalingam:
> @gramalingam does this look familiar to you?

It is evaluating a type-annotation. The only type-annotation with `[]` is `Sequence[int]` in the example. It is not clear to me why evaluating that expression should produce this error.

gramalingam:
Well, there is also `tuple[TReal, TReal, TReal]`. May be that is different in Python 3.8 and not supported? Try `Tuple`?

justinchuby:
I see. that must be it

justinchuby:
`Tuple` did it

titaiwangms (approved):
Is it because right now ONNXFunction doesn't support `is None` statement?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2282,7 +2295,14 @@ def aten_index_reduce(
 
 # FIXME(#277): Script when attributes can come before inputs
 @torch_op("aten::index_select", trace_only=True)
-def aten_index_select(self: TTensor, dim: int, index: TInt) -> TTensor:
+def aten_index_select(self: TTensor, dim: int, index: IntType) -> TTensor:
```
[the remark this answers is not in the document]
justinchuby: It swapped the arguments since we have an attribute before an input

justinchuby:
> Is it because right now ONNXFunction doesn't support `is None` statement?

More of onnxruntime doesn’t support it but we want to be able to run these ops now

BowenBao:
PR contains many new 'overload' ops. Are they part of change w.r.t the 'trace only ops' in title, or are they unrelated?

justinchuby:
They are “overload” since they registered to the same aten name. So far this doesn’t affect their use and is just for meta info generation

justinchuby:
Ah: so yes they are related and are a part of this refactoring

PR #352, opened 2023-01-20
feat(eager): move user mode adaptation to inside evaluators

justinchuby:
Move user mode adaptation to inside Evaluator to allow evaluators to override the value adaptation behavior. For example, `TorchScriptEvaluator` prefers inputs to be raw python values and handles constant creation inside its graph.


justinchuby:
@gramalingam could you take a look? thanks!

gramalingam (approved):
Looks okay for now as a way to enable the torchscript evaluator.

PR #353, opened 2023-01-20
Fix conversion between onnxscript and ort representation

gramalingam:
Handle conversion of lists between onnxscript and ort.

Fixes https://github.com/microsoft/onnx-script/issues/351

### DIFF
--- a/onnxscript/evaluator.py
+++ b/onnxscript/evaluator.py
@@ -147,7 +147,7 @@ def _os_to_ort_value(v):
     if isinstance(v, tensor.Tensor):
         return v.value
     if isinstance(v, list):
-        return v
+        return [_os_to_ort_value(x) for x in v]
     if v is None:
         # Treated as a static-optional value.
         # Dynamic optional None not yet supported.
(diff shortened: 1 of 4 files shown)


justinchuby (approved):
This is awesome, thanks! 🙌

justinchuby:
FAILED onnxscript/test/function_libs/torch_aten/ops_correctness_test.py::TestOutputConsistencyCPU::test_output_match_cat_cpu_float32

Unexpected success (which proves this fix works!)

May need to remove an xfail(“cat”) in the test file

gramalingam:
There seem to be some flakiness with the CIs, again relating to tox.

gramalingam:
Why does my local lintrunner not catch this blank space/line that CI lint is complaining about?

gramalingam:
Better still, why doesn't `linrtunner format` do it?

justinchuby:
> There seem to be some flakiness with the CIs, again relating to tox.

I think that’s an actual failure. You may expand the error message to see it

justinchuby:
Try lintrunner f —all-files or lintrunner f -m main? Lintrunner by default only operates on the file diff against HEAD

gramalingam:
well, lintrunner init seemed to have helped. Not sure when exactly that is required.

justinchuby:
there were still errors - pushing a forward fix

PR #356, opened 2023-01-21
Fix types and address comments

justinchuby:


onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -444,3 +480,7 @@ def to_model_proto(
         onnx.checker.check_model(onnx_model, full_check=True)
         print("[Success] ONNX model exported")
         return onnx_model
+
+    def apply(self, graph_pass: Callable, *args, **kwargs):
```
titaiwangms, line 484: Future usage?

titaiwangms (approved):
Please make sue the changes are tested with torch branch.

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: Yes. I updated to fx side to reflect this

PR #358, opened 2023-01-22
chore(test): move tests to the source dir

justinchuby:
Rename test/ to tests/ to contain only integration tests and common test utilities (`common/`). Module unit tests should be created next to the module source with the name `<module name>_test.py`.

This makes them easy to discover and obvious when unit tests are missing.

Added README to describe the test structure.


gramalingam:
What should we do if we need to create multiple test-files for a single module? Would it be appropriate to create a folder `module_name_tests` to group them?

gramalingam:
Just want to get the design details right. We can do the refactoring of other tests gradually or in a later PR

justinchuby:
> What should we do if we need to create multiple test-files for a single module? Would it be appropriate to create a folder `module_name_tests` to group them?

I think we should have a single file for unit tests related to the model, grouping different tests with test classes, and place integration tests in the tests folder

justinchuby:
> What should we do for test-files containing onnxscript-code, especially to avoid running lint on those code-fragments?

Current way of having a models folder and importing from it seems to work. Otherwise we can disable a few mypy lints on those test files? Are there type of lints you are more concerned about?

gramalingam:
> Current way of having a models folder and importing from it seems to work. Otherwise we can disable a few mypy lints on those test files? Are there type of lints you are more concerned about?

I was just wondering whether having a single folder for such files simplifies disabling lint for them ... what's the level of granularity for disabling such checks, can we do it at a folder level, file level (apart from disabling specific instances)?

gramalingam:
> > What should we do if we need to create multiple test-files for a single module? Would it be appropriate to create a folder `module_name_tests` to group them?
> 
> I think we should have a single file for unit tests related to the model, grouping different tests with test classes, and place integration tests in the tests folder

Ok, let me rephrase the question: would there be any technical issue preventing the use of a folder in such a case? Just wondering in case we run into a situation requiring multiple test files for a single module.

PR #361, opened 2023-01-24
feat(atenlib): conv2d

justinchuby:
Also remove `_arange_input_wrangler` because it is not needed after #355 

Co-authored-by: @fatcat-z

Replace #333

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -997,32 +997,67 @@ def aten_conv1d(
     raise NotImplementedError()
 
 
+@torch_op("aten::conv2d", trace_only=True)
 def aten_conv2d(
-    input: TensorType,
-    weight: TensorType,
-    bias: Optional[TensorType] = None,
+    input: TFloat,
+    weight: TFloat,
+    bias: Optional[TFloat] = None,
     stride: Sequence[int] = (1, 1),
     padding: Sequence[int] = (0, 0),
     dilation: Sequence[int] = (1, 1),
     groups: int = 1,
-) -> TensorType:
+) -> TFloat:
     # conv2d(Tensor input, Tensor weight, Tensor? bias=None, int[2] stride=1, int[2] padding=0, int[2] dilation=1, int groups=1) -> Tensor
 
-    raise NotImplementedError()
+    if not isinstance(padding, Sequence):
+        padding = (padding, padding)
+    pads = [*padding, *padding]
+
+    if not isinstance(dilation, Sequence):
+        dilation = (dilation, dilation)
+    dilations = list(dilation)
+
+    if not isinstance(stride, Sequence):
+        stride = (stride, stride)
+    strides = list(stride)
+
+    if bias is None:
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -997,18 +997,68 @@ def aten_conv1d(
...
+
+    result = aten_conv2d_onnx(
+        input, weight, bias, strides=strides, pads=pads, dilations=dilations, groups=groups
+    )
+
+    return result
+
+
+@torch_op("aten::conv2d", overload=True)
+def aten_conv2d_onnx(
```
titaiwangms, line 1039: So if I am building graph from scratch, I should use `aten_conv2d_onnx` instead?

titaiwangms (approved):

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: Should we rename aten_conv2d_onnx to _aten_conv2d_onnx to avoid the confusion like Titai had?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -997,18 +997,68 @@ def aten_conv1d(
...
+
+    if not isinstance(dilation, Sequence):
+        dilation = (dilation, dilation)
+    dilations = list(dilation)
+
+    if not isinstance(stride, Sequence):
+        stride = (stride, stride)
+    strides = list(stride)
+
+    if bias is None:
```
fatcat-z, line 1025: Why do we need to prepare a default value for bias if it's None? Will these operations be more efficient than adding the bias to result in the end?

justinchuby: That's because conv has a bias input slot we should make use of. In the future we can use the optional feature within functions

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: SG

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -997,18 +997,68 @@ def aten_conv1d(
...
+        bias = op.Expand(zero, bias_shape)
+
+    result = aten_conv2d_onnx(
+        input, weight, bias, strides=strides, pads=pads, dilations=dilations, groups=groups
+    )
+
+    return result
+
+
+@torch_op("aten::conv2d", overload=True)
```
BowenBao, line 1038: Is it necessary to register with `torch_op` if user should not directly use `aten_conv2d_onnx`?

justinchuby: Good point. Prob not

justinchuby: I changed my mind. torch_op also sets the proper domain etc for the functions. For now I would like to go with torch_op and figure out a better way to categorize these functions soon

onnxscript/function_libs/torch_aten/ops/core.py

PR #362, opened 2023-01-24
chore(lint): improve with refurb

justinchuby:
Use https://github.com/dosisod/refurb to modernize and simplify code


justinchuby:
```
  E   ValueError: ERROR
  E   thenGraph_7:12 ...        inner = op.Mul(op.Sqrt(op.Div(2.0, math.pi)), inner)
  E       Unsupported expression type <class 'ast.Attribute'>.
```

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -335,7 +336,7 @@ def aten_gelu(self: TReal, approximate: str = "none") -> TReal:
         cubed = op.Pow(self, 3)
         inner = op.Mul(0.044715, cubed)
         inner = op.Add(self, inner)
-        inner = op.Mul(op.Sqrt(op.Div(2.0, 3.141592653589793)), inner)
+        inner = op.Mul(op.Sqrt(op.Div(2.0, math.pi)), inner)
```
justinchuby, line 339: This should be a python expression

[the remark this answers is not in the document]
gramalingam: For now, I suggest the following, which is supported by onnxscript : define any such constant outside the script using a symbolic name (whether it is `pi` or `sqrt_2_over_pi`) and use the symbolic name inside the script.

Using the expressions inside the script introduces some design questions ("When/what expressions should we evaluate at script-time"), so would suggest that we consider them in the future.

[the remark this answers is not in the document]
gramalingam: Eg, see: https://github.com/microsoft/onnx-script/blob/main/onnxscript/tests/functions/ort_custom_ops.py#L15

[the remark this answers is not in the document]
justinchuby: Sounds good! Done.

onnxscript/irbuilder.py
```diff
@@ -171,7 +171,7 @@ def to_node_proto(self, node_name: str) -> onnx.NodeProto:
     @property
     def output_names(self) -> Sequence[str]:
         """Returns the list of variables assigned to by this statement."""
-        return [str(x) for x in self.result]
+        return list(self.result)
```
gramalingam, line 174: Same as above.

onnxscript/irbuilder.py
```diff
@@ -160,7 +160,7 @@ def to_node_proto(self, node_name: str) -> onnx.NodeProto:
         n = helper.make_node(
             self.callee.opname,
             [_opt_var_to_str(x) for x in self.args],
-            [str(x) for x in self.result],
+            list(self.result),
```
gramalingam, line 163: Why is this correct? It seems to rely on the type of `self.result` and may fail if that is changed from string to something else (like an IRVar), right?

justinchuby: I think self.result was annotated as Sequence[str], should we update the annotation?

gramalingam: My point was that this change is not robust. I don't think there is a good reason to make this change even if the annotation is correct today.

justinchuby: OK. I will change this back

justinchuby: Looking at this again, I still think we should maintain the invariant at the setter/initializer if we annotate a field as a certain type. This way users can assume the invariant and not have to worry about the actual type being different than what's claimed.

gramalingam: Okay. Feel free to update the annotation to indicate that it is string-like: something that can be converted to a string via "str(.)".

But if that annotation is "Any", that is probably not a great annotation.

I think the normal subtyping safety principle is: callers should absolutely respect the annotation on parameters, but if the called code handles a superset of the annotated-type, that is perfectly fine and safe.

justinchuby: I recall all Python objects have __str__ defined so that basically equates to Any

gramalingam: Also: I believe that it is unsafe for `refurb` to do such a transformation, in general, because python does not *require* that type-annotations be satisfied, so the transformation can change the behavior of python-code, which is generally not a good choice for such tools. Any "cleanup" tool should play it safe and avoid transformations that might change the behavior.

justinchuby: I think it makes sense to annotate as any if that’s the most accurate type. That said I will revert this change for now!

justinchuby: I did the change because I saw the annotation. Refurb didn’t do it :D

justinchuby: done

onnxscript/irbuilder.py
justinchuby: Reverted

gramalingam (approved):

PR #366, opened 2023-01-24
chore(codeql): disable py/import-and-import-from

justinchuby:
Disable the `py/import-and-import-from` error. E.g. https://github.com/microsoft/onnx-script/security/code-scanning/6491

cc @jcwchen

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,3 @@
+query-filters:
+- exclude:
+    id: py/import-and-import-from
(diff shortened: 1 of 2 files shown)


justinchuby:
It should look like

```
query-filters:
# This should run js/path-injection and js/zipslip
- include:
    tags contain: external/cwe/cwe-022

# Removes js/path-injection
- exclude:
    id: js/path-injection
```

but what should the config file be?

justinchuby:
this works

jcwchen (approved):
Thanks for figuring this out 🚀

titaiwangms (approved):

Issue #320, opened 2023-01-14
Unify interface for accessing custom function and op schemas


justinchuby:
Provide a common interface for converters to construct nodes based on custom OnnxFunctions and Ops.

justinchuby:
This is done with #594

PR #368, opened 2023-01-25
feat(atenlib): separate inputs from attributes

justinchuby:
Create `param_manipulation` for separate inputs and attributes from python args and kwargs.
Raname `typing.py` to `onnxscript/function_libs/torch_aten/tensor_typing.py`

- #305 
- #287
- #320


onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,46 @@
...
+        """Check if the parameter is an ONNX attribute."""
+        return not self.is_input
+
+
+def separate_input_attributes_from_arguments(
+    param_schemas: Sequence[ParamSchema],
+    args,
+    kwargs,
+) -> Tuple[List[str, Any], OrderedDict[str, Any]]:
+    pass
```
justinchuby, line 46: TODO: Implement

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,46 @@
...
+import collections
+import dataclasses
+from typing import Any, List, OrderedDict, Sequence, Tuple
+
+# A special value to indicate that the default value is not specified
+EmptyDefault = object()
+
+
+@dataclasses.dataclass(frozen=True)
+class ParamSchema:
```
justinchuby, line 13: TODO: create two functions that takes from OnnxFunction or Op and produce a list of `ParamSchema`s

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,58 @@
...
+        param_schemas: The parameter schemas of an Op or a OnnxFunction.
+        args: The Python positional arguments supplied by the caller.
+        kwargs: The Python keyword arguments supplied by the caller.
+
+    Returns:
+        A tuple of two elements:
+        - A list of ONNX inputs.
+        - An ordered dictionary of ONNX attribute names and values.
+    """
+    raise NotImplementedError()
```
justinchuby, line 58: TODO: Implement

titaiwangms (approved):

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,171 @@
...
+        default: The default value of the parameter.
+        is_input: Whether the parameter is an ONNX input.
+    """
+
+    name: str
+    type: Any = None  # Op input does not have a type, for now
+    default: Any = _EmptyDefault
+    is_input: bool = True
+
+    def __repr__(self) -> str:
```
gramalingam, line 32: Is the return value supposed to be valid python or just a description?

gramalingam: Just wondering whether __str__ is more appropriate or __repr__ ?

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,171 @@
...
+        param_kind = "INPUT" if self.is_input else "ATTRIBUTE"
+        text = f"{self.name}<{param_kind}>: {self.type}"
```
gramalingam, line 34: nit (optional): f"{self.name} : {param_kind}({self.type})" or f"{param_kind} {self.name} : {self.type}" would be more readable?

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,171 @@
...
+_EmptyDefault = object()
+
+
+@dataclasses.dataclass(frozen=True)
+class ParamSchema:
+    """A schema for a parameter of an Op or a OnnxFunction.
+
+    Attributes:
+        name: The name of the parameter.
+        type: The type of the parameter.
```
gramalingam, line 22: Since we have multiple type representations, which one is this supposed to be? Eg., is this supposed to be python-types (like "int" and "List[FLOAT]") ?

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,171 @@
...
+            text += f" = {self.default}"
+        return text
+
+    @property
+    def is_attribute(self) -> bool:
+        """Returns True if the parameter is an ONNX attribute."""
+        return not self.is_input
+
+
+def extract_param_schema_from_function(onnx_func: values.OnnxFunction) -> List[ParamSchema]:
```
gramalingam, line 45: The current `OnnxFunction` IR loses the positional-order information between inputs and attributes (since we didn't care about it then given our assumptions that all inputs come first). We can't reconstruct it subsequently.

Perhaps we should consider updating `OnnxFunction` IR to store a `List[ParamSchema]` in the first place ... just thinking out aloud, haven't read the rest of the PR yet.

onnxscript/function_libs/torch_aten/param_manipulation.py
```diff
@@ -0,0 +1,171 @@
...
+            else:
+                onnx_attributes[param.name] = args[i]
+        elif param.name in kwargs:
+            if not param.is_attribute:
+                onnx_inputs.append(kwargs[param.name])
+            else:
+                onnx_attributes[param.name] = kwargs[param.name]
+        else:
+            # input doesn't have default
+            onnx_attributes[param.name] = param.default
```
gramalingam, line 169: I haven't seen the usage of this function. Perhaps some usage contexts would not want default-values to be filled in. Anyway, I guess we can cross that bridge when we come to it.

onnxscript/function_libs/torch_aten/param_manipulation.py
justinchuby: I like this idea! I was going to propose exposing ParamSchema somewhere in the function and op as the next thing to do so we have a common interface as #320

onnxscript/function_libs/torch_aten/param_manipulation.py
justinchuby: SG!

onnxscript/function_libs/torch_aten/param_manipulation.py
justinchuby: Good catch, thanks!

onnxscript/function_libs/torch_aten/param_manipulation.py
justinchuby: a<INTPUT>: INT64

vs

INPUT a : INT64
a : INTPUT(INT64)

justinchuby: I will go with `a: INTPUT[INT64]`

onnxscript/function_libs/torch_aten/param_manipulation.py
justinchuby: I added a fill_defaults option to the function

PR #370, opened 2023-01-25
Update documentation about Scan op and graph-attributes

gramalingam:
Move the documentation about use of nested-functions to represent graph-attributes in the tutorial.


justinchuby (approved):
Much needed for https://github.com/microsoft/onnx-script/blob/0a046645bf62176cc2fc7a3865c127e9831df483/onnxscript/function_libs/torch_aten/ops/core.py#L2690-L2701. Thanks!

gramalingam:
> Much needed for

I don't understand. This example doesn't use Scan or any such operation. Are you considering defining a lower-level definition of this op than the existing one? Standard if-then-else doesn't require use of nested-functions.

gramalingam:
Okay, I understand ... you are referring to defining ops like CumSum.

justinchuby:
Currently op.Log(op.CumSum(op.Exp(self), dim)) easily overflows so I was thinking I needed to use Scan to implement https://www.tensorflow.org/api_docs/python/tf/math/cumulative_logsumexp

PR #382, opened 2023-01-27
chore(evaluator): add and fix typing annotations

justinchuby:
* #383
* __->__ #382


onnxscript/evaluator.py
```diff
@@ -418,8 +442,3 @@ def default_as(temp_default: Evaluator):
         yield
     finally:
         set_default(old_default)
-
-
-def eval(schema, inputs, attributes):
```
justinchuby, line 423: I removed this because I thought it duplicates with `evaluator.default().eval`. LMK if this `eval` is intended to be a public api?

titaiwangms: cc @gramalingam

justinchuby:
@gramalingam could you take a look? Thanks!

onnxscript/evaluator.py
```diff
@@ -115,59 +113,74 @@ class Evaluator(abc.ABC):
     supported by onnxscript to those expected by a particular backend.
     """
 
-    def eval(self, schema, inputs, attributes):
-        closure = self.adapt_attributes(schema, attributes)
+    def eval(
+        self, schema: onnx.defs.OpSchema, inputs: Sequence[Any], attributes: Mapping[str, Any]
```
gramalingam, line 117: Is it clear that "Any" here means that we have not yet bothered to more precisely define the allowed types of values? As, opposed to asserting that any type of values can be passed? 

Concretely, would it make sense to introduce a type-alias that captures the intention, like
```py
InputValueType  = Any  # TODO: elaborate this.
```
and use it?

onnxscript/evaluator.py
```diff
@@ -115,59 +113,74 @@ class Evaluator(abc.ABC):
...
+    def adapt_attributes(
+        self, schema: onnx.defs.OpSchema, attributes: Mapping[str, Any]
+    ) -> tuple[dict[str, Any], dict[str, Any]]:
         """Transform attributes (in-place) to the expected format for the evaluator.
 
         Returns a closure that can be used to evaluate graph-valued attributes.
         """
         use_graph_attribute = self.use_graph_attribute(schema)
         closure = {}
+        adapted_attributes = {}
```
gramalingam, line 141: May be worth checking if the caller uses the returned value or passed-in parameter. If it is correct, we should update the above comment in line 135 which says attributes are updated in place as well.

onnxscript/evaluator.py
gramalingam: Looks fine either way.

onnxscript/evaluator.py
justinchuby: Sounds good!

justinchuby: I updated to use ExtendedModeValue

onnxscript/evaluator.py
justinchuby: Updated comment to remove `in-place`

gramalingam (approved):

PR #392, opened 2023-01-28
feat(docs): use the furo theme

### LINKED ISSUE #391: ghpage: use the Furo theme
https://pradyunsg.me/furo/. It’s pretty and more modern. Used by black https://black.readthedocs.io/en/stable/#

justinchuby:
Use the furo theme (which is also recommended by the pydata theme site)

## snapshots 

Features

- More compact layout for function signatures and improved color palette
- Easy to navigate side bar 
- Copy code block button
- Sphinx 6.0 support

| Before | After |
| - | - |
| image | image |
| image | image |

Fixes #391


justinchuby:
After

image


Before

image

docs/conf.py
```diff
@@ -44,13 +45,12 @@
 # -- Options for HTML output -------------------------------------------------
 
 html_static_path = ["_static"]
-html_theme = "pydata_sphinx_theme"
+html_theme = "furo"
```
xadupre, line 48: That would be great if we would have the same theme accross all the packages related to onnx (onnx, onnxruntime, onnx-script, ...).

justinchuby: Is furo a theme we like? If so I can replicate it in onnx. ORT so far is using the default theme when I checked. Or is there a desire to create our own theme like pytorch and tf did?

gramalingam:
LGTM ... I am fine with it.

abock (approved):

PR #395, opened 2023-01-31
chore(test): use nox instead of tox to run tests

justinchuby:
* #396
* __->__ #395

`nox` provides a more flexible and accurate way of installing dependencies than `tox` and is used by [major python official packages](https://nox.thea.codes/en/stable/index.html#projects-that-use-nox)

cc @jcwchen


titaiwangms (approved):

justinchuby:
--append --cov-branch

PR #403, opened 2023-02-02
fix(atenlib): remove **kwargs in aten_embedding.

### LINKED ISSUE #387: Remove **kwargs in aten_embedding
Not supported by onnx script. We should add all positional args in the signature.

fatcat-z:
**kwargs is not supported by ONNX Script yet, so restore the signatures.

Fix #387 .


justinchuby:
I don’t know if max norm is part of the aten spec. If not we can filter them out in the test. I would check native functions yaml

fatcat-z:
> I don’t know if max norm is part of the aten spec. If not we can filter them out in the test. I would check native functions yaml

Accoding to the definition [here](https://pytorch.org/docs/master/generated/torch.nn.Embedding.html), PyTorch embedding function should accept these parameters. Will update the signatures.

justinchuby:
We don't really care about the torch function. Only the aten signature is meaningful to us

fatcat-z:
> We don't really care about the torch function. Only the aten signature is meaningful to us, aka. native functions yaml

According to [here](https://github.com/pytorch/pytorch/blob/d996acfbc2e7536b0438eb491628a328bdf59f05/aten/src/ATen/native/native_functions.yaml#L2141), max_norm does not exist in the signature.

No idea why it doesn't exist in navtive_functions.yaml but provided in sample input of test. In this case, adding this parameter doesn't hurt and it helps reducing the customizations for our tests. 

Thoughts?

justinchuby:
I think we should adhere to the aten signature and tweak the test inputs to fit it. That is because the tests are defined for the `torch.*` ops and we are simply borrowing them for testing aten. In cases when the signatures of the two differ like we see here, aten's signature is usually simplified for a good reason. Adhering to it reduces confusion. 

I would add a comment in the test wrangler to explain why it is needed

justinchuby:
also if we added those arguments, users would expect them to work when they actually don't.

fatcat-z:
> also if we added those arguments, users would expect them to work when they actually don't.

But for embedding() method, there are already some arguments will cause such impression already, so I think for this case, adding max_norm and norm_type into the signature doesn't hurt.

justinchuby:
I still don’t think that’s a good idea

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -179,6 +179,16 @@ def _cat_input_wrangler(
     return args, kwargs
 
 
+def _embedding_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    if "max_norm" in kwargs:
```
justinchuby, line 185:
```suggestion
) -> tuple[list[Any], dict[str, Any]]:
    """Remove arguments not present in the aten op signature."""
    if "max_norm" in kwargs:
```

justinchuby (approved):
LGTM!

PR #404, opened 2023-02-03
feat(eager): generalized logic for splitting inputs and attributes

### LINKED ISSUE #287: Inputs and attributes could not be separated by _adapt_to_eager_mode method.

justinchuby:
Generalize #368 to all eager backends

split happens in the evaluator instead of the Op or OnnxFunction call methods so that evaluators can control behaviors like whether to fill in default values for attributes.

Fixes #287, partially #305 (still need the scripted version)


titaiwangms (approved):

onnxscript/values.py
```diff
@@ -104,49 +176,54 @@ def __init__(
...
-        return new_kwargs
-
-    def __call__(self, *args, **kwargs):
-        # FIXME(after #225): Move import to the top of the file.
-        from onnxscript import evaluator  # pylint: disable=import-outside-toplevel
-
-        return evaluator.default().eval(
-            self.opschema, args, self._convert_kwargs_to_numpy(kwargs)
-        )
+    def param_schemas(self) -> tuple[ParamSchema, ...]:
```
gramalingam, line 200: May be introducing a type-alias `Schema = tuple[ParamSchema, ...]` would be helpful.

justinchuby: Would it be confused with OpSchema? I was thinking if there's a more accurate name... Any suggestions?

onnxscript/values.py
```diff
@@ -104,49 +176,54 @@ def __init__(
...
+        """Returns the parameter schemas for this op."""
+        if self._param_schemas is not None:
+            return self._param_schemas
+
+        op_schema = self.get_schema()
+        schemas = []
+        for input_ in op_schema.inputs:
+            param_schema = ParamSchema(name=input_.name, is_input=True)
+            if input_.option == onnx.defs.OpSchema.FormalParameterOption.Optional:
```
gramalingam, line 209: Nit: perhaps
```py
param_schema = ParamSchema(name=input_.name, is_input=True,
   required = (input_.option != onnx.defs.OpSchema.FormalParameterOption.Optional),
   is_variadic = (input_.option == onnx.defs.OpSchema.FormalParameterOption.Variadic))
```

justinchuby: thanks! done

onnxscript/evaluator.py
```diff
@@ -116,10 +128,25 @@ class Evaluator(abc.ABC):
...
     ):
+        """Evaluates an ONNX op.
+
+        Args:
+            op: The op to evaluate.
+            args: The positional arguments to the op.
+            kwargs: The keyword arguments to the op.
+        """
+        param_schemas = op.param_schemas()
+        # Split happens in the evaluator instead of the Op __call__ method
```
gramalingam, line 143: It just struck me that perhaps we are likely getting this for free already, because of the current generation of python functions from the operator schema (at least for ONNX ops)? I mean the generated code such as https://github.com/microsoft/onnx-script/blob/9624c4e177351fe49243c1106532dc79a0e66301/onnxscript/onnx_opset/_impl/opset18.py#L143

justinchuby: I think eager calls follow a different route?

justinchuby: Aka we start from an Op object: `op(*self._prepare_inputs(schema, X))`?

onnxscript/values.py
gramalingam: I was thinking of that too. One possibility is a name like `IRSchema` since this is really part of the IR we are using (which has counterparts to many things in ONNX).

onnxscript/evaluator.py
justinchuby: Ohh I may be seeing what you are saying

justinchuby: In that case we will only care about functions and not built in ops?

justinchuby: Done. Removed the call in `eval()` and resumed its signature. Added eager tests.

justinchuby:
@gramalingam I am merging this now. Please feel free to continue to add comments if there are and I will update in another PR.

PR #408, opened 2023-02-04
chore(lint): replace flake8 with ruff

justinchuby:
Use ruff https://github.com/charliermarsh/ruff to lint files, replacing flake8 and its plugins. It is faster and has auto-fix.

Also enforce PEP8-naming


jcwchen (approved):
Thanks! If it goes well here, we can probably enable it to replace flake8 in onnx/onnx repo as well.

PR #409, opened 2023-02-06
Fix _attribute_value type handling | fix(onnx_export)

abock:
It appears `AttributeProto.HasField` is returning returning false in some cases. Change to always return the value through the appropriate field based on `AttributeProto.type`.

### Repro

```python
from onnxscript import script
from onnxscript.onnx_opset import opset15 as op
from onnxscript.onnx_types import FLOAT
from onnxscript.backend.onnx_export import export2python

@script()
def CustomArgMax(X: FLOAT, int_attr: int = 0) -> FLOAT:
    return op.ArgMax(X, axis=int_attr)

print(export2python(CustomArgMax.to_model_proto()))
```

### Exception

```
...

  File "/Users/aaron/src/onnx-script/onnxscript/backend/onnx_export.py", line 172, in _attribute_value
    raise NotImplementedError(f"Unable to return a value for attribute {attr!r}.")
NotImplementedError: Unable to return a value for attribute name: "axis"
type: INT
ref_attr_name: "int_attr"
.
```


onnxscript/backend/onnx_export.py
```diff
@@ -150,25 +150,34 @@ def _to_str(s):
     return s
 
 
-def _attribute_value(attr):
-    if attr.HasField("f"):
+def _attribute_value(attr: onnx.AttributeProto):
```
justinchuby, line 153: Possible to reuse onnx.helper.get_attribute_value https://onnx.ai/onnx/api/helper.html#id2?

gramalingam: Yes, that sounds good

PR #412, opened 2023-02-07
Extend documentation for optional inputs support

shubhambhokare1:
Add documentation for the optional inputs checking as per https://github.com/microsoft/onnx-script/pull/91


docs/tutorial/examples/clip.py
```diff
@@ -0,0 +1,14 @@
+from onnxscript import opset15 as op
+from onnxscript import script
+
+
+@script()
+def Clip(input, min, max):
```
liqunfu, line 6: Please add test to cover this function. Or just enable the test at https://github.com/microsoft/onnx-script/blob/b2fd55bd7d8a22383c15127ffaeb074756ac3563/onnxscript/tests/functions/onnxfns1A_test.py#L75

docs/tutorial/examples/clip.py
```diff
@@ -0,0 +1,14 @@
...
+    # The inputs min and max are optional inputs
+    result = input
+    # OptionalHasElement is used to check if an optional input is not None
+    if op.OptionalHasElement(min):
+        result = op.Where(result < min, min, result)
```
justinchuby, line 11: Does it mean we don’t need to use optional get element?

docs/tutorial/examples/clip.py
```diff
@@ -0,0 +1,14 @@
+from onnxscript import opset15 as op
```
gramalingam, line 1: I think you need to use opset18 for the extended Optional ops.

docs/tutorial/examples/clip.py
gramalingam: I do think we need to use OptionalGetElement. Otherwise, the checker will likely complain.

docs/tutorial/index.rst
```diff
@@ -36,6 +36,18 @@ omitted in the case of trailing inputs:
...
+
+The example below shows a definition of ``Clip`` as an *onnxscript* function.
+The ``Clip`` operator has input arguments that are *optional*: the *min* and *max* inputs.
+``OptionalHasElement`` and ``OptionalGetElement`` operators can be used to utilize optional
+inputs within an *onnxscript* function, as shown below:
+
+.. literalinclude:: examples/clip.py
+
+This pattern will not work in eager-mode execution as
+``OptionalHasElement`` and ``OptionalGetElement`` operators do not take None as inputs.
```
gramalingam, line 49: In opset 18 they do. But there is likely a different issue, which is to do with the two kinds of optional inputs we have in ONNX. Optional inputs of the first kind are supported only by ops/nodes, not by graphs. we may need some extra code in onnxscript when we turn a single-op invocation into a graph/model

gramalingam: It may be simpler to provide a python implementation in the ORTMixedEvaluator, like [here](https://github.com/microsoft/onnx-script/blob/main/onnxscript/evaluator.py#L450) and use it

gramalingam: I took a look at the onnxscript code, and it does seem to do what's necessary for optional inputs (so ignore my comment above). It may be worth trying using opset18 and see if it works.

PR #413, opened 2023-02-07
integrate with fake_gpt.py

titaiwangms:
[fake_gpt.py](https://github.com/pytorch/pytorch/blob/onnx-team/dynamo-exporter/torch/onnx/_internal/docs/fake_gpt.py) might add None as input for graph builder. Further looking into it, it seems just a bug, as we already defined a if branch for None value.

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -306,7 +306,7 @@ class TorchScriptGraph:
 
     @beartype
     def add_input(
-        self, input_name: str, input_value: Union[torch.Tensor, None]
+        self, input_name: str, input_value: Optional[torch.Tensor] = None
     ) -> TorchScriptTensor:
         # TODO: Take in a TorchScriptTensor?
         # TODO: Support dynamic shapes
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -305,7 +305,9 @@ def torch_graph(self):
         return self._torch_graph
 
     @beartype
-    def add_input(self, input_name: str, input_value: torch.Tensor) -> TorchScriptTensor:
+    def add_input(
+        self, input_name: str, input_value: Union[torch.Tensor, None]
```
justinchuby, line 309: Optional[]

justinchuby (approved):

BowenBao (approved):

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -305,7 +305,9 @@ def torch_graph(self):
         return self._torch_graph
 
     @beartype
-    def add_input(self, input_name: str, input_value: torch.Tensor) -> TorchScriptTensor:
+    def add_input(
+        self, input_name: str, input_value: Optional[torch.Tensor] = None
```
justinchuby, line 309: We prob always want users to specify the input value. So I would remove the ` = None` part.

titaiwangms: I will put this in https://github.com/microsoft/onnx-script/pull/416

Issue #388, opened 2023-01-28
Function authoring guide for ATen lib 


justinchuby:
- How to find the signature
- Guide on typing 
    - broadest supported type by the ONNX ops used in the function
- How to test
- #402 
- #412 
- #407 
- #403 
- #400
- #375 
- #293 
- #350 
- #359 
- #411
- Scan
- Avoid trace only. Always add a reason; split the trace only part and make decomposition logic to a function; when to use trace only: dtype attribute, attribute manipulation; trace only requires a static graph.
- Private functions for ONNX
- #419
- https://github.com/microsoft/onnx-script/pull/416
- https://github.com/microsoft/onnx-script/pull/301
- how we handle scalars
- https://github.com/microsoft/onnx-script/pull/428 tricky boolean attributes
- Don’t raise errors #454 
- How to handle ATen overloads; Testing: verify the OpInfo coverage https://github.com/microsoft/onnx-script/pull/463#discussion_r1125165927

BowenBao:
topic suggestion: mutation/aliasing

justinchuby:
the number (0.0) being scripted as constant + castlike https://github.com/microsoft/onnx-script/pull/515#discussion_r1132733650

PR #416, opened 2023-02-08
[atenlib] Integrate symbolic bloom export

titaiwangms:
Add ops

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -138,11 +138,11 @@ def aten_affine_grid_generator_backward(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::alias", trace_only=True)
 def aten_alias(self: TensorType) -> TensorType:
     # alias(Tensor(a) self) -> Tensor(a)
 
-    raise NotImplementedError()
+    return self
 
 
 def aten_alias_copy(self: TensorType) -> TensorType:
@@ -526,7 +526,7 @@ def aten_avg_pool1d(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::baddbmm")
 def aten_baddbmm(
     self: TensorType,
     batch1: TensorType,
@@ -535,8 +535,12 @@ def aten_baddbmm(
     alpha: float = 1.0,
 ) -> TensorType:
     # baddbmm(Tensor self, Tensor batch1, Tensor batch2, *, Scalar beta=1, Scalar alpha=1) -> Tensor
-
-    raise NotImplementedError()
+    batch_mul = op.MatMul(batch1, batch2)
+    alpha_cast = op.CastLike(alpha, self)
+    mul_a = op.Mul(batch_mul, alpha_cast)
+    beta_cast = op.CastLike(beta, self)
+    mul_b = op.Mul(self, beta_cast)
+    return op.Add(mul_a, mul_b)
 
 
 def aten_bartlett_window(window_length: int) -> TensorType:
@@ -1477,11 +1481,14 @@ def aten_cumprod_backward(
 
     raise NotImplementedError()
 
-
-def aten_cumsum(self: TensorType, dim: int, dtype: Optional[int] = None) -> TensorType:
+@torch_op("aten::cumsum")
+def aten_cumsum(self: TensorType, dim: int, dtype: int = -1) -> TensorType:
     # cumsum(Tensor self, int dim, *, ScalarType? dtype=None) -> Tensor
-
-    raise NotImplementedError()
+    if dtype == -1:
+        cast = self
+    else:
+        cast = op.Cast(self, to=dtype)
+    return op.CumSum(cast, dim)
 
 
 def aten_data(self: TensorType) -> TensorType:
@@ -2935,11 +2942,12 @@ def aten_margin_ranking_loss(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::masked_fill", trace_only=True)
 def aten_masked_fill(self: TensorType, mask: TensorType, value: TensorType) -> TensorType:
     # masked_fill.Tensor(Tensor self, Tensor mask, Tensor value) -> Tensor
-
-    raise NotImplementedError()
+    mask_cast = op.Cast(mask, to=BOOL.dtype)
+    value_cast = op.CastLike(value, self)
+    return op.Where(mask_cast, value_cast, self)
 
 
 def aten_masked_scatter(self: TensorType, mask: TensorType, source: TensorType) -> TensorType:
@@ -4523,6 +4531,13 @@ def aten_segment_reduce(
 
     raise NotImplementedError()
 
+@torch_op("aten::select")
+def aten_select(
+    self: TensorType, dim: int, index: int
+) -> TensorType:
+    # select(Tensor self, int dim, int index) -> Tensor
+
+    return op.Gather(self, index, axis=dim)
 
 def aten_select_backward(
     grad_output: TensorType, input_sizes: INT64, dim: int, index: int
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -306,7 +306,7 @@ class TorchScriptGraph:
 
     @beartype
     def add_input(
-        self, input_name: str, input_value: Union[torch.Tensor, None]
+        self, input_name: str, input_value: Optional[torch.Tensor] = None
     ) -> TorchScriptTensor:
         # TODO: Take in a TorchScriptTensor?
         # TODO: Support dynamic shapes
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -305,7 +305,9 @@ class TorchScriptGraph:
         return self._torch_graph
 
     @beartype
-    def add_input(self, input_name: str, input_value: torch.Tensor) -> TorchScriptTensor:
+    def add_input(
+        self, input_name: str, input_value: Union[torch.Tensor, None]
+    ) -> TorchScriptTensor:
         # TODO: Take in a TorchScriptTensor?
         # TODO: Support dynamic shapes
         if input_value is None:


justinchuby:
I will do another review later today.

BowenBao (commented):
Add unittests?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -138,11 +138,11 @@ def aten_affine_grid_generator_backward(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::alias", trace_only=True)
```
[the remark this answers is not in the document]
BowenBao: cc @justinchuby

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4523,6 +4531,13 @@ def aten_segment_reduce(
 
     raise NotImplementedError()
 
+@torch_op("aten::select")
+def aten_select(
```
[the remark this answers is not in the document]
titaiwangms: I actually had no idea of writing op here. I am trying to write down necessary ops and wait @justinchuby to the rescue.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2935,11 +2942,12 @@ def aten_margin_ranking_loss(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::masked_fill", trace_only=True)
```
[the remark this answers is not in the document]
titaiwangms: There is a bug if I tried to make it an onnx function. Looking into this.

onnxscript/function_libs/torch_aten/ops/core.py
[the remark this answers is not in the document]
titaiwangms: Would love to know more about the mutation with symbolic_fn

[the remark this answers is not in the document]
justinchuby: We don’t have mutations in onnx and assumes fx handle them all. Identity sounds like the right thing to do.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -95,10 +95,13 @@ def aten_addmm(
 ) -> TFloat:
     # addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor
 
+    # TODO(titaiwang): op.Gemm seems needed to take care of corner case according to old symbolic_fn.
+    # Currently, it shows op-level validation failing on bloom.
```
justinchuby, line 99: What were the cases? Would replacing the body with Gemm make more sense?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -526,7 +529,7 @@ def aten_avg_pool1d(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::baddbmm")
 def aten_baddbmm(
     self: TensorType,
```
justinchuby, line 534:
```suggestion
    self: TFloat,
```

Does TFloat seem to be an appropriate type for this? We want a type that is the broadest supported type by the ONNX ops used in the function. You can refer to tensor_typing.py to find or create a type.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2935,11 +2945,12 @@ def aten_margin_ranking_loss(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::masked_fill")
 def aten_masked_fill(self: TensorType, mask: TensorType, value: TensorType) -> TensorType:
```
justinchuby, line 2949: I would update the types like above here and for all new functions. LMK if you needed more info.

justinchuby (commented):
You may use https://github.com/microsoft/onnx-script/pull/301/files#diff-db423ee0dcd2d4a7c3dddb5141d8b3e2b4cc3f60b32354e36ec14ac3f61dbc01 as an example for enabling tests.

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: I see. I will try to replace annotations

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: In aten_mul, why don't we care about uint8?
https://github.com/microsoft/onnx-script/blob/78ea55b888de88bfdadce7c3f6f3f83fa1404c7f/onnxscript/function_libs/torch_aten/ops/core.py#L3440
https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Mul-14

justinchuby: We probably should. It must be a mistake. Since we are only testing with float inputs rn I expect we miss corner cases here and there. #417

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: The idea comes from https://github.com/pytorch/pytorch/pull/46570/files#diff-6616e68665d29ba8ab30d14fc99a8459f974fb58368d71daf7bca89701e557db. But it seems that even though we don't use Gemm, we should still be fine. THe op-level validation is failing with mismatch rn.

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: > Does TFloat seem to be an appropriate type for this? We want a type that is the broadest supported type by the ONNX ops used in the function. You can refer to tensor_typing.py to find or create a type.

I think you mean the intersection of supported types across all used ONNX ops in that function?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -425,6 +425,11 @@ def _where_input_wrangler(
         matcher=lambda sample: sample.input[0].equal(torch.tensor([])),
         reason="cat does not support zero-dim tensors yet",
     ),
+    # skip(
```
justinchuby, line 428: You may refer to https://github.com/microsoft/onnx-script/blob/72da830c0e85d8afa1c94f1af8b7c4876090b38b/onnxscript/function_libs/torch_aten/ops/core.py#L2841 to see how we support scalar (tldr check rank and return)

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1477,11 +1481,14 @@ def aten_cumprod_backward(
...
     raise NotImplementedError()
 
-
-def aten_cumsum(self: TensorType, dim: int, dtype: Optional[int] = None) -> TensorType:
+@torch_op("aten::cumsum")
+def aten_cumsum(self: TensorType, dim: int, dtype: int = -1) -> TensorType:
     # cumsum(Tensor self, int dim, *, ScalarType? dtype=None) -> Tensor
-
-    raise NotImplementedError()
+    if dtype == -1:
```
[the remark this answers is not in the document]
justinchuby: I would extract the dtype logic out, mark as trace_only, then wrap only the scalar handling and cumsum part into _aten_cumsum_onnx as a function

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1487,7 +1487,7 @@ def aten_cumprod_backward(
     raise NotImplementedError()
 
 @torch_op("aten::cumsum", trace_only=True)
-def aten_cumsum(self: TRealUnlessInt16OrInt8, dim: int, dtype: int = -1) -> TRealUnlessInt16OrInt8:
+def aten_cumsum(self: TRealUnlessInt16OrInt8, dim: Union[INT32, INT64], dtype: int = -1) -> TRealUnlessInt16OrInt8:
```
justinchuby, line 1490: Can dim be INT32 (supported by cumsum)? Do we need (a) an explicit cast to INT64, or (b) just assume it’s INt64 and let caller handle type conversion?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: You said it better. That’s the idea. I will put it in #388

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
titaiwangms: Nice

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Ahh, I get it now. Making a subfunction covering as much as content we could from the origin enables a relative compact function layer still, instead of tracing the whole aten function? Which means the curren solution for dtype if branch would be also like this?

titaiwangms: > Can dim be INT32 (supported by cumsum)? Do we need (a) an explicit cast to INT64, or (b) just assume it’s INt64 and let caller handle type conversion?

Get this from https://github.com/onnx/onnx/blob/main/docs/Changelog.md#CumSum-14

T2 accepts both INT32 and INT64 though

justinchuby: > Ahh, I get it now. Making a subfunction covering as much as content we could from the origin enables a relative compact function layer still, instead of tracing the whole aten function? Which means the curren solution for dtype if branch would be also like this?

Precisely! What do you think?

titaiwangms: I think it's great for a temp solution, as at the end of the day, it's a onnx-script issue to support these pywriting anyway. cc @BowenBao This is the discussion about how to deal with unsupported control flow.

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Done

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -138,11 +138,11 @@ def aten_affine_grid_generator_backward(
 
     raise NotImplementedError()
 
-
+@torch_op("aten::alias", trace_only=True)
 def aten_alias(self: TensorType) -> TensorType:
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Track by https://github.com/microsoft/onnx-script/issues/424

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -361,6 +365,8 @@
     "argmax": core_ops.aten_argmax,
     "argmin": core_ops.aten_argmin,
     "clamp": core_ops.aten_clamp,
+    "cumsum": core_ops.aten_cumsum,
+    "nn.functional.gelu": nn_ops.aten_gelu,
```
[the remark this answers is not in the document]
justinchuby: this

justinchuby (approved):
🎆

justinchuby (approved):

PR #421, opened 2023-02-08
fix(atenlib): make functions with dtype attributes trace_only

### LINKED ISSUE #419: Make functions with if and dtype attributes trace_only
According to Rama, if branches needs to produce the same type and this conflicts with some logic we have in the atenlib

justinchuby:
ONNX requires if branches to produce the same type. This requirement conflicts with some logic we have in the atenlib. This PR fixes the logic.

cc @fatcat-z @xiaowuhu @titaiwangms @BowenBao @gramalingam 

Fixes #419


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -247,9 +247,6 @@ def _where_input_wrangler(
     "acosh": core_ops.aten_acosh,
     "add": core_ops.aten_add,
     "addmm": core_ops.aten_addmm,
-    "arange_start_step": core_ops.aten_arange_start_step,
```
BowenBao, line 250: The issue was not raised by unit tests earlier. Is it because we are using reference evaluator? Is it a bug in reference evaluator that it overlooks the if-branch-same-dtype requirement?

justinchuby: onnx.checker.check_function missed to check it I think.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,44 +269,80 @@ def aten_any(self: TensorType) -> TensorType:
...
+        # is casted to the specified dtype.
+        end_ = end
+        end = op.Cast(end, to=FLOAT.dtype)
+        zero = op.Constant(value_float=0.0)
+        one = op.Constant(value_float=1.0)
+        result = op.Range(zero, end, one)
+        if dtype != -1:
+            result = op.Cast(result, to=dtype)
+        else:
+            result = op.CastLike(result, end_)
```
gramalingam, line 308: May be it would be better to handle the `dtype == -1` case separately, since it seems doable without the final cast (eg., from float to some int-type). It is almost the same as the then-branch, except we need `CastLike(_, end_)`.

justinchuby: there are types Range doesn't support, commented in code

```
        # Cast input to float if dtype is not supported by Range,
        # because the input dtype may be e.g. bool
        # which Range does not support. The output type is ensured because the output
        # is casted to the specified dtype.
```

Does that sound reasonable?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,44 +269,80 @@ def aten_any(self: TensorType) -> TensorType:
...
-    if dtype != -1:
-        result = op.Cast(result, to=dtype)
+    if _range_supported(dtype):
+        end = op.Cast(end, to=dtype)
+        zero = op.Cast(0, to=dtype)
+        one = op.Cast(1, to=dtype)
+        result = op.Range(zero, end, one)
+    else:
+        # Cast input to float if dtype is not supported by Range,
+        # because the input dtype may be e.g. bool
```
gramalingam, line 297: Is this used, eg., for 16-bit floats? Just wondering, since bool seems like a very special odd-case, though I guess a user could use it.

gramalingam: Hopefully not for complex, which is tricky

justinchuby: Not complex (and not bool actually):

image

justinchuby: uint8, int8, float16, bfloat16

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,44 +269,80 @@ def aten_any(self: TensorType) -> TensorType:
...
+        if dtype != -1:
+            result = op.Cast(result, to=dtype)
+        else:
+            result = op.CastLike(result, end_)
 
     return result
 
 
-@torch_op("aten::arange", overload=True)
+@torch_op("aten::arange", overload=True, trace_only=True)
```
gramalingam, line 313: (Not part of this PR, but wondering): What does the name `TRealUnlessFloat16OrInt8` mean? The name seems unconnected to the actual definition.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
BowenBao: Filed https://github.com/microsoft/onnx-script/issues/423

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: `DOUBLE, FLOAT, INT16, INT32, INT64`. I think it's accurate? Maybe I am missing something?

onnxscript/function_libs/torch_aten/ops/core.py
gramalingam: Yes. But in the case where `end` has a supported type and `dtype == -1` (which may be a common case) you will end up with more convoluted code with extra casts. So, I should refine my suggestion to this case.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: the evaluator actually executes line by line so its behavior may be different when executing a function. @gramalingam should we allow a function mode? e.g. `ORTEvaluator(compile_functions=True)` But that will not allow break points any more...

justinchuby: We do run both the reference and ort in test.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: But we don't know if end has a supported type I don't think?

justinchuby: Unless we create an overload and dispatch from the exporter

onnxscript/function_libs/torch_aten/ops/core.py
gramalingam: Ok. There's bfloat16 missing in the name. And the word `Unless` is misleading. Perhaps it is supposed to be `Less` or `Minus` or `ButNot` or `AndNot`.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,44 +269,80 @@ def aten_any(self: TensorType) -> TensorType:
...
+        # which Range does not support. The output type is ensured because the output
+        # is casted to the specified dtype.
+        end_ = end
+        end = op.Cast(end, to=FLOAT.dtype)
+        start = op.Cast(start, to=FLOAT.dtype)
+        one = op.Constant(value_float=1.0)
+        result = op.Range(start, end, one)
+        if dtype != -1:
+            result = op.Cast(result, to=dtype)
+        else:
```
gramalingam, line 339: In this case, it looks like all input types are supported by range. Special-casing for `dtype == -1` may be useful.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,44 +269,80 @@ def aten_any(self: TensorType) -> TensorType:
...
+            result = op.CastLike(result, end_)
 
     return result
 
 
-@torch_op("aten::arange", overload=True)
+@torch_op("aten::arange", overload=True, trace_only=True)
```
gramalingam, line 345: The function name and parameter name don't match. Is it `step` or `end` ?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: I thought about this again. You are right. I will update later.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: It's step from the aten signature: arange.start_step

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
gramalingam: Yes, as Justin says, execution of control-flow constructs uses python (for better debugging) and will not invoke the ONNX checker (for an IF or LOOP). 

One possibility is to add a checker API, that looks like a normal eager-mode call, but invokes the checker (with an appropriate context). Sort of similar to Justin's suggestion above, except it is just a checker with no-execution.

BowenBao: Yes I don't see that as mutually exclusive (do checker first, then eager execute). Maybe a performance concern?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Done.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: How about `TRealNotFloat16OrBFloat16OrInt8`

onnxscript/function_libs/torch_aten/ops/core.py

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Created https://github.com/microsoft/onnx-script/issues/425 to track

justinchuby:
I am going to test this fix with ORT.

titaiwangms:
What happened to CI / test (ubuntu-latest, py310-onnx-weekly) (pull_request) ?

justinchuby:
onnx weekly didn't upload a linux wheel last week. we expect this to be fixed next week.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -275,44 +275,77 @@ def aten_any(self: TensorType) -> TensorType:
...
-        end = op.Cast(end, to=DOUBLE.dtype)
+    # NOTE: trace_only because both if branches need to be the same type, but we have
+    # a cast in the if branch.
 
-    result = op.Range(start, end, 1)
-    if dtype != -1:
-        result = op.Cast(result, to=dtype)
+    if dtype == -1:
+        one = op.CastLike(1.0, end)
+        result = op.Range(start, end, one)
```
gramalingam, line 329: I assume that inputs start and end are required to be of same type?

justinchuby: Yes. We express that in the function signature (mostly true for all functions in atenlib)

gramalingam (approved):

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
gramalingam: Agree. The infrastructure should enable the users to decide whether they want separate tests or always run the checker in eager-mode, etc.

PR #428, opened 2023-02-11
feat(atenlib): enable model checker in TorchScriptGraph

justinchuby:
- Support adding BOOL constants in graph_builder
- Fix type error in `_aten_convolution_onnx` where the `cond` to if is `int64` if `transposed` were an attribute
- Enable model checker in TorchScriptGraph

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -348,6 +348,14 @@ class TorchScriptGraph:
         return
 
     def _add_constant_to_graph(self, constant) -> torch.Value:
+        if isinstance(constant, bool):
+            # Be sure to put bool before int, because bool is a subclass of int
+            return _create_op_call_in_torch_graph(
+                self._torch_graph,
+                "onnx::Constant",
+                inputs=(),
+                attributes=dict(value=torch.tensor(constant, dtype=torch.bool)),
+            )[0]
         if isinstance(constant, float):
             return _create_op_call_in_torch_graph(
                 self._torch_graph,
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):
LGTM!

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -347,6 +348,14 @@ def register_outputs(
         return
 
     def _add_constant_to_graph(self, constant) -> torch.Value:
+        if isinstance(constant, bool):
+            # Be sure to put bool before int, because bool is a subclass of int
+            return _create_op_call_in_torch_graph(
```
BowenBao, line 353: nit: repeated code pattern.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1230,15 +1231,20 @@ def _aten_convolution_onnx(
...
     strides: Sequence[int],
     pads: Sequence[int],
     dilations: Sequence[int],
-    transposed: bool = False,
     output_padding: Sequence[int] = (0,),
     groups: int = 1,
 ) -> TFloat:
     """ConvXd with attributes pre-computed to fit the ONNX spec."""
 
+    # NOTE: transposed must be an input because when provided as an attribute,
```
BowenBao, line 1243: Is it correct to understand this is also a limitation on onnxscript side? (or onnx because no bool type attribute?)

BowenBao (approved):

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: It's because onnx has no bool type attribute. We _may_ be able to allow the translator to recognize this as a special case and add a "cast to BOOL" when it's promoting it into an input? @gramalingam

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby: #429

PR #431, opened 2023-02-11
Add several new ops relative to convolution.

fatcat-z:
Add below aten ops and their tests into library:

conv1d
conv3d
conv_transpose1d


justinchuby (approved):
Nice, thanks! Was there issue with conv transpose?

fatcat-z:
> Nice, thanks! Was there issue with conv transpose?

Yes, let's discuss it in today's meeting.

PR #436, opened 2023-02-14
feat(atenlib): add ops (new_empty, new_empty_strided)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3930,16 +3930,27 @@ def aten_negative(self: TensorType) -> TensorType:
...
-def aten_new_empty_strided(self: TensorType, size: INT64, stride: INT64) -> TensorType:
+@torch_op("aten::new_empty_strided")
+def aten_new_empty_strided(
+    self: TTensor, size: INT64, stride: INT64  # pylint: disable=unused-argument
+) -> TTensor:
     # new_empty_strided(Tensor self, SymInt[] size, SymInt[] stride, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor
 
-    raise NotImplementedError()
+    # using zero to simulate empty array
+    result = op.ConstantOfShape(size)
```
justinchuby, line 3951: naming: consider more accurate names for intermediate variables?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3930,16 +3930,27 @@ def aten_negative(self: TensorType) -> TensorType:
...
+    result = op.Expand(zero, size)
+    result = op.CastLike(result, self)
+    return result
 
 
-def aten_new_empty_strided(self: TensorType, size: INT64, stride: INT64) -> TensorType:
+@torch_op("aten::new_empty_strided")
+def aten_new_empty_strided(
+    self: TTensor, size: INT64, stride: INT64  # pylint: disable=unused-argument
+) -> TTensor:
```
justinchuby, line 3947: Looks like it can take a dtype

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3930,16 +3930,27 @@ def aten_negative(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_new_empty(self: TensorType, size: INT64) -> TensorType:
+@torch_op("aten::new_empty")
+def aten_new_empty(self: TTensor, size: INT64) -> TTensor:
```
justinchuby, line 3934: take dtype?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -396,6 +398,8 @@ def _where_input_wrangler(
     skip("empty_like", reason="Using zeros_like to simulate empty_like"),
     xfail("logcumsumexp", reason="naive implementation not numerically stable"),
     xfail("logsumexp", reason="ONNX Runtime 1.13 does not support ReduceLogSumExp-18"),
+    skip("new_empty", reason="Using zeros to simulate empty"),
```
justinchuby, line 401: Use xfail instead?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3931,25 +3931,31 @@ def aten_negative(self: TensorType) -> TensorType:
...
     # new_empty(Tensor self, SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor
 
     # using zero to simulate empty array
     zero = op.Constant(value_float=0.0)
     result = op.Expand(zero, size)
-    result = op.CastLike(result, self)
+    if dtype == -1:
+        result = op.CastLike(result, self)
+    else:
+        result = op.Cast(result, to=dtype)
```
justinchuby, line 3943: dtype needs to be the same for both branches. this op may need to be trace only

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3931,25 +3931,31 @@ def aten_negative(self: TensorType) -> TensorType:
...
     return result
 
 
 @torch_op("aten::new_empty_strided")
```
justinchuby, line 3947: potentially trace only

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: not got your point

justinchuby: So ONNX requires that both branches of an If node’s type to be the same. Since we are casting here the graph violates this constraint. I think Rama is looking at potential solutions, but for now we will need to mark the function traceonly, or else the graph would be invalid.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -396,6 +398,8 @@ def _where_input_wrangler(
     skip("empty_like", reason="Using zeros_like to simulate empty_like"),
     xfail("logcumsumexp", reason="naive implementation not numerically stable"),
     xfail("logsumexp", reason="ONNX Runtime 1.13 does not support ReduceLogSumExp-18"),
+    xfail("new_empty", reason="Using zeros to simulate empty"),
```
justinchuby, line 401: @xiaowuhu I also realized this can succeed unexpectedly. In that case a skip may be a better choice than xfail.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: More info in https://github.com/onnx/onnx/issues/4872

PR #437, opened 2023-02-14
chore: remove autopep8

justinchuby:
autopep8 is slow. We remove it and eventually should move to use Aaron's code gen module.


BowenBao (approved):
LGTM, autopep8 has been very slow on even a very small model (360s+ for gpt2 tiny). IIUC we could run lintrunner on generated code if we'd need to clean up the code style and format.

cc @gramalingam

justinchuby:
Personally since this codegen is going to be user facing lintrunner may be too heavy. black itself may be a good choice already?

PR #438, opened 2023-02-14
fix(atenlib): add domain as part of key to onnx function store

justinchuby:
Previously we don't use domain to deduplicate functions. This may cause functions that share the same name to overwrite each other.


titaiwangms (approved):
Could you give an example of what was overwritten? I thought we only had one domain.

PR #439, opened 2023-02-15
[WIP] Unblock bloom model

titaiwangms:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -765,7 +767,7 @@ def aten_bitwise_left_shift(self: TInt, other: TInt) -> TInt:
 @torch_op("aten::bitwise_not")
 def aten_bitwise_not(self: TInt) -> TInt:
     # bitwise_not(Tensor self) -> Tensor
-
+    self = op.Cast(self, to=INT64.dtype)
     return op.BitwiseNot(self)
```
justinchuby, line 771: Curious what the input type from bloom to this op was?

titaiwangms: bool

justinchuby: If we casted bool to int64, will there be more bits coming out from nowhere?

justinchuby: Is bitwise_not only used on bool types in this model? If so, I would create an overload as we can't handle dtypes in a function.

```python
@torch_op("aten::bitwise_not", overload=True)
def aten_bitwise_not_bool(self: BOOL) -> BOOL:
    # bitwise_not(Tensor self) -> Tensor

    return op.Not(self)
```

and then use this `aten_bitwise_not_bool` in place of the original in the dictionary for now.

titaiwangms: BitwiseNot expects INT anyway, so I thought we can temp cast it inside. The input is from prims_convert_element_type. Casting there makes regression on GPT.

justinchuby: I just don't think the behavior will remain the same?

titaiwangms: Got it now!

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -765,10 +767,18 @@ def aten_bitwise_left_shift(self: TInt, other: TInt) -> TInt:
...
     # bitwise_not(Tensor self) -> Tensor
-
+    # TODO(titaiwang): Support BOOL input
     return op.BitwiseNot(self)
 
 
+@torch_op("aten::bitwise_not", overload=True)
+def aten_bitwise_not_bool(self: BOOL) -> BOOL:
+    # bitwise_not(Tensor self) -> Tensor
+    # FIXME(titaiwang): This is a hack to get around the fact that we don't have op.BitwiseNot supporting bool now.
```
justinchuby, line 777: We will have to go with this I think. Not a hack as we need a dispatcher to handle different dtypes

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3146,9 +3156,8 @@ def aten_margin_ranking_loss(
 @torch_op("aten::masked_fill")
 def aten_masked_fill(self: TTensor, mask: BOOL, value: TTensor) -> TTensor:
```
justinchuby, line 3157: Could you create another PR for the non-hacks we identified, including changes to mask fill and bitwise not (both versions) so that we can merge into main? The rest can stay here for reference but we are probably not going to merge?

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Need if for `aten_bitwise_not`?

justinchuby: I think so

justinchuby: Oh we can do if because the dtypes are different, _and_ we don’t know the dtype within a function

titaiwangms: So how should we know the dtype beforehand?

titaiwangms: Maybe there are only certain kinds of ops can be input of BitWise operations that we can do the processing in those ops?

justinchuby: > So how should we know the dtype beforehand?

the exporter knows about dtypes so it can choose which function to use

titaiwangms:
Replace by #442

PR #440, opened 2023-02-15
feat(atenlib): add ops(normal, narrow)

xiaowuhu:


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -455,6 +459,11 @@ def _where_input_wrangler(
         matcher=lambda sample: sample.kwargs.get("as_tuple") is not None,
         reason="as_tuple=True is not supported",
     ),
+    skip(
```
justinchuby, line 462: What are some other possible types? How can we account for them? Is this a case where the OpInfo does not match the aten op exactly? If that’s the case I would put in reason “non float values for the argument 'mean' are not expected by the aten op”

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4036,12 +4051,21 @@ def aten_norm_except_dim(v: TensorType, pow: int = 2, dim: int = 0) -> TensorTyp
...
 
+@torch_op("aten::normal")
 def aten_normal(
-    self: TensorType, mean: float = 0.0, std: float = 1.0, generator: Optional[str] = None
-) -> TensorType:
+    self: TTensor,
+    mean: float = 0.0,
+    std: float = 1.0,
+    generator: str = "none",  # pylint: disable=unused-argument
+) -> TTensor:
```
justinchuby, line 4060: Looks like the output type can only be float, according to the doc?

```suggestion
) -> FloatType:
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4036,12 +4051,21 @@ def aten_norm_except_dim(v: TensorType, pow: int = 2, dim: int = 0) -> TensorTyp
...
 
 
+@torch_op("aten::normal")
 def aten_normal(
-    self: TensorType, mean: float = 0.0, std: float = 1.0, generator: Optional[str] = None
-) -> TensorType:
+    self: TTensor,
+    mean: float = 0.0,
+    std: float = 1.0,
+    generator: str = "none",  # pylint: disable=unused-argument
```
justinchuby, line 4059:
```suggestion
```

safe to remocet

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: the torch's test case also have Tensor type for 'mean', this is not supported by ort's op.RandomNormalLike().

xiaowuhu: ORT only accept float type for args[0] 'mean'

justinchuby (approved):
thanks!

PR #442, opened 2023-02-15
fix(atenlib): bitwise_not and maskd_fill

titaiwangms:
Create a bool overload for bitwise_not and fix an incorrect cast in maskd_fill

Discovered when running bloom export.

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -769,6 +769,12 @@ def aten_bitwise_not(self: TInt) -> TInt:
     return op.BitwiseNot(self)
 
 
+@torch_op("aten::bitwise_not", overload=True)
+def aten_bitwise_not_bool(self: BOOL) -> BOOL:
+    # bitwise_not(Tensor self) -> Tensor
+    return op.Not(self)
+
+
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3146,9 +3152,8 @@ def aten_margin_ranking_loss(
 @torch_op("aten::masked_fill")
 def aten_masked_fill(self: TTensor, mask: BOOL, value: TTensor) -> TTensor:
     # masked_fill.Tensor(Tensor self, Tensor mask, Tensor value) -> Tensor
-    mask_cast = op.Cast(mask, to=BOOL.dtype)
     value_cast = op.CastLike(value, self)
```
justinchuby, line 3155: maybe add a note along this line?

```
# NOTE: Do not attempt to cast `mask` to BOOL because mask should not take any other types.
# `mask` coming in as other types is often an error and should fail the model.
```

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Done

PR #444, opened 2023-02-16
feat(atenlib): add ops(cross_entropy_loss)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
-    target: TensorType,
-    weight: Optional[TensorType] = None,
-    reduction: int = 1,
-    ignore_index: INT64 = -100,
+@torch_op("aten::cross_entropy", trace_only=True)
+def aten_cross_entropy(
+    self: TFloatOrBFloat16,
+    target: Sequence[int],
+    weight: Optional[TFloatOrBFloat16] = None,
+    reduction: str = "mean",
```
fatcat-z, line 247: Still suggest to keep the original signature where reduction is an integer. It's very possible that the aten op cross_entropy will receive an integer as the value of 'reduction'.

We might rename this function to _aten_cross_entropy() and keep it as is. Then add a new aten_cross_entropy() function who can accept 'reduction' as an integer.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
+    ignore_index: int = -100,
     label_smoothing: float = 0.0,
-) -> TensorType:
+) -> TFloatOrBFloat16:
     # cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor
 
-    raise NotImplementedError()
+    if weight is None:
```
fatcat-z, line 253: Could we try OptionalHasElement op from now on?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -184,6 +184,16 @@ def _cat_input_wrangler(
     return args, kwargs
 
 
+def _cross_entropy_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    # Remove the self argument
```
fatcat-z, line 190: What does this comment mean? I see the code is trying to update the data type of weight, not removing any other ones.

xiaowuhu: fixed the comment

onnxscript/function_libs/torch_aten/ops/nn.py
xiaowuhu: fixed. seems works.

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: We can just set it to int I think. No need for a new function.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
+    reduction: str = "mean",
+    ignore_index: int = -100,
     label_smoothing: float = 0.0,
-) -> TensorType:
+) -> TFloatOrBFloat16:
     # cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor
 
-    raise NotImplementedError()
+    if op.OptionalHasElement(weight):
+        result, _ = op.SoftmaxCrossEntropyLoss(
```
justinchuby, line 254:
```suggestion
        result = op.SoftmaxCrossEntropyLoss(
```

I think we can just do this as ONNX knows to output just one thing when there is only one output connected

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
+            self, target, reduction=reduction, ignore_index=ignore_index
+        )
+    else:
+        result, _ = op.SoftmaxCrossEntropyLoss(
```
justinchuby, line 258:
```suggestion
        result = op.SoftmaxCrossEntropyLoss(
```

as above

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -481,6 +492,11 @@ def _where_input_wrangler(
         matcher=lambda sample: isinstance(sample.kwargs.get("padding"), str),
         reason="String padding is not accepted by aten::conv2d",
     ),
+    skip(
```
justinchuby, line 495: I am trying to follow why this skip is needed. Could you elaborate?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
+    weight: Optional[TFloatOrBFloat16] = None,
+    reduction: str = "mean",
+    ignore_index: int = -100,
     label_smoothing: float = 0.0,
-) -> TensorType:
+) -> TFloatOrBFloat16:
     # cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor
 
-    raise NotImplementedError()
+    if op.OptionalHasElement(weight):
```
justinchuby, line 253: Was the condition reversed?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
...
+        result, _ = op.SoftmaxCrossEntropyLoss(
+            self, target, reduction=reduction, ignore_index=ignore_index
+        )
+    else:
+        result, _ = op.SoftmaxCrossEntropyLoss(
+            self, target, weight, reduction=reduction, ignore_index=ignore_index
```
justinchuby, line 259: Do we need optional get element?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,26 @@ def aten_conv_depthwise3d(
     raise NotImplementedError()
 
 
-def aten_cross_entropy_loss(
-    self: TensorType,
-    target: TensorType,
-    weight: Optional[TensorType] = None,
-    reduction: int = 1,
-    ignore_index: INT64 = -100,
+@torch_op("aten::cross_entropy")
```
titaiwangms, line 242: Isn't it `aten::cross_entropy_loss`?

onnxscript/function_libs/torch_aten/ops/nn.py
titaiwangms: I think it serves `check None` situation for weight according to what we have in symbolc_opset12.py.

```python
@_onnx_symbolic("aten::cross_entropy_loss")
@_beartype.beartype
def cross_entropy_loss(
    g: jit_utils.GraphContext,
    self,
    target,
    weight,
    reduction,
    ignore_index,
    label_smoothing,
):
    # none reduction : onnx::Constant[value={0}]
    # mean reduction : onnx::Constant[value={1}]
    # sum reduction : onnx::Constant[value={2}]
    reduction = symbolic_helper._maybe_get_const(reduction, "i")
    reduction_vals = ["none", "mean", "sum"]
    reduction = reduction_vals[reduction]

    label_smoothing = symbolic_helper._maybe_get_const(label_smoothing, "f")
    if label_smoothing is not None and label_smoothing > 0.0:
        raise errors.SymbolicValueError(
            "Unsupported: ONNX does not support label_smoothing", self
        )

    # in onnx SoftmaxCrossEntropyLoss specification, ignore_index is optional without default value.
    # therefore we need to set ignore_index attribute even if it is not specified (e.g. ignore_index=-100).
    ignore_index = symbolic_helper._maybe_get_const(ignore_index, "i")
    if weight.node().mustBeNone():
        celoss = g.op(
            "SoftmaxCrossEntropyLoss",
            self,
            target,
            reduction_s=reduction,
            ignore_index_i=ignore_index,
        )
    else:
        celoss = g.op(
            "SoftmaxCrossEntropyLoss",
            self,
            target,
            weight,
            reduction_s=reduction,
            ignore_index_i=ignore_index,
        )

    return celoss
```

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: in op.SoftmaxCrossEntropyLoss(), we only accept rarget as INT64 type. but the torch's test case can have FLOAT type which means something else.

onnxscript/function_libs/torch_aten/ops/nn.py
xiaowuhu: yes, originally it is cross_entropy_loss, but I cannot find any torch document, only find cross_entropy().

onnxscript/function_libs/torch_aten/ops/nn.py
xiaowuhu: I cannot find any document for cross_entropy_loss(), but only cross_entropy(). and in the document about cross_entropy(), the 'reduction' parameter is str, not int.
from my understanding, reduction(int) signature is the original API definition.  while reduction(str) is a helper function which user always want to use.

justinchuby: Again we don't worry about what the torch document says (which is a documentation of torch op, not aten ops.). The aten signature is the only source of truth we care about. If it says int, we should use int. This is because torch will internally convert user input into what the then op expects.

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: image

justinchuby: I speculate there's probably no documentation for the aten op if it couldn't be found. I would refer to the corresponding symbolic function implementation in torch.onnx as a hint

xiaowuhu: so we will keep cross_entropy_loss() as the function name?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: That's curious. What does it mean? Do we need to create an overload to implement the FLOAT case?

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby: Yeah we should keep the original.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: from the document: 

**target ([Tensor](https://pytorch.org/docs/master/tensors.html#torch.Tensor)) – Ground truth class indices or class probabilities;**

onnxscript/function_libs/torch_aten/ops/nn.py
xiaowuhu: OK

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: I see! I guess we can worry about this later

onnxscript/function_libs/torch_aten/ops/nn.py
xiaowuhu: OptionalGetElement() will throw exception on local box testing

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -313,6 +313,7 @@ def _where_input_wrangler(
     "nn.functional.adaptive_avg_pool1d": nn_ops.aten_adaptive_avg_pool1d,
     "nn.functional.adaptive_avg_pool2d": nn_ops.aten_adaptive_avg_pool2d,
     "nn.functional.adaptive_avg_pool3d": nn_ops.aten_adaptive_avg_pool3d,
+    # "nn.functional.cross_entropy_loss": nn_ops.aten_cross_entropy_loss,  # not is OPS_DB
```
justinchuby, line 316: Can we create an OpInfo for it? I feel that this is an op we need to test.

justinchuby: Jay has an example https://github.com/microsoft/onnx-script/blob/818d75c9e142f4fa9d082a2af75f05726e41c6dd/onnxscript/tests/function_libs/torch_aten/extra_opinfo.py

justinchuby: https://github.com/pytorch/pytorch/blob/da41003b5f2f1709c1e1e524165c000e4f6e413e/torch/testing/_internal/common_methods_invocations.py#L11520

Is this relavant or reusable?

xiaowuhu: yes I'm using the 'cross_entropy' as test case to test cross_entropy_loss

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -699,6 +716,10 @@ def test_output_match(self, device: str, dtype: torch.dtype, op):
                 inputs=repr(inputs),
                 kwargs=repr(cpu_sample.kwargs),
             ):
+
+                if i == 19:
```
fatcat-z, line 720: Is this necessary?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -378,6 +389,7 @@ def _where_input_wrangler(
     "nn.functional.conv1d": core_ops.aten_conv1d,
     "nn.functional.conv2d": core_ops.aten_conv2d,
     "nn.functional.conv3d": core_ops.aten_conv3d,
+    "nn.functional.cross_entropy": (nn_ops.aten_cross_entropy_loss, _cross_entropy_input_wrangler),
```
fatcat-z, line 392: Can we add a comment to confirm this is by design, not a bug?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,28 @@ def aten_conv_depthwise3d(
...
-
-    raise NotImplementedError()
+    self: TFloatOrBFloat16,
+    target: Sequence[int],
+    weight: Optional[TFloatOrBFloat16] = None,
+    reduction: int = 1,  # default is 'mean'
+    ignore_index: int = -100,
+    label_smoothing: float = 0.0,  # this was ignored due to ONNX not support
+) -> TFloatOrBFloat16:
+    # cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor
```
justinchuby, line 251: Nit docstring

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,28 @@ def aten_conv_depthwise3d(
...
+
+    reduction_vals = ["none", "mean", "sum"]
```
justinchuby, line 253: Can we write some of branches and condition on the int value instead?

justinchuby: So that the function is not trace only, even though the logic may become more complex

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -239,17 +239,45 @@ def aten_conv_depthwise3d(
     raise NotImplementedError()
 
 
+@torch_op("aten::cross_entropy_loss", trace_only=True)
```
fatcat-z, line 242: Is trace_only required here?

xiaowuhu: Yes. because 
1) there has one bug in onnx-script  when parse Optional argument.
2) this function contains no op() in logic, so have to use trace_only.

fatcat-z (approved):
LGTM, thanks!

onnxscript/function_libs/torch_aten/ops/nn.py
justinchuby, line 242: I would add a comment on why trace only is needed

justinchuby: Possible to combine the functions?

justinchuby: https://github.com/microsoft/onnx-script/pull/555/files

PR #448, opened 2023-02-17
feat(eager): ignore_unknown_function_kwargs option

justinchuby:
Allow ORTEvaluator to ignore unknown function kwargs so we can handle kwargs we don't care in the exporter.


onnxscript/evaluator.py
```diff
@@ -262,7 +274,11 @@ def eval_function(
         # Split happens in the evaluator instead of the OnnxFunction __call__ method
         # so that evaluators can control behaviors like whether to fill in default values for attributes.
         inputs, attributes = param_manipulation.separate_input_attributes_from_arguments(
-            param_schemas, args, kwargs, fill_defaults=False, allow_extra_kwargs=False
+            param_schemas,
+            args,
+            kwargs,
+            fill_defaults=False,
+            allow_extra_kwargs=self._ignore_unknown_function_kwargs,
```
titaiwangms, line 281: Not so sure how this works:

https://github.com/microsoft/onnx-script/blob/24b153ae2083527ad928c9be54d0b6b8dfb129a1/onnxscript/_internal/param_manipulation.py#L41-L42

Shouldn't it just raise an error if the differences are detected?

justinchuby: So if allow_extra_kwargs and there is extra args then we don't raise; if allow_extra_kwargs is false and there is extra args then we raise. Was I missing something?

gramalingam: (1) Are the extra kwargs silently ignored when no error is raised?
(2) What is the motivating use-case?

titaiwangms: I was looking for filter, but this seems just allowing it.

justinchuby: (1) Yes
(2) One use case is in the TorchScriptEvaluator, where there is an unknown and changing number of kwargs in the torch ops, e.g. `memory_format` etc. that does not change the op's numerical behavior and thus not applicable in onnx. We initially use a block like to filter out those kwargs, but as torch and introduce new kwargs, I think it is safer for us to just ignore them so atenlib doesn't break as torch introduces these kwargs.

titaiwangms (approved):

justinchuby:
@gramalingam merging this to unblock the exporter. LMK if there are changes you want me to make and I will follow up with a PR

titaiwangms:
@justinchuby Should we adapt this into TorchScriptEvaluator?

justinchuby:
We have already. The torchscript evaluator always allow unknown args for practicality

titaiwangms:
It seems that diagnostics relying on accurate kwargs for some reasons:

```
  File "/home/titaiwang/pytorch/symbolically_gpt.py", line 28, in <module>
    ) = fx_onnx.export_without_parameters_and_buffers(
  File "<@beartype(torch.onnx._internal.fx.exporter.export_without_parameters_and_buffers) at 0x7f9c5c80c280>", line 96, in export_without_parameters_and_buffers
  File "/home/titaiwang/pytorch/torch/onnx/_internal/fx/exporter.py", line 899, in export_without_parameters_and_buffers
    _export(
  File "<@beartype(torch.onnx._internal.fx.exporter._export) at 0x7f9c5c804940>", line 31, in _export
  File "/home/titaiwang/pytorch/torch/onnx/_internal/fx/exporter.py", line 638, in _export
    onnxscript_graph, initializers = _export_fx_to_onnxscript(
  File "/home/titaiwang/pytorch/torch/onnx/_internal/diagnostics/infra/decorator.py", line 124, in wrapper
    return_values = fn(*args, **kwargs)
  File "/home/titaiwang/pytorch/torch/onnx/_internal/fx/exporter.py", line 527, in _export_fx_to_onnxscript
    _export_fx_node_to_onnxscript(
  File "<@beartype(torch.onnx._internal.fx.exporter._export_fx_node_to_onnxscript) at 0x7f9c5c804430>", line 145, in _export_fx_node_to_onnxscript
  File "/home/titaiwang/pytorch/torch/onnx/_internal/diagnostics/infra/decorator.py", line 124, in wrapper
    return_values = fn(*args, **kwargs)
  File "/home/titaiwang/pytorch/torch/onnx/_internal/fx/exporter.py", line 439, in _export_fx_node_to_onnxscript
    ] = symbolic_fn(*onnx_args, **onnx_kwargs)
  File "/home/titaiwang/pytorch/torch/onnx/_internal/diagnostics/infra/decorator.py", line 115, in wrapper
    format_function_signature_in_markdown(
  File "<@beartype(torch.onnx._internal.diagnostics.infra.decorator.format_function_signature_in_markdown) at 0x7f9c5cb9b4c0>", line 84, in format_function_signature_in_markdown
  File "/home/titaiwang/pytorch/torch/onnx/_internal/diagnostics/infra/decorator.py", line 39, in format_function_signature_in_markdown
    state = utils.function_state(fn, args, kwargs)
  File "<@beartype(torch.onnx._internal.diagnostics.infra.utils.function_state) at 0x7f9c5f9b9b80>", line 67, in function_state
  File "/home/titaiwang/pytorch/torch/onnx/_internal/diagnostics/infra/utils.py", line 65, in function_state
    bind = inspect.signature(fn).bind(*args, **kwargs)
  File "/home/titaiwang/anaconda3/envs/pytorch-dev/lib/python3.9/inspect.py", line 3045, in bind
    return self._bind(args, kwargs)
  File "/home/titaiwang/anaconda3/envs/pytorch-dev/lib/python3.9/inspect.py", line 3034, in _bind
    raise TypeError(
TypeError: got an unexpected keyword argument 'layout'
```

BowenBao:
Hmm I feel silently ignoring extra kwargs is prone to error. For example a typo won't be caught. Besides, I feel like there should always be a good reason whenever we ignore a certain kwarg from aten, and it should be made explicit.

If continuing with this PR, @titaiwangms it sounds like a good opportunity to capture a diagnostics when this is hit. You can add logic to mitigate this situation for `function_state`, log the kwargs being ignored, and alter the diagnostics level to warning.

justinchuby:
> Hmm I feel silently ignoring extra kwargs is prone to error. For example a typo won't be caught. Besides, I feel like there should always be a good reason whenever we ignore a certain kwarg from aten, and it should be made explicit.
> 
> If continuing with this PR, @titaiwangms it sounds like a good opportunity to capture a diagnostics when this is hit. You can add logic to mitigate this situation for `function_state`, log the kwargs being ignored, and alter the diagnostics level to warning.

I agree. That's currently the best I can think of to account for an unknown set of kwargs. Would love to hear ideas too how we can do this better and more robustly.

PR #449, opened 2023-02-17
fix: replace deprecated onnx helpers

justinchuby:
Remove all onnx.mapping reference


gramalingam (approved):
Thanks for fixing these

jcwchen (approved):
LGTM. Thanks!

PR #450, opened 2023-02-17
fix(atenlib): flaky test for new_empty

justinchuby:
Test for new_empty was flaky because it doesn't always fail. Changed to skip


BowenBao (approved):
Is this related to the onnx fix for empty tensor?

xiaowuhu:
> Is this related to the onnx fix for empty tensor?

I think this is because new_empty always failed even we use xfail, but re-run it will passed.

justinchuby:
like @xiaowuhu says. new_emtpy is not a deterministic op but it _can_ be zeros sometime.

BowenBao:
ah it's uninitialized data.

PR #451, opened 2023-02-18
chore(atenlib): convert aten signatures from comments to docstrings

justinchuby:
So they show up as doc and in the onnx model.

### DIFF
--- a/onnxscript/function_libs/tools/torch_aten/generate_aten_signatures.py
+++ b/onnxscript/function_libs/tools/torch_aten/generate_aten_signatures.py
@@ -237,7 +237,7 @@ def create_signature(func: torchgen.model.NativeFunction) -> cg.FunctionDef:
         *py_args,
         return_type=create_return_type(func.func.returns),
         body=[
-            cg.ThunkStmt(f"# {func.func}"),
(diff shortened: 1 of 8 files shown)


onnxscript/function_libs/tools/torch_aten/generate_aten_signatures.py
```diff
@@ -237,7 +237,7 @@ def create_signature(func: torchgen.model.NativeFunction) -> cg.FunctionDef:
         *py_args,
         return_type=create_return_type(func.func.returns),
         body=[
-            cg.ThunkStmt(f"# {func.func}"),
+            cg.ThunkExpr(f'"""{func.func}"""'),
```
BowenBao, line 240: I thought the CI typing issue is with the next line (Raise not being Stmt)?

justinchuby: Yeah... The thing should have been an expression but the body was typed to only accept statements. I changed it back.

justinchuby:
ONNX Weekly CI tracked: https://github.com/microsoft/onnx-script/issues/455

PR #454, opened 2023-02-21
feat(atenlib): add ops(contiguous, dropout)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1075,10 +1075,14 @@ def aten_constant_pad_nd(self: TensorType, pad: INT64, value: float = 0.0) -> Te
     raise NotImplementedError()
 
 
-def aten_contiguous(self: TensorType, memory_format: str = "contiguous_format") -> TensorType:
+@torch_op("aten::contiguous", trace_only=True)
+def aten_contiguous(self: TTensor, memory_format: str = "contiguous_format") -> TTensor:
     # contiguous(Tensor(a) self, *, MemoryFormat memory_format=contiguous_format) -> Tensor(a)
 
-    raise NotImplementedError()
+    if memory_format in ["contiguous_format", "preserve_format"]:
```
justinchuby, line 1082: I wonder if we can always return self? This way we reduce the amount of trace_only ops

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -490,6 +501,11 @@ def _where_input_wrangler(
         matcher=lambda sample: isinstance(sample.kwargs.get("padding"), str),
         reason="String padding is not accepted by aten::conv2d",
     ),
+    skip(
+        "nn.functional.dropout",
+        matcher=lambda sample: len(sample.kwargs) == 0 or sample.kwargs.get("p", 0.0) > 0.0,
+        reason="dropout is randomly so the result not match",
```
justinchuby, line 507:
```suggestion
        reason="dropout is random so the result not match",
```

nit.

At some point we should match only the dtype and shape for these random ops. cc @fatcat-z do you have thoughts on how we should do this?

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: We can leave such data validation work out of aten lib functions. If the string is wrong, when use inference the ONNX graph, ort is supposed to throw an error.

justinchuby: good point as @fatcat-z says. thanks for pointing out.

xiaowuhu: Offline discussed with Jay, finally we think we need break the pipeline at this time, because ort doesn't support the other 2 format string.  
If we always return op.Identity(self), but when the memory_format in ["channels_last", "channels_last_3d"] in torch model, this will leading value errors in onnx model.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: fixed

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: @justinchuby 
Yes, this is different than what I thought before. In current case, the value of memory_format won't be passed to ORT meaning ORT would not know the memory_format might be an unsupported one (channels_last/channels_last_3d). And this will cause a wrong inference result and it could only be found when we check this value in ONNX graph which is unfriendly.

Based on this, probably the best solution is to reminder users during the export by raising an exception or printing a warning.

Thoughts?

justinchuby: I think that’s ok. The functions will be much clearer if we restrict them to not raise any errors. Also for consistency, because only the trace only functions can raise errors. 

I would leave a todo in-line for this case, and handle them in the future when we can annotate constraints for argument, as part of the function meta data structure

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1075,10 +1075,14 @@ def aten_constant_pad_nd(self: TensorType, pad: INT64, value: float = 0.0) -> Te
     raise NotImplementedError()
 
 
-def aten_contiguous(self: TensorType, memory_format: str = "contiguous_format") -> TensorType:
-    """contiguous(Tensor(a) self, *, MemoryFormat memory_format=contiguous_format) -> Tensor(a)"""
+@torch_op("aten::contiguous", trace_only=True)
+def aten_contiguous(self: TTensor, memory_format: str = "contiguous_format") -> TTensor:
+    # contiguous(Tensor(a) self, *, MemoryFormat memory_format=contiguous_format) -> Tensor(a)
```
justinchuby, line 1080: Docstring

justinchuby (commented):
I would return self in aten_contiguous and add a todo explaining what is not supported. Otherwise lgtm. Thanks!

onnxscript/function_libs/torch_aten/ops/core.py
fatcat-z: @gramalingam 

Do you have any suggestions for such case?

gramalingam: I think it is good to make sure we don't export models that will have a different behavior. It is better that users see failure at export-time instead of getting an erroneous model exported. I don't quite understand Justin's last comment. Are you suggesting some other mechanism for catching such "unsupported cases"? It seems to me an exception, trace-only mode, etc. seems preferable to a todo comment.

gramalingam: I don't understand the op spec in aten. I guess it is supposed to alter the layout representation in some cases, but not others? May be I am reading too much into the names, but I am puzzled how both "contiguous_format" and "preserve_format" get mapped to identity.

justinchuby: > It is better that users see failure at export-time instead of getting an erroneous model exported

I agree that we should get failures at export-time. I think it would be nice to keep the functions purely descriptive (logic, signature, expected input) so that trace_only and onnxfunctions can be seem the same. I am playing with the idea that we annotate the arguments (via Annotated, or via a decorator) to specify any additional constraints which the exporter can then pick up and handle errors accordingly. The TODO comment was suggested for adding this annotation later when we implement this mechanism.

> I am puzzled how both "contiguous_format" and "preserve_format" get mapped to identity.

That's a good point. Maybe the wise thing to do for this op in particular is to drop it in the FX exporter and never send it to ONNX? Does onnx have a notion of memory format?

fatcat-z: So probably the best way for now is:
1) Keep raising this exception if the value of memory_format is invalid for ONNX.
2) Add a TODO comment to mention Justin's idea so we might come back once that idea has been proved is feasible.

fatcat-z: > > It is better that users see failure at export-time instead of getting an erroneous model exported
> 
> I agree that we should get failures at export-time. I think it would be nice to keep the functions purely descriptive (logic, signature, expected input) so that trace_only and onnxfunctions can be seem the same. I am playing with the idea that we annotate the arguments (via Annotated, or via a decorator) to specify any additional constraints which the exporter can then pick up and handle errors accordingly. The TODO comment was suggested for adding this annotation later when we implement this mechanism.
> 
> > I am puzzled how both "contiguous_format" and "preserve_format" get mapped to identity.
> 
> That's a good point. Maybe the wise thing to do for this op in particular is to drop it in the FX exporter and never send it to ONNX? Does onnx have a notion of memory format?

Current PyTorch exporter could export it so it'd better we don't lose it.

justinchuby: > So probably the best way for now is:
> 
> 1. Keep raising this exception if the value of memory_format is invalid for ONNX.
> 2. Add a TODO comment to mention Justin's idea so we might come back once that idea has been proved is feasible.

ok

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1825,10 +1829,20 @@ def aten_dot(self: TFloat, tensor: TFloat) -> TFloat:
...
+@torch_op("aten::dropout")
+def aten_dropout(input: TFloat, p: float = 0.5, train: bool = True) -> TFloat:
+    # dropout(Tensor input, float p, bool train) -> Tensor
 
-    raise NotImplementedError()
+    rank_input = op.Size(op.Shape(input))
+    if rank_input == 0:
+        input = op.Reshape(input, op.Constant(value_ints=[-1]))
+
+    result, _ = op.Dropout(input, p, train)
```
justinchuby, line 1840:
```suggestion
    result = op.Dropout(input, p, train)
```

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: add below comment:
'''
# TODO: find some fundamental solutoin to solve this kind of error
'''

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: op.Dropout() return (result, mask), but aten_dropout only need result without mask.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -1077,12 +1077,12 @@ def aten_constant_pad_nd(self: TensorType, pad: INT64, value: float = 0.0) -> Te
...
 @torch_op("aten::contiguous", trace_only=True)
 def aten_contiguous(self: TTensor, memory_format: str = "contiguous_format") -> TTensor:
-    # contiguous(Tensor(a) self, *, MemoryFormat memory_format=contiguous_format) -> Tensor(a)
+    """contiguous(Tensor(a) self, *, MemoryFormat memory_format=contiguous_format) -> Tensor(a)"""
 
     if memory_format in ["contiguous_format", "preserve_format"]:
         return op.Identity(self)
     else:
-        raise NotImplementedError()
+        raise NotImplementedError()  # TODO: find some fundamental solutoin to solve this kind of error
```
fatcat-z, line 1085:
```suggestion
        # TODO: Find out a way to annotate constraints for argument, as part of the function meta data structure.
        raise NotImplementedError() 
```

xiaowuhu: good

fatcat-z (approved):
LGTM, thanks!

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: @gramalingam i think this is another example when eager mode differs with the function, as Dropout can have one or two outputs but eager mode returns two no matter what?

gramalingam: [This](https://github.com/microsoft/onnx-script/blob/main/onnxscript/evaluator.py#L319) needs to be fixed.

Either a two-line short-term patch for specifically Dropout to indicate the expected number of outputs, or a longer term fix of invoking ONNX inference to actually figure this out. (The longer-term solution is a bit non-trivial.)

gramalingam: (So, I recommend the short-term fix for immediate use.)

PR #457, opened 2023-02-21
chore(release)!: bump python version to 3.8

justinchuby:
Specify python 3.8 as the lowest supported version, now that numpy and pytorch are supporting only 3.8+.

1. Bumped the ruff config to target python 3.8.
2. Enabled pyupgrade in ruff to update code.


pyproject.toml
```diff
@@ -141,3 +142,7 @@ unfixable = [
 
 [tool.ruff.pydocstyle]
 convention = "google"
+
+[tool.ruff.pyupgrade]
+# Preserve types, even if a file imports `from __future__ import annotations`.
```
justinchuby, line 147:
```suggestion
# Preserve types, even if a file imports `from __future__ import annotations`.
# This is needed to support ONNX Script function annotation syntaxes.
```

gramalingam (approved):

PR #459, opened 2023-02-22
feat(atenlib): add ops(layer_norm)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2876,17 +2876,25 @@ def aten_kthvalue(
...
-    bias: Optional[TensorType] = None,
+    weight: Optional[TReal] = None,
+    bias: Optional[TReal] = None,
     eps: float = 1e-05,
     cudnn_enable: bool = True,
 ) -> TensorType:
     """layer_norm(Tensor input, int[] normalized_shape, Tensor? weight=None, Tensor? bias=None, float eps=1e-05, bool cudnn_enable=True) -> Tensor"""
 
-    raise NotImplementedError()
+    axes = [-i for i in range(len(normalized_shape), 0, -1)]
```
gramalingam, line 2890: So, the actual value specified in normalized_shape is irrelevant for aten_layer_norm? That sounds odd, so checking.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -400,6 +400,7 @@ def _where_input_wrangler(
     "convolution": core_ops.aten_convolution,
     "empty_like": core_ops.aten_empty_like,
     "index_select": core_ops.aten_index_select,
+    # "native_layer_norm": core_ops.aten_layer_norm,  # FIXME: 3 outputs != 1 output
```
justinchuby, line 403: I would duplicate the opinfo and update its op

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2893,17 +2893,25 @@ def aten_kthvalue(
...
+    if weight is None:
+        weight = op.CastLike(1, input)
+    if bias is None:
+        bias = op.CastLike(0, input)
+
+    result, _, _ =  _aten_native_layer_norm_onnx(input, weight, bias, axes=axes, eps=eps)
```
justinchuby, line 2913: I wonder if we should implement an function only for layer_norm, so we don’t have unused outputs in the onnx graph and also make it clear to users that the op is a different op. 

Also since we have opset 18 now I wonder if some None checking can be done with optional get element? For this I do want to check if the resulting function works well on ORT first

onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
```diff
@@ -184,4 +222,16 @@ def sample_inputs_convolution(op_info, device, dtype, requires_grad, **kwargs):
         skips=(),
         supports_out=False,
     ),
+    opinfo_core.OpInfo(
```
fatcat-z, line 225: Let's move this on top of "nn.functional.conv3d" for alphabetical order.

fatcat-z (approved):
LGTM, thanks!

titaiwangms:
There is actually an opset for LayerNorm

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3984,18 +3995,18 @@ def _aten_native_layer_norm_onnx(
 ) -> Tuple[TReal, TReal, TReal]:
```
titaiwangms, line 3995: axes should be input now.

```python
axes: Sequence[INT64],
```

And other places using ReduceMax/ReduceMean/ReduceMin should all be updated, otherwise, the model would pop errors saying ReduceXXX having unexpected input/attribute axes depending on opset_verison 17/18.

gramalingam: Not necessarily. The interface of an aten op/function shouldn't change because of its implementation. Since an attribute can be promoted to be an input, it should be okay to leave the interface as is. But it is important to ensure that the implementation works correctly. Eg., use the `noop_with_empty_axes` attribute as appropriate to ensure correct behavior for when axes is empty, etc.

gramalingam: I am saying that the annotation should ideally depend on the aten specification (and not on the onnx opset). The call to onnx ops like ReduceMean, of course, will depend on the onnx opset.

titaiwangms: Make sense, but for onnx function (aten interface), I think we still rely on function_ir to differentiate attrs/inputs. I guess I am not sure if function_ir changed due to the opset version bump 17 -> 18 in this case?

titaiwangms: I think the idea might be related to https://github.com/microsoft/onnx-script/issues/443

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3984,18 +3995,18 @@ def _aten_native_layer_norm_onnx(
...
 
     # FIXME(justinchuby): Use opset18 when it is supported by onnxruntime
```
justinchuby, line 3997: nit: Remove

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: > I am saying that the annotation should ideally depend on the aten specification (and not on the onnx opset). The call to onnx ops like ReduceMean, of course, will depend on the onnx opset.

I agree. We _could_ change it to `INT64` (instead of `Sequence[INT64]`) prepare for symint inputs. The evaluator should be able to handle this (maybe already)

PR #463, opened 2023-02-23
feat(atenlib): add ops (any, expand_as)

xiaowuhu:


gramalingam:
A couple of questions: why not use `ReduceMax` instead of `ReduceSum` ? Is it because of opset dependence?

Also: does ReduceSum not work for scalar inputs? I am wondering where exactly the logic breaks (with onnx.ORT) if a zero-dimensional tensor is used.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,60 @@ def aten_angle(self: TensorType) -> TensorType:
...
+        dims = op.Cast(dim, to=INT64.dtype)
+        result_sum = op.ReduceSum(result_int, dims, keepdims=keepdim, noop_with_empty_axes=0)
+        result = op.Greater(result_sum, zero)
+    else:  # kwarg{dim,keepdim} was not given
+        result = op.Greater(self, zero)
+        result_rank = op.Size(op.Shape(result))
+        if result_rank > 0:  # need return rank=0 result
+            result_int = op.Cast(result, to=FLOAT.dtype)
+            result_sum = op.ReduceSum(result_int, keepdims=0, noop_with_empty_axes=0)
+            result = op.Greater(result_sum, zero)
```
justinchuby, line 296: I wonder if we can just cast everything to BOOL then reduce

justinchuby: Hopefully only `0.0` will be casted to `0`?

justinchuby: that means -0.0000001 or 0.00000001 stays `True`

justinchuby: maybe something like `Not(Equal(0)) -> Cast(INT) -> ReduceMax -> Cast(BOOL)` 😅

justinchuby: Or maybe this is better:

`Cast(BOOL) -> Cast(UINT8) -> ReduceMax -> Cast(BOOL)` so there is not equality and simulation of the casting.

gramalingam: Yes, that was my suggestion. (There is a minor downside/implication, since ReduceMax switched from attribute to input, for axes, in opset 18 ... while ReduceSum has been uniformly so, since opset13. So, this will complicate the opsets that can be targetted potentially.)

justinchuby: Do we worry that ReduceSum may overflow?

xiaowuhu:
> A couple of questions: why not use `ReduceMax` instead of `ReduceSum` ? Is it because of opset dependence?
> 
> Also: does ReduceSum not work for scalar inputs? I am wondering where exactly the logic breaks (with onnx.ORT) if a zero-dimensional tensor is used.

thanks for reminder. I changed to use ReduceMax now.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,33 @@ def aten_angle(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_any(self: TensorType) -> TensorType:
+@torch_op("aten::any", trace_only=True)
+def aten_any(self: BOOL, dim: Optional[int] = None, keepdim: bool = True) -> BOOL:
```
fatcat-z, line 273: Why do we specify the input is BOOL? Should not it be TTensor?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,33 @@ def aten_angle(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_any(self: TensorType) -> TensorType:
+@torch_op("aten::any", trace_only=True)
```
fatcat-z, line 272: Why do we need trace_only here? Looks like all of operations are safe.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,33 @@ def aten_angle(self: TensorType) -> TensorType:
...
 
-    raise NotImplementedError()
+    minus_1 = op.Constant(value_ints=[-1])
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, minus_1)
+
+    self = op.Abs(self)  # change negative to positive then do op.Greater()
+    zero = op.Constant(value_float=0.0)
+    result = op.Greater(self, zero)
```
fatcat-z, line 283: Why do we need this step?

xiaowuhu: Because test case has both positive number and neg number. Abs to convert them all to positive numbers, then using Greater to check if it is zero(False) or positive numbers(True).

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: any() means to check if any element in the tensor is not equal to 0, so finally it is a BOOL.

fatcat-z (approved):
Thanks!

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,32 @@ def aten_angle(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_any(self: TensorType) -> TensorType:
+@torch_op("aten::any", trace_only=True)
+def aten_any(self: TTensor, dim: Optional[int] = None, keepdim: bool = True) -> BOOL:
```
gramalingam, line 273: Is `dim` meant to be an attribute or input? I guess it is meant to be an attribute?

xiaowuhu: yes, it is an attribute from the test case.

BowenBao, line 273: I think keepdim should default to False, or set as `keepdim: Optional[bool] = None`, and explicitly set to `False` if `None`.

That way we can avoid hard coding `keepdims=0` later.

https://github.com/microsoft/onnx-script/blob/ec7c26471e0ef1896ba492549f33d0367b0301e4/onnxscript/function_libs/torch_aten/ops/core.py#L292

gramalingam: Shouldn't the default value be determined by the spec of the aten op?

BowenBao: Looking at it again, I figured we are supporting 2 overloads of `aten::any` with this function. So for `- func: any(Tensor self) -> Tensor` we need actually figuring out what the behavior is since both `dim` and `keepdim` are not provided. So I think this is fine.

And for `- func: any.dim(Tensor self, int dim, bool keepdim=False) -> Tensor`, yea you are right, the current default is wrong and it should be False.

```
- func: any(Tensor self) -> Tensor
  device_check: NoCheck   # TensorIterator
  structured_delegate: any.all_out
  variants: method, function
  dispatch:
    SparseCPU, SparseCUDA: any_sparse

- func: any.dim(Tensor self, int dim, bool keepdim=False) -> Tensor
  device_check: NoCheck   # TensorIterator
  structured_delegate: any.out
  variants: function, method
```

BowenBao: cc @justinchuby it is weird the wrong default is not caught in op test, I wonder if it is just not enough coverage in original opinfo or something else.

justinchuby: I think we should not have combined the two overloads, and yes - you have a great point that we should verify the opinfo to assert how much confidence it can provide us for each op before trusting it. I will add this to the authoring guide

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,32 @@ def aten_angle(self: TensorType) -> TensorType:
...
     """any(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    minus_1 = op.Constant(value_ints=[-1])
```
justinchuby, line 276: nit: negative_one

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -269,10 +269,32 @@ def aten_angle(self: TensorType) -> TensorType:
...
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, minus_1)
+
+    zero = op.Constant(value_float=0.0)
+    result = op.Not(op.Equal(self, zero))
```
justinchuby, line 282: Why not just cast to BOOL?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Could you say more on what optional argument parsing is? I would also add a comment on why trace_only was needed in code

xiaowuhu: if not use trace_only, it will throw exception:

---
E   TypeError: Required input 'dim: Attribute[None]' was not provide
---

gramalingam: I think this is the same as the issue we discussed in today's meeting. `if (OptionalHasElement(x)) then use(x)` is an issue. We need to find a solution to this. One possibility is to extend OptionalGetElement, but wonder if that's enough for all cases.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Address by https://github.com/microsoft/onnx-script/issues/498 @xiaowuhu i would link the new prs here for reference

PR #464, opened 2023-02-23
feat(atenlib): test random ops

justinchuby:
Check shape and dtype only for ops that are known to be indeterministic.
Also added the skip_if mechanism to skip tests based on environment.

### DIFF
--- a/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
@@ -407,7 +407,7 @@ OPINFO_FUNCTION_MAPPING_TRACE_ONLY: dict[
 }
 
 # These ops are not deterministic, so we check shape and dtype only
-UNDETERMINISTIC_OPS: frozenset[str] = frozenset(
+INDETERMINISTIC_OPS: frozenset[str] = frozenset(
     (
         "empty_like",
         "empty",
@@ -570,9 +570,9 @@ OP_WITH_SKIPPED_SUBTESTS = frozenset(meta.op_name for meta in SKIP_SUBTESTS)
 ALL_OPS_IN_DB = frozenset(op_info.name for op_info in OPS_DB)
 # Assert all ops in OPINFO_FUNCTION_MAPPING are in the OPS_DB
 assert TESTED_OPS.issubset(ALL_OPS_IN_DB), f"{TESTED_OPS - ALL_OPS_IN_DB} not in OPS_DB"
-assert UNDETERMINISTIC_OPS.issubset(
+assert INDETERMINISTIC_OPS.issubset(
     TESTED_OPS
-), f"{UNDETERMINISTIC_OPS - TESTED_OPS} not in OPS_DB"
+), f"{INDETERMINISTIC_OPS - TESTED_OPS} not in OPS_DB"
 
 TORCH_TYPE_TO_ONNX = {
     torch.bool: onnx.TensorProto.BOOL,
@@ -772,9 +772,9 @@ class TestOutputConsistency(unittest.TestCase):
                         else torch.tensor(torch_output)
                     )
 
-                    if op.name in UNDETERMINISTIC_OPS:
+                    if op.name in INDETERMINISTIC_OPS:
                         # Check shape and dtype only for ops that are known to be
-                        # undeterministic
+                        # indeterministic
                         self.assertEqual(actual.shape, expected.shape)
                         self.assertEqual(actual.dtype, expected.dtype)
                         continue
--- a/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
@@ -572,7 +572,7 @@ ALL_OPS_IN_DB = frozenset(op_info.name for op_info in OPS_DB)
 assert TESTED_OPS.issubset(ALL_OPS_IN_DB), f"{TESTED_OPS - ALL_OPS_IN_DB} not in OPS_DB"
 assert INDETERMINISTIC_OPS.issubset(
     TESTED_OPS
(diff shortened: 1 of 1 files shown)


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -402,25 +430,17 @@ def _where_input_wrangler(
...
-                "nn.functional.upsample_nearest2d",
-                reason="ONNX Runtime 1.13 does support opset18",
-            ),
-        )
-        if version_utils.onnxruntime_older_than("1.14")
-        else ()
+    xfail(
+        "logsumexp",
+        reason="ONNX Runtime 1.13 does not support ReduceLogSumExp-18",
+        active_if=version_utils.onnxruntime_older_than("1.14"),
```
BowenBao, line 436: Do we need this or should we just remove the xfail?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -389,6 +406,17 @@ def _where_input_wrangler(
     "zeros_like": core_ops.aten_zeros_like,
 }
 
+# These ops are not deterministic, so we check shape and dtype only
+INDETERMINISTIC_OPS: frozenset[str] = frozenset(
```
gramalingam, line 410: Nit: "Indeterministic" => "Nondeterministic"

justinchuby: Thanks 😅

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: For now we still need it because the ort version with function default attribute support is still a variant of 1.13. @gramalingam i think we can move to use nightly after liqun s PR?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Done

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -743,14 +765,27 @@ def test_output_match(self, device: str, dtype: torch.dtype, op):
...
+                    expected = (
+                        torch_output
+                        if isinstance(torch_output, torch.Tensor)
+                        else torch.tensor(torch_output)
+                    )
+
+                    if op.name in NONDETERMINISTIC_OPS:
+                        # Check shape and dtype only for ops that are known to be
+                        # nondeterministic
+                        self.assertEqual(actual.shape, expected.shape)
```
fatcat-z, line 778: It'd better to verify the mean value between them are very close with an acceptable tolerance. [Here](https://github.com/pytorch/pytorch/blob/a641d6075761050599bf972c0b6bc3823182914e/test/onnx/test_pytorch_onnx_onnxruntime.py#L11792)is an example.

justinchuby: Ops like empty are creating non initialized values. Or dropout which does not expect a matching mean. Should we distinguish them?

justinchuby: I think I am inclined to create individual tests for distributions if desired

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -97,16 +101,18 @@ def skip(
...
 
     Args:
         op_name: The name of the operator.
         variant_name: Optional OpInfo variant_test_name.
-        dtypes: The dtypes to skip.
         reason: The reason for skipping.
+        dtypes: The dtypes to skip.
         matcher: A function that matches the test sample input. It is used only when
             xfail is in the SKIP_SUBTESTS list.
+        active_if: Whether the skip is active.
```
fatcat-z, line 115: A little confused: I see this parameter was used in xfail() function, while this description is saying about skip.

Is this a general parameter, or related to skip or xfail?

justinchuby: Sorry that was a typo. It’s general and inherited from PyTorch

justinchuby:
I removed merge at lgtm because there seems to be more comments to address

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Updated. I didn't create the name but it makes sense: when used it becomes something like `active_if=onnxruntime_older_than("1.14"),`, which is good for redability.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Created https://github.com/microsoft/onnx-script/issues/467 to track

justinchuby:
I think I addressed everything.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
fatcat-z: Nit:
I agree that "active_if" also makes sense. 

I'm thinking this in terms of unified style of our code. While I go through our code base, we didn't use "active" to describe if something will work or not. Most of time, we will use "enabled" for such purpose, like cudnn_enabled, logger.isEnabledFor. 

So probably "enabled_if" would be better.

justinchuby: SG!

justinchuby: Done

fatcat-z (approved):
LGTM, thanks!

PR #466, opened 2023-02-23
Support none-valued attribute parameter

gramalingam:
Extend the converter to support attribute-values that are None (in actual parameter list). Currently, it requires users to omit such attribute values instead of specifying a value of None.

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -443,8 +443,11 @@ class Converter:
 
                 # Create GraphProto attribute
                 val = val.to_graph_proto()
-            return self.ir_builder.make_attr(attr_name, val)
-        return self.ir_builder.make_attr(attr_name, self.eval_constant_expr(expr))
+        else:
+            val = self.eval_constant_expr(expr)
+        if val is None:
+            return None
+        return self.ir_builder.make_attr(attr_name, val)
 
     def translate_docstring(self, node):
         if hasattr(node.value, "value"):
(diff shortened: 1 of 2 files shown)


onnxscript/converter.py
```diff
@@ -443,8 +443,11 @@ def translate_attr(self, attr_name, expr):
 
                 # Create GraphProto attribute
                 val = val.to_graph_proto()
-            return self.ir_builder.make_attr(attr_name, val)
-        return self.ir_builder.make_attr(attr_name, self.eval_constant_expr(expr))
+        else:
+            val = self.eval_constant_expr(expr)
+        if val is None:
```
justinchuby, line 448: Would it be helpful to add a comment on the case this is handling?

justinchuby (approved):

onnxscript/converter.py
```diff
@@ -753,6 +756,7 @@ def translate_call_expr(self, node):
         args = [self.translate_opt_expr(x) for x in node.args]
         args = autocast.static_cast_inputs(self, callee.get_schema(), *args)
         attrs = [self.translate_attr(x.arg, x.value) for x in node.keywords]
+        attrs = [attr for attr in attrs if attr is not None]
```
justinchuby, line 759: This reminds me of how torch.onnx handles it. I think they did the same thing.

justinchuby, line 759: It'd be nice to explain the filter with a comment

onnxscript/converter.py
gramalingam: Done

onnxscript/converter.py
gramalingam: Done

PR #468, opened 2023-02-24
fix(docs): page titles and toc

justinchuby:
Remove furo theme's error message for TOCs. Updated page titles


docs/api/converter.rst
```diff
@@ -1,4 +1,4 @@
-Class Converter
+Converter
 ===============
```
xadupre, line 2: you should shorten the underline too

docs/api/opsets.rst
```diff
@@ -1,25 +1,28 @@
-opsets and types
+Opsets and Types
 ================
 
-.. contents::
```
xadupre, line 4: I did not expect that to cause a warning.

justinchuby: It was from the theme because how it was designed, they decided this is redundant and need special treatment if we really want it

docs/api/converter.rst
justinchuby: Done!

justinchuby:
@xadupre I think this is ready

docs/api/opsets.rst
justinchuby: there are some discussions in https://github.com/pradyunsg/furo/discussions/125

gramalingam (approved):

PR #469, opened 2023-02-24
fix(ci): pin onnx-weekly version

justinchuby:
Pin onnx-weekly version to avoid CI breaks. The requirements file should be updated by dependabot.


gramalingam (approved):

justinchuby:
Not sure if dependabot supports pre release versions. We will see how it goes next week.

jcwchen:
Just curious: why fixing onnx-weekly version is needed here? I was waiting for this update in ONNX: https://github.com/onnx/onnx/pull/4928 by upcoming onnx-weekly package to fix the pipeline failure [here](https://github.com/microsoft/onnx-script/issues/455).

justinchuby:
I was hoping that once the fix is in and we bump the version pin, we will have the CI green. Then in the future dependabot will create prs and merge them only when there is no regression. We keep the CI green with this mechanism.

But I need to find out if dependabot will bump versions when they are prerelease.

PR #474, opened 2023-02-25
Extend converter to support flexible input, attribute ordering

gramalingam:
Extend the converter to support flexible input and attribute ordering, using param_manipulation to classify inputs and attributes.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -12,9 +12,7 @@ nox.options.error_on_missing_interpreters = False
 COMMON_TEST_DEPENDENCIES = (
     "click",
     "jinja2",
-    'numpy==1.23.5; python_version>="3.8"',
-    'numpy; python_version<"3.8"',
-    "protobuf<4",
+    "numpy==1.23.5",
     "typing_extensions",
     "beartype",
     "types-PyYAML",
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -3992,7 +3992,7 @@ def aten_negative(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::new_empty")
+@torch_op("aten::new_empty", trace_only=True)
 def aten_new_empty(self: TTensor, size: INT64, dtype: int = -1) -> TTensor:
     """new_empty(Tensor self, SymInt[] size, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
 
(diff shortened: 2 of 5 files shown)


onnxscript/converter_test.py
```diff
@@ -615,14 +616,25 @@ def inc_alpha_expanded(A: FLOAT[...], alpha: int) -> FLOAT[...]:
...
+            return op11.Squeeze(X, axes=None)
 
         @script()
         def implicit_none(X):
-            return op.Squeeze(X)
+            return op11.Squeeze(X)
 
         self.assertSame(explicit_none, implicit_none)
 
+    def test_input_and_attr1(self):
```
justinchuby, line 627: I would clarify the behavior being tested in the test name. e.g. test_<behavior>

justinchuby:
Suprising that mypy doesn't ask us to assert not None in call sites: https://github.com/microsoft/onnx-script/blob/eb3accb076d45b69109fbe7ef222ea01081c041f/onnxscript/evaluator.py#L283 and https://github.com/microsoft/onnx-script/blob/eb3accb076d45b69109fbe7ef222ea01081c041f/onnxscript/function_libs/torch_aten/graph_building.py#L218. Should we do it?

justinchuby:
Oh I see this is changing Op, not OnnxFunction, maybe we don't need the asserts

gramalingam:
Yes. The idea is to be able to handle unknown ops (that is, ops without a known schema) during conversion. It can be useful and is possible to handle them for now.

onnxscript/converter_test.py
```diff
@@ -615,14 +616,25 @@ def inc_alpha_expanded(A: FLOAT[...], alpha: int) -> FLOAT[...]:
     def test_none_attribute(self):
         @script()
         def explicit_none(X):
-            return op.Squeeze(X, axes=None)
+            return op11.Squeeze(X, axes=None)
```
justinchuby, line 619: Was there a reason the change the opset?

justinchuby: If it’s important, would a comment help to remind future readers?

justinchuby (approved):

justinchuby:
Am I understanding it right that this is what enables custom ops? Will we need similar changes the the graph capturing evaluators? cc @titaiwangms

titaiwangms (approved):

onnxscript/converter_test.py
gramalingam: It's not critical. I could have used some dummy op for this test, since it was testing the use of None as an attribute value. "axes" was changed from an attribute to input for Squeeze in a later opset. Previously we were not using the opschema to identify attributes, but now we are, so it was getting mapped to an input instead of attribute, defeating the purpose of the test-case.

gramalingam: Done

PR #477, opened 2023-02-26
feat(atenlib): modify ops(new_empty, new_empty_strided)

xiaowuhu:
Due to using if-else in the function but return different type.


fatcat-z (approved):
LGTM, thanks!

justinchuby:
Fixes https://github.com/microsoft/onnx-script/issues/445

PR #480, opened 2023-02-27
Merge 'initializers' into 'TorchScriptGraph'

BowenBao:
Reduce friction.

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -296,10 +296,15 @@ class TorchScriptGraph:
         # All the functions used, deduplicated by name
         # key: (name, domain)
         self._function_store: Dict[Tuple[str, str], onnxscript.OnnxFunction] = {}
+        self._initializers: Dict[str, torch.Tensor] = {}
 
     @property
     def torch_graph(self):
         return self._torch_graph
+
+    @property
+    def initializers(self) -> Mapping[str, torch.Tensor]:
+        return self._initializers
 
     @beartype
     def add_input(
@@ -326,6 +331,12 @@ class TorchScriptGraph:
         )
         tensor_value = _wrap_torch_value_to_tensor(torch_value)
         return tensor_value  # type: ignore[return-value]
+
+    @beartype
+    def add_initializer(
+        self, input_name: str, input_value: torch.Tensor
+    ) -> None:
+        self._initializers[input_name] = input_value
 
     @beartype
     def register_outputs(
@@ -447,7 +458,7 @@ class TorchScriptGraph:
 
     @beartype
     def to_model_proto(
-        self, initializers: Mapping[str, torch.Tensor], opset_version: Optional[int]
+        self, opset_version: int
     ) -> onnx.ModelProto:
         (
             proto,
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):

titaiwangms:
I think torch-nightly is broken by https://github.com/microsoft/onnx-script/pull/460

BowenBao:
@titaiwangms thx, hard to figure out what are known failures...

Does the CI here depend on code from torch exporter? This api change requires modification from both sides, I was planning to merge this first hoping CI could be all green.

titaiwangms:
> @titaiwangms thx, hard to figure out what are known failures...
> 
> Does the CI here depend on code from torch exporter? This api change requires modification from both sides, I was planning to merge this first hoping CI could be all green.

I don't think onnx-script relies on torch exporter, but what part in this PR needs torch exporter modification? As far as I know, the experimental branch is broken due to lack of maintain on ort-experimental branch, and torch-nightly fails because of squeeze op. There are other two erroring with ORT that I don't quite understand, but this PR seems to be unrelated...

BowenBao:
The api changed, `initializers` no longer an argument to `TorchScripGraph.to_model_proto`, and pt exporter need to adjust for it, which is in https://github.com/pytorch/pytorch/pull/95676. It does seem that pt exporter is not involved in any of the CI tests here.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -296,11 +296,16 @@ def __init__(self):
...
         # key: (name, domain)
         self._function_store: Dict[Tuple[str, str], onnxscript.OnnxFunction] = {}
+        self._initializers: Dict[str, torch.Tensor] = {}
 
     @property
     def torch_graph(self):
         return self._torch_graph
 
+    @property
+    def initializers(self) -> Mapping[str, torch.Tensor]:
```
justinchuby, line 306: Would a brief docstring for this property help?

onnxscript/function_libs/torch_aten/graph_building_test.py
```diff
@@ -17,23 +17,22 @@
 @unittest.skipIf(version_utils.torch_older_than("2.0"), "torchscript in 1.13 not supported")
 class TestTorchScriptTracingEvaluator(unittest.TestCase):
     def setUp(self):
+        # FIXME: Currently this must match with the import line
+        # `from onnxscript import opset17 as op`, which restricts opset to be 17 in these
```
justinchuby, line 21: The model opset can actually be different from the local functions' opsets I think. Do you know if that's true?

BowenBao: Yea it could be different under constraint.
https://github.com/onnx/onnx/blob/4b2d50334914621835cc1e8dadd4fe82b6b9876c/onnx/onnx.in.proto#L824-L828
```
  // The operator sets imported by FunctionProto should be compatible with the ones
  // imported by ModelProto. Example, if same operator set say 'A' is imported by FunctionProto
  // and ModelProto then versions for the operator set may be different but,
  // the operator schema returned for op_type, domain, version combination
  // for both the versions should be same.
```

PR #482, opened 2023-03-01
feat(ateblib): add ops(stack)

titaiwangms:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5132,10 +5132,14 @@ def aten_sspaddmm(
     raise NotImplementedError()
 
 
-def aten_stack(tensors: Sequence[TensorType], dim: int = 0) -> TensorType:
+@torch_op("aten::stack")
+def aten_stack(
+    tensors: Union[TTensorOrString, Sequence[TTensorOrString]], dim: int = 0
```
gramalingam, line 5137: What does it do if the input is a tensor, instead of a sequence? Does it act like an identity op then?

Inference for `ConcatFromSequence` will fail if the input is not a sequence, I think.

titaiwangms: Ahh Yes, there might be a chance it's a Tensor input for this ATen op, I should either change the annotation, or add an TypeError.

onnxscript/function_libs/torch_aten/tensor_typing.py
```diff
@@ -54,6 +60,7 @@
 ]
 
 TTensor = TypeVar("TTensor", bound=_TensorType)
+TTensorOrString = TypeVar("TTensorOrString", bound=Union[_TensorType, str])
```
gramalingam, line 63: Don't understand this. Do you mean a tensor of strings? If so, it should be `STRING` not `str`.

titaiwangms: Done

onnxscript/function_libs/torch_aten/tensor_typing.py
```diff
@@ -39,6 +43,9 @@
     INT32,
     INT64,
     UINT8,
+    UINT16,
```
fatcat-z, line 46: Any reason why we add these unsigned types?

titaiwangms: Following the spec: https://github.com/onnx/onnx/blob/main/docs/Changelog.md#ConcatFromSequence-11

fatcat-z: We didn't add these types and had above comments based on [this](https://pytorch.org/docs/stable/tensors.html). Probably this is not necessary, but anyway , it doesn't hurt.

fatcat-z (approved):
LGTM, thanks!

onnxscript/function_libs/torch_aten/tensor_typing.py
titaiwangms: I see. I will revert it back

PR #483, opened 2023-03-01
feat(atenlib): aten_expand supports -1 dim

titaiwangms:


gramalingam (approved):

titaiwangms:
> Are there tests we can enable for this?

Thinking about parametrizing fx-export into runtime test.

justinchuby:
Was there an OpInfo covering this case?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2039,6 +2039,8 @@ def aten_expand(self: TTensor, size: TInt) -> TTensor:
     """expand(Tensor(a) self, SymInt[] size, *, bool implicit=False) -> Tensor(a)"""
 
     size = op.Cast(size, to=INT64.dtype)
+    # To support -1 dim.
+    size = op.Abs(size)
```
justinchuby, line 2043: I understand -1 becomes 1. Was that intended or was I missing something?

titaiwangms: Only -1 becomes 1. Do you think comment is redundant?

PR #484, opened 2023-03-01
feat(graph_building): support dynamic inputs

### LINKED ISSUE #390: Torchscript Graph: Support dynamic axes
Build support for dynamic axes in inputs

### LINKED ISSUE #518: Fixing support for SymInts
In PyTorch, SymInts or symbolic ints is a way of supporting dynamic attributes. In atenlib we tend to represent them as `INT64` tensors to maintain their dynamic nature. However in fx, they are usually passed in as a list of nodes (a list of tensors).

**It may be better if we use `Sequence[INT64]` to represent them to maintain consistency with the fx graph.**

This representation further enables `trace_only` support for SymInts when we need to obtain a rank. (#517)
As an example in the `native_layer_norm`:

```python
def aten_native_layer_norm(
    input: TReal,
    normalized_shape: Sequence[INT64],
    weight: Optional[TReal],
    bias: Optional[TReal],
    eps: float,
) -> Tuple[TReal, TReal, TReal]:
    """native_layer_norm(Tensor input, SymInt[] normalized_shape, Tensor? weight, Tensor? bias, float eps) -> (Tensor, Tensor, Tensor)"""
```

Here `normalized_shape` is a SymInt sequence. If we annotate it as `Sequence[int]`, it will be considered an attribute by onnxscript and fail with a TypeError because we reject `TorchScriptTensor` types as attributes; if we annotate it as `INT64`, the Python line for taking the rank [`start_axis = -len(normalized_shape)`](https://github.com/microsoft/onnx-script/pull/517/files#diff-e94a882bd9dcdfc75120b0571c970a28d35ec65bf9b8ba02b68a8b442f2853a7R4039) will fail because `TorchScriptTensor` does not have `len` defined, and we cannot use `op.Neg(op.Size(op.Shape(normalized_shape)))` because `start_axis` is subsequently used as an attribute (so not dynamic) in `LayerNormalization`.

## Change required

For this change, we need to update how we convert python scalar inputs to tensors: Whereas before we convert `List[int]` to `INT64`, we should now convert them to `List[INT64]` tensors (convert each element to INT64 then `SequenceConstruct`). Functions that take SymInts should use `ConcatFromSequence` to concatenate the sequence into a single tensor for its own use.

## Potential issues

1. I am not sure if all `SymInt` sequences are passed as lists? Can the length (aka. rank) be dynamic too which will force us to use `INT64`, or use `INT64` -> `SplitToSequence`?
2. We may make things slower for static int sequences. For this we need to constant fold `ConcatFromSequence`.

cc @titaiwangms @BowenBao @xiaowuhu @fatcat-z @gramalingam

titaiwangms:
Used by https://github.com/pytorch/pytorch/pull/96350

fixes #518 fixes #390


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5304,6 +5304,15 @@ def aten_swapdims(self: TensorType, dim0: int, dim1: int) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::sym_size")
+def aten_sym_size(self: TReal, dim: int = 0) -> TReal:
```
justinchuby, line 5308: no op info available I assume?

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: No. It doesn't even have Overload.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5304,6 +5304,17 @@ def aten_swapdims(self: TensorType, dim0: int, dim1: int) -> TensorType:
...
+    """sym_size(Tensor self, int dim) -> Tensor"""
+    # NOTE: onnx-script doesn't support attribute process,
+    # so op.Shape(self, start=dim, end=dim + 1) is not supported.
+    shape = op.Shape(self)
+    start = op.Reshape(dim, [1])
+    end = op.Reshape(dim + 1, [1])
+    return op.Slice(shape, start, end)
```
BowenBao, line 5315: Shape + Gather?

titaiwangms: It seems I would have to put dim into a list? 
```python
@torch_op("aten::sym_size")
def aten_sym_size(self: TReal, dim: int = 0) -> TReal:
    """sym_size(Tensor self, int dim) -> Tensor"""
    # NOTE: onnx-script doesn't support attribute process,
    # so op.Shape(self, start=dim, end=dim + 1) is not supported.
    shape = op.Shape(self)
    return op.Gather(shape, [dim])
```

some expression errors

BowenBao: a bit more like 
https://github.com/pytorch/pytorch/blob/be220690d9d8b7e8d58b81694c6cbf0958f94845/torch/onnx/symbolic_helper.py#L1322

You can reference https://github.com/pytorch/pytorch/blob/be220690d9d8b7e8d58b81694c6cbf0958f94845/torch/onnx/symbolic_opset9.py#L1019
There are a lot edge case handling which are probably not needed.

titaiwangms: Revisit on this, and found that in this op, we should follow this line https://github.com/pytorch/pytorch/blob/be220690d9d8b7e8d58b81694c6cbf0958f94845/torch/onnx/symbolic_opset9.py#L1188, but it doesn't seem right.

According to [Gather](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Gather), axis should be int, and `dim` in aten_sym_size should be equivalent to `indices` which is tensor(INT32/INT64). Either way, I need Reshape to convert dim from int to tensor(INT64) I think. So I prefer to keep this function. LMK if I miss anything!

Although I tried like
```python
@torch_op("aten::sym_size")
def aten_sym_size(self: TReal, dim: int = 0) -> TReal:
    """sym_size(Tensor self, int dim) -> Tensor"""
    # NOTE: onnx-script doesn't support attribute process,
    # so op.Shape(self, start=dim, end=dim + 1) is not supported.
    shape = op.Shape(self)
    idx= op.Reshape(dim, [1])
    return op.Slice(shape, idx)
```
It errors with SegFault.

BowenBao: > I need Reshape to convert dim from int to tensor(INT64)

Can you use Constant?

> It errors with SegFault.

Does it segfault in onnxscript or onnxruntime? Could you open an issue?

titaiwangms: > Can you use Constant?

No, I remember @gramalingam told me that `dim` is an expression that we can't prcoess it in attributes of ONNX op. I am not so familiar with the root cause. Maybe @gramalingam or @justinchuby can share more here?

> Does it segfault in onnxscript or onnxruntime? Could you open an issue? 

It seems the same error I mentioned yesterday. The protobuf can't be called on `onnx_model.SerializeToString()`. It's from protobuf API, but I will post it and see if ONNX team would know more about this.

titaiwangms: > It seems the same error I mentioned yesterday. The protobuf can't be called on onnx_model.SerializeToString(). It's from protobuf API, but I will post it and see if ONNX team would know more about this.

A second thought on this, thinking this might be the pass: `self._torch_graph._export_onnx()` being incompatible with new graph?

justinchuby: Yeah Constant requires a compile time constant

gramalingam: Ok, there are a couple of different issues here: the SegFault is mysterious and needs to be debugged. The other issue is straightforward: if we want to convert a scalar attribute into a 1D tensor, we need to do it in two steps, first converting it into a 0D tensor (via Constant, but OnnxScript will insert this for you), and then by reshaping it into a 1D tensor.

titaiwangms: https://github.com/microsoft/onnx-script/issues/523

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -312,27 +312,34 @@ def initializers(self) -> Mapping[str, torch.Tensor]:
...
-                input_value
-            )
+        torchscript_input_value = torch.TensorType.create_from_tensor(  # pylint: disable=c-extension-no-member,protected-access
+            input_value
         )
+        if dynamic_axes:
+            rank = torchscript_input_value.dim()
+            # We set input sizes all to be dynamic to initialize the model conversion,
+            # and nodes are also set to be dynamic in exporter function: _fill_tensor_meta.
+            torch_value.setType(torchscript_input_value.with_sizes([None] * rank))
```
BowenBao, line 340: I think we should grab the sizes from fx graph input? I'm ok with this for now since we have not hooked up dynamic export yet, but once that happens, we should just need to copypaste shape info from there. Do you see any concerns/issues? @titaiwangms @justinchuby

titaiwangms: For current solution (faster), for each input FakeTensor, we keep static part of shapes, and replace the SymInts with None, and still use `with_sizes()` to set it on TorchScriptGraph. The downside I guess it's losing information that "same SymInts indicates same shape", for example, [symint0, symint0, symint1] would all be [None, None, None]

onnxscript/function_libs/torch_aten/ops/core.py
gramalingam: As discussed in another PR, `dim+1` also doesn't work if dim is -1. Hence, Gather seems a better solution.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -313,21 +313,26 @@ def initializers(self) -> Mapping[str, torch.Tensor]:
...
-            tensor_value = _wrap_torch_value_to_tensor(torch_value)
-            return tensor_value  # type: ignore[return-value]
-        torch_value = self._torch_graph.addInput(input_name)
-        torch_value.setType(torch.TensorType.create_from_tensor(input_value))
+        else:
+            torch_value = self._torch_graph.addInput(input_name)
+            torch_value.setType(torch_value.type().with_dtype(dtype))  # type: ignore[arg-type]
+            torch_value.setType(
+                torch_value.type().with_sizes(
+                    [dim if isinstance(dim, int) else None for dim in shape]  # type: ignore[union-attr]
```
BowenBao, line 333: Can you add a TODO or NOTE or FIXME that this step loses the information of individual symbols?

titaiwangms:
> Is this ready to merge?

Not yet, I am waiting for torch side, because of the add_input API changes... SHould be these two days!

onnxscript/function_libs/torch_aten/graph_building.py
titaiwangms: OK

titaiwangms:
Merged this, and will update it with https://github.com/microsoft/onnx-script/pull/532#issuecomment-1477189487 into https://github.com/pytorch/pytorch/pull/96350

PR #486, opened 2023-03-01
feat(atenlib): add op(squeeze)

xiaowuhu:
It breaks nightly build, because there have different test case than PR build pipeline.


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5139,10 +5139,25 @@ def aten_square(self: TensorType) -> TensorType:
...
+def aten_squeeze(self: TTensor, dim: Optional[int] = None) -> TTensor:
     """squeeze(Tensor(a) self) -> Tensor(a)"""
 
-    raise NotImplementedError()
+    if op.OptionalHasElement(dim):
+        rank = op.Size(op.Shape(self))
+        if rank == 0:
+            self = op.Reshape(self, op.Constant(value_ints=[-1]))
+        # check if specified dimension equal to 1
+        shape = op.Shape(self, start=dim, end=dim + 1)  # type: ignore[operator]
```
gramalingam, line 5151: We can't use computed values like `dim + 1` as an attribute. We will need to get the shape, and use a slice op to extract something like this.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5139,10 +5139,25 @@ def aten_square(self: TensorType) -> TensorType:
...
+        if shape != 1:
```
gramalingam, line 5152: Is this an error-condition or acceptable in aten's squeeze op?

xiaowuhu: aten_squeeze definition: when value_of_specified_dim != 1, don't do the squeeze, just return self.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: got it. changed.

justinchuby:
I would see if there is a way to make it not trace_only. There's too many loose ops otherwise in the graph.

justinchuby:
cc @shubhambhokare1

justinchuby:
So squeeze has three aten overloads we care about. I think we should implement three (or two, combining dim and dims if it is clean enough) functions likewise. This way we avoid default values on inputs (not supported by onnx), create OnnxFunctions, and avoids the has element check.

```yaml
- func: squeeze(Tensor(a) self) -> Tensor(a)
  variants: function, method

- func: squeeze.dim(Tensor(a) self, int dim) -> Tensor(a)
  variants: function, method
  tags: core

- func: squeeze.dims(Tensor(a) self, int[] dim) -> Tensor(a)
  variants: function, method
  tags: core

# We don't care about the dimname overload
- func: squeeze.dimname(Tensor(a) self, Dimname dim) -> Tensor(a)
  variants: function, method
```

becomes

```python
def aten_squeeze(self: TTensor) -> TTensor:
    ....

def aten_squeeze_dims(self: TTensor, dim: INT64) -> TTensor:
    ....
```

justinchuby:
The same would apply for any (https://github.com/microsoft/onnx-script/pull/463#discussion_r1125417498) and min https://github.com/microsoft/onnx-script/pull/491

justinchuby:
We can duplicate and fit the OpInfos for each of these overloads. The goal is to make them all onnx functions and not trace only functions.

xiaowuhu:
> So squeeze has three aten overloads we care about. I think we should implement three (or two, combining dim and dims if it is clean enough) functions likewise. This way we avoid default values on inputs (not supported by onnx), create OnnxFunctions, and avoids the has element check.
> 
> ```yaml
> - func: squeeze(Tensor(a) self) -> Tensor(a)
>   variants: function, method
> 
> - func: squeeze.dim(Tensor(a) self, int dim) -> Tensor(a)
>   variants: function, method
>   tags: core
> 
> - func: squeeze.dims(Tensor(a) self, int[] dim) -> Tensor(a)
>   variants: function, method
>   tags: core
> 
> # We don't care about the dimname overload
> - func: squeeze.dimname(Tensor(a) self, Dimname dim) -> Tensor(a)
>   variants: function, method
> ```
> 
> becomes
> 
> ```python
> def aten_squeeze(self: TTensor) -> TTensor:
>     ....
> 
> def aten_squeeze_dims(self: TTensor, dim: INT64) -> TTensor:
>     ....
> ```

done

justinchuby (approved):

PR #490, opened 2023-03-02
feat(atenlib): refine native_layer_norm

xiaowuhu:


onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
```diff
@@ -199,35 +199,35 @@ def sample_inputs_layer_norm(
 
 OP_DB: List[opinfo_core.OpInfo] = [
     opinfo_core.OpInfo(
-        "convolution",
-        aliases=("convolution",),
-        aten_name="convolution",
+        "layer_norm",
```
fatcat-z, line 202: "convolution" should be in the front of "layer_norm".

onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
```diff
@@ -211,23 +211,23 @@ def sample_inputs_layer_norm(
         supports_out=False,
     ),
     opinfo_core.OpInfo(
-        "nn.functional.conv3d",
```
BowenBao, line 214: The diff in this file seems just swapping the test upside down? Should be able to revert since there is no real change right?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4003,37 +4003,20 @@ def aten_native_layer_norm(
     # where D is the dimension of normalized_shape. For example, if normalized_shape is
     # (3, 5) (a 2-dimensional shape), the mean and standard-deviation are computed
     # over the last 2 dimensions of the input (i.e. input.mean((-2, -1))).
+
     axes_list = [-i for i in range(len(normalized_shape), 0, -1)]
-    axes = op.Constant(value_ints=axes_list)
+    start_axis = axes_list[0]
     if not op.OptionalHasElement(weight):
-        weight = op.CastLike(1, input)
```
BowenBao, line 4009: Would `CastLike` still be needed?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4003,37 +4003,20 @@ def aten_native_layer_norm(
...
+        one = op.Constant(value_float=1.0)
+        weight = op.Expand(one, op.Shape(input, start=start_axis))
```
BowenBao, line 4011: Can it be rewritten with `ConstantOfShape` to keep in one line?

BowenBao (approved):
LGTM w/ comments inline

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3990,7 +3990,7 @@ def aten_native_group_norm_backward(
 @torch_op("aten::native_layer_norm", trace_only=True)
 def aten_native_layer_norm(
     input: TReal,
-    normalized_shape: Sequence[int],
+    normalized_shape: Sequence[INT64],
```
gramalingam, line 3993: Same question here. Is this supposed to be an int64 tensor, or a sequence of int64 tensors? I doubt it is meant to be the second. Is there a pointer to the corresponding aten op documentation?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4003,37 +4003,20 @@ def aten_native_layer_norm(
...
     if not op.OptionalHasElement(weight):
-        weight = op.CastLike(1, input)
+        one = op.Constant(value_float=1.0)
+        weight = op.Expand(one, op.Shape(input, start=start_axis))
     if not op.OptionalHasElement(bias):
-        bias = op.CastLike(0, input)
-    return _aten_native_layer_norm_onnx(input, weight, bias, axes, eps)
-
+        zero = op.Constant(value_float=0.0)
+        bias = op.Expand(zero, op.Shape(input, start=start_axis))
```
gramalingam, line 4014: ONNX's LayerNorm also allows bias to be optional. It might be possible to avoid creating a zero-valued bias as above. (May be worth trying, but there could be other complications.)

onnxscript/function_libs/torch_aten/ops/core.py
gramalingam: I also think it might be possible to just use the scalar constant one (without expanding it), assuming the ONNX LayerNormalization op permits a weight that is broadcastable to the input. Sadly, the documentation of the op in ONNX and ORT do not clarify this. But the ORT implementation seems to allow it, when I tried it.

onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
xiaowuhu: yes :)

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: yes, should be int64 tensor. fixed.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: if only use
···
weight = op.Constant(value_float=1.0)
···
will get below error:
----
E   onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running LayerNormalization node. Name:'' Status Message: Size of X.shape()[axis:] == 6. Size of scale and bias (if provided) must match this. Got scale size of 1 and bias size of 6

xiaowuhu: > Can it be rewritten with `ConstantOfShape` to keep in one line?

```
op.ConstantOfShape(op.Shape(input), value=op.Constant(value_ints=[1]))
```
always failed.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: add back

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: good point. changed the code to allow bias = None

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4003,37 +4003,24 @@ def aten_native_layer_norm(
     # where D is the dimension of normalized_shape. For example, if normalized_shape is
     # (3, 5) (a 2-dimensional shape), the mean and standard-deviation are computed
     # over the last 2 dimensions of the input (i.e. input.mean((-2, -1))).
+
     axes_list = [-i for i in range(len(normalized_shape), 0, -1)]
```
justinchuby, line 4007: We will not be able to get len() of a Tensor. Can this be refined?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3990,7 +3990,7 @@ def aten_native_group_norm_backward(
 @torch_op("aten::native_layer_norm", trace_only=True)
 def aten_native_layer_norm(
     input: TReal,
-    normalized_shape: Sequence[int],
+    normalized_shape: INT64,
     weight: Optional[TReal],
```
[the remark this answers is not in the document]
gramalingam: Restating the question, as I understand it: ONNX has two kinds of "optional inputs". I think the essential question is whether we need to distinguish between these two when we define an onnxscript function. Currently, we don't yet have a way to do so. We currently use `Optional[...]`.

I think the ideal approach would be to write (and translate) onnxscript functions in a way that works for both kinds of optional inputs (so, it is polymorphic over this dimension). This was a motivating factor for extending OptionalHasElement to work for both types of optional inputs. However, this is not yet complete, since the extension to OptionalGetElement achieves this only partially.

If we ever need to support the case where users explicitly identify which kind of optional input a function expects/supports, then we could do that by introducing new constructs `Optional1[...]` and `Optional2[...]`. But I am not yet sure we need that. If you see a scenario where it is needed, please let me know.

PR #491, opened 2023-03-02
feat(atenlib): add op(min)

xiaowuhu:
The logic is same wth aten_max()


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3425,10 +3425,58 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_min(self: TensorType) -> TensorType:
+@torch_op("aten::max", trace_only=True)
```
justinchuby, line 3428:
```suggestion
@torch_op("aten::min", trace_only=True)
```

typo?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3425,10 +3425,58 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
...
+
+
+@torch_op("aten::max", overload=True)
+def _aten_min_with_other(self: TReal, other: TReal) -> TReal:
+    result = op.Min(self, other)
+    return result
+
+
+@torch_op("aten::max", overload=True)
+# def _aten_max_with_dim(self: TReal, dim: int, keepdim: bool) -> tuple[TReal, TInt]:
```
justinchuby, line 3474: dead code

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3425,10 +3425,58 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_min(self: TensorType) -> TensorType:
+@torch_op("aten::max", trace_only=True)
+def aten_min(
+    self: TReal, dim_or_other: Union[TReal, INT64] = None, keepdim: BOOL = None
+) -> TReal:
     """min(Tensor self) -> Tensor"""
```
justinchuby, line 3432: Is this signature accurate from native_functions? Should we create overloads for different aten overloads?

justinchuby:
I will do a more in-depth review after the initial comments are addressed

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: should be overload for "aten::min"

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: yes, we have 3 overload

justinchuby: I meant public overloads

xiaowuhu: sorry, not got your idea ~

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3456,10 +3455,57 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
...
+
+    if output == 2:
+        if self_rank == 0:
+            indices = op.Squeeze(indices)  # type: ignore[has-type]
+        return result, indices
+    return result
+
+
+@torch_op("aten::min", overload=True)
+def _aten_min_with_no_dim(self: TReal) -> TReal:
```
justinchuby, line 3492: - func: min(Tensor self) -> Tensor
- func: min.other(Tensor self, Tensor other) -> Tensor
  device_check: NoCheck   # TensorIterator
  variants: method, function
  tags: pointwise
- func: min.dim(Tensor self, int dim, bool keepdim=False) -> (Tensor values, Tensor indices)
  device_check: NoCheck   # TensorIterator
  structured_delegate: min.dim_min
  variants: function, method
  dispatch:
    QuantizedCPU, QuantizedCUDA: qmin
  tags: core

- func: min.dim_min(Tensor self, int dim, bool keepdim=False, *, Tensor(a!) min, Tensor(b!) min_indices) -> (Tensor(a!) values, Tensor(b!) indices)
  device_check: NoCheck   # TensorIterator
  structured: True

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3456,10 +3455,57 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
...
+    result = op.ReduceMin(self, keepdims=0)
+    return result
+
+
+@torch_op("aten::min", overload=True)
+def _aten_min_with_other(self: TReal, other: TReal) -> TReal:
```
justinchuby, line 3498:
```suggestion
def aten_min_other(self: TReal, other: TReal) -> TReal:
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3456,10 +3456,52 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_min(self: TensorType) -> TensorType:
+@torch_op("aten::min")
+def aten_min(self: TReal) -> TReal:
     """min(Tensor self) -> Tensor"""
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
```
justinchuby, line 3463: I remember Rama saying reduce ops can handle scalar inputs. Did we try this without the rank conditions?

xiaowuhu: "scalar" means : 0 rank Tensor(scalar), or scalar something in in attribute?

justinchuby: I meant 0 rank Tensor sorry

xiaowuhu: works

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3456,10 +3456,52 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
...
+
+    return result, indices
+
+
+@torch_op("aten::min", overload=True)
+def aten_min_other(self: TReal, other: TReal) -> TReal:
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:
+        self = op.Reshape(self, op.Constant(value_ints=[-1]))
+
```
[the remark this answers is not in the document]
xiaowuhu: works

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3456,10 +3456,37 @@ def aten_meshgrid(tensors: Sequence[TensorType]) -> TensorType:
...
     """min(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.ReduceMin(self, keepdims=0)
+
+
+@torch_op("aten::min", overload=True)
+def aten_min_dim(self: TReal, dim: int, keepdim: bool = False) -> Tuple[TReal, TInt]:
+
+    neg_1 = op.Constant(value_ints=[-1])
```
justinchuby, line 3469: If input is rank 0, can we directly return self and index 0? Because there is only one element anyways? Better would there be a way to remove if? I see it’s a little tricky because we need to return index as well. Still thinking. Other parts LGTM

justinchuby:
Nice! So they do handle rank 0 tensors. I would do the same to max.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby:
```python
    self_rank = op.Size(op.Shape(self))
    if self_rank == 0:
        result = self
    else:
        ...
```

justinchuby: This if condition is known compile time so I think its fine

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -502,6 +505,23 @@ def _where_input_wrangler(
         matcher=lambda sample: sample.kwargs.get("rounding_mode") is not None,
         reason="rounding_mode is not yet supported",
     ),
+    skip(
+        "min",  # aten_mean
+        matcher=lambda sample: len(sample.args) > 0,
+        reason="only support one tensor as input",
```
justinchuby, line 511:
```suggestion
        reason="this ATen overload only supports one tensor as input by design",
```

for clarity so that we know it's by design. here and below

justinchuby (approved):
Looking great, thanks!

PR #506, opened 2023-03-07
Extend Hardmax example in docs to be end-to-end

shubhambhokare1:
- Show how to serialize an ONNX implementation to an ONNX graph
- Automatically update example based on changes made in example.py files


docs/tutorial/examples/hardmax_e2e.py
```diff
@@ -0,0 +1,36 @@
+import onnx
```
justinchuby, line 1: I would consider `hardmax_end_to_end.py` a good name with better readability

README.md
```diff
@@ -89,17 +89,18 @@ pytest onnxscript
...
 ```python
-from onnx import TensorProto
-from onnx.helper import make_tensor
-from onnxscript import script, INT64, FLOAT
+from onnx
 
 # We use ONNX opset 15 to define the function below.
-from onnxscript import opset15 as op
+from onnxscript.onnx_opset import opset15 as op
```
gramalingam, line 96: Why this change? I think it is better that we expose commonly used names (like opsets and types) in the top-level namespace. The internal organization is more of an implementation-issue that we can hide from external users.

docs/tutorial/examples/hardmax_e2e.py
```diff
@@ -0,0 +1,36 @@
...
+    zero = op.Constant(value_ints=[0])
+    depth = op.GatherElements(xshape, zero)
+    empty_shape = op.Constant(value_ints=[0])
+    depth = op.Reshape(depth, empty_shape)
+    values = op.Constant(value_ints=[0, 1])
+    cast_values = op.CastLike(values, X)
+    return op.OneHot(argmax, depth, cast_values, axis=axis)
+
+# onnx_model is an in-memory ModelProto
+onnx_model = onnx_hardmax.to_model_proto()
```
gramalingam, line 33: Some nits:
* This will produce a model with inputs/outputs of unknown shape (rank), which will fail the model-checker.
* This will also potentially fail because of the presence of the attribute-parameter `axis`.
* The above two are the key differences between a "function" and a "graph/model" in ONNX.

I think it would be better for a tutorial-example to avoid this confusion. So, the example needs to be modified to produce a valid model that will pass the model-checker.

docs/tutorial/examples/hardmax_end_to_end.py
```diff
@@ -0,0 +1,51 @@
...
+    empty_shape = op.Constant(value_ints=[0])
+    depth = op.Reshape(depth, empty_shape)
+    values = op.Constant(value_ints=[0, 1])
+    cast_values = op.CastLike(values, X)
+    return op.OneHot(argmax, depth, cast_values, axis=axis)
+
+# We use the script decorator to indicate that
+# this is meant to be translated to ONNX.
+@script()
+def SampleModel(X : FLOAT[64, 128] , Wt: FLOAT[128, 10], Bias: FLOAT[10]) -> FLOAT[64, 10]:
```
justinchuby, line 35: Use snake_case to adhere to pep8 maybe?

docs/tutorial/examples/hardmax_end_to_end.py
```diff
@@ -0,0 +1,51 @@
...
+    matmul = op.MatMul(X, Wt) + Bias
+    return onnx_hardmax(matmul, axis=1)
+
+# onnx_model is an in-memory ModelProto
+onnx_model = SampleModel.to_model_proto()
+
+# Save the ONNX model at a given path
+onnx.save(onnx_model, "hardmax.onnx")
```
gramalingam, line 43: nit: filename: "hardmax" => "SampleModel"  or "sample_model"?

docs/update_readme.py
```diff
@@ -0,0 +1,20 @@
+# Script to update end-to-end example in README.md.
```
gramalingam, line 1: I am a bit unclear how this is supposed to work. I assume this is intended to be run by a person creating a PR, and the actual file in the repo should be the *updated* file?

docs/update_readme.py
```diff
@@ -0,0 +1,20 @@
...
+    readme = f.readlines()
+    for line in readme:
+        if not in_stub:
+            updated_readme.append(line)
+        if line == "```python update-readme\n":
+            in_stub = True
+            with open('docs/tutorial/examples/hardmax_end_to_end.py', encoding="utf-8") as example_f:
+                example_code = example_f.readlines()
+                updated_readme += example_code
+        if line == "```\n" and in_stub:
```
gramalingam, line 15: Maybe more robust to have some clear end-marker (rather than an empty-line) ... something that will NOT be in the replacement code ... so that running this update twice has the same behavior as running it once

docs/update_readme.py
shubhambhokare1: This makes sense, I add a separate workflow to check diffs and instruct the user to run update_readme if the ci fails

docs/tutorial/examples/hardmax_e2e.py
shubhambhokare1: Would it be a good idea to demonstrate the relevant difference between inputs and attributes here?

docs/update_readme.py
shubhambhokare1: I guess a code block is of the format,
``` python code-block-name
code_stub
```
And the update_readme script only updates the code_stub section.

gramalingam: Sorry, this looks fine. Agree that three backticks should be fine, I misread it.

shubhambhokare1:
@gramalingam Anything else needed for the PR approval?

gramalingam (approved):

PR #507, opened 2023-03-07
better comment on expand node

titaiwangms:
Better comment on Expand node adjustment.
More detail can be found: https://github.com/pytorch/pytorch/pull/34069


justinchuby (approved):
Thanks!

PR #509, opened 2023-03-07
fix(test): bump onnxruntime, onnx, torch versions in noxfile.py

jcwchen:
I saw an error comes from certain Python 3.9 CI pipeline: `'/Users/runner/work/onnx-script/onnx-script/.nox/test/lib/python3.9/site-packages/onnxruntime/capi/onnxruntime_pybind11_state.so' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64)))`, which is fixed by the latest ONNX Runtime 1.14.1: https://github.com/microsoft/onnxruntime/issues/14663. Bumping onnxruntime version here should help. Meanwhile, also try to bump onnx and torch to the latest patch release, which should be harmless.

### DIFF
--- a/onnxscript/backend/onnx_export_test.py
+++ b/onnxscript/backend/onnx_export_test.py
@@ -14,7 +14,6 @@ from typing import Pattern
 import onnxruntime as ort
 import parameterized
 from onnxruntime.capi import onnxruntime_pybind11_state
-from packaging.version import Version
 
 import onnxscript
 from onnxscript.backend import onnx_backend, onnx_export
--- a/onnxscript/backend/onnx_export_test.py
+++ b/onnxscript/backend/onnx_export_test.py
@@ -160,9 +160,7 @@ class TestOnnxBackEnd(unittest.TestCase):
             onnxruntime_pybind11_state.InvalidArgument,  # pylint: disable=c-extension-no-member
         ) as e:
             self.skipTest(f"Unable to load the model: {e}")
-        except (
-            onnxruntime_pybind11_state.RuntimeException
-        ) as e:  # pylint: disable=c-extension-no-member
+        except onnxruntime_pybind11_state.RuntimeException as e:  # pylint: disable=c-extension-no-member
             self.skipTest(f"Unable to run the model: {e}")
         except AssertionError as e:
             self.skipTest(f"ORT result mismatches with the expected: {e}")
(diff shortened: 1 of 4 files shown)


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -465,12 +465,12 @@ def _where_input_wrangler(
     xfail(
         "logsumexp",
         reason="ONNX Runtime 1.13 does not support ReduceLogSumExp-18",
-        enabled_if=version_utils.onnxruntime_older_than("1.14"),
+        enabled_if=version_utils.onnxruntime_older_than("1.14.1"),
```
justinchuby, line 468: Do we need these two, since they are for 1.13?

onnxscript/tests/functions/onnxfns2_test.py
```diff
@@ -116,7 +116,7 @@ def test_onnxfns_hardmax(self):
     #         skip_test_names=[])
 
     @unittest.skipIf(
-        Version(onnxruntime.__version__) == Version("1.14"),
+        Version(onnxruntime.__version__) == Version("1.14.1"),
```
justinchuby, line 119: Check the major, minor version to make it more robust maybe?

onnxscript/backend/onnx_export_test.py
```diff
@@ -63,7 +63,7 @@ def skip(pattern: str | Pattern, reason: str, *, condition: bool = True):
     skip(
         r"^test_optional_get_element_tensor",
         "ORT Unable to create onnxruntime InferenceSession for executing .OptionalGetElement op with onnx model",
-        condition=ort.__version__ == "1.14.0",
+        condition=ort.__version__ == "1.14.1",
```
justinchuby, line 66: Maybe check major, minor versions?

justinchuby (approved):
thanks! just some nits

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
jcwchen: Good catch. Removing them seems good.

onnxscript/tests/functions/onnxfns2_test.py
jcwchen: I used `onnxruntime.__version__[:4] == "1.14",` instead. PTAL. Feel free to merge this PR if you are good with the updates.

PR #511, opened 2023-03-09
feat(atenlib): add op(chunk)

xiaowuhu:


justinchuby (approved):
LGTM! I would wait for Rama to review the parts in the evaluator

xiaowuhu:
> LGTM! I would wait for Rama to review the parts in the evaluator

@gramalingam pls help to review. The original implementation doesn't return valid output numbers and based on opset17(split).

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -955,10 +955,30 @@ def aten_choose_qparams_optimized(
...
-    raise NotImplementedError()
+    neg_1 = op.Constant(value_ints=[-1])
+    # get size of specified dim
+    self_shape = op.Shape(self)
+    dim_size = op.Gather(self_shape, dim, axis=0)
+    # cal size/chunk, get number of data in one chunk
+    num_per_chunk = op.Div(dim_size, chunks)
+    if op.Mod(dim_size, chunks) > 0:  # type: ignore[operator]
+        # cannot be divisible, then num_in_chunk + 1
+        num_per_chunk = op.Add(num_per_chunk, 1)
```
justinchuby, line 970: I wonder if using Where will be more efficient or not

justinchuby: Even better: `op.Cast(op.Mod(dim_size, chunks) > 0, to=INT64.value) + num_per_chunk`

justinchuby: The take away for me was `If` tends to be slow and should be replaced with other things when possible.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -955,10 +955,30 @@ def aten_choose_qparams_optimized(
...
+
+    # cal real chunk number
```
justinchuby, line 972:
```suggestion
    # calculate real chunk number
```

typo?

justinchuby: Trying to understand: why was the above not "real"?

justinchuby:
Also cc @BowenBao for the logic

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: I think it can also work on BOOL value

justinchuby:
We can merge if there's no movement

PR #513, opened 2023-03-10
feat(ateblib):Add Ops (all, isfinite, split_with_sizes)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -176,10 +176,30 @@ def aten_align_to(self: TensorType, names: Sequence[str]) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_all(self: TensorType) -> TensorType:
+@torch_op("aten::all")
```
justinchuby, line 179: Is this the same with any where we can create a few overloads? Since there isn’t a good way of determining whether an input is a static optional yet in I function I feel it’s better for us to avoid using optionalhaselement on static optional inputs now and should rather create overloads like we have in eg min

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -759,6 +762,9 @@ def test_output_match(self, device: str, dtype: torch.dtype, op):
                 inputs=repr(inputs),
                 kwargs=repr(cpu_sample.kwargs),
             ):
+                if i == 9:
```
justinchuby, line 765: Debug code

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2863,10 +2883,11 @@ def aten_isclose(
     raise NotImplementedError()
 
 
+@torch_op("aten::isfinite")
 def aten_isfinite(self: TensorType) -> TensorType:
     """isfinite(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.Not(op.IsInf(self))
```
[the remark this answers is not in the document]
xiaowuhu: fixed.  NaN is finite.

justinchuby (approved):
LGTM🚀

PR #514, opened 2023-03-10
chore(ci): run tests in random order

justinchuby:
https://pypi.org/project/pytest-randomly/

Tests are still reproducible locally by supplying the random seed from CI:

image

> Randomness in testing can be quite powerful to discover hidden flaws in the tests themselves, as well as giving a little more coverage to your system.
>
> By randomly ordering the tests, the risk of surprising inter-test dependencies is reduced - a technique used in many places, for example Google’s C++ test runner googletest. Research suggests that “dependent tests do exist in practice” and a random order of test executions can effectively detect such dependencies [1]. Alternatively, a reverse order of test executions, as provided by pytest-reverse, may find less dependent tests but can achieve a better benefit/cost ratio.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -28,6 +28,7 @@ COMMON_TEST_DEPENDENCIES = (
 ONNX = "onnx==1.13.1"
 ONNX_RUNTIME = "onnxruntime==1.14.1"
 PYTORCH = "torch==1.13.1"
+PYTEST_RANDOMLY = "pytest-randomly"
 
 
 @nox.session(tags=["build"])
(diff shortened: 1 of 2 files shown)


BowenBao:
Do our test cases have `det setUp` that fix the seed or no?

justinchuby:
We do. the random seed for numpy and torch is still fixed as intended. Only the order of the test run is random

noxfile.py
```diff
@@ -22,6 +22,7 @@
     "pytest-cov",
     "pytest-subtests",
     "pytest-xdist",
+    "pytest-randomly",
```
jcwchen, line 25: nit: make pytest-randomly in alphabetical order as below?

jcwchen (approved):

noxfile.py
justinchuby: Done!

PR #522, opened 2023-03-14
feat(atenlib): add op(index put)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2726,15 +2726,73 @@ def aten_index_copy(
...
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    # change array([1,3]) to array([[1,1,1,1,1],[3,3,3,3,3]])
+    self_rank_1 = op.Gather(op.Shape(self), 1)
+    index_rank_0 = op.Gather(op.Shape(index), 0)
+    neg_1 = op.Constant(value_ints=[-1])
+    shape = op.Concat(op.Reshape(self_rank_1, neg_1), op.Reshape(index_rank_0, neg_1), axis=0)
+    new_ind = op.Expand(index, shape)
+    new_ind_t = op.Transpose(new_ind)
+
+    if accumulate:
```
justinchuby, line 2747: This is a little tricky: We need to cast accumulate to BOOL when the attribute is used as a condition in if statements. This is because the attribute is internally represented as int, and becomes INT64 when implicitly casted to Input when participating in the logic. We may be able to update onnx script to do this cast automatically? @gramalingam

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2726,15 +2726,73 @@ def aten_index_copy(
...
-    self: TensorType,
-    indices: Optional[Sequence[TensorType]],
-    values: TensorType,
+    self: TReal,
+    indices: Sequence[TReal],
+    values: TReal,
     accumulate: bool = False,
-) -> TensorType:
+) -> TReal:
     """index_put(Tensor self, Tensor?[] indices, Tensor values, bool accumulate=False) -> Tensor"""
```
justinchuby, line 2736: Will indices be empty?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2726,15 +2726,73 @@ def aten_index_copy(
...
 
-    raise NotImplementedError()
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    # change array([1,3]) to array([[1,1,1,1,1],[3,3,3,3,3]])
+    self_rank_1 = op.Gather(op.Shape(self), 1)
```
justinchuby, line 2740: For my education, what do rank_1 and rank_0 mean / do?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2726,15 +2726,73 @@ def aten_index_copy(
     raise NotImplementedError()
 
 
+@torch_op("aten::index_put")
 def aten_index_put(
-    self: TensorType,
-    indices: Optional[Sequence[TensorType]],
-    values: TensorType,
+    self: TReal,
+    indices: Sequence[TReal],
```
justinchuby, line 2732: Does indices need to be INT64?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2726,15 +2726,73 @@ def aten_index_copy(
...
+def aten_index_put_bool(
+    self: TReal,
+    indices: Sequence[BOOL],
+    values: TReal,
+    accumulate: bool = False,
+) -> TReal:
+    """index_put(Tensor self, Tensor?[] indices, Tensor values, bool accumulate=False) -> Tensor"""
+
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    ind_float = op.Cast(index, to=FLOAT.dtype)
```
justinchuby, line 2767: Would it make sense to cast to INT64?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Looks like the doc says tensor(int32), tensor(int64)

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -327,6 +327,8 @@ def _where_input_wrangler(
     "full_like": core_ops.aten_full_like,
     "ge": core_ops.aten_ge,
     "gt": core_ops.aten_gt,
+    "index_put_bool": core_ops.aten_index_put_bool,
```
justinchuby, line 330: Nice tests. Thanks!

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: op.ReduceSum(ind_float) CANNOT accept INT64 type, although the doc says it can support INT/FLOAT.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: if indices is empty, this function will be meaningless. PyTorch will return error if it is empty.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: yes. you are right.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: rank_1 mean the size for dimension 1. rank_0 mean the size for dimension 0. 
shape(2,3),  rank_1 is 3, rank_0 is 2.
if you have a better name, pls share to me. :)

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: changed  to op.Cast(to=INT64.dtype)

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Maybe call it dim? Rank to me can only be of a tensor but not a dimension

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: I see. Should we include a comment about this in code? I think it’s something onnx runtime needs to support.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2701,10 +2701,10 @@ def aten_index_put(
 
     index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
     # change array([1,3]) to array([[1,1,1,1,1],[3,3,3,3,3]])
-    self_rank_1 = op.Gather(op.Shape(self), 1)
-    index_rank_0 = op.Gather(op.Shape(index), 0)
+    self_dim_1 = op.Gather(op.Shape(self), 1)
```
justinchuby, line 2704: Just out of curiosity: would Slice work too, or do we have to use Gather here?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2728,19 +2728,19 @@ def aten_index_put_bool(
     """index_put(Tensor self, Tensor?[] indices, Tensor values, bool accumulate=False) -> Tensor"""
 
     index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
-    ind_float = op.Cast(index, to=FLOAT.dtype)
+    ind_float = op.Cast(index, to=FLOAT.dtype)  # TODO: ReduceSum() cannot accept INT input
```
justinchuby, line 2731:
```suggestion
    # FIXME: ORT ReduceSum fails on INT input even though ONNX allows it
    ind_float = op.Cast(index, to=FLOAT.dtype
```

maybe something like

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Slice need more steps than Gather. And if the starts = -1, you don't know how to set the ends value.

justinchuby:
LMK if you are able to fix the test errors? I can take a look tomorrow too

xiaowuhu:
> LMK if you are able to fix the test errors? I can take a look tomorrow too

Just have a rest and go back to review this tmr.no hurry.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: I see. thanks!

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2690,15 +2690,74 @@ def aten_index_copy(
...
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    # change array([1,3]) to array([[1,1,1,1,1],[3,3,3,3,3]])
+    self_dim_1 = op.Gather(op.Shape(self), 1)
+    index_dim_0 = op.Gather(op.Shape(index), 0)
+    neg_1 = op.Constant(value_ints=[-1])
+    shape = op.Concat(op.Reshape(self_dim_1, neg_1), op.Reshape(index_dim_0, neg_1), axis=0)
+    new_ind = op.Expand(index, shape)
+    new_ind_t = op.Transpose(new_ind)
+
+    if op.Cast(accumulate, to=INT64.dtype):
```
gramalingam, line 2711: Is that meant to be `BOOL.dtype` ?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2690,15 +2690,74 @@ def aten_index_copy(
...
 
-    raise NotImplementedError()
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    # change array([1,3]) to array([[1,1,1,1,1],[3,3,3,3,3]])
+    self_dim_1 = op.Gather(op.Shape(self), 1)
+    index_dim_0 = op.Gather(op.Shape(index), 0)
+    neg_1 = op.Constant(value_ints=[-1])
+    shape = op.Concat(op.Reshape(self_dim_1, neg_1), op.Reshape(index_dim_0, neg_1), axis=0)
+    new_ind = op.Expand(index, shape)
+    new_ind_t = op.Transpose(new_ind)
```
gramalingam, line 2709: The `perm` attribute seems to be missing

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2690,15 +2690,74 @@ def aten_index_copy(
...
+    self: TReal,
+    indices: Sequence[INT64],
+    values: TReal,
     accumulate: bool = False,
-) -> TensorType:
+) -> TReal:
     """index_put(Tensor self, Tensor?[] indices, Tensor values, bool accumulate=False) -> Tensor"""
 
-    raise NotImplementedError()
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
```
gramalingam, line 2702: I am bit unclear about the assumption above. Is this just for now? Is the input type expected to be changed to `INT64` ?

justinchuby: Looks like indices is a list of 1d tensors, so INT64 may not do it?

justinchuby: On this line, was there a reason we assume indices only has one element?

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: I think it was optional?

justinchuby:
Graph builder fixed in https://github.com/microsoft/onnx-script/pull/528

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: From the test case, all the indices have only one element. If it contains multiple elements, we need to use for/while to do the logic. 
But for this aten op, the multiple elements can be combined to one element, for example, indices=[[0,1], [2,3]], can be combined to indices=[[0,1,2,3],].

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: got it

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Would we do a sequenceConcat or something like that?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2690,15 +2690,74 @@ def aten_index_copy(
...
+    values: TReal,
+    accumulate: bool = False,
+) -> TReal:
+    """index_put(Tensor self, Tensor?[] indices, Tensor values, bool accumulate=False) -> Tensor"""
+
+    index = op.SequenceAt(indices, 0)  # assume indices only have 1 element
+    # FIXME: ORT ArgMax fails on INT input even though ONNX allows it
+    ind_float = op.Cast(index, to=FLOAT.dtype)
+    # if all False, return self
+    if op.ReduceSum(ind_float) == 0:
```
justinchuby, line 2734: I think I was able to do ReduceMin on INT8. May be worth a try?

xiaowuhu: The ind_float will be used in ArgMax(ind_float) and it will return error on INT64, not ReduceMin().  I gave a wrong message last time.
I tried INT32, it works for ArgMax.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2690,15 +2690,74 @@ def aten_index_copy(
...
+            op.Reshape(self_dim_1, neg_1), op.Reshape(index_dim_0, neg_1), axis=0
+        )
+        new_ind = op.Expand(index, shape)
+        new_ind_t = op.Transpose(new_ind)
+
+        # values must have same rank with input(self)
+        if op.Size(op.Shape(values)) < op.Size(op.Shape(self)):  # type: ignore[operator]
+            values = op.Unsqueeze(values, op.Constant(value_ints=[0]))
+
+        if op.Cast(accumulate, to=INT64.dtype):
```
justinchuby, line 2753: Here as well

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: I tried in Torch, actually it cannot work like this:
torch.index_put( self=tensor(5x5), indices=(tensor([0]), tensor([1]),), values=tensor(2,5))
so we don't know how to use more than one element in indices tuple.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: on the way.

justinchuby (approved):

PR #528, opened 2023-03-16
fix(atenlib): test split and support Sequence input

justinchuby:
This PR adjusts the test code to be able to handle the output of `torch.split` and adds support for sequence as input. Note that this is different from the exporter's concat for symints, because at the point when we have inputs processed and still get list of Tensors as a single argument, we know it should be an ONNX Sequence tensor in the graph.

We also need to keep the addition in here and not in the exporter to match behavior with onnxscript eager: we accept a list of tensors as input when the op takes a Sequence.

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -397,7 +397,7 @@ class TorchScriptGraph:
         graph_inputs = []
         assert isinstance(unwrapped_inputs, Sequence)
         for input in unwrapped_inputs:
-            if isinstance(input, Sequence) and any(
+            if isinstance(input, Sequence) and all(
                 isinstance(elem, torch.Value) for elem in input
             ):
                 input_sequence = _create_op_call_in_torch_graph(
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -397,7 +397,17 @@ class TorchScriptGraph:
         graph_inputs = []
         assert isinstance(unwrapped_inputs, Sequence)
         for input in unwrapped_inputs:
-            if not isinstance(input, torch.Value):
+            if isinstance(input, Sequence) and any(
+                isinstance(elem, torch.Value) for elem in input
+            ):
+                input_sequence = _create_op_call_in_torch_graph(
+                    self._torch_graph,
+                    "onnx::SequenceConstruct",
+                    inputs=input,
+                    attributes={},
+                )[0]
+                graph_inputs.append(input_sequence)
+            elif not isinstance(input, torch.Value):
                 graph_inputs.append(self._add_constant_to_graph(input))
             else:
                 graph_inputs.append(input)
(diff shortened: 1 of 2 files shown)


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1148,11 +1130,20 @@ def test_output_match(self, device: str, dtype: torch.dtype, op):
...
+                    # value. So we wrap it into a tuple.
+                    # TODO(justinchuby): Find a more general solution
+                    flattened_torch_outputs = (flattened_torch_outputs,)
+
                 function_output = self.function_executor(flattened_torch_outputs)(
                     onnx_function, input_onnx, kwargs_onnx
                 )
+                # Finally we re-flatten everything
+                # TODO: add pytree structure comparison.
+                flattened_torch_outputs, _ = pytree.tree_flatten(torch_output)
```
xiaowuhu, line 1146: same line at 1133, is it by design?

xiaowuhu (approved):
Leaving a comments. others LGTM. Thanks!

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Yes, because 1139 changed the shape for the function_executor. I should probably rename it

justinchuby:
Should I make split trace only instead? It seems suitable with its dynamic output nature. But then we would be able to support the symint. To me that seems ok

justinchuby:
> Should I make split trace only instead? It seems suitable with its dynamic output nature. But then we would be able to support the symint. To me that seems ok

No because when there is `sizes` the output number is always dynamic. I wonder what the traced graph from fx looks like? @BowenBao

justinchuby:
I will merge to unblock other prs and create an issue for discussions.

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -397,7 +397,19 @@ def _add_torchscript_op_call(
...
         for input in unwrapped_inputs:
-            if not isinstance(input, torch.Value):
+            if isinstance(input, Sequence) and all(
+                isinstance(elem, torch.Value) for elem in input
+            ):
+                # If all elements in the Sequence are torch.Values we know it
+                # should be a Sequence input in ONNX.
+                input_sequence = _create_op_call_in_torch_graph(
+                    self._torch_graph,
+                    "onnx::SequenceConstruct",
```
titaiwangms, line 407: I think it would make more sense to move this to PyTorch side, like we did on op.Concat? We should process all the list related stuff in the same place. Especially, fx.node might not always be tensors.

justinchuby: I think this is different because we annotate the type as `Sequence[Tensor]`, whereas for the Concat case we annotated as `INT64`. We can chat more

justinchuby: Well, the current type system we use shows that the functions take a (python) Sequence. If we do SequenceConstruct outside, the input will not be a Sequence. However, we can also extend the return value of SequenceConstruct to be a proper Sequence so that we can do what you suggested

PR #530, opened 2023-03-16
feat(atenlib):add ops(isnan,issamesize,isneginf,isposinf,isnonzero)

xiaowuhu:


justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2850,10 +2850,16 @@ def aten_is_neg(self: TensorType) -> bool:
     raise NotImplementedError()
 
 
-def aten_is_nonzero(self: TensorType) -> bool:
+@torch_op("aten::is_nonzero")
+def aten_is_nonzero(self: Union[TReal, BOOL]) -> BOOL:
```
justinchuby, line 2854:
```suggestion
def aten_is_nonzero(self: Union[RealType, BOOL]) -> BOOL:
```

I feel taking the union of a TypeVar with another type is going to create complications for us when we parse them

PR #531, opened 2023-03-17
chore(ci): reenable mypy

justinchuby:
Hopefully it works this time


justinchuby:
So it is still very slow

PR #532, opened 2023-03-17
fix(graph_building): change `add_initializer` to return a Tensor; make shape inference optional

wschin:
1. Initializer must be an input. Otherwise, there is no torch._C.Value to repreent it in torch._C.Graph.
2. Shape inference fails for custom op, so it should be optional until we fix it.

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -472,7 +472,7 @@ class TorchScriptGraph:
         return result
 
     @beartype
-    def to_model_proto(self, opset_version: int, run_onnx_shape_inference: bool = True) -> onnx.ModelProto:
+    def to_model_proto(self, opset_version: int) -> onnx.ModelProto:
         (
             proto,
             _,
@@ -498,11 +498,11 @@ class TorchScriptGraph:
         for onnx_function in self._function_store.values():
             function_proto_list.append(onnx_function.to_function_proto())
         onnx_model.functions.extend(function_proto_list)
-        if run_onnx_shape_inference:
+
+        try:
             onnx_model = onnx.shape_inference.infer_shapes(
                 onnx_model, check_type=True, strict_mode=False, data_prop=True
             )
-        try:
             onnx.checker.check_model(onnx_model, full_check=True)
         except (onnx.checker.ValidationError, onnx.shape_inference.InferenceError) as e:
             warnings.warn(f"ONNX model is invalid: {e}")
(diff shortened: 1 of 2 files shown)


justinchuby:
Curious when would an initializer be an input?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -332,8 +332,9 @@ def add_input(
         return tensor_value  # type: ignore[return-value]
 
     @beartype
-    def add_initializer(self, input_name: str, input_value: torch.Tensor) -> None:
+    def add_initializer(self, input_name: str, input_value: torch.Tensor) -> TorchScriptTensor:
         self._initializers[input_name] = input_value
+        return self.add_input(input_name, input_value)
```
BowenBao, line 337: Looking at this again. It would make more sense to keep this code and do the "initializer and input" refactoring later when we have a dedicated onnx IR, than doing it on torch script exporter `export_onnx`. I think a comment here for clarification would be nice.

It is also a good opportunity to clean up related code in exporter side. https://github.com/pytorch/pytorch/blob/b97ce3774a403444d21845b0f42e2b98d8faa0f3/torch/onnx/_internal/fx/passes/fx_to_onnxscript.py#L360-L382

At least 3 issues:
1. Initializers are duplicated every time for a `get_attr` on same parameter.
2. `add_input` call separately outside of `add_initializer`.
3. `current_attr` type confusion..

BowenBao: On a higher level, this requirement means you are operating on 'TorchScript' IR level. If you are operating on ModelProto you won't need this API. Curious if there is a particular need for the process to happen on 'TorchScript' graph? @wschin

BowenBao:
It'd be nice to have a more informative title

onnxscript/function_libs/torch_aten/graph_building.py
justinchuby, line 337:
```suggestion
        if input_name in self._initializers:
            raise ValueError(f"Initializer {input_name} exists already")
        self._initializers[input_name] = input_value
        torch_value = self._torch_graph.addInput(input_name)
        torch_value.setType(torch.TensorType.create_from_tensor(input_value))
        tensor_value = _wrap_torch_value_to_tensor(torch_value)
        return tensor_value
```

justinchuby:
I am going to work on the same PR.

justinchuby:
@BowenBao ptal

justinchuby:
unexpected success 😃

BowenBao:
@justinchuby We still need the exporter side changes. Input is added twice for initializer. 

This PR LGTM. It'd be great to add unittest covering the desired use case of adding initializer and using its returned value. Non blocking.

BowenBao (approved):

justinchuby:
@titaiwangms could you help making this change in https://github.com/pytorch/pytorch/pull/96350?

> This PR LGTM. It'd be great to add unittest covering the desired use case of adding initializer and using its returned value. Non blocking.

PR #534, opened 2023-03-18
feat(atenlib): add op(unfold)

xiaowuhu:
Close this one, see https://github.com/microsoft/onnxscript/pull/893

### DIFF
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -5767,7 +5767,8 @@ def aten_unfold(self: TTensor, dimension: int, size: int, step: int) -> TTensor:
         # perm = [0,1,2,3,4]
         # perm.append(perm.pop(dimension+1))
         rank_result = op.Squeeze(op.Size(op.Shape(concat_result)))
-        seq_perm = op.SequenceEmpty()  # FIXME: op.SequenceEmpty(dtype=INT64.dtype) doesn't work
+        # FIXME: op.SequenceEmpty(dtype=INT64.dtype) doesn't work
+        seq_perm = op.SequenceEmpty()
         dim_plus_1 = dimension + 1
         for i in range(0, rank_result):
             if i != dim_plus_1:
--- a/onnxscript/function_libs/torch_aten/ops/core.py
+++ b/onnxscript/function_libs/torch_aten/ops/core.py
@@ -5766,18 +5766,15 @@ def aten_unfold(self: TTensor, dimension: int, size: int, step: int) -> TTensor:
         # below logic equal to:
         # perm = [0,1,2,3,4]
         # perm.append(perm.pop(dimension+1))
+
         rank_result = op.Squeeze(op.Size(op.Shape(concat_result)))
-        # FIXME: op.SequenceEmpty(dtype=INT64.dtype) doesn't work
-        seq_perm = op.SequenceEmpty()
         dim_plus_1 = dimension + 1
-        for i in range(0, rank_result):
-            if i != dim_plus_1:
-                idx_int = op.Constant(value_int=i)
-                idx_float = op.Cast(idx_int, to=FLOAT.dtype)
-                seq_perm = op.SequenceInsert(seq_perm, idx_float)
-        seq_perm = op.SequenceInsert(seq_perm, op.Cast(dim_plus_1, to=FLOAT.dtype))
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5740,10 +5740,43 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
...
+    if self_rank == 0:
+        result = op.Unsqueeze(self, 0)
+    else:
+        dims = op.Constant(value_ints=[dimension])
+        dim_size = op.Gather(self_shape, dims, axis=0)
+        # target = ((i-size)/step + 1) * step != i-size+step
+        target_end = op.Squeeze(((dim_size - size) / step + 1) * step)
+        seq_result = op.SequenceEmpty()
+
+        for i in range(0, target_end, step):
```
justinchuby, line 5758: @gramalingam Do you have recommendation on how this loop should be created?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5756,10 +5756,44 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
...
+def aten_unfold(self: TTensor, dimension: int, size: int, step: int) -> TTensor:
     """unfold(Tensor(a) self, int dimension, int size, int step) -> Tensor(a)"""
 
-    raise NotImplementedError()
+    self_shape = op.Shape(self)
+    self_rank = op.Size(self_shape)
+    if self_rank == 0:
+        result = op.Unsqueeze(self, 0)
+    else:
+        dims = op.Constant(value_ints=[dimension])
```
justinchuby, line 5768: We may need to use Expand because Constant requires a compile time constant

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5756,10 +5756,44 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
...
+        dim_size = op.Gather(self_shape, dims, axis=0)
+        # target = ((i-size)/step + 1) * step != i-size+step
+        target_end = op.Squeeze(((dim_size - size) / step + 1) * step)
+        seq_result = op.SequenceEmpty()
+
+        for i in range(target_end):
```
justinchuby, line 5774: Using trace only will be tricky because we don’t know what target end is at trace time

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5815,10 +5815,44 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
...
+            if op.Mod(i, step) == 0:
+                starts = op.Constant(value_ints=[i])
```
justinchuby, line 5835: Maybe assign this outside of the for loop first

gramalingam, line 5835: This won't work: attribute-values can NOT depend on runtime-values (like `i`). But this is not necessary. We can use `i`, perhaps with a `Unsqueeze` or `Reshape` to convert a 0D tensor into a 1D tensor.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5815,10 +5815,44 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_unfold(self: TensorType, dimension: int, size: int, step: int) -> TensorType:
+@torch_op("aten::unfold", trace_only=True)  # FIXME: Seems ast.For was not supported
+def aten_unfold(self: TTensor, dimension: int, size: int, step: int) -> TTensor:
```
gramalingam, line 5819: I don't understand what the op is supposed to do. Is there a description I can read somewhere? Looks like it is a variant of a slice-like op with params (size, step) in (dimension), followed by some form of transpose?

xiaowuhu: The unfold do below things:
x = [1,2,3,4,5,6], 
unfold it to [1,2],[2,3],[3,4]... when stride=2,step=1
unfold it to [1,2],[3,4],[5,6] when stride=2,step=2
it is a Core aten op.

xiaowuhu:
> It doesn't seem like unfold is really being used these days. https://pytorch.org/docs/stable/ir.html Maybe we can just skip it?

Agree.

justinchuby:
Ok so it turns out we still need it in #783

PR #543, opened 2023-03-21
Add type promotion on certain nodes to enable bloom

titaiwangms:
I had to add these every time I made graph-level change...


justinchuby (approved):
sure

PR #551, opened 2023-03-23
feat(atenlib):add ops(scatter-add, floor)

xiaowuhu:
ORT failed on FullGraph test mode, because if-else branch have different shape (rank 0 vs. rank N)

### DIFF
--- a/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
@@ -416,6 +416,7 @@ OPINFO_FUNCTION_MAPPING_SCRIPTED: dict[
     "expand_as": core_ops.aten_expand_as,
     "erf": core_ops.aten_erf,
     "fill": core_ops.aten_fill,
+    "floor": core_ops.aten_floor,
     "fmod": core_ops.aten_fmod,
     "full": core_ops.aten_full,
     "full_like": core_ops.aten_full_like,
--- a/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
+++ b/onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
@@ -1204,11 +1204,6 @@ def run_test_output_match(
             inputs=repr(inputs),
             kwargs=repr(cpu_sample.kwargs),
         ):
-            if i == 0:
-                print(i)
-            else:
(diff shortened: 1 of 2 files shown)


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -337,6 +337,13 @@ def _permute_input_wrangler(
     return args, kwargs
 
 
+def _scatter_add_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    kwargs["dim"] = args.pop(1)
```
justinchuby, line 343: Curious why this is needed?

xiaowuhu: The original test case is:  self(tensor), dim( [int, int, ...]), index(tensor), src(tensor). 
so the dim in the second place, cannot not be parsed.
if we use wrangler to change it to Tensor, the follow op.ScatterElements(self, index, src, axis=dim) will be failed, due to dim is not an attribute.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -671,6 +680,11 @@ def _where_input_wrangler(
     xfail(
         "round", variant_name="decimals_neg_3", reason="The op does not support decimals yet"
     ),
+    xfail(
+        "scatter_add",
+        reason="fixme: ORT failed due to different rank for if-else branch",
```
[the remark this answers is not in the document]
xiaowuhu: skipped

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -868,6 +863,11 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.args[0]) == 0,
         reason="Empty perm is not supported",
     ),
+    skip(
+        "scatter_add",
+        matcher=lambda sample: len(sample.input.shape) == 0,
+        reason="Rank(0) input will lead ORT failed due to different rank(result) in if-else branch",
```
justinchuby, line 880: fixme:

justinchuby (approved):

PR #552, opened 2023-03-23
feat(atenlib): add ops(flip)

xiaowuhu:


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -587,6 +588,11 @@ def _where_input_wrangler(
     xfail(
         "chunk", reason="fixme: ORT error", test_class_name="TestOutputConsistencyFullGraph"
     ),
+    xfail(
```
justinchuby, line 591: @titaiwangms i think your fix will handle this

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2313,10 +2313,18 @@ def aten_fix(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_flip(self: TensorType, dims: Sequence[int]) -> TensorType:
+@torch_op("aten::flip")
+def aten_flip(self: TTensor, dims: Sequence[int]) -> TTensor:
```
justinchuby, line 2317:
```suggestion
def aten_flip(self: TTensor, dims: INT64) -> TTensor:
```

i would take dims as INT64

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2313,10 +2313,18 @@ def aten_fix(self: TensorType) -> TensorType:
...
     """flip(Tensor self, int[] dims) -> Tensor"""
 
-    raise NotImplementedError()
+    dims_tensor = op.Constant(value_ints=dims)
```
justinchuby, line 2320: This should fail as Constant requires a compile time constant.

xiaowuhu: So what should we do for this scenario?

titaiwangms: Kind of like this: https://github.com/microsoft/onnx-script/blob/1e8d764a9be04323d7171e4d5f511332790cb809/onnxscript/function_libs/torch_aten/ops/core.py#L5503-L5513

justinchuby: We can use expand instead

justinchuby: But here we should actually expend an INT64 tensor

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: The original test case defined: self is input, dims is kwargs {"dims": (0,1,2)}, it is about dimension list, not a tensor.
[Here](https://pytorch.org/docs/master/generated/torch.flip.html?highlight=flip#torch.flip) shows the torch document, like this:
```
torch.flip(x, [0, 1])
```
So I think "attribute" is more appropriate.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: > 

Thanks @titaiwangms , but, seems the aten_sym_size is not enabled the testing, because no test case in OPS_DB.
I test it with below code:
```
import numpy as np
a = np.random.rand(2,3)
dim = 1
r=aten_sym_size(a, dim)
print(r)
exit(0)
```
It will failed at line 5511 and 5512.  The should be op.Reshape(dim, op.Constant(value_ints=[1])).

But for my case: the dim is a tuple (0,1,2), op.Reshape() or op.Expand() both cannot work.

xiaowuhu: > We can use expand instead

op.Expand(A, shape) cannot work when A is a list.

justinchuby: Can we annotate dims as INT64? This way it will be converted to a tensor

xiaowuhu: I skip the test case when dims=(empty) because it is meaningless, torch also return error for this case.

xiaowuhu: fixed

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -2313,10 +2313,17 @@ def aten_fix(self: TensorType) -> TensorType:
...
+@torch_op("aten::flip")
+def aten_flip(self: TTensor, dims: INT64) -> TTensor:
     """flip(Tensor self, int[] dims) -> Tensor"""
 
-    raise NotImplementedError()
+    shape_dim = op.Shape(dims)
+    neg_1 = op.Constant(value_int=-1)
+    starts = op.Expand(neg_1, shape_dim)  # something like [-1, -1, -1]
+    steps = op.Expand(neg_1, shape_dim)  # something like [-1, -1, -1]
+    ends = starts * 65535  # something like [-65535, -65535, -65535]
```
titaiwangms, line 2324: @xiaowuhu @justinchuby I wonder if we want INT64_MAX here.

justinchuby: We do. I remember something like this was updated recently. let me check.

justinchuby: ok so it wasn't

justinchuby: https://github.com/microsoft/onnx-script/issues/607

PR #558, opened 2023-03-26
Add ops(hardtanh, slice_scatter) | feat(atenlib)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -529,10 +529,12 @@ def aten_hardswish_backward(grad_output: TensorType, self: TensorType) -> Tensor
...
 
 
-def aten_hardtanh(self: TensorType, min_val: float = -1.0, max_val: float = 1.0) -> TensorType:
+@torch_op("aten::hardtanh")
+def aten_hardtanh(self: TReal, min_val: float = -1.0, max_val: float = 1.0) -> TReal:
     """hardtanh(Tensor self, Scalar min_val=-1, Scalar max_val=1) -> Tensor"""
 
-    raise NotImplementedError()
+    result = op.Clip(self, min_val, max_val)
+    return result
```
justinchuby, line 537:
```suggestion
    return op.Clip(self, min_val, max_val)
```

simply ^

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5336,17 +5336,52 @@ def aten_slice_copy(
...
-    start: Optional[INT64] = None,
-    end: Optional[INT64] = None,
-    step: INT64 = 1,
-) -> TensorType:
+    self: TTensor,
+    src: TTensor,
+    dim: int,
+    start: int,
+    end: int,
+    step: int = 1,
```
justinchuby, line 5346:
```suggestion
    dim: int=0,
    start: Optional[INT64]=None,
    end: Optional[INT64]=None,
    step: INT64 = 1,
```

start etc can be symint (dynamic) it looks like. Do we also want to consider None as inputs?

xiaowuhu: Althouth 'start' and 'end' can be None in signature, but actually 'start' must be specified
Assert(start is not None)
And, 'end' also must be specified, and end-start must be equal to the size of 'src'
Assert(end-start == shape(src) > 0)
Try torch sample to get more information:
https://pytorch.org/docs/master/generated/torch.slice_scatter.html?highlight=slice_scatter#torch.slice_scatter

justinchuby: Why do we need them to be int?

xiaowuhu: From signature, they can be INT64, both INT64 and int can works well.
From fact intention, they are all represent to scalar(index).
Even we set dim: Optional[INT64] in the public function, we still need to set dim: int in the private function, because op.ScatterElements(... axis=dim) still need dim as int type.

justinchuby: ok dim needs to be int. Can we set start and end to be INT64 since they are SymInts in the signature? In PyTorch SymInts are dynamic.

xiaowuhu: OK.

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5365,17 +5365,58 @@ def aten_slice_copy(
...
 
 
+@torch_op("aten::slice_scatter", trace_only=True)
 def aten_slice_scatter(
-    self: TensorType,
-    src: TensorType,
-    dim: int = 0,
+    self: TTensor,
+    src: TTensor,
+    dim: Optional[int] = 0,
```
justinchuby, line 5372:
```suggestion
    dim: int = 0,
```

Looks like dim is not optional

PR #559, opened 2023-03-27
feat(atenlib): add ops(avg pool2d)

xiaowuhu:
Have to use the trace_only = True because of complicated parameters input.


onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -676,6 +677,11 @@ def _where_input_wrangler(
         reason="fixme: ORT fails with invalid model: 'ONNX Schema aten_adaptive_avg_pool3d: failed validating the check: !(it.GetName().empty())'",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
+        "nn.functional.avg_pool2d",
```
justinchuby, line 681: is this needed since it’s skipped below already?

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -125,18 +125,62 @@ def aten_adaptive_max_pool3d_backward(
...
 
+@torch_op("aten::avg_pool2d", trace_only=True)
 def aten_avg_pool2d(
-    self: TensorType,
+    self: TFloat,
     kernel_size: Sequence[int],
-    stride: Optional[Sequence[int]] = None,
-    padding: Sequence[int] = (0, 0),
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = 0,
```
titaiwangms, line 133:
```suggestion
    padding: Sequence[int] = (0,),
```

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -682,6 +683,11 @@ def _where_input_wrangler(
...
+        reason="fixme: graph_building.py assert not isinstance(value, TorchScriptTensor)",
+        test_class_name="TestOutputConsistencyFullGraph",
+    ),
```
justinchuby, line 690:
```suggestion
```

Removing to see what happens. This need to work.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -125,18 +125,62 @@ def aten_adaptive_max_pool3d_backward(
     raise NotImplementedError()
 
 
+@torch_op("aten::avg_pool2d", trace_only=True)
 def aten_avg_pool2d(
-    self: TensorType,
+    self: TFloat,
     kernel_size: Sequence[int],
```
justinchuby, line 131: We may need to use a wrangler to change the np arrays back to a list, if `kernel_size`, `stride` and `padding` are supplied as np.arrays. I suspect this was why there was an assertion error in the full graph test - we passed them as inputs mistakenly because they were preprocessed as np arrays (since they are supplied as positional arguments).

We will soon be able to distinguish between inputs and arguments even better after the OpSchema PR (https://github.com/onnx/onnx/pull/5020) in onnx is merged.

justinchuby (approved):

PR #560, opened 2023-03-27
Add ops(maxpool2d with index) | feat(atenlib)

xiaowuhu:


justinchuby:
Was this ready for review?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,87 @@ def aten_max_pool1d_with_indices(
...
 def aten_max_pool2d(
     self: TensorType,
     kernel_size: Sequence[int],
-    stride: Optional[Sequence[int]] = None,
-    padding: Sequence[int] = (0, 0),
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = 0,
     dilation: Sequence[int] = (1, 1),
     ceil_mode: bool = False,
+    return_indices: bool = False,
```
justinchuby, line 3583:
```suggestion
```

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -564,6 +564,7 @@ def _where_input_wrangler(
     "empty_like": core_ops.aten_empty_like,
     "index_select": core_ops.aten_index_select,
     "layer_norm": core_ops.aten_layer_norm,
+    "nn.functional.max_pool2d": core_ops.aten_max_pool2d,
```
fatcat-z, line 567: Let's move this to line 577, after nn.functional.linear op.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,87 @@ def aten_max_pool1d_with_indices(
     raise NotImplementedError()
 
 
+@torch_op("aten::max_pool2d", trace_only=True)
 def aten_max_pool2d(
     self: TensorType,
```
titaiwangms, line 3577: Should be: tensor(float16), tensor(float), tensor(double), tensor(int8), tensor(uint8)

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,87 @@ def aten_max_pool1d_with_indices(
...
     kernel_size: Sequence[int],
-    stride: Optional[Sequence[int]] = None,
-    padding: Sequence[int] = (0, 0),
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = 0,
     dilation: Sequence[int] = (1, 1),
```
titaiwangms, line 3581: @justinchuby Found a conflict. In this case, we should annotate dilation as Sequnece[int], but aten signature has default dilation=1. Do we annotate it as `dilation: sequence[int] = (1,)` ?

```yaml
max_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, int[2] dilation=1, bool ceil_mode=False) -> Tensor
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,87 @@ def aten_max_pool1d_with_indices(
...
 
 
+@torch_op("aten::max_pool2d", trace_only=True)
 def aten_max_pool2d(
     self: TensorType,
     kernel_size: Sequence[int],
-    stride: Optional[Sequence[int]] = None,
-    padding: Sequence[int] = (0, 0),
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = 0,
```
justinchuby, line 3580:
```suggestion
    padding: Sequence[int] = (0, 0),
```

ok to keep (0, 0)?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,87 @@ def aten_max_pool1d_with_indices(
...
     dilation: Sequence[int] = (1, 1),
     ceil_mode: bool = False,
+    return_indices: bool = False,
 ) -> TensorType:
     """max_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, int[2] dilation=1, bool ceil_mode=False) -> Tensor"""
 
-    raise NotImplementedError()
+    self_len_org = len(self.shape)
```
justinchuby, line 3587: For this function I suggest using a trace_only to handle all python manipulation, then use a orivate, non-trace_only function to handle the bulk of the logic.

onnxscript/function_libs/torch_aten/ops/nn.py
```diff
@@ -573,6 +573,7 @@ def aten_max_pool2d_with_indices(
     padding: Sequence[int] = (0, 0),
     dilation: Sequence[int] = (1, 1),
     ceil_mode: bool = False,
+    return_indices: bool = True,
```
justinchuby, line 576:
```suggestion
```

because it is not part of the aten signature

justinchuby: We should filter the tests to take `return_indices=True` tests only for this op

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -573,6 +574,7 @@ def _where_input_wrangler(
     "nn.functional.conv3d": core_ops.aten_conv3d,
     "nn.functional.gelu": nn_ops.aten_gelu,
     "nn.functional.linear": nn_ops.aten_linear,
+    # "nn.functional.max_pool2d_with_indices": nn_ops.aten_max_pool2d_with_indices,  # no test case in OPS_DB
```
justinchuby, line 577: Duplicate `nn.functional.max_pool2d` and filter on `return_indices=True` tests

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -676,6 +678,11 @@ def _where_input_wrangler(
         reason="fixme: ORT fails with invalid model: 'ONNX Schema aten_adaptive_avg_pool3d: failed validating the check: !(it.GetName().empty())'",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
+        "nn.functional.max_pool2d",
+        reason="fixme: no idea how to get length in scalar",
+        test_class_name="TestOutputConsistencyFullGraph",
+    ),
```
justinchuby, line 685:
```suggestion
```

Graph mode needs to work

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: No matter what it is, but should be align with padding's format.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: No matter what it is, but should be align with dialtion's format.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
+    kernel_size: Sequence[int],
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = (0, 0),
+    dilation: Sequence[int] = (1, 1),
+    ceil_mode: bool = False,
+) -> Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8]:
+    """max_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, int[2] dilation=1, bool ceil_mode=False) -> (Tensor, Tensor)"""
+
+    self_len = len(self.shape)
+    expand_size = 4
```
justinchuby, line 3663: A comment explaining this logic for deciding `expand_size` would be great!

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
+    )
+
+    if self_rank == 3:
+        pool_result = op.Squeeze(pool_result, op.Constant(value_ints=[0]))
+
+    # Torch use relative position number for the second Channel data
+    # If align, need reduce size(Channel)
+    # e.g. [[8,3,10],[30,32,23]]-[0,18] -> [[8,3,10],[12,14,5]]
+    # 18 = H x W = 3 x 6
+    N = op.Shape(self, start=0, end=1)
```
justinchuby, line 3728:
```suggestion
    batches = op.Shape(self, start=0, end=1)
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
+    C = op.Shape(self, start=1, end=2)
```
justinchuby, line 3729:
```suggestion
    channels = op.Shape(self, start=1, end=2)
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
     raise NotImplementedError()
 
 
+@torch_op("aten::max_pool2d", trace_only=True)
 def aten_max_pool2d(
-    self: TensorType,
+    self: Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8],
```
justinchuby, line 3577: I would create a TypeVar for this (like TFloat, etc ... that we have) so we know the input type and the output type are the same

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
     padding: Sequence[int] = (0, 0),
     dilation: Sequence[int] = (1, 1),
     ceil_mode: bool = False,
-) -> TensorType:
+) -> Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8]:
     """max_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, int[2] dilation=1, bool ceil_mode=False) -> Tensor"""
 
-    raise NotImplementedError()
+    self_len = len(self.shape)
+    expand_size = 4
```
fatcat-z, line 3587: I  think we don't need a default value for expand so that it can handle an input whose length is not in [3, 4].
This function is specifically designed for max_pool2d which requires an input whose length is 3 or 4.  For any other case, I think we should fail the conversion as soon as possible instead of having a default value to make the conversion pass but ort fail later.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
+    kernel_shape: Sequence[int],
+    strides: Sequence[int],
+    pads: Sequence[int],
+    dilations: Sequence[int],
+    ceil_mode: bool,
+) -> Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8]:
+
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 3:
+        self = op.Unsqueeze(self, axes=0)
```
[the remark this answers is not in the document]
xiaowuhu: It works. But change to axes=op.Constant(value_ints=[0]) is better.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3572,175 @@ def aten_max_pool1d_with_indices(
...
+    self: Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8],
+    kernel_size: Sequence[int],
+    stride: Sequence[int] = None,
+    padding: Sequence[int] = (0, 0),
+    dilation: Sequence[int] = (1, 1),
+    ceil_mode: bool = False,
+) -> Union[DOUBLE, FLOAT, FLOAT16, INT8, UINT8]:
+    """max_pool2d(Tensor self, int[2] kernel_size, int[2] stride=[], int[2] padding=0, int[2] dilation=1, bool ceil_mode=False) -> (Tensor, Tensor)"""
+
+    self_len = len(self.shape)
```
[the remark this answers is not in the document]
xiaowuhu: Yes, changed.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: TFloatOrUInt8 is created, 'UInt8' means Uint8 and Int8.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: set expand_size = 2

justinchuby (approved):

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -3572,17 +3573,180 @@ def aten_max_pool1d_with_indices(
...
+    else:
+        strides = stride
+
+    return _aten_max_pool2d_with_indices_onnx(
+        self, expand_size, kernel_shape, strides, pads, dilations, ceil_mode
+    )
+
+
+@torch_op("aten::max_pool2d", private=True)
+def _aten_max_pool2d_with_indices_onnx(
```
titaiwangms, line 3705: This function is breaking CI @justinchuby @xiaowuhu. It seems like the ONNXFunction is invalid. Maybe `if self.rank == 3` makes this function trace_only?

justinchuby: Good catch. So instead of making it trace_only, we should find ways to script the function. In this case rank = op.Size(op.Shape())

justinchuby:
https://github.com/microsoft/onnx-script/pull/606

PR #569, opened 2023-03-28
feat(atenlib): addOp(var,vardim,varcorrection)

xiaowuhu:
Temporarily disabled shape_inference and check_model checking.


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+        correction = 1
+
+    if dim is None:
+        var, mean = _aten_var_mean(self, correction, keepdim)
+    else:
+        var, mean = _aten_var_mean_dim(self, dim, correction, keepdim)
+    return var, mean
+
+
+@torch_op("aten::var_mean", private=True)
```
fatcat-z, line 6059: As we discussed before, let's use "internal" as the namespace of such function which was set private=True.

xiaowuhu: I think private=True and "internal" are the same meaning.  Do we already have internal attribute for decorator?

onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -508,18 +508,18 @@ def to_model_proto(self, opset_version: int) -> onnx.ModelProto:
...
+        #         onnx_model, check_type=True, strict_mode=False, data_prop=True
+        #     )
+        #     onnx.checker.check_model(onnx_model, full_check=True)
+        # except (onnx.checker.ValidationError, onnx.shape_inference.InferenceError) as e:
+        #     warnings.warn(f"ONNX model is invalid: {e}", stacklevel=1)
+        #     logging.debug(
+        #         "ONNX model:\n%s\n\nTorchScript graph:\n%s",
+        #         onnxscript.proto2text(onnx_model),
+        #         self.torch_graph,
+        #     )
```
justinchuby, line 522: I would revert these changes

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+@torch_op("aten::var_mean", overload=True, trace_only=True)
+def aten_var_mean_dim(
+    self: TReal, dim: int, unbiased: bool = True, keepdim: bool = False
+) -> tuple[TReal, TReal]:
+    """var_mean.dim(Tensor self, int[1]? dim, bool unbiased=True, bool keepdim=False) -> (Tensor, Tensor)"""
+
+    if unbiased:
+        correction = 1
+    else:
+        correction = 0
```
justinchuby, line 6038: Can we pass this in directly? bool type is a subclass of int iirc

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Discussed again in teams. The current name as is is good

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+
+    if dim is None:
+        var, mean = _aten_var_mean(self, correction, keepdim)
+    else:
+        var, mean = _aten_var_mean_dim(self, dim, correction, keepdim)
+    return var, mean
+
+
+@torch_op("aten::var_mean", private=True)
+def _aten_var_mean(
```
justinchuby, line 6060:
```suggestion
def _aten_var_mean_onnx(
```

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: Otherwise just do `bool(unbiased)`

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+        correction = 1
+    else:
+        correction = 0
+    var, mean = _aten_var_mean(self, correction, keepdim=False)
+    return var, mean
+
+
+@torch_op("aten::var_mean", overload=True, trace_only=True)
+def aten_var_mean_dim(
+    self: TReal, dim: int, unbiased: bool = True, keepdim: bool = False
```
justinchuby, line 6031:
```suggestion
    self: TReal, dim: Optional[int], unbiased: bool = True, keepdim: bool = False
```

according to the signature below?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+        correction = 1
+    else:
+        correction = 0
+    var, mean = _aten_var_mean_dim(self, dim, correction, keepdim)
+    return var, mean
+
+
+@torch_op("aten::var_mean", overload=True, trace_only=True)
+def aten_var_mean_correction(
+    self: TReal, dim: int = None, correction: int = 1, keepdim: bool = False
```
justinchuby, line 6045:
```suggestion
    self: TReal, dim: Optional[int] = None, correction: Optional[int] = None, keepdim: bool = False
```

Is correction float or int, since it is "Scalar" below?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+    if dim is None:
+        var, mean = _aten_var_mean(self, correction, keepdim)
+    else:
+        var, mean = _aten_var_mean_dim(self, dim, correction, keepdim)
+    return var, mean
+
+
+@torch_op("aten::var_mean", private=True)
+def _aten_var_mean(
+    self: TReal, correction: int = 1, keepdim: bool = False
```
justinchuby, line 6061: Should correction be float?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+        sub = op.Sub(numel_int, correction)
+        var = op.Div(mul, op.Cast(sub, to=FLOAT.dtype))
+
+    return var, mean
+
+
+@torch_op("aten::var_mean", private=True)
+def _aten_var_mean_dim(
+    self: TReal, dim: int, correction: int = 1, keepdim: bool = False
+) -> tuple[TReal, TReal]:
```
justinchuby, line 6083:
```suggestion
) -> Tuple[TReal, TReal]:
```

for python3.8 compatibility

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6014,10 +6014,93 @@ def aten_var(self: TensorType, unbiased: bool = True) -> TensorType:
...
+        mul = op.Mul(var, numel_float)
+        sub = op.Sub(numel_int, correction)
+        var = op.Div(mul, op.Cast(sub, to=FLOAT.dtype))
+
+    return var, mean
+
+
+@torch_op("aten::var_mean", private=True)
+def _aten_var_mean_dim(
+    self: TReal, dim: int, correction: int = 1, keepdim: bool = False
```
justinchuby, line 6082:
```suggestion
    self: TReal, dim: INT64, correction: int = 1, keepdim: bool = False
```

Update dim to input to avoid the manual casting below?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1502,14 +1535,14 @@ def _capture_graph_and_evaluate_torch_script_evaluator(function: Callable, args,
...
-                f"{onnxscript.proto2text(onnx_model)}"
-            ) from e
+        # try:
+        #     onnx.checker.check_model(onnx_model, full_check=True)
+        # except onnx.checker.ValidationError as e:
+        #     raise AssertionError(
+        #         f"ONNX model is invalid: {e}. "
+        #         f"Model:\n"
+        #         f"{onnxscript.proto2text(onnx_model)}"
+        #     ) from e
```
justinchuby, line 1545: Once the tests pass, I would uncomment this, mark xfails for the tests for now, and submit an issue to onnx. Hopefully we can get the fix in a newer weekly release then enable the xfail'ed tests again.

justinchuby (approved):

justinchuby (changes requested):
I clicked approve prematurely. Resetting for now

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: https://pytorch.org/docs/master/generated/torch.var_mean.html?highlight=var_mean#torch.var_mean
PyTorch document say correction is int.
I tried to change the signature to float, the FullGraph mode has some trouble for:
```
if correction != 0.0
```
Because the input correction is 2 instead of 2.0

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Changed to INT64 but need upstream code to explicitly do the conversion, such as:
```
        if isinstance(dim, Tuple):
            dim_tensor = op.Constant(value_ints=dim)
        else:
            dim_tensor = op.Constant(value_int=dim)
        var, mean = _aten_var_mean_dim_onnx(self, dim_tensor, correction, keepdim)
```
Because sometimes the dim in test case is (1,), sometimes is 1.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: FullGraph mode can passed without check_model. So marked as xfails.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: https://pytorch.org/docs/master/generated/torch.var_mean.html?highlight=var_mean#torch.var_mean
Pytorch document say the correction is int.
I tried to set it to float, but FullGraph mode will be failed due to:
```
if correction != 0.0
```
Because the input value is int(2), not float(2.0), so the type mismatched.
So I would like to keep it to int.

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: I would like to keep it as int. The reason see above.

justinchuby (approved):

PR #570, opened 2023-03-28
Set up github to build and publish python packages | chore(release)

liqunfu:


justinchuby (changes requested):
We should not make pyproject changes because it is the recommended way of configuring new Python projects

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+    - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+      with:
+        name: wheels
+        path: ./onnx-script/dist
+
+    - name: Upload onnx-script-weekly wheel to TestPyPI/PyPI weekly
+      if: (github.event_name == 'schedule') # Only triggered by weekly event
+      run: |
+        twine upload --verbose onnx-script/dist/*.whl --repository-url https://upload.pypi.org/legacy/ -u ${{ secrets.ONNXSCRIPTWEEKLY_USERNAME }} -p ${{ secrets.ONNXSCRIPTWEEKLY_TOKEN }}
+        twine upload --verbose onnx-script/dist/*.whl --repository-url https://test.pypi.org/legacy/ -u ${{ secrets.TESTPYPI_USERNAME }} -p ${{ secrets.TESTPYPI_PASSWORD }}
```
justinchuby, line 87:
```suggestion
```

I don't think we need test pypi. ONNX is moving away from it too

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+      with:
+        name: wheels
+        path: ./onnx-script/dist
+
+    - name: Upload onnx-script-weekly wheel to TestPyPI/PyPI weekly
+      if: (github.event_name == 'schedule') # Only triggered by weekly event
+      run: |
+        twine upload --verbose onnx-script/dist/*.whl --repository-url https://upload.pypi.org/legacy/ -u ${{ secrets.ONNXSCRIPTWEEKLY_USERNAME }} -p ${{ secrets.ONNXSCRIPTWEEKLY_TOKEN }}
```
justinchuby, line 86: Consider using pypa's official task:

https://github.com/justinchuby/lintrunner-adapters/blob/7a3a7b409db1d2146ef98f56a3406930e8270dbd/.github/workflows/ci.yml#L116-L119

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+          echo "Build weekly TestPyPI package"
+          python setup.py bdist_wheel --weekly_build
+        } else {
+          python setup.py bdist_wheel
+        }
+        Get-ChildItem -Path dist/*.whl | foreach {python -m pip install --upgrade $_.fullname}
+
+    - name: Test the installed wheel
+      run: |
+        cd onnx-script
```
justinchuby, line 75: Does this test the wheel or the source?

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+
+        echo "Install ONNX-Script"
+
+        if ('${{ github.event_name }}' -eq 'schedule') {
+          echo "Build weekly TestPyPI package"
+          python setup.py bdist_wheel --weekly_build
+        } else {
+          python setup.py bdist_wheel
+        }
+        Get-ChildItem -Path dist/*.whl | foreach {python -m pip install --upgrade $_.fullname}
```
justinchuby, line 71: Prefer linux. It's faster and we don't need to write powershell

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+
+        cd onnx-script
+
+        echo "Install ONNX-Script"
+
+        if ('${{ github.event_name }}' -eq 'schedule') {
+          echo "Build weekly TestPyPI package"
+          python setup.py bdist_wheel --weekly_build
+        } else {
+          python setup.py bdist_wheel
```
justinchuby, line 69: Create a conditional step? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+        python -m pip install -q --upgrade pip
+        cd onnx-script
+        python -m pip install -q -r requirements-dev.txt
+
+    - name: Build ONNX-Script wheel
+      run: |
+        $arch = 'x64'
+        if ('${{ matrix.architecture }}' -eq 'x86') {
+          $arch = 'Win32'
+        }
```
justinchuby, line 59: I don't think we need to build for different architectures. In fact I think we only need to create one universal build?

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+  contents: read
+
+jobs:
+  build:
+    if: github.event_name != 'pull_request' || startsWith( github.base_ref, 'rel-') || contains( github.event.pull_request.labels.*.name, 'run release CIs')
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        architecture: ['x64', 'x86']
```
justinchuby, line 23:
```suggestion
```

I think we just need one

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+      uses: actions/setup-python@d27e3f3d7c64b4bbf8e4abfb9b63b83e846e0435 # v4.5.0
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: ${{ matrix.architecture }}
+
+    - name: Add msbuild to PATH
+      uses: microsoft/setup-msbuild@1ff57057b5cfdc39105cd07a01d78e9b0ea0c14c # v1.3.1
+      with:
+        msbuild-architecture: ${{ matrix.architecture }}
+
```
justinchuby, line 47:
```suggestion
```

pyproject.toml
```diff
@@ -1,45 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
 
-[build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
```
justinchuby, line 5: These can stay

pyproject.toml
```diff
@@ -1,45 +1,5 @@
...
-[tool.setuptools.dynamic]
-version = {attr = "onnxscript.__version__"}
-
-[tool.setuptools.packages.find]
-include = ["onnxscript*"]
-# Tests are by default excluded: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html
-
-[tool.setuptools.package-data]
-onnxscript = ["py.typed"]
-onnx = ["py.typed"]
```
justinchuby, line 40: I think these should stay too

setup.py
```diff
@@ -6,6 +6,36 @@
 
 NOTE: Put all metadata in pyproject.toml.
 """
+import os
+import pathlib
 import setuptools
+import sys
 
-setuptools.setup()
+from collections import namedtuple
```
justinchuby, line 14:
```suggestion
```

setup.py
```diff
@@ -6,6 +6,36 @@
...
+from datetime import date
```
justinchuby, line 15: Import modules only: `import datetime` and use `datetime.date`

setup.py
```diff
@@ -6,6 +6,36 @@
...
+
+TOP_DIR = os.path.realpath(os.path.dirname(__file__))
+PACKAGE_NAME = "onnx_script"
```
justinchuby, line 18:
```suggestion
PACKAGE_NAME = "onnx-script"
```

setup.py
```diff
@@ -6,6 +6,36 @@
...
+
+with open(os.path.join(TOP_DIR, "VERSION_NUMBER")) as version_file:
+    VERSION_NUMBER = version_file.read().strip()
+    if "--weekly_build" in sys.argv:
+        today_number = date.today().strftime("%Y%m%d")
+        VERSION_NUMBER += ".dev" + today_number
+        PACKAGE_NAME = "onnx_script_weekly"
+        sys.argv.remove("--weekly_build")
+
+packages = setuptools.find_packages() + setuptools.find_namespace_packages(
```
justinchuby, line 28:
```suggestion
```

setup.py
```diff
@@ -6,6 +6,36 @@
...
+    include="onnxscript"
+)
+
```
justinchuby, line 31:
```suggestion
```

setup.py
```diff
@@ -6,6 +6,36 @@
...
+setuptools.setup(
+    name=PACKAGE_NAME,
+    version=VERSION_NUMBER,
+    description="ONNX Script",
+    long_description=pathlib.Path("README.md").read_text(),
+    long_description_content_type="text/markdown",
+    packages=packages,
```
justinchuby, line 38:
```suggestion
```

noxfile.py
```diff
@@ -35,7 +35,7 @@
 def build(session):
     """Build package."""
     session.install("build", "wheel")
-    session.run("python", "-m", "build")
```
justinchuby, line 38: I prefer this continue to work. We should not drop `python -m build` support as it is the new/right way to build.

justinchuby: Having

```

[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"
```

below should enable support for this

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        architecture: ['x64', 'x86']
+    steps:
+    - name: Checkout ONNX-Script
+      uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
+      with:
+         path: ./onnx-script
```
justinchuby, line 28: Is there a reason why we clone to a sub folder?

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+    - name: Checkout ONNX-Script submodules
+      shell: bash
```
justinchuby, line 30:
```suggestion
```
nit

.github/workflows/release.yml
```diff
@@ -0,0 +1,87 @@
...
+      run: |
+         cd onnx-script
+         auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+         git submodule sync --recursive
+         git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive --depth=1
```
justinchuby, line 35:
```suggestion
```

we don't have submodules

setup.py
justinchuby: nit: This is done by pyproject.toml. Anything that doesn't need to happen in setup.py shouldn't happen in it imo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,84 @@
+name: Release
+
+on:
+  schedule:
+    # Run weekly on Monday 00:00
+    - cron:  '00 00 * * MON'
+  push:
+    branches: [main, rel-*]
+  pull_request:
+    branches: [main, rel-*]
```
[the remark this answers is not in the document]
justinchuby: Do we not want to create release branch?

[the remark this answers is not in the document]
jcwchen: Ah good catch. I didn't see any now, but it makes sense we should have them.

justinchuby:
We need to exclude opgen

image

justinchuby:
Looks like the version in metadata is not dev
image

justinchuby (approved):

.github/workflows/main.yaml
```diff
@@ -91,3 +95,55 @@ jobs:
...
+      - name: Install ONNX Script wheel
+        run: |
+          python -m pip install dist/*.whl
+
+  publish:
+    needs: [release]
+    runs-on: ubuntu-latest
+    environment: production
+    # Publish only when it is a scheduled run (weekly dev builds) or a tag push with version number
+    if: (github.event_name == 'schedule') || (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v'))
```
justinchuby, line 139: Need to complete the if by adding that from release

PR #574, opened 2023-03-29
Move opgen into onnxscript | chore(release)

justinchuby:
Move opgen into onnxscript so that it will be part of the package. Also clean up the import lines of the scripts.

Tested locally


tools/generate_ops.py
```diff
@@ -0,0 +1,54 @@
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
```
justinchuby, line 5: Usage to be added in another pr

abock (changes requested):
Why does this need to be part of the package? This is infrastructure for building ONNX Script? It intentionally lives outside.

justinchuby:
I think there was a plan to use pygen to replace jinja? Moving this in so we can start using it.

justinchuby:
Maybe I’ll just move [onnx_opset_builder.py](https://github.com/microsoft/onnx-script/pull/574#diff-03b85a11b0f585e696caf1b8813e9ec9c13c52f0b9104f8509a281649e59aef0) back out?

justinchuby:
Updated to only move pygen.

justinchuby:
Closing because Aaron is working on some new cool stuff

PR #579, opened 2023-03-29
Make onnx-script PEP561 compliant

### LINKED ISSUE #572: mypy complains ONNX Script is not PEP 561 compliant
It seems we need to explicitly declare ONNX Script as [PEP 561](https://peps.python.org/pep-0561/) compliant in order for mypy to follow into it from an external package (e.g. usage inside PyTorch).

In PyTorch we see the [following error](
https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker):

```
  Error (MYPY) [import]
    Skipping analyzing "onnxscript": module is installed, but missing library
    stubs or py.typed marker

        198  |        #   supported opset version Y
        199  |        # - etc.
        200  |        import onnx
    >>> 201  |        import onnxscript
        202  |
        203  |        assert onnx.defs.onnx_opset_version() >= opset_version
        204  |        assert isinstance(
```

ONNX Script is installed, but because mypy cannot determine if it's meant to be typed, it does not descend into it, resulting in `onnxscript` being treated as `Any`.

shubhambhokare1:
Fixes #572

### DIFF
--- a/onnxscript/py.typed
+++ b/onnxscript/py.typed
@@ -1,0 +1,1 @@
+# Marker file for PEP-561 (inline types)


justinchuby (approved):
Thanks!

PR #582, opened 2023-03-30
Add op(scatter reduce) | feat(atenlib)

xiaowuhu:
Opened feature request: https://github.com/onnx/onnx/issues/5100
To ask support include_self=True mode, and reduce='mean' option.


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,28 @@ def aten_scatter_add(
...
 
 
+@torch_op("aten::scatter_reduce", trace_only=True)
+def aten_scatter_reduce(
+    self: TReal,
+    index: TInt,
+    src: TReal,
+    reduce: str,
+    dim: int,
+    include_self: bool = True,  # pylint: disable=unused-argument
```
justinchuby, line 5136: I would match the order from the signature

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,28 @@ def aten_scatter_add(
...
+):
+    # FIXME(xiaowuhu): Support include_self == False
+    dict = {  # convert torch string name to onnx string name
```
justinchuby, line 5139: Avoid redefining dict, which is a builtin name. consider `reduce_modes`

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -752,6 +761,18 @@ def _where_input_wrangler(
...
+        variant_name="mean",
+        reason="ORT doesn't support reduce='mean' option",
+        test_class_name="TestOutputConsistencyEager",
+    ),
+    xfail(
+        "scatter_reduce",
+        variant_name="mean",
+        reason="ORT doesn't support reduce='mean' option",
+        test_class_name="TestOutputConsistencyFullGraph",
+    ),
```
justinchuby, line 775:
```suggestion
    xfail(
        "scatter_reduce",
        variant_name="mean",
        reason="ORT doesn't support reduce='mean' option",
    ),
```

to enable it for both tests

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -981,6 +1002,12 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.input.shape) == 0,
         reason="fixme: Rank(0) input will lead ORT failed due to different rank(result) in if-else branch",
     ),
+    skip(
+        "scatter_reduce",
+        matcher=lambda sample: sample.kwargs.get("include_self") is False
+        or len(sample.input.shape) == 0,  # skip rank 0 case
```
justinchuby, line 1008: Does it fail when rank is 0?

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: It is hard to use original signature, because the 'dim' input is int, not INT64. It will be used in the following op.ScatterElements( ..... axis=dim.....). So it must be moved down as attribute.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: Yes. Added more information.
After convert the self/index/src to rank 1 tensor,  will still failed.

onnxscript/function_libs/torch_aten/ops/core.py
justinchuby: We can mix inputs and attributes when trace only because it’s just a plain Python function. We also must match the order because that is what the exporter will expect. In cases when we do want to script parts of the function we can create a trace only interface that matches the argument order with the ATen, which then calls an internal implementation with order adjusted to fit ONNX

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -981,6 +995,13 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.input.shape) == 0,
         reason="fixme: Rank(0) input will lead ORT failed due to different rank(result) in if-else branch",
     ),
+    skip(
+        "scatter_reduce",
+        matcher=lambda sample: sample.kwargs.get("include_self") is False
+        or len(sample.input.shape)
+        == 0,  # rank 0 case failed even after converted to rank 1 tensor
```
justinchuby, line 1002: What was the failure? Was it value mismatch/ shape inference/ ORT crashes etc.?

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: got your point.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
xiaowuhu: The reason should be: the graph is not correct, missing the logic under op.If().

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -986,6 +999,12 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.input.shape) == 0,
         reason="fixme: Rank(0) input will lead ORT failed due to different rank(result) in if-else branch",
     ),
+    skip(
+        "scatter_reduce",
+        matcher=lambda sample: sample.kwargs.get("include_self") is False,  # ORT only support include_self=True mode
+        # or len(sample.input.shape) == 0,
+        reason="ORT does't support include_self=False option, and skip rank 0 case because failed in FullGraph mode",
```
justinchuby, line 1006:
```suggestion
        reason="ORT does't support include_self=False option",
```

looks like len(sample.input.shape) == 0 was commented out?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,38 @@ def aten_scatter_add(
...
+    onnx_reduce = reduce_mode[reduce]
+    self_rank = op.Size(op.Shape(self))
+    if self_rank == 0:  # assert (index_rank == 0 and rank_src == 0)
+        neg_1 = op.Constant(value_ints=[-1])
+        self = op.Reshape(self, neg_1)
+        index = op.Reshape(index, neg_1)
+        src = op.Reshape(src, neg_1)
+    result = op.ScatterElements(self, index, src, axis=dim, reduction=onnx_reduce)
+    if self_rank == 0:
+        result = op.Squeeze(result)
```
justinchuby, line 5157: I would wrap this part into a separate function.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,38 @@ def aten_scatter_add(
...
+    index: TInt,
+    src: TReal,
+    reduce: str,
+    include_self: bool = True,  # pylint: disable=unused-argument
+):
+    """scatter_reduce.two(Tensor self, int dim, Tensor index, Tensor src, str reduce, *, bool include_self=True) -> Tensor"""
+
+    # FIXME(xiaowuhu): Support include_self == False
+    reduce_mode = {  # convert torch string name to onnx string name
+        "mean": "none",  # mean doesn't support in ORT
```
justinchuby, line 5142:
```suggestion
        "mean": "none",
```

nit. Also: just to clarify, was it not supported in ORT or by ONNX? Should we open an issue in any case?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,38 @@ def aten_scatter_add(
...
+        "sum": "add",
+        "prod": "mul",
+        "amin": "min",
+        "amax": "max",  # default none = amax
```
justinchuby, line 5146: What does `default none = amax` mean? Is `reduce` ever `"none"`?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -383,6 +383,13 @@ def _scatter_add_input_wrangler(
     return args, kwargs
 
 
+def _scatter_reduce_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    kwargs["reduce"] = args.pop(4)  # put the string into kwargs, otherwise FullGraph mode will cannot find get 'reduce' argument
```
justinchuby, line 389:
```suggestion
    # Put the string into kwargs, otherwise FullGraph mode will cannot find get 'reduce' argument
    kwargs["reduce"] = args.pop(4)
```

nit: Prefer a separate line for comments. Also capitalize the first letter

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5126,6 +5126,38 @@ def aten_scatter_add(
     return op.ScatterElements(self, index, src, axis=dim, reduction="add")
 
 
+@torch_op("aten::scatter_reduce", trace_only=True)
+def aten_scatter_reduce(
+    self: TReal,
+    dim: int,  # we have to use int here becuase ScatterElements() will use this attribute
```
justinchuby, line 5132: nit: "becuase" is a misspelling of "because" (from misspell)

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Opened a feature request: https://github.com/onnx/onnx/issues/5100

justinchuby (approved):
Thanks!

PR #588, opened 2023-04-04
Simplify signal function | test

gramalingam:
Simplify the function definition for signal ops (FFT) using syntactic simplifications (operator overloading, constant promotion etc.).


justinchuby (approved):

onnxscript/tests/models/signal_dft.py
```diff
@@ -53,59 +49,54 @@ def blackman_window(window_length):
...
+    ni = op.Cast(op.Range(0, window_length, 1), to=1)
+    cos2 = op.Cos((ni * TWO_PI) / N_minus_1)
+    cos4 = op.Cos((ni * FOUR_PI) / N_minus_1)
+    return (0.42 - (cos2 * 0.5)) + (cos4 * 0.08)
 
 
 @script()
 def switch_axes(x: FLOAT[...], axis1: INT64[1], axis2: INT64[1]) -> FLOAT[...]:
-    """Switches two axis. The function assumes `axis1 < axis2`."""
+    """Switches two axis. The function assumes `axis1 < axis2`.
```
justinchuby, line 61: Compare and swap if needed so we don’t need the assumption?

onnxscript/tests/models/signal_dft.py
```diff
@@ -145,7 +136,7 @@ def dft_last_axis(
     where the last axis indicates whether it is the real or the imaginary part.
 
     Args:
-        x: float tensor, the last dimension is the complex one, if has 1
+        x: float tensor, the last dimension is the complex one, if has 1dim1_size
```
justinchuby, line 139: 1dim1_size Is intended or typo?

onnxscript/tests/models/signal_dft.py
```diff
@@ -162,16 +153,14 @@ def dft_last_axis(
...
-    n = op.Reshape(n0, shape1)
+    range = op.Range(zero, fft_length, one)  # fft_length or dim
+    range_float = op.Cast(range, to=1)
+    shape1 = op.Constant(value=make_tensor("shape1", TensorProto.INT64, [2], [-1, 1]))
+    n = op.Reshape(range_float, shape1)
 
-    kar = op.Range(zero, fft_length, one)
-    k0 = op.Cast(kar, to=1)
-    k = op.Reshape(k0, shape2)
+    shape2 = op.Constant(value=make_tensor("shape2", TensorProto.INT64, [2], [1, -1]))
```
justinchuby, line 162: It would be nice if we can make tensor w/o needing to supply a name

onnxscript/tests/models/signal_dft.py
gramalingam: Typo, thanks for catching it. Well, it was mis-directed typing, I was searching for `dim1_size`, I think :-)

onnxscript/tests/models/signal_dft.py
gramalingam: Agree. Ideally, we should be able to use just `[1, -1]` and have the converter automatically do the rest. It almost works now, but there are a couple of issues. Eager-mode doesn't seem to fully handle it. But, there's also a design-issue: how to map a value such as `[1, -1]` to an ONNX value. A 1-dimensional tensor is the most-common and useful choice, but there could be rare cases where we want a Sequence of scalar tensors. So, a complete solution requires a bit more thought.

justinchuby: Personally I like an np.array() like interface. So we don't confuse a tensor with a python sequence

PR #591, opened 2023-04-04
Remove __main__

gramalingam:
The `__main__.py` file is very old and outdated. Removing it. (If anyone wants this, it should probably be updated/redone.)


justinchuby (approved):
That’s a strange failure

justinchuby:
Ah, there's also `cli_test` that needs to be removed

onnxscript/__init__.py
```diff
@@ -63,6 +63,12 @@
     # package is not installed
     pass
 
+def const (pyvalue: Any) -> Any:
```
justinchuby, line 66: typing: T -> T?

onnxscript/tests/cli_test.py
```diff
@@ -7,15 +7,10 @@
 
 import click.testing
 
-from onnxscript.__main__ import onnx2script, translate
+from onnxscript.__main__ import translate
```
justinchuby, line 10: So there’s still translate? Just curious where it was

onnxscript/__init__.py
gramalingam: Sorry, this crept in accidentally into this PR. It was (meant to be) part of a different branch for an in-progress PR. Will remove it.

onnxscript/tests/cli_test.py
gramalingam: That was incomplete, sorry. Removed it entirely.

justinchuby:
I suppose we no longer need click as a requirement?

justinchuby:
https://github.com/microsoft/onnx-script/pull/593

PR #598, opened 2023-04-06
TODO improvement on aten_sym_size op when ORT==1.15 | feat(atenlib)

titaiwangms:
https://github.com/microsoft/onnx-script/pull/484#discussion_r1136105039

Not sure why another file was linted...


onnxscript/backend/onnx_export_test.py
```diff
@@ -168,7 +168,9 @@ def test_export2python_produces_correct_onnx_script_model(
             onnxruntime_pybind11_state.InvalidArgument,  # pylint: disable=c-extension-no-member
         ) as e:
             self.skipTest(f"Unable to load the model: {e}")
-        except onnxruntime_pybind11_state.RuntimeException as e:  # pylint: disable=c-extension-no-member
+        except (
+            onnxruntime_pybind11_state.RuntimeException
+        ) as e:  # pylint: disable=c-extension-no-member
```
justinchuby, line 173: You may revert this change. We prob need to bump black. I will send a pr

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5540,6 +5540,14 @@ def aten_sym_size(self: TReal, dim: int = 0) -> TReal:
     """sym_size(Tensor self, int dim) -> Tensor"""
     # NOTE: onnxscript doesn't support attribute process,
     # so op.Shape(self, start=dim, end=dim + 1) is not supported.
+
+    # TODO(titaiwang): ORT==1.15 fixes SegFault
+    # https://github.com/microsoft/onnx-script/pull/484#discussion_r1136105039
+    # Change the op to:
+    # shape = op.Shape(self)
+    # idx= op.Reshape(dim, [1])
+    # return op.Gather(shape, idx)
```
justinchuby, line 5549: Why gather when we can slice? Just for my learning

justinchuby:
fyi we just decided to change the PR title format following https://github.com/microsoft/onnx-script/issues/577

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: From Rama: As discussed in another PR, dim+1 also doesn't work if dim is -1. Hence, Gather seems a better solution.

justinchuby (approved):

PR #600, opened 2023-04-06
Upgrade lint dependencies | chore(lint)

justinchuby:
Upgrade all lint dependencies and fix new lint errors


justinchuby:
@jcwchen I wonder if we should have a requirements-lint file and let dependabot take over. What do you think?

jcwchen (approved):
>@jcwchen I wonder if we should have a requirements-lint file and let dependabot take over. What do you think?

I am exactly thinking the same thing! but would it be one PR or multiple PRs if there are few dependency updates in a period? If it would be just one PR, it seems like a good idea.

jcwchen:
> I am exactly thinking the same thing! but would it be one PR or multiple PRs if there are few dependency updates in a period? If it would be just one PR, it seems like a good idea.

After consideration, ideally these dependencies should not update too often at the same time? Then I am fine with using dependabot anyway.

justinchuby:
I will need to update the lintrunner adapters package to support using a requirements.txt file. I will actually move that script to here

justinchuby:
I think we can configure the frequency

justinchuby:
Merged. Figures crossed dependabot works

PR #602, opened 2023-04-06
Handle ORT segfaults in tests | test(atenlib)

justinchuby:
Provide an option set by the environment variable `CATCH_ORT_SEGFAULT` to handle ORT segfaults in tests by running an inference session in a subprocess and handling the SIGSEGV signal.

Bumped ORT nightly version


requirements-ort-nightly.txt
```diff
@@ -1,2 +1,2 @@
 --index-url=https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
-ort-nightly==1.15.0.dev20230320005
+ort-nightly==1.15.0.dev20230405002
```
justinchuby, line 2: Apparently dependabot doesn't understand custom indices

justinchuby: It can be configured

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1132,6 +1150,70 @@ def _should_skip_test_sample(op_name: str, sample) -> Optional[str]:
...
+
+    Returns:
+        The inference result.
+
+    Raises:
+        OrtAbortedError if the process did not execute successfully.
+    """
+    manager = multiprocessing.Manager()
+    return_dict = manager.dict()
+    process = multiprocessing.Process(
```
BowenBao, line 1196: Do you have a measure of performance impact, if any?

justinchuby: Looks like it's running 10x slower on windows and mac in some cases. I don't know if it was github actions or this change. Linux was fine

justinchuby: No good. I will create a local option instead

justinchuby:
@BowenBao PTAL. I was only able to create an option for local testing as a compromise. Python cannot fork on windows. Is fork much faster than spawn?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
BowenBao: Looks like `spawn` is the only option for windows and mac :( https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

I'm not sure if it is applicable to utilize process pool or some way to persist the process. If it's too much effort I guess local is good for now.

BowenBao (approved):

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1132,6 +1154,71 @@ def _should_skip_test_sample(op_name: str, sample) -> Optional[str]:
...
+
+
+def _ort_session_run(serialized_model, ort_inputs, return_dict) -> None:
+    """Run a model with ONNX Runtime and store the results in return_dict."""
+
+    # Handle SIGSEGV so Python does not abort
+    def sig_handler(signum, frame):
+        warnings.warn(f"Segmentation error ({signum}) detected in ORT:\n{frame}")
+
+    signal.signal(signal.SIGSEGV, sig_handler)
```
BowenBao, line 1168: do we need to reset this somewhere?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1132,6 +1154,71 @@ def _should_skip_test_sample(op_name: str, sample) -> Optional[str]:
...
+
+    # Disable all ORT optimizations
+    session_options = onnxruntime.SessionOptions()
+    session_options.graph_optimization_level = (
+        onnxruntime.GraphOptimizationLevel.ORT_DISABLE_ALL
+    )
+    try:
+        session = ort.InferenceSession(serialized_model, session_options)
+        results = session.run(None, ort_inputs)
```
BowenBao, line 1177: what happens when `SIGSEGV` is received?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1264,12 +1351,17 @@ def _capture_graph_and_evaluate_torch_script_evaluator(function: Callable, args,
...
-            onnxruntime.GraphOptimizationLevel.ORT_DISABLE_ALL
-        )
+
         try:
+            if os.environ.get("CATCH_ORT_SEGFAULT"):
+                # Use an individual process to run ONNX Runtime to catch segfaults
+                return _safe_ort_session_run(onnx_model.SerializeToString(), ort_inputs)
+
+            # Disable all ORT optimizations
+            session_options = onnxruntime.SessionOptions()
```
BowenBao, line 1361: nit: this part is repeated

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Good question. I am not sure. Let me check

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: It will abort in this case.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: It is. Thoughts on combining? I felt it was not worth breaking down `_ort_session_run` at the time because it also handles signals

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
BowenBao: Would it behave differently without calling `signal.signal(signal.SIGSEGV, sig_handler)` (other than the warning)?

justinchuby: The whole thing will then crash. It looks like it was propagated to the parent process?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: I removed them

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: Merged with _ort_session_run

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: I am still trying to figure out how to manage a pool when some of the workers can fail

PR #606, opened 2023-04-07
Improve `max_pool2d_with_indices` | fix(atenlib)

justinchuby:
Fix typing annotation to be compatible with python 3.8, because `tuple[]` cannot be a generic type until 3.9. Also updated the aten name to match the signature


titaiwangms (approved):
Thanks!

PR #612, opened 2023-04-07
Add ops(unflatten) | feat(atenlib)

titaiwangms:
Fixes https://github.com/microsoft/onnx-script/issues/599
xfail graph test to wait https://github.com/microsoft/onnxruntime/pull/15409


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -6031,6 +6031,30 @@ def aten_type_as(self: TensorType, other: TensorType) -> TensorType:
...
+
+    self_size = op.Shape(self)
+
+    if dim < 0:
+        # PyTorch accepts negative dim as reversed counting
+        self_rank = op.Size(self_size)
+        dim = self_rank + dim
+
+    head_start_idx = op.Constant(value_ints=[0])
+    head_end_idx = op.Reshape(dim, op.Constant(value_ints=[1]))
```
titaiwangms, line 6046: @gramalingam @justinchuby 
ORTEvaluator doesn't seem to understand:

```python
head_end_idx = op.Reshape(dim, [1]))
```

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1063,6 +1076,11 @@ def _where_input_wrangler(
         matcher=lambda sample: not (len(sample.args) > 0 and isinstance(sample.args[0], int)),
         reason="this Aten overload only support one tensor as input and one int as args by design",
     ),
+    skip(
+        "unflatten",
+        matcher=lambda sample: any(dim == 0 for dim in sample.input.shape),
+        reason="0 dim in ONNX is undefined behavior.",
+    ),
```
justinchuby, line 1083: Can we create a condition to handle this case?

justinchuby (approved):

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
titaiwangms: To raise a warning? Not sure what we should give back to users.

justinchuby: What is pytorch's behavior?

titaiwangms: PyTorch takes this because 0-d is in their spec.
For example,
```python
a = torch.randn(4, 0)
# tensor([], size=(4, 0))

torch.unflatten(a, 0, (2, 2))
# tensor([], size=(2, 2, 0))
```

titaiwangms: I wonder if torch_libs has a certain way to inform users when the difference between specs is hit?

justinchuby: I see. We could skip this if it is too complicated and come back to it as need. I would update the skip reason to say something to the effect of "fixme: logic not implemented for size 0 inputs"

justinchuby: We don't raise errors in atenlib. The exporter can raise errors when it sees a difference after OpSchema is implemented

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -100,9 +100,6 @@ def aten_addmm(
 ) -> TFloat:
     """addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor"""
 
-    # TODO(titaiwang): op.Gemm seems needed to take care of corner case according to old symbolic_fn.
-    # Currently, it shows op-level validation failing on bloom.
-
```
justinchuby, line 105: Remove line 110 down below as well?

PR #619, opened 2023-04-11
Move release pipeline to a separate file | chore(release)

justinchuby:
I moved the release pipeline out because the tests are sometimes flaky but we may not want them to block the scheduled release. I separated the dev and official releases to use different environments so the official releases can be manually reviewed before uploading to pypi.

Updated requirements-dev.txt


.github/workflows/release.yaml
```diff
@@ -0,0 +1,82 @@
...
+          name: wheels
+          path: dist
+      - name: Install ONNX Script wheel
+        run: |
+          python -m pip install dist/*.whl
+
+  publish-dev:
+    needs: [release]
+    runs-on: ubuntu-latest
+    environment: PyPI Dev
```
justinchuby, line 53: Different environments are used here

liqunfu (approved):

jcwchen (commented):
Shall we still add some tests after publishing wheels? Testing the built wheel can still be useful.

Just curious: which test is flaky? https://github.com/microsoft/onnx-script/issues/567 this issue?

justinchuby:
> Shall we still add some tests after publishing wheels? Testing the built wheel can still be useful.

Sure

> Just curious: which test is flaky? #567 this issue?

Yes. They are tracked in https://github.com/microsoft/onnx-script/issues/539

jcwchen (approved):

PR #622, opened 2023-04-11
Compute the correct number of output for Split | fix(graph_building)

wschin:
When number of outputs is known, Split should not a sequence of tensors.


gramalingam:
Looks reasonable to me. But will leave it to those more familiar with the exporter comment/review.

justinchuby (approved):

justinchuby:
I Will go ahead to merge this

justinchuby:
@wschin could you rebase on main? Thanks!

justinchuby:
We can use https://github.com/microsoft/onnxscript/blob/85fece18dea2635dffd3aa49df4aaa38c7fb418f/onnxscript/evaluator.py#L319 instead

PR #625, opened 2023-04-12
Merge `attrs` and `attr_protos` in `IRFunction` | chore(irbuilder)

justinchuby:
* #626
* #631
* __->__ #625

Merge the two list in `IRFunction` by changing its type to `IRAttributeParameter`. This way we retain type information for all attributes. It is useful for creating correct `OpSchema`s and `ParamSchema`s in the next PR.

Also Include `typeinfo` in `add_attr_parameter`.


titaiwangms:
The changes aren't stacked? It seems being piled up to the next commit?

justinchuby:
> The changes aren't stacked? It seems being piled up to the next commit?

That's the thing I talked about with ghstack and how we use github to merge which is annoying. I changed the base branch back. Try now?

titaiwangms:
> > The changes aren't stacked? It seems being piled up to the next commit?
> 
> That's the thing I talked about with ghstack and how we use github to merge which is annoying. I changed the base branch back. Try now?

Working!

justinchuby:
onnxscript/converter_test.py::TestConverter::test_signal - ValueError: A function with attributes cannot be exported as a model.

onnxscript/irbuilder.py
```diff
@@ -108,10 +108,16 @@ def _opt_var_to_str(x):
 
 
 class IRAttributeValue:
-    """An attribute value (representing an actual parameter)."""
+    """An attribute value (representing an actual parameter).
```
gramalingam, line 111: With the current changes, this class is being used to cover both *actual* parameters (which are values) as well as *formal* parameters (with or without a default-value). I am not 100% sure if we should unify all of these into a single type/class. But if we do, may be the name should be changed to reflect this? May be `IRAttributeParameter` ? And, of course, the above documentation line should be changed to actual+formal parameter.

gramalingam: For example, should we add an assertion/check in the IRStmt __init__ method that all attrs have a default-value?

justinchuby: Thanks! I added IRAttributeParameter. I wonder if it should be a subclass of IRAttributeValue or be its own class? For now I subclassed but semantically should it be that?

gramalingam: Yes, semantically it should not be a subclass. The current form is preferable.

onnxscript/irbuilder.py
```diff
@@ -120,14 +125,48 @@ def __str__(self):
...
         return self.attr_proto.type
 
 
+@dataclasses.dataclass(frozen=True)
+class IRAttributeParameter:
+    """An attribute parameter (representing a formal parameter).
+
+    It may or may not carry a default value.
+
+    Attributes:
```
gramalingam, line 142: Is this out-dated, or is this the accepted convention? That is, should this list be "name, type, default_value" or as below?

onnxscript/irbuilder.py
```diff
@@ -120,14 +125,48 @@ def __str__(self):
...
+    def __str__(self):
+        return helper.printable_attribute(self.attr_proto)
+
+    @property
+    def has_default(self):
+        return self.default_value is not None
+
+    @property
+    def attr_proto(self) -> onnx.AttributeProto:
+        if self.default_value is None:
```
gramalingam, line 162: Is this method used in any context where the default_value is None? Wonder if we should throw an exception instead, since the returned value is an incomplete AttributeProto.

onnxscript/irbuilder.py
```diff
@@ -191,9 +230,7 @@ def __init__(self, name: str, domain: str = "") -> None:
         self.outputs: list[IRVar] = []
         self.stmts: list[IRStmt] = []
         # attribute parameters
-        self.attrs: list[str] = []
-        # attribute parameters with default value
-        self.attr_protos: list[IRAttributeValue] = []
+        self.attrs: list[IRAttributeParameter] = []
```
gramalingam, line 233: Think this should be `IRAttributeValue`

onnxscript/irbuilder.py
justinchuby: I should add the other attributes. thanks for catching

onnxscript/irbuilder.py
justinchuby: OK!

onnxscript/irbuilder.py
justinchuby: Done

onnxscript/irbuilder.py
justinchuby: Done

gramalingam (approved):

onnxscript/irbuilder.py
justinchuby: image

Is this the same view you saw?

gramalingam: Yes

justinchuby: Wait so was this good or did you mean something else?

gramalingam: I guess I am trying to understand the conventions here. The current form makes sense as the "public interface" exposed. Technically, however, `default_value` is also an attribute. Is the idea that it is "private"? Would renaming `default_value` to `_default_value` be appropriate?

justinchuby: Added default_value. Thanks!

PR #626, opened 2023-04-12
Auto generate OpSchema for functions | feat(op_schema)

justinchuby:
* #692
* #674
* __->__ #626
* #684

This change adds the capability to auto generate `OpSchema`.

### Changes

- Implement the `opschema` property in `OnnxFunction`
- Test on all torch_lib functions

### Next PR

Support trace_only functions

## Example

```python
from onnxscript.function_libs.torch_aten.ops import core, nn


print("core.aten_abs.opschema: ", core.aten_abs.opschema)

print("nn.aten_cross_entropy_loss.opschema: ", nn.aten_cross_entropy_loss.opschema)
```

Results

```
core.aten_abs.opschema:  OpSchema(
    name='aten_abs',
    domain='onnxscript.atenlib',
    since_version=1,
    doc='abs(Tensor self) -> Tensor',
    type_constraints=[OpSchema.TypeConstraintParam(type_param_str='TReal', allowed_type_strs=['tensor(float)', 'tensor(int8)', 'tensor(int16)', 'tensor(int32)', 'tensor(int64)', 'tensor(float16)', 'tensor(double)', 'tensor(bfloat16)'], description='')],
    inputs=[OpSchema.FormalParameter(name='self', type_str='TReal', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    outputs=[OpSchema.FormalParameter(name='return_val', type_str='TReal', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    attributes={}
)
nn.aten_cross_entropy_loss.opschema:  OpSchema(
    name='aten_cross_entropy_loss',
    domain='onnxscript.atenlib',
    since_version=1,
    doc='cross_entropy_loss(Tensor self, Tensor target, Tensor? weight=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0) -> Tensor',
    type_constraints=[OpSchema.TypeConstraintParam(type_param_str='TFloatOrBFloat16', allowed_type_strs=['tensor(float)', 'tensor(float16)', 'tensor(double)', 'tensor(bfloat16)'], description=''), OpSchema.TypeConstraintParam(type_param_str='T1', allowed_type_strs=['tensor(float)', 'tensor(float16)', 'tensor(double)', 'tensor(bfloat16)'], description='')],
    inputs=[OpSchema.FormalParameter(name='self', type_str='TFloatOrBFloat16', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>), OpSchema.FormalParameter(name='weight', type_str='T1', description='', param_option=<FormalParameterOption.Optional: 1>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    outputs=[OpSchema.FormalParameter(name='result_10', type_str='TFloatOrBFloat16', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
    attributes={'ignore_index': OpSchema.Attribute(name='ignore_index', type=<AttrType.INT: 2>, description='', default_value=name: "ignore_index"
i: -100
type: INT
, required=False), 'label_smoothing': OpSchema.Attribute(name='label_smoothing', type=<AttrType.FLOAT: 1>, description='', default_value=name: "label_smoothing"
f: 0.0
type: FLOAT
, required=False), 'reduction': OpSchema.Attribute(name='reduction', type=<AttrType.INT: 2>, description='', default_value=name: "reduction"
i: 1
type: INT
, required=False), 'target': OpSchema.Attribute(name='target', type=<AttrType.INTS: 7>, description='', default_value=, required=True)}
)
```

Fixes https://github.com/microsoft/onnx-script/issues/476


onnxscript/type_annotation.py
```diff
@@ -146,3 +147,52 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
     if get_origin(typeinfo) is tuple:
         return get_args(typeinfo)
     return (typeinfo,)
+
+
+def _reduce_type_var_to_union(hint: typing.TypeVar):
+    """Reduce a TypeVar to a Union type on which we can use issubclass to check membership."""
```
gramalingam, line 153: change "Union type" to "type" or "type bound" ? It doesn't always seem to be a "Union" below.

justinchuby:
Example errors 

```
onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py:1436: in test_script_function_has_op_schema
    op_schema = func.opschema
onnxscript/values.py:333: in opschema
    allowed_types=type_annotation.get_supported_input_types(type_),
onnxscript/type_annotation.py:194: in get_supported_input_types
    elif issubclass(tensor_type, pytype):
../../anaconda3/envs/onnx/lib/python3.10/typing.py:997: in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
E   TypeError: Subscripted generics cannot be used with class and instance checks
```

fatcat-z (approved):

fatcat-z:
Please make sure those CI failures have no relations with your changes.

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -1495,6 +1495,20 @@ def test_script_function_passes_checker(self, _, func_with_wrangler):
...
+        "OpSchema is not writable before ONNX 1.14",
+    )
+    def test_script_function_has_op_schema(self, _, func_with_wrangler):
+        if isinstance(func_with_wrangler, tuple):
+            func = func_with_wrangler[0]
+        else:
+            func = func_with_wrangler
+        op_schema = func.opschema
+        assert op_schema is not None
+        self.assertEqual(op_schema.name, func.name)
```
titaiwangms, line 1510: Is there ONNXFunction with wrangler?

onnxscript/type_annotation.py
justinchuby: Removed the function instead

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: That was how we setup the tuple list in test: `Tuple[Tuple[<the function>, <wrangler function>]]

justinchuby:
```
onnxscript/type_annotation_test.py:214: in TypeConversionFunctionsTest
    def test_pytype_to_input_strings(self, _, pytype: Any, expected: list[str]):
```

python3.8 need future annotations

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
titaiwangms: Yeah, I was thinking if the goal here is to test all ONNXFunction, we could isinstance(func, ONNXFunction) rather than see if it has wrangler or not.

justinchuby: So this logic is just taking the onnx function out from the tuple. We are testing all onnx functions already I think?

titaiwangms: I misread the whole thing. NVM

onnxscript/values.py
```diff
@@ -271,12 +299,93 @@ def __init__(
...
         self._param_schemas: Optional[tuple[ParamSchema, ...]] = None
+        self._opschema: Optional[onnx.defs.OpSchema] = None
 
     @property
     def name(self):
         """Returns the function name."""
         return self.opname
 
+    @property
+    def opschema(self) -> Optional[onnx.defs.OpSchema]:
```
gramalingam, line 310: Is this being used anywhere yet?

onnxscript/values.py
```diff
@@ -271,12 +299,93 @@ def __init__(
...
+                name=type_annotation.get_type_constraint_name(type_) or name,
+                allowed_types=type_annotation.pytype_to_input_strings(type_),
+            )
+
+        formal_inputs = [
+            onnx.defs.OpSchema.FormalParameter(
+                arg.name,
+                type_to_constraint[arg.typeinfo].name,
+                param_option=onnx.defs.OpSchema.FormalParameterOption.Optional
+                if typing.get_origin(arg.typeinfo) is Union and typing.get_args(arg.typeinfo)
```
gramalingam, line 337: May be worth abstracting this into a function and moving it into `type_annotation`. Doesn't it need to check if `None` is one the values in the union? We may also need to figure out how to distinguish the two kinds of optionals ONNX supports. (That is, extend the type-annotations to allow users to indicate which they intend.)

onnxscript/values.py
```diff
@@ -271,12 +299,93 @@ def __init__(
...
+                is_homogeneous=True,
+            )
+            for arg in function_ir.inputs
+        ]
+        formal_outputs = [
+            onnx.defs.OpSchema.FormalParameter(
+                arg.name,
+                type_to_constraint[arg.typeinfo].name,
+                param_option=onnx.defs.OpSchema.FormalParameterOption.Optional
+                if typing.get_origin(arg.typeinfo) is Union and typing.get_args(arg.typeinfo)
```
gramalingam, line 349: Same as above.

onnxscript/values.py
justinchuby: No. It will be used by the exporter. For now I added tests to make sure all torch_lib functions have opschemas defined

onnxscript/values.py
```diff
@@ -271,12 +299,93 @@ def __init__(
...
+    def opschema(self) -> Optional[onnx.defs.OpSchema]:
+        """Construct an OpSchema from function_ir."""
+        if self._opschema is not None:
+            return self._opschema
+
+        if not _ONNX_OP_SCHEMA_WRITABLE:
+            return None
+
+        function_ir = self.function_ir
+        # Find all distinct types in the inputs and outputs
```
gramalingam, line 319: What's the notion of equality (of types) here? We need to be careful about the following: two parameters of type which is the same TypeVar can be mapped to the same typevar in ONNX. However, Python is more expressive. It allows us to define two parameters of type "Optional[TypeVar[X]]", right? ONNX doesn't allow us to express something like this. This is a partial constraint since, it allows one parameter to be None and the other to be non-None. But if both are non-None, they are expected to be of the same type. We can't say that in ONNX.

justinchuby: Good catch 🤔 Do you have any suggestions? Should we be more relaxed about the optional inputs? Or maybe fix it when we need those scenarios? For now we mostly just handle static optionals

justinchuby:
@gramalingam When is a function not homogeneous?

onnxscript/values.py
justinchuby: Done. Thanks for catching the bug too

onnxscript/values.py
justinchuby: Done

onnxscript/values.py
justinchuby: > We may also need to figure out how to distinguish the two kinds of optionals ONNX supports. (That is, extend the type-annotations to allow users to indicate which they intend.)

Merging to the discussion below

onnxscript/values.py
justinchuby: Sorry the tests were left out for some reason. I added them back

onnxscript/values.py
justinchuby: I will add a comment that points to an issue for refinement in a separate PR.

gramalingam: Ok. I think splitting the logic into the loop here and the utility `get_type_constraint_name` is problematic. One approach would be to combine/integrate both: that is, define a single function that maps a collection of types into a corresponding collection of type-var-name, plus bindings for those type-vars (if needed). Then, it would be able to map use of same python type-variable to the same onnx type-var, while renaming them into unique instances when used as a parameter to some other type-constructor like optional/list.

gramalingam: Or, may be the renaming can be done even with the current split (may be integrating them is not necessary).

justinchuby: I am not following 100%. Could you explain why it's problematic? Or I can chat with you offline

gramalingam: It is problematic for the renaming part, since we would need some state-like mechanism to generate distinct names over several different inputs/outputs. But, on reflection, renaming won't be a perfect solution either. We can't capture the python type-constraint exactly in ONNX (in the problematic case). I was suggesting relaxing the type-constraint instead of tightening the type-constraint. But perhaps the preferable solution is to raise an error and let the use know about the issue.

justinchuby: Currently, the logic is

- TypeVars -> use the name because it is unique for the function.
- Other types -> always create a new name, as intended in the signature
- Optional[TypeVar] -> Trickier: If input is Optional[TypeVar] and output is TypeVar, they are assigned to different type-names but should _sometimes_ be the same like you suggested. If we drop support for dynamic optionals, then we can assign the same name for `Optional[TypeVar]` and `TypeVar`, which is accurate. Nevertheless, assigning a different name for `Optional[TypeVar]` and `TypeVar` is a more relaxed constraint.

justinchuby: I created #696. I am planning to merge this PR and refine based on what falls out from the discussion in the future.

PR #631, opened 2023-04-14
Analyze type annotation for input types | feat(op_schema)

justinchuby:
* #692
* #674
* #626
* #684
* __->__ #631
* #688


onnxscript/type_annotation.py
```diff
@@ -146,3 +147,56 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+            supported_types.append(tensor_type.to_string())
+        elif issubclass(tensor_type, pytype):
+            supported_types.append(tensor_type.to_string())
+        # TODO(justinchuby): Handle sequence types
+
+    return supported_types
+
+
+def get_type_constraint_name(pytype: TypeAnnotationValue) -> Optional[str]:
+    pass
```
justinchuby, line 202: TODO

onnxscript/type_annotation_test.py
```diff
@@ -14,7 +14,7 @@
 from onnxscript.tests.common import testutils
 
 
-class TypeAnnotationTester(testutils.TestBase):
+class TypeAnnotationTest(testutils.TestBase):
```
justinchuby, line 17: Revert

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
     if get_origin(typeinfo) is tuple:
         return get_args(typeinfo)
     return (typeinfo,)
+
+
+# A sorted list of all type strings used in an OpSchema
+ALL_TYPE_STRINGS = (
```
fatcat-z, line 170: Duplicated code?

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+        pytype: A type annotation.
+
+    Returns:
+        A list of all supported input types for the given type annotation.
+        Ensures that the list is sorted in the same order as ALL_TYPE_STRINGS.
+    """
+    if pytype is None:
+        return list(ALL_TYPE_STRINGS)
+    if pytype is type(None):
+        return list(ALL_TYPE_STRINGS)
```
titaiwangms, line 201: Why would we see NoneType as all type? Looks like we prefer NoneType to wrong type.

titaiwangms: Is it for custom functions?

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+    if typing.get_origin(pytype) is Union:
+        subtypes = typing.get_args(pytype)
+        if len(subtypes) == 2 and type(None) in subtypes:
+            for subtype in subtypes:
+                if isinstance(subtype, TypeVar):
+                    return f"Optional_{subtype.__name__}"
+    if typing.get_origin(pytype) in _LIST_CONSTRUCTORS:
+        subtypes = typing.get_args(pytype)
+        if len(subtypes) == 1 and isinstance(subtypes[0], TypeVar):
+            return f"Sequence_{get_type_constraint_name(subtypes[0])}"
```
gramalingam, line 264: Nit: make lines 260 and 264 consistent? Either use recursive call in both places or neither? Recursion may be better in case the function is further refined in the future. Perhaps
```py
type_param_name = get_type_constraint_name(subtypes[0])
return None if type_param_name is None else "Sequence_" + type_param_name
```

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+        return pytype.__name__
+    if typing.get_origin(pytype) is Union:
+        subtypes = typing.get_args(pytype)
+        if len(subtypes) == 2 and type(None) in subtypes:
+            for subtype in subtypes:
+                if isinstance(subtype, TypeVar):
+                    return f"Optional_{subtype.__name__}"
+    if typing.get_origin(pytype) in _LIST_CONSTRUCTORS:
+        subtypes = typing.get_args(pytype)
+        if len(subtypes) == 1 and isinstance(subtypes[0], TypeVar):
```
gramalingam, line 263: minor point/question: Should `len(subtypes) == 1` be an assert?

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+        - Prefixes the name with "Sequence_" if the type annotation is a Sequence[].
+        - Returns None if the type annotation does not have a type constraint.
+    """
+    if isinstance(pytype, TypeVar):
+        return pytype.__name__
+    if typing.get_origin(pytype) is Union:
+        subtypes = typing.get_args(pytype)
+        if len(subtypes) == 2 and type(None) in subtypes:
+            for subtype in subtypes:
+                if isinstance(subtype, TypeVar):
```
gramalingam, line 259: May be changing to `subtype is not type(None)` with a recursive call (see discussion below) would be better. Eg., what if we want to use `Optional[List[...]]` ?

onnxscript/type_annotation.py
```diff
@@ -146,3 +164,102 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+            return pytype_to_input_strings(Union.__getitem__(constraints))
+        bound = pytype.__bound__
+        if bound is None:
+            return list(ALL_TYPE_STRINGS)
+        return pytype_to_input_strings(bound)
+    if typing.get_origin(pytype) is Union:
+        options = []
+        subtypes = typing.get_args(pytype)
+        # A None type in a Union is equivalent to an optional type
+        is_optional = any(subtype is type(None) for subtype in subtypes)
```
gramalingam, line 220: We need some annotation to distinguish the two kinds of optionals ONNX supports.

onnxscript/type_annotation.py
justinchuby: Good catch thanks

onnxscript/type_annotation.py
justinchuby: Does `ALL_TENSOR_TYPE_STRINGS` sound good to you? That's more what's intended

justinchuby: > Why would we see NoneType as all type? Looks like we prefer NoneType to wrong type.

When a parameter is not annotated at all, we get NoneType but it will be treated as a tensor (actually, do we just get None? I will remove NoneType to see what happens)

onnxscript/type_annotation.py
justinchuby: I didn't want the code to abort so I was more relaxed in checking the cases

onnxscript/type_annotation.py
justinchuby: For now I am just mentally treating everything to be static optional. Any suggestions?

onnxscript/type_annotation.py
justinchuby: Removed

onnxscript/type_annotation.py
```diff
@@ -59,7 +77,7 @@ def _is_primitive_attr_type(typeinfo: TypeAnnotationValue) -> bool:
 
 def pytype_to_attrtype(
     pytype: TypeAnnotationValue,
-) -> onnx.AttributeProto.AttributeType:
+) -> typing.Optional[onnx.AttributeProto.AttributeType]:
```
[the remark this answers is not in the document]
justinchuby: Updated

onnxscript/type_annotation.py
justinchuby: Updated

onnxscript/type_annotation.py
justinchuby: Done. PTAL

onnxscript/type_annotation.py
```diff
@@ -143,6 +166,85 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
     if isinstance(typeinfo, typing.Sequence):
         return typeinfo
-    if get_origin(typeinfo) is tuple:
-        return get_args(typeinfo)
+    if typing.get_origin(typeinfo) is tuple:
+        return typing.get_args(typeinfo)
     return (typeinfo,)
+
+
+def pytype_to_input_strings(pytype: TypeAnnotationValue) -> list[str]:
```
gramalingam, line 174: nit: perhaps `pytype_to_type_strings` would be a better name.

onnxscript/type_annotation.py
```diff
@@ -143,6 +166,85 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+    """Returns a list of all supported input types in string representation for a given type annotation.
```
gramalingam, line 175: Nit: maybe `Returns a list of type-strings corresponding to a given type annotation`? It could represent even an output-type, not just input-type.

onnxscript/type_annotation.py
justinchuby: Removed.

onnxscript/type_annotation.py
justinchuby: Done

onnxscript/type_annotation.py
justinchuby: Done

gramalingam (approved):
I think your latest changes have probably not been pushed yet!

justinchuby:
> I think your latest changes have probably not been pushed yet!

Yes. ghstack is trying to push 6 changes and it was rejected by github because of our repo settings. Could you help relax that limit?

onnxscript/type_annotation.py
```diff
@@ -43,13 +43,31 @@
 
 _LIST_CONSTRUCTORS = frozenset([list, typing.List, typing.Sequence, collections.abc.Sequence])
 
+# A sorted list of all type strings used in an OpSchema
+ALL_TENSOR_TYPE_STRINGS = (
```
BowenBao, line 47: nit: must this be hardcoded?

onnxscript/type_annotation.py
```diff
@@ -143,6 +166,85 @@ def get_return_types(typeinfo: type | typing.Sequence[type]) -> typing.Sequence[
...
+        # Remove duplicates
+        return sorted(set(options))
+    if typing.get_origin(pytype) in _LIST_CONSTRUCTORS:
+        subtypes = typing.get_args(pytype)
+        return [f"seq({s})" for s in pytype_to_input_strings(subtypes[0])]
+
+    raise ValueError(f"Unsupported type: {pytype}")
+
+
+def get_type_constraint_name(pytype: TypeAnnotationValue) -> Optional[str]:
```
BowenBao, line 225: When is this function used/useful?

BowenBao (approved):
Non-blocking minor comments/questions.

onnxscript/type_annotation.py
justinchuby: It’s used for converting Python function signatures into onnx type constraints. The tests show usage. You can find actual usage in #626 and #674

onnxscript/type_annotation.py
justinchuby: Not sure where else we can find it?

gramalingam: I guess you could compute it by iterating over [tensor_type_registry](https://github.com/microsoft/onnxscript/blob/eb58e24b81b84c1a1d54a66deca6c8ebba8576e4/onnxscript/onnx_types.py#L35) ... it would be convenient to have the `to_string` method defined in one of the other PRs.

gramalingam: But do you want the 8-bit floats included in this?

justinchuby: Done

PR #635, opened 2023-04-14
Import onnx_model transformer utilities into onnxscript

shubhambhokare1:
- [ ] Initial import
- [ ] Add type annotations
- [ ] Add helper functions for OnnxFunction
- [ ] Integrate shape type inference utils
- [ ] Flesh out test example/create test suite

### DIFF
--- a/onnxscript/onnx_model.py
+++ b/onnxscript/onnx_model.py
@@ -1,0 +1,1128 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation.  All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import logging
+import os
+import sys
+from collections import deque
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple
+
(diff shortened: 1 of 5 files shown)


justinchuby:
I would move any new files to /_internal and selectively expose apis. Also consider names that are not "helper" where appropriate?

onnxscript/onnx_model.py
```diff
@@ -0,0 +1,1021 @@
...
+            return graph_input.type.tensor_type.elem_type
+
+        graph_output = self.find_graph_output(input_or_output)
+        if graph_output:
+            return graph_output.type.tensor_type.elem_type
+
+        return None
+
+    @staticmethod
+    def get_node_attribute(node: NodeProto, attribute_name: str):
```
justinchuby, line 494: nit: (knowing this will evolve) I would avoid static methods:

... instead make them module-level functions. Putting the functions in the module global namespace is a preferred way to organize them.

> Never use staticmethod unless forced to in order to integrate with an API defined in an existing library. Write a module-level function instead.

https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

PR #641, opened 2023-04-16
Update sphinx to use markdown | docs

justinchuby:
Markdowns are easier to author and work with.

Replaces #516


justinchuby:
https://github.com/onnx/onnx/pull/5149

PR #644, opened 2023-04-17
add op(native group norm) | feat(atenlib)

xiaowuhu:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,50 @@ def aten_native_dropout_backward(
...
+    bias: Optional[TFloat],
+    N: INT64 = None,  # pylint: disable=unused-argument
+    C: INT64 = None,  # pylint: disable=unused-argument
+    HxW: INT64 = None,  # pylint: disable=unused-argument
+    group: int = None,
+    eps: float = None,
+) -> TFloat:
+    # FIXME: for the return, we can only return one TReal instead of [x,y,z]
+    # Because we don't how to computer the running_var and running_mean
+    # No native_group_norm test case, and the group_norm function in torch only return one output
```
justinchuby, line 4224: We do need to return three things because that is what the aten op does. I would skip the tests if we need to

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
     raise NotImplementedError()
 
 
+@torch_op("aten::native_group_norm", trace_only=True)
```
justinchuby, line 4211: Add a comment on why trace_only is needed?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
...
+    N: INT64 = None,  # pylint: disable=unused-argument
+    C: INT64 = None,  # pylint: disable=unused-argument
+    HxW: INT64 = None,  # pylint: disable=unused-argument
+    group: int = None,
+    eps: float = None,
+) -> Tuple[TFloat, TFloat, TFloat]:
     """native_group_norm(Tensor input, Tensor? weight, Tensor? bias, SymInt N, SymInt C, SymInt HxW, int group, float eps) -> (Tensor, Tensor, Tensor)"""
 
-    raise NotImplementedError()
+    # Assert(weight is not None, and, bias is not None)
```
justinchuby, line 4224: I think they can be None? If they are it will just be input unchanged?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
...
+    # Create weight_instance_norm and bias_instance_norm
+    weight_inst = op.Constant(value_floats=[1.0] * group)
+    bias_inst = op.Constant(value_floats=[0.0] * group)
+    # 0 in the shape list keeps dimension value unchanged, for InstanceNorm need
+    shape = op.Constant(value_ints=[0, group, -1])
```
justinchuby, line 4229: What happens when group is None?

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
...
+
+    norm = _aten_native_group_norm_onnx(
+        input, weight, bias, weight_inst, bias_inst, shape, eps
+    )
+    # weight_inst, bias_inst are fake output, because we must return 3 outpurs
```
justinchuby, line 4234:
```suggestion
    # weight_inst, bias_inst are fake output, because we must return 3 outputs
```

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
...
+    group: int = None,
+    eps: float = None,
+) -> Tuple[TFloat, TFloat, TFloat]:
     """native_group_norm(Tensor input, Tensor? weight, Tensor? bias, SymInt N, SymInt C, SymInt HxW, int group, float eps) -> (Tensor, Tensor, Tensor)"""
 
-    raise NotImplementedError()
+    # Assert(weight is not None, and, bias is not None)
+    # Create weight_instance_norm and bias_instance_norm
+    weight_inst = op.Constant(value_floats=[1.0] * group)
+    bias_inst = op.Constant(value_floats=[0.0] * group)
```
justinchuby, line 4227: Can this be `Expand` and be absorbed into the function?

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Adding checking some None input so need trace_only=True

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: Set to default 1.0 and 0.0 if None

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: This must not be None. Set to 1 as default, will equal to LayerNorm.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,54 @@ def aten_native_dropout_backward(
...
+def _aten_native_group_norm_onnx(
+    input: TFloat,
+    weight: Optional[TFloat],
+    bias: Optional[TFloat],
+    weight_inst: TFloat,
+    bias_inst: TFloat,
+    shape: INT64,
+    eps: float = None,
+) -> TReal:
+    # Using InstanceNorm to simulate GroupNorm, because GroupNorm need weight[group] and bias[group]
```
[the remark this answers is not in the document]
xiaowuhu: For ONNX GroupNorm op, when group_num=2, the weight and bias should be 1x2  shape. 
But aten_native_group_norm(), when input is (N,C,xxx)=(1,6,xxx), the weight and bias should be (1x6) shape.
So we cannot use ONNX GroupNorm directly.
Torch Converter use InstanceNorm, so I copied it.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,70 @@ def aten_native_dropout_backward(
...
-    N: INT64,
-    C: INT64,
-    HxW: INT64,
-    group: int,
-    eps: float,
-) -> tuple[TensorType, TensorType, TensorType]:
+    input: TFloat,
+    weight: Optional[TFloat],
+    bias: Optional[TFloat],
+    N: Optional[INT64] = None,  # pylint: disable=unused-argument
```
BowenBao, line 4216: For my education. How do we know if an argument can be optional or not?

From aten signature below `native_group_norm(Tensor input, Tensor? weight, Tensor? bias, SymInt N, SymInt C, SymInt HxW, int group, float eps) -> (Tensor, Tensor, Tensor)`

`weight` and `bias` have `?` tagged, hence it's natural. But why are `N`, `C`, etc also optional? (And somewhat related, why does `eps` alone has default value?)

justinchuby: Good question. From the signature I don't think they should be optional. I am guessing it is due to the discrepancy in the OpInfo definition? @xiaowuhu 

I am thinking if we should create our opinfo for this op.

onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -4208,19 +4208,70 @@ def aten_native_dropout_backward(
...
+
+    if weight is None:  # Set to 1.0 as default, the shape is Channel size
+        weight = op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2))
+
+    if bias is None:  # Set to 0.0 as default, the shape is Channel size
+        bias = op.Expand(op.Constant(value_floats=[0.0]), op.Shape(input, start=1, end=2))
+
+    norm, fake_mean, fake_var = _aten_native_group_norm_onnx(input, weight, bias, group, eps)
+    # FIXME: return fake value because we must return 3 outputs(norm, mean, var)
+    # We know how the 'mean' was computed in Torch, but don't know the 'var'
```
BowenBao, line 4236: why is it called `weight_inst_norm, bias_inst_norm` in `_aten_native_group_norm_onnx`, but called `fake_mean, fake_var` here?

onnxscript/function_libs/torch_aten/ops/core.py
xiaowuhu: @justinchuby can explain more here :)
For in ONNX, we don't explicitly need N,C,HxW, because the information are already in the input tensor shape. They cannot have default value.
But, aten_native_group_norm() function must need these values as input.
For 'group' and 'eps', they should be not None. I will change the signature in next commit.

titaiwangms: If aten signature doesn't specify the default, should we find it's upper level torch function for default? If that's the case, I think eps=1-e5 would be reasonable.

onnxscript/function_libs/torch_aten/ops/core.py
titaiwangms: Looks like it;s here:
https://github.com/pytorch/pytorch/blob/8d3dc2131d6a866f826e4796c7402156aedf29b5/aten/src/ATen/native/group_norm.cpp#L64-L105

And
https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cpu/group_norm_kernel.cpp

xiaowuhu: OK, I got the formula about Torch returned var (actually it is rstd) = ```1 / (sqrt( sum/n + eps))```, where ```n = HxW*C/N```

justinchuby (approved):
Fantastic!

PR #645, opened 2023-04-17
Fix empty list handling in maxpool | fix(atenlib)

justinchuby:
Previously it was passed in as an empty list and graph_builder fails when the list is empty.

Discovered in https://github.com/pytorch/pytorch/pull/99284

- Added maxpool_2d/3d custom OpInfo tests. Skipping 3d for now due to shape mismatch.
- Pin torch-nightly version to a requirements file


BowenBao (commented):
Should we add extra custom opinfo for coverage, or contribute to pt opinfo?

justinchuby:
We should. An in this case it looks like the OpInfo for torch and the aten signature doesn't match very well.

titaiwangms:
> We should. And in this case it looks like the OpInfo for torch and the aten signature doesn't match very well.

We should add this case into our own OpInfo test.

justinchuby:
> > We should. And in this case it looks like the OpInfo for torch and the aten signature doesn't match very well.
> 
> We should add this case into our own OpInfo test.

Done

titaiwangms (approved):
LGTM

titaiwangms:
I wonder if there is a way that exporter can also use extra_Opinfo

justinchuby:
> I wonder if there is a way that exporter can also use extra_Opinfo

I can think of coping this there. Any suggestions?

justinchuby:
Somethings off with convolution

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -799,6 +801,11 @@ def _where_input_wrangler(
         reason="fixme: current implementation gets shape inference error",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
```
fatcat-z, line 804: Is this a bug?

justinchuby: It is. A bug I haven't figured out yet

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
```diff
@@ -701,6 +701,8 @@ def _where_input_wrangler(
     "index_select": core_ops.aten_index_select,
     "layer_norm": core_ops.aten_layer_norm,
     "max": core_ops.aten_max,
+    "max_pool2d": nn_ops.aten_max_pool2d,  # Custom from extra_opinfo
```
fatcat-z, line 704: We've already had nn.functional.max_pool2d and nn.functional.max_pool3d tests. If we want to use this custom testing to replace them, let's remove those 2 ones from the test file.

onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
```diff
@@ -201,6 +202,95 @@ def sample_inputs_layer_norm(op_info, device, dtype, requires_grad, **kwargs):
         )
 
 
+class _TestParamsMaxPoolEmptyStrideBase:
```
fatcat-z, line 205: Why do we keep a local custom test data here?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
fatcat-z: > I wonder if there is a way that exporter can also use extra_Opinfo

What's the scenario of using these op information during exporting?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: The new tests just test for when stride is empty

onnxscript/tests/function_libs/torch_aten/extra_opinfo.py
justinchuby: I just copied the class and edited it to contain empty strides. Any suggestions?

onnxscript/tests/function_libs/torch_aten/ops_correctness_test.py
justinchuby: We can probably enrich the pytorch opinfo db @titaiwangms

PR #648, opened 2023-04-17
Display name/shape info in TorchScriptTensor | feat(graph_building)

justinchuby:
Create repr so it doesn't show as `TorchScriptTensor(None)`

### DIFF
--- a/onnxscript/function_libs/torch_aten/graph_building.py
+++ b/onnxscript/function_libs/torch_aten/graph_building.py
@@ -76,7 +76,7 @@ class TorchScriptTensor(onnxscript_tensor.Tensor):
         self._name: Optional[str] = None
 
     def __repr__(self):
-        return f"TorchScriptTensor({self._torch_value!r})"
+        return f"TorchScriptTensor('{self._torch_value!r}')"
 
     @property  # type: ignore[override]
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_aten/graph_building.py
```diff
@@ -75,6 +75,9 @@ def __init__(self, value: torch.Value):
         self._shape: Optional[Tuple[int | None, ...]] = None
         self._name: Optional[str] = None
 
+    def __repr__(self):
+        return f"TorchScriptTensor('{self._torch_value!r}')"
```
BowenBao, line 79: Should it be composed from other properties such as shape/dtype/name other than `_torch_value`?

justinchuby: sure, I was just lazy lol

justinchuby: Actually, I think it's appropriate since it was initialized this way. Granted if we eval the string it wouldn't give us back the same thing.

BowenBao: > since it was initialized this way

This makes sense.

BowenBao (approved):

PR #650, opened 2023-04-18
Add dtype -1 condition in aten_scalar_tensor | feat(torch_lib)

titaiwangms:


onnxscript/function_libs/torch_aten/ops/core.py
```diff
@@ -5141,11 +5141,13 @@ def aten_rsub(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
...
-def aten_scalar_tensor(s: float, dtype: int = FLOAT.dtype) -> TTensor:  # type: ignore[type-var]
+@torch_op("aten::scalar_tensor", trace_only=True)
+def aten_scalar_tensor(s: float, dtype: int = -1) -> TTensor:  # type: ignore[type-var]
     """scalar_tensor(Scalar s, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
-
-    return op.Cast(s, to=dtype)
+    # If dtype is not specified in torch, aten would put -1 as default.
+    if dtype != -1:
+        return op.Cast(s, to=dtype)
+    return op.Cast(s, to=FLOAT.dtype)
```
titaiwangms, line 5150: Following original behavior of this function. However, I am not sure what we should do with -1. Is CastLike itself better?

justinchuby: Why is this change needed?

titaiwangms: Because when dtype is None, aten doesn't give us None, they give us -1

titaiwangms:
Closed as we should disable dtype -1 default assignment in exporter instead.

justinchuby:
For future reference, we need to strive to preserve existing scripted functions and not make them trace_only when possible. trace_only should be seen as a compromise and not preferred.

PR #657, opened 2023-04-21
Split ops_correctness_test.py into 3 files | chore(atenlib)

xiaowuhu:
1) ops_test_common.py, contains basic definition and functions
2) ops_test_data.py, contains the manually changed part when adding new operators. Depends on 1)
3) ops_test.py, contains test only stuff. Depends on 1) and 2).

So next time when we want to add new ops, only change 2) is good enough.


onnxscript/tests/function_libs/torch_aten/ops_base.py
justinchuby, line 1: We can call this ops_test_common or something like that

onnxscript/tests/function_libs/torch_lib/ops_correctness_test.py
justinchuby, line 1: You may delete this file because pytest will still run it at its current state

fatcat-z (approved):

Issue #418, opened 2023-02-08
Prims op support


justinchuby:
Support prims ops in torch. Need 

1. Code gen from torch._prims.
2. OpInfos for testing.

BowenBao:
Is this for "Prim IR", or are we observing prim ops for fx graphs in "ATen IR"? (IR reference https://pytorch.org/docs/master/ir.html)

justinchuby:
it's for Prims IR. Do we need/ not need them?

fatcat-z:
By calling torch._C._dispatch_get_all_op_names we can filter 189 prims ops in total. 151 of them have corresponding Opinfos in common_methods_invocations.py file while the rest of 38 don't exist.

So I think we can leverage current ops test mechanism and common_methods_invocations.py file for those 151 prims ops. Probably some of them are not implemented under aten namespace and we need to move them to prims.py file.

For the left 38 prims ops, we just need to leverage current extra_opinfo.py to customize the tests. 

So seems we don't need to do any extra work for item 2. Make sense?

PR #658, opened 2023-04-21
Create code gen tool for Prims op | feat(atenlib)

fatcat-z:
1. Implement a code gen tool for all prims op.
2. Upload the first prims.py file which contains all prims op.
3. Make sure the generated prims.py file could pass our lintrunner check.
4. Finish the item 1 in #418.


onnxscript/function_libs/tools/torch_prims/generate_prims_signatures.py
```diff
@@ -0,0 +1,341 @@
...
+        {dashline}
+        # mypy: disable-error-code="misc,arg-type,type-arg,valid-type,assignment,return-value"
+        """
+    )
+
+
+def _get_func_schema_in_namespace(namespaces: List[_OpNamespace]) -> Dict[str, FunctionSchema]:
+    table: Dict[str, FunctionSchema] = {}
+    for op_namespace in namespaces:
+        for attr_name in dir(op_namespace):
```
titaiwangms, line 248: Not sure if torch.ops.prims has missing ops issue like we had torch.ops.aten (https://github.com/pytorch/pytorch/issues/99681). But reverse look up is generally preferred I think. Please take a look at https://github.com/pytorch/pytorch/pull/99682

fatcat-z: So far, there is no such problem for prims ops. But I agree we should find a way to guarantee the op list after all prims ops are stable. In the [doc](https://pytorch.org/docs/master/ir.html#prims-ir), pirms ops are still under development so we don't have a full list for them.

xiaowuhu (approved):

justinchuby (changes requested):
Avoid rushing PRs on weekends

fatcat-z:
Create another [PR](https://github.com/microsoft/onnxscript/pull/660) which contains only 1 comment change. It has the failures as same as this one, which is caused by test_onnx_func_expe function. So merge this into main branch.

onnxscript/function_libs/tools/torch_aten/generate_prims_signatures.py
justinchuby, line 1: Is the intended path torch_aten?

justinchuby (approved):
Amazing work. Thank you!

onnxscript/function_libs/tools/torch_aten/generate_prims_signatures.py
justinchuby: Looks like I forgot to rename. I can do that later.

PR #663, opened 2023-04-24
Add Op(Tile) | feat(torchlib) 

titaiwangms:


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5819,8 +5819,8 @@ def aten_sym_size(self: TReal, dim: int = 0) -> TReal:
     shape = op.Shape(self)
     # Reshape helps dim from int to tensor, and
     # input arguments support attribute processing.
-    start = op.Reshape(dim, [1])
-    end = op.Reshape(dim + 1, [1])
+    start = op.Reshape(dim, op.Constant(value_ints=[1]))
+    end = op.Reshape(dim + 1, op.Constant(value_ints=[1]))
```
titaiwangms, line 5823: enable eager mode

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -79,6 +79,7 @@ def xfail(
     *,
     reason: str,
     dtypes: Optional[Collection[torch.dtype]] = None,
+    matcher: Optional[Callable[[Any], Any]] = None,
```
justinchuby, line 82: Extend how we use info in xfail so that we actually assert that it fails?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -723,6 +724,12 @@ def _where_input_wrangler(
         reason="ORT Graph attribute inferencing failed on rank-1 input",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
+        "tile",
+        reason="Shape inference error. Remove after ONNX 1.14 release",
```
titaiwangms, line 729: Has same xfail as aten::repeat. Locally tested on 1.15 and it works.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
titaiwangms: Could you give more details? I just thought skipping subtests with reason fixme doesn't feel right, so would like to enable xfail for those cases.

justinchuby (approved):

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
titaiwangms: https://github.com/microsoft/onnxscript/issues/669

titaiwangms:
> Be sure to check the failing jobs: https://github.com/microsoft/onnxscript/actions/runs/4799146894/jobs/8538385377?pr=663 potentially skipping ort older than 1.15

I suspect the reason this happening is that xfail doesn't go away when ort==1.15 is in environment?

PR #673, opened 2023-04-25
Implement `prims::convert_element_type` | feat(torchlib)

justinchuby:


titaiwangms (approved):
Guess we don't need a test for this guy.

justinchuby:
Still good to test but I will need to see how to construct an opinfo for this type of ops

PR #677, opened 2023-04-26
Split arg{max,min} into arg{max,min}_dim and improve function signatures | feat(atenlib)

xiaowuhu:
To remove trace_only and allow the dispatcher to select the correct function, we explicitly split argmax with argmax_dim and fixed the function return type.

---

One potential problem as below code:
```python
@torch_op("aten::argmax", overload=True)
def aten_argmax_dim(self: TReal, dim: int, keepdim: bool = False) -> TInt:
    self_is_scaler = op.Size(op.Shape(self)) == 0
    if self_is_scaler:
        self = op.Reshape(self, op.Constant(value_ints=[-1]))

    result = op.ArgMax(self, axis=dim, keepdims=keepdim)
    if self_is_scaler:
        result = op.Squeeze(result)

    return result
```
The above code works well. But when it was rewritten to below (reduce one if/else), the ShapeInference will fail:

```python
@torch_op("aten::argmax", overload=True)
def aten_argmax_dim(self: TReal, dim: int, keepdim: bool = False) -> TInt:
    self_is_scaler = op.Size(op.Shape(self)) == 0
    if self_is_scaler:
        self = op.Reshape(self, op.Constant(value_ints=[-1]))
        result = op.ArgMax(self, axis=dim, keepdims=keepdim)
        result = op.Squeeze(result)
    else:
        result = op.ArgMax(self, axis=dim, keepdims=keepdim)
    return result
```


justinchuby:
I suggest we wait a little bit to remove trace only for more commonly used ops, since the dispatcher on the torch side isn’t ready yet and currently we rely on a common target to handle different inputs. 

Eventually we will still want to combine these functions back when ONNX has its optional support in place. So we may need to be super aggressive at eliminating trace only functions for now

justinchuby:
OK. Maybe this _is_ desirable since we have now a good dispatcher. @titaiwangms is this something the dispatcher will handle?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -444,20 +444,21 @@ def aten_arctanh(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::argmax", trace_only=True)
-def aten_argmax(self: TReal, dim: Optional[int] = None, keepdim: bool = False) -> TReal:
+@torch_op("aten::argmax")
+def aten_argmax(self: TReal, keepdim: bool = False) -> TInt:
```
justinchuby, line 448:
```suggestion
def aten_argmax(self: TTensor, keepdim: bool = False) -> INT64:
```

good catch; to refine the types further:

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -444,20 +444,21 @@ def aten_arctanh(self: TensorType) -> TensorType:
...
 
-    return aten_argmax_dim(self, dim=dim, keepdim=keepdim)
+    return result
 
 
 @torch_op("aten::argmax", overload=True)
-def aten_argmax_dim(self: TReal, dim: int, keepdim: bool = False) -> TReal:
-    """argmax(Tensor self, int? dim=None, bool keepdim=False) -> Tensor"""
-
+def aten_argmax_dim(self: TReal, dim: int, keepdim: bool = False) -> TInt:
```
justinchuby, line 461:
```suggestion
def aten_argmax_dim(self: TTensor, dim: int, keepdim: bool = False) -> INT64:
```

as well as those below

justinchuby (approved):

fatcat-z (approved):

titaiwangms:
> OK. Maybe this _is_ desirable since we have now a good dispatcher. @titaiwangms is this something the dispatcher will handle?

It will. But we should be aware that dim=None is not a case anymore, which means if dim=None, we expect PyTorch ignores the keyword argument dim, instead of pass it in as None. It's different.

justinchuby:
> > OK. Maybe this _is_ desirable since we have now a good dispatcher. @titaiwangms is this something the dispatcher will handle?
> 
> It will. But we should be aware that dim=None is not a case anymore, which means if dim=None, we expect PyTorch ignores the keyword argument dim, instead of pass it in as None. It's different.

Do you know which is torch's behavior? Will it pass None?

titaiwangms:
Given the current implementation, I would say most likely it passes None. Alternatively, we can have it filtered by param schema? There is no benefit to pass in None IMO.

PR #678, opened 2023-04-26
Refine op(chunk) | feat(atenlib)

xiaowuhu:
The previous implementation is based on op17. For op18, using op.Split() by specified num_outputs attribute.

This change updates the graph builder to create a correct number of outputs for Split when num_outputs is specified.

Closes #622

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -1028,27 +1028,10 @@ def aten_choose_qparams_optimized(
 
 
 @torch_op("aten::chunk")
-def aten_chunk(self: TTensor, chunks: INT64, dim: int = 0) -> TTensor:
+def aten_chunk(self: TTensor, chunks: int, dim: int = 0) -> TTensor:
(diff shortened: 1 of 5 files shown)


justinchuby:
this may be the fix: https://github.com/microsoft/onnxscript/pull/622

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1028,27 +1028,10 @@ def aten_choose_qparams_optimized(
 
 
 @torch_op("aten::chunk")
-def aten_chunk(self: TTensor, chunks: INT64, dim: int = 0) -> TTensor:
+def aten_chunk(self: TTensor, chunks: int, dim: int = 0) -> TTensor:
```
justinchuby, line 1031:
```suggestion
def aten_chunk(self: TTensor, chunks: int, dim: int = 0) -> Tuple[TTensor, …]:
```

potentially this?

justinchuby (approved):

justinchuby:
I moved to use SplitToSequence because it is closer to what torch produces.

justinchuby:
https://github.com/onnx/onnx/issues/5192

PR #684, opened 2023-04-27
Refactor converter to isolate translate_function_signature logic | feat(converter)

justinchuby:
* #692
* #674
* #626
* __->__ #684

This change refactors the `translate_function_def` method in `Converter` to isolate the signature handling logic to `translate_function_signature`. `translate_function_signature` is used in #674 to handle function signatures so we do not need to translate the function body for general python functions incompatible with ONNX.


onnxscript/converter.py
```diff
@@ -1352,9 +1361,15 @@ def translate_function_def(self, fn: ast.FunctionDef) -> irbuilder.IRFunction:
                 self.returntype = None
         else:
             self.returntype = None
+
+        return self._current_fn
+
+    def translate_function_def(self, fn: ast.FunctionDef) -> irbuilder.IRFunction:
```
titaiwangms, line 1367: This function now looks awkwardly short. Maybe leave function descriptions on `translate_function_signature`, or comment here?

titaiwangms (commented):
The purpose of extracting the code out from an existing function should be

gramalingam (approved):

onnxscript/converter.py
justinchuby: I added a docstring. I think its quite natural to handle the signature and the body in two steps like we do now?

PR #685, opened 2023-04-27
Add an environment variable to disable ort session cache if necessary | fix(atenlib)

### LINKED ISSUE #614: Windows CI has MemoryError
First seen: https://github.com/microsoft/onnx-script/actions/runs/4641478064/jobs/8214491813,
Also https://github.com/microsoft/onnx-script/actions/runs/4642961358/jobs/8217281756

fatcat-z:
Try to disable ort session cache to avoid Windows memory exception caused by CI tests.
Fix #614

### DIFF
--- a/onnxscript/_internal/feature_switch.py
+++ b/onnxscript/_internal/feature_switch.py
@@ -1,0 +1,10 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Switches to determine if the corresponding feature of onnxscript is enabled or not."""
+
+import os
+
+# By default: Enable
+cache_ort_session = os.environ.get("CACHE_ORT_SESSION", "1")
--- a/onnxscript/_internal/feature_switch.py
+++ b/onnxscript/_internal/feature_switch.py
@@ -7,4 +7,4 @@
 import os
 
 # By default: Enable
-cache_ort_session = os.environ.get("CACHE_ORT_SESSION", "1")
+CACHE_ORT_SESSIONS: bool = os.getenv("CACHE_ORT_SESSIONS", "1") != "0"
--- a/onnxscript/evaluator.py
+++ b/onnxscript/evaluator.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 
 import abc
 import contextlib
-import os
 import pprint
 import typing
 from typing import (
@@ -28,7 +27,7 @@ import onnx.helper
 from typing_extensions import TypeAlias
 
 from onnxscript import autocast, irbuilder, onnx_opset, tensor, utils, values
-from onnxscript._internal import param_manipulation
+from onnxscript._internal import feature_switch, param_manipulation
 
 if typing.TYPE_CHECKING:
     import onnxruntime as ort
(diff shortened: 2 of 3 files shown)


justinchuby:
How about removing the cache? Any cons?

fatcat-z:
> 


> How about removing the cache? Any cons?

@gramalingam , any comments about adding this cache in the beginning?

onnxscript/evaluator.py
```diff
@@ -346,6 +347,7 @@ def _compute_num_outputs(schema: onnx.defs.OpSchema, *args: Any, **kwargs: Any):
 
 
 _cache_models: dict[Any, ort.InferenceSession] = {}
+cache_ort_session = os.environ.get("CACHE_ORT_SESSION", "1")
```
justinchuby, line 350: As discussed, we can move this to flags.py:

```python
CACHE_ORT_SESSIONS: bool = os.getenv("CACHE_ORT_SESSIONS", "1") == "1"
```

fatcat-z: Done.

onnxscript/_internal/feature_switch.py
```diff
@@ -0,0 +1,10 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Switches to determine if the corresponding feature of onnxscript is enabled or not."""
+
+import os
+
+# By default: Enable
+CACHE_ORT_SESSIONS: bool = os.getenv("CACHE_ORT_SESSIONS", "1") != "0"
```
justinchuby, line 10:
```suggestion
cache_ort_sessions: bool = os.getenv("CACHE_ORT_SESSIONS", "1") != "0"
```

sorry - I changed my mind because it looks like it is a flag that can be changed in runtime (not constant), so snake case instead?

onnxscript/_internal/feature_switch.py
justinchuby, line 1: consider `flags.py` as in "feature flags"

justinchuby (approved):
Thanks! Last round of minor suggestions.

gramalingam (approved):

onnxscript/_internal/feature_switch.py
fatcat-z: > Is this in `_internal` by design? I can see arguments both ways. It should ideally be exposed to the user, but if we think this API will change, I can see a point in leaving it here.

My opinion is: Such switches should not be exposed to users calling onnx-script ops or torch_libs to construct a model. Because these switches control only some internal features of onnx-script. These internal features are not necessary to be known by these users.

fatcat-z: > sorry - I changed my mind because it looks like it is a flag that can be changed in runtime (not constant), so snake case instead?

I think a variable in upper case is more like a constant which should not be changed in runtime. Isn't it? And I think this is aligned with the assumption that these switches are designed to be controlled by the environment variable only.

onnxscript/_internal/feature_switch.py
fatcat-z: In my past projects, such things which determines if a feature is enable or not were named like "xxxx_switch", so I followed them.
In addition, I think 'switch' implies this can control something while 'flag' is more like a static label.

PR #690, opened 2023-04-27
RefineOp(mse_loss) | feat(atenlib)

xiaowuhu:
1. Change if/elif to if-if
2. Remove xfail

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
@@ -679,12 +679,6 @@ EXPECTED_SKIPS_OR_FAILS = (
         test_class_name="TestOutputConsistencyFullGraph",
         enabled_if=version_utils.onnxruntime_older_than("1.15"),
     ),
-    xfail(
-        "nn.functional.mse_loss",
-        reason="Shape inference error. Remove after ONNX 1.14 release",
-        test_class_name="TestOutputConsistencyFullGraph",
-        enabled_if=version_utils.onnx_older_than("1.14"),
-    ),
     skip(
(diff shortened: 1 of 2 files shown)


justinchuby:
Could you compare how the graph will change? Curious if this leads to a more simple or complex graph

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -679,12 +679,6 @@ def _where_input_wrangler(
         test_class_name="TestOutputConsistencyFullGraph",
         enabled_if=version_utils.onnxruntime_older_than("1.15"),
     ),
-    xfail(
-        "nn.functional.mse_loss",
-        reason="Shape inference error. Remove after ONNX 1.14 release",
-        test_class_name="TestOutputConsistencyFullGraph",
-        enabled_if=version_utils.onnx_older_than("1.14"),
```
justinchuby, line 686: Ops like this are OK and I wouldn’t too much about if it is already fixed in ONNX 1.14. All we need to do is to wait a bit for the version bump. We are running ONNX 1.14 in our CI

xiaowuhu: The ort inference was fixed in ONNX 1.14, but the check_model method still failed.

xiaowuhu: The previous graph is:
```
if then
    ReduceMean
else
    if then
        ReduceSum
    else
        Identity
```

The new graph is:
```
if then
    ReduceMean
else
    Identity

if then
    ReduceSum
else
    Identity
```

justinchuby (approved):

PR #692, opened 2023-04-27
Create the OpLike protocol and refactor Op | feat(values)

justinchuby:
* __->__ #692
* #674
* #698

## Changes

- Creates the `OpLike` protocol that defines common attributes and methods for `Op`, `OnnxFunction` and `TraceOnlyFunction` so we can assume a common interface.
- Implement `param_schemas` for `TraceOnlyFunction`
- Refactor `param_schemas` to extract common logic.
- Removes `is_single_op` from `Op` because it is unused.

### DIFF
--- a/onnxscript/function_libs/torch_lib/tracing.py
+++ b/onnxscript/function_libs/torch_lib/tracing.py
@@ -5,13 +5,17 @@ import ast
 import inspect
 import textwrap
 import types
-from typing import Optional
+import typing
+from typing import Optional, Tuple
 
 import onnx
 
 import onnxscript
 from onnxscript import converter as ons_converter
 from onnxscript._internal import version_utils
+
+if typing.TYPE_CHECKING:
+    from onnxscript import irbuilder
 
 _ONNX_OP_SCHEMA_WRITABLE = not version_utils.onnx_older_than("1.14")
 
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/tracing.py
```diff
@@ -33,7 +37,7 @@ def _get_src_and_ast(f: types.FunctionType) -> tuple[str, ast.FunctionDef]:
     return src, f_ast
 
 
-class TraceOnlyFunction:
+class TraceOnlyFunction(onnxscript.values.OpLike):
```
gramalingam, line 40: What are the differences between a TraceOnlyFunction and an OnnxFunction? Eg., would it make sense for an OnnxFunction to be an extension (or derived class) of a TraceOnlyFunction?

titaiwangms (commented):
I suggest we move `TraceOnlyFunction` to values.py. It would make more sense to the existence of OpLike. And if we decide to go with the current setting, we should change the file name tracing to something more specific. Tracing seems general... Codes are all good!

onnxscript/function_libs/torch_lib/tracing.py
justinchuby: A TraceOnlyFunction is simply a python function with added attributes to make it more like an onnx function. To me it is a semantically different thing that happens to implement the same protocol for convenience. But I could be wroing

titaiwangms (approved):

onnxscript/function_libs/torch_lib/tracing.py
justinchuby: Let me see if a TraceOnlyFunction can be derived from Op

PR #695, opened 2023-04-28
Duplicate OpInfos for prims ops | test(torchlib)

justinchuby:
Create mechanism to duplicate OpInfos for prims ops.

Caveat: since we changed all ops to use torch.ops.prims ops, some test samples many no longer fit the signature and may need modification, in which case we should remove from this list and add the extra_opinfos.


onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -201,6 +201,43 @@ def duplicate_opinfo(opinfos: list[opinfo_core.OpInfo], name: str, new_names: tu
...
+    Args:
+        opinfos: The list of opinfo_core.OpInfo to add the new opinfo to.
+        name: The name of the opinfo to duplicate.
+        prims_name: The name of the prims op. If None, it will be set to `name`.
+    """
+    if prims_name is None:
+        prims_name = name
+    # The name of the new OpInfo
+    new_name = f"prims_{prims_name}"
+    assert prims_name.startswith("prims_")
```
fatcat-z, line 222: Just wondering if we have such limitation, probably we should not allow uses passing prims_name into this function. We just need the name and then add the prefix directly to avoid possible confusion to caller. IMO, The function name has a postfix "_prims" which is clear enough.

fatcat-z, line 222: In addition, do we need prims_name to begin with 'prims_'? If so, new_name might be "prims_prims_abs"?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1191,6 +1191,103 @@ def _where_input_wrangler(
     ),
 )
 
+PRIMS_OPS_WITH_OP_INFO = (
```
fatcat-z, line 1194:
```suggestion
# Calling dir(torch.ops.prims) to get these ops. 
# It needs to be updated before the prims op list is stable.
PRIMS_OPS_WITH_OP_INFO = (
```

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1191,6 +1191,103 @@ def _where_input_wrangler(
...
+    "tanh",
+    "transpose",
+    "trunc",
+    "uniform",
+    "var",
+    "where",
+)
+
+for op in PRIMS_OPS_WITH_OP_INFO:
+    # Duplicate opinfo for prim ops. The names all start with "prim_". E.g. "abs" -> "prim_abs"
```
fatcat-z, line 1288:
```suggestion
    # Duplicate opinfo for prim ops. The new names all start with "prim_". E.g. "abs" -> "prim_abs".
```

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -201,6 +201,43 @@ def duplicate_opinfo(opinfos: list[opinfo_core.OpInfo], name: str, new_names: tu
...
+    all_info_names = {opinfo.name for opinfo in opinfos}
+    for opinfo in opinfos:
+        if opinfo.name == name:
+            if new_name in all_info_names:
+                # NOTE: Avoid duplicating an opinfo that already exists in the database.
+                # New opinfos are expected to be added in torch-nightly.
```
fatcat-z, line 228: What does it mean by "to be added in torch-nightly"?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: Sorry that was a mistake

justinchuby: Yeah good catch

justinchuby: Removed

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: Done

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: Done

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: Removed this line. It applies to other op infos where new op infos are constantly being added so opinfos that didn't exist and needed to be created using duplication may now be available in nightly.

xiaowuhu (approved):

fatcat-z (approved):
LGTM, thanks!

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1180,6 +1180,103 @@ def _where_input_wrangler(
...
+    "transpose",
+    "trunc",
+    "uniform",
+    "var",
+    "where",
+)
+
+for op in PRIMS_OPS_WITH_OP_INFO:
+    # Duplicate opinfo for prim ops. The new names all start with "prims_". E.g. "abs" -> "prims_abs".
+    ops_test_common.duplicate_opinfo_for_prims(OPS_DB, op)
```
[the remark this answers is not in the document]
fatcat-z: Probably we don't need to throw the exception of "OpInfo {name} not found in the database.". Or update the list as below which meets the above criteria:

['log', 'normal', 'digamma', 'hypot', 'slice', 'gt', 'asin', 'remainder', 'as_strided_scatter', 'view', 'asinh', 'full', 'scalar_tensor', 'where', 'erf', 'round', 'atan2', 'amax', 'sqrt', 'where', 'tan', 'fmin', 'empty', 'amin', 'real', 'sign', 'div', 'trunc', 'log1p', 'bitwise_and', 'amin', 'trunc', 'rsqrt', 'full', 'bitwise_or', 'lt', 'log10', 'pow', 'cat', 'cosh', 'add', 'sin', 'eq', 'igamma', 'eq', 'asin', 'exp2', 'conj_physical', 'lgamma', 'pow', 'ge', 'acos', 'transpose', 'as_strided', 'full_like', 'gt', 'imag', 'gcd', 'real', 'cosh', 'fmod', 'ge', 'cos', 'minimum', 'sub', 'igammac', 'conj', 'transpose', 'tanh', 'log10', 'tanh', 'le', 'atanh', 'var', 'clone', 'fmax', 'exp', 'atan', 'amax', 'bitwise_not', 'fmod', 'acosh', 'sum', 'floor', 'tan', 'native_batch_norm', 'bitwise_xor', 'sum', 'mul', 'squeeze', 'fill', 'acos', 'log2', 'sign', 'abs', 'neg', 'rsqrt', 'log2', 'floor', 'sub', 'uniform', 'signbit', 'mul', 'sin', 'sinh', 'expm1', 'ceil', 'neg', 'erfc', 'sinh', 'atan2', 'maximum', 'bitwise_not', 'imag', 'le', 'reciprocal', 'abs', 'erfc', 'clone', 'bitwise_xor', 'ne', 'log', 'prod', 'round', 'svd', 'reciprocal', 'exp', 'squeeze', 'atan', 'remainder', 'nextafter', 'lt', 'cos', 'erf', 'rand_like', 'isfinite', 'atanh', 'ceil', 'bitwise_or', 'var', 'var_mean', 'add', 'div', 'expm1', 'log1p', 'ne', 'isfinite', 'sqrt', 'reshape', 'lgamma', 'bitwise_and']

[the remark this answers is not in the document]
justinchuby: Yeah, I updated the list to ensure all of them exist in the op_db. There's also some additional ones in https://github.com/pytorch/pytorch/issues/100223 that are named differently I will need to add.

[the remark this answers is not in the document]
justinchuby: Also it looks like torch-nightly has more infos than 2.0 does. How should we handle this difference?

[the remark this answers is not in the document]
fatcat-z: > Also it looks like torch-nightly has more infos than 2.0 does. How should we handle this difference?

If there are more opinfos meet the criteria like above ops I list in torch-nightly, it won't fail our tests so I think we don't need to care about them. We just need regularly check if any new ops in torch latest version meet above criteria and update the list accordingly.

If there are less opinfos meet the criteria like above ops I list in torch-nightly, torch-nightly CI tests will fail. Then we need to update the code to copy the opinfos of missed prims op into extra_opinfo.py file so CI tests pass.

Thoughts?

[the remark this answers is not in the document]
justinchuby: Looks like the newly added one is `empty_permuted`. I removed it for now. If we need it we can copy from torch main to extra_opinfo.py.

I think it's good to keep the "not found in the database" error, otherwise we may think we have something we don't.

PR #699, opened 2023-04-28
Fix autocast bug when schema is missing

gramalingam:
Autocast wasn't handling the case where a schema is missing correctly. Fix it.

### DIFF
--- a/onnxscript/autocast.py
+++ b/onnxscript/autocast.py
@@ -62,7 +62,7 @@ def cast_inputs(
     return tuple(cast_args)
 
 
-def dynamic_cast_inputs(op_schema: OpSchema, *args):
+def dynamic_cast_inputs(op_schema: OpSchema, args):
     """Used for autocast during eager-mode execution."""
 
     def get_type_info(x):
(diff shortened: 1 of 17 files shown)


onnxscript/autocast.py
```diff
@@ -83,13 +83,11 @@ def cast(x, typeinfo) -> tensor.Tensor:
...
 
-    return cast_inputs(get_type_info, cast, op_schema, *args)
+    return cast_inputs(get_type_info, cast, op_schema, args)
 
 
-def static_cast_inputs(converter, op_schema: Optional[OpSchema], *args):
+def static_cast_inputs(converter, op_schema: Optional[OpSchema], *args) -> tuple[str, ...]:
     """Used for autocast during script-translation."""
-    if op_schema is None:
-        return args
```
justinchuby, line 92: Do we need a comment to remind future readers?

justinchuby (approved):
Thanks! 🚀

onnxscript/autocast.py
```diff
@@ -83,13 +83,11 @@ def cast(x, typeinfo) -> tensor.Tensor:
             return tensor.Tensor(np.array(x, dtype=dtype))
         return x
 
-    return cast_inputs(get_type_info, cast, op_schema, *args)
+    return cast_inputs(get_type_info, cast, op_schema, args)
 
 
-def static_cast_inputs(converter, op_schema: Optional[OpSchema], *args):
+def static_cast_inputs(converter, op_schema: Optional[OpSchema], *args) -> tuple[str, ...]:
```
justinchuby, line 89: Should this take `args` instead of `*args` too?

gramalingam: Done

PR #698, opened 2023-04-28
Retire `get_schema` in Op | chore!(api)

justinchuby:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #698

Remove `get_schema` in Op because (1) if an op does not have schema, it will not be in its opset either. (2) If an op does have schema, we don't need to consult with the opset. This change ensures a single correct way of accessing OpSchema.

Rename `opschema` to `op_schema` for naming consistency.

PR #700, opened 2023-04-29
Skip incompatible dtypes in op_test using OpSchema | test(torchlib)

justinchuby:
This requires ONNX 1.14. I am skipping the op test for ONNX<1.14 altogether.

I tested locally to make sure no float32 test is skipped with this mechanism, and that it skips properly according to the type annotation.

Currently no new dtypes are enabled because there are too many failures.

Requires 
- #701 
- #698

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/ops_test.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test.py
@@ -162,6 +162,11 @@ def run_test_output_match(
     # An example is nn.functional.upsample_nearest2d, which has a different signature
     # than the aten operator upsample_nearest2d
     onnx_function, input_wrangler = _split_function_and_wrangler(onnx_function_and_wrangler)
+    if not ops_test_common.dtype_op_schema_compatible(dtype, onnx_function.opschema):
+        test_suite.skipTest(
+            f"dtype '{dtype}' is not supported by the op '{op.name}'. "
+            f"Type constraints: {onnx_function.opschema.type_constraints}"
+        )
 
     for i, cpu_sample in enumerate(samples):
         inputs = (cpu_sample.input, *cpu_sample.args)
--- a/onnxscript/tests/function_libs/torch_lib/ops_test.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test.py
@@ -165,7 +165,7 @@ def run_test_output_match(
     if not ops_test_common.dtype_op_schema_compatible(dtype, onnx_function.opschema):
         test_suite.skipTest(
             f"dtype '{dtype}' is not supported by the op '{op.name}'. "
-            f"Type constraints: {onnx_function.opschema.type_constraints}"
+            f"Type constraints: {onnx_function.op_schema.type_constraints}"
         )
 
     for i, cpu_sample in enumerate(samples):
--- a/onnxscript/tests/function_libs/torch_lib/ops_test.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test.py
@@ -162,7 +162,7 @@ def run_test_output_match(
     # An example is nn.functional.upsample_nearest2d, which has a different signature
     # than the aten operator upsample_nearest2d
     onnx_function, input_wrangler = _split_function_and_wrangler(onnx_function_and_wrangler)
-    if not ops_test_common.dtype_op_schema_compatible(dtype, onnx_function.opschema):
+    if not ops_test_common.dtype_op_schema_compatible(dtype, onnx_function.op_schema):
         test_suite.skipTest(
             f"dtype '{dtype}' is not supported by the op '{op.name}'. "
             f"Type constraints: {onnx_function.op_schema.type_constraints}"
--- a/onnxscript/tests/function_libs/torch_lib/ops_test.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test.py
@@ -256,6 +256,10 @@ def run_test_output_match(
                         raise
 
 
+@unittest.skipIf(
+    version_utils.onnx_older_than("1.14"),
+    "OpSchema not available for functions before ONNX 1.14",
+)
 class TestOutputConsistencyEager(unittest.TestCase):
     """Test output consistency between the ONNX op run with ONNX eager mode and PyTorch eager mode.
 
@@ -284,6 +288,10 @@ class TestOutputConsistencyEager(unittest.TestCase):
         run_test_output_match(self, device, dtype, op, ops_test_common.eager_executor)
 
 
+@unittest.skipIf(
+    version_utils.onnx_older_than("1.14"),
+    "OpSchema not available for functions before ONNX 1.14",
+)
 class TestOutputConsistencyFullGraph(unittest.TestCase):
     """Test output consistency between exported ONNX op run as a graph and PyTorch eager mode.
 
--- a/onnxscript/tests/function_libs/torch_lib/ops_test_common.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test_common.py
@@ -397,6 +397,27 @@ def dtype_op_schema_compatible(dtype: torch.dtype, schema: onnx.defs.OpSchema) -
         # has the first input as `size`, which is an integer, but it can support
         # any dtype.
         return True
+
+    # Otherwise we check the type constraints of the first input.
+    # For example, when dtype=torch.float32, and the op being tested has the schema
+    # ```
+    # OpSchema(
+    #     name='aten_abs',
+    #     domain='onnxscript.atenlib',
+    #     since_version=1,
+    #     doc='abs(Tensor self) -> Tensor',
+    #     type_constraints=[OpSchema.TypeConstraintParam(type_param_str='TReal', allowed_type_strs=['tensor(float)', 'tensor(int8)', 'tensor(int16)', 'tensor(int32)', 'tensor(int64)', 'tensor(float16)', 'tensor(double)', 'tensor(bfloat16)'], description='')],
+    #     inputs=[OpSchema.FormalParameter(name='self', type_str='TReal', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
+    #     outputs=[OpSchema.FormalParameter(name='return_val', type_str='TReal', description='', param_option=<FormalParameterOption.Single: 0>, is_homogeneous=True, min_arity=1, differentiation_category=<DifferentiationCategory.Unknown: 0>)],
+    #     attributes={}
+    # )
+    # ```
+    # we see the first input type is "TReal", corresponding to the type constraint
+    # with allowed types ['tensor(float)', 'tensor(int8)', 'tensor(int16)',
+    # 'tensor(int32)', 'tensor(int64)', 'tensor(float16)', 'tensor(double)',
+    # 'tensor(bfloat16)'].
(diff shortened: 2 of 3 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -355,6 +355,41 @@ def _format_model_and_input_information(onnx_model, inputs):
...
+}
+
+
+def dtype_op_schema_compatible(dtype: torch.dtype, schema: onnx.defs.OpSchema) -> bool:
+    if not schema.inputs:
+        # If there are no inputs, we can't check compatibility. Assume it is compatible.
+        # e.g. aten_randn has only attributes.
+        return True
+    if schema.inputs[0].name not in {"self", "input"}:
+        # If the first input is not self or input, it is usually another input that is not
```
BowenBao, line 380: I'm not following this comment section... I don't get the example too.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -355,6 +355,41 @@ def _format_model_and_input_information(onnx_model, inputs):
     )
 
 
+TORCH_DTYPE_TO_ONNX_STRING = {
+    torch.bool: "tensor(bool)",
```
BowenBao, line 359: nit: Commented elsewhere before but don't recall precisely. Does it has to be plain string for onnx types?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -355,6 +355,41 @@ def _format_model_and_input_information(onnx_model, inputs):
...
+    torch.float16: "tensor(float16)",
+    torch.float32: "tensor(float)",
+    torch.float64: "tensor(double)",
+    torch.complex64: "tensor(complex64)",
+    torch.complex128: "tensor(complex128)",
+    torch.bfloat16: "tensor(bfloat16)",
+}
+
+
+def dtype_op_schema_compatible(dtype: torch.dtype, schema: onnx.defs.OpSchema) -> bool:
```
BowenBao, line 374: I think it's worth commenting to explain the context for `dtype`. It is the "dtype setting" introduced by opinfo tests, so not any average input dtype. Feels like it holds the assumption that to some extent, all arguments outputs should have the same `dtype`.

I think it helps also to brief what it means if a schema is compatible to a dtype. Do all arguments, outputs have to share that same type?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -355,6 +355,41 @@ def _format_model_and_input_information(onnx_model, inputs):
...
+    if not schema.inputs:
+        # If there are no inputs, we can't check compatibility. Assume it is compatible.
+        # e.g. aten_randn has only attributes.
+        return True
+    if schema.inputs[0].name not in {"self", "input"}:
+        # If the first input is not self or input, it is usually another input that is not
+        # the same type as the output. Assume it is compatible.
+        # e.g. aten_add has inputs "self" and "other".
+        return True
+    first_input_dtype = schema.inputs[0].type_str
```
BowenBao, line 384: Related to the comment on this function, I'm not quite following the logic of below lines. Could you briefly explain?

BowenBao: Thoughts are clearer after realizing `first_input_dtype` is a type param str like `TReal`. Is it guaranteed that `schema.inputs[0].type_str` is always a type param str?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: Yes, it needs to be a plain string in onnx's type constraints

BowenBao: I was wondering if there is anything like `str(onnx.BoolTensorType)`/`onnx.BoolTensorType.to_string()`

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: Added When a dtype is "compatible" with the schema, it means we can use the dtype to create sample inputs by OpInfo to test the ONNX function and expect outputs to match.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: > Is it guaranteed that schema.inputs[0].type_str is always a type param str?

Do you mean it can be `sequence(tensor(int64))`? I realized that's a case I didn't handle.

justinchuby: But it's also just ops like `def aten_stack(tensors: Sequence[TTensorOrString], dim: int = 0)` whose first input is name "self" or "input" so we always test it on all dtypes.

BowenBao: Cool for finding an edge case, but to your question slightly no, I was just wondering if it is legal to set `tensor(float32)` for `schema.inputs[0].type_str`.  I think it shouldn't, otherwise it becomes weird for `schema.type_constraints`.

BowenBao: Or maybe it's legal when that's the only type supported, and type_constraints could be empty?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
BowenBao: I'm very curious if there are edge cases when sample inputs are different type tensors and how opinfo handles it. Thanks for comments. They are very clear now.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: type_str will be a name (`T1`, `TFloat` etc.) We use this name to find the allowed types in type_constraints

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -162,6 +162,11 @@ def run_test_output_match(
     # An example is nn.functional.upsample_nearest2d, which has a different signature
     # than the aten operator upsample_nearest2d
     onnx_function, input_wrangler = _split_function_and_wrangler(onnx_function_and_wrangler)
+    if not ops_test_common.dtype_op_schema_compatible(dtype, onnx_function.op_schema):
+        test_suite.skipTest(
+            f"dtype '{dtype}' is not supported by the op '{op.name}'. "
+            f"Type constraints: {onnx_function.op_schema.type_constraints}"
+        )
```
titaiwangms, line 169: I have a concern on skipping tests with the manual annotated type constraints. If we wrongly annotate an Op inputs, it could skip the tests it shouldn't, which makes us miss the very first opportunity to catch it.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: That's guaranteed to not be `tensor(float32)` if that was the question

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -355,6 +355,58 @@ def _format_model_and_input_information(onnx_model, inputs):
...
+    torch.float16: "tensor(float16)",
+    torch.float32: "tensor(float)",
+    torch.float64: "tensor(double)",
+    torch.complex64: "tensor(complex64)",
+    torch.complex128: "tensor(complex128)",
+    torch.bfloat16: "tensor(bfloat16)",
+}
+
+
+def dtype_op_schema_compatible(dtype: torch.dtype, schema: onnx.defs.OpSchema) -> bool:
```
BowenBao, line 374: I think this is good for a quick and general guard for tests and evolve from there.

What are your thoughts on more fine-grained type checking that validates type of each concrete input to the type of the matching input in schema?

BowenBao (approved):

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: > if it is legal to set `tensor(float32)` for `schema.inputs[0].type_str`. 

It is not. Or I should say we _can_ (we may be able to use it as a name), but it still needs to correspond to an entry in type_constraints. Otherwise there will an initialization error in OpSchema I believe.

onnxscript/tests/function_libs/torch_lib/ops_test.py
justinchuby: That's true. Any suggestions?

BowenBao: wonder if we can xfail

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: I think we will need to build the same logic in the dispatcher. We can evolve that can use it here. For now I think it's good to use logic that is easy to understand and get right for running test cases. It is always better to say inputs are compatible than not so we don't miss test cases. We can always put an xfail on them.

justinchuby: Also this logic is for reducing the amount of tests we need to manually skip, to accuracy isn't super important here (as long as we don't skip more than we should)

onnxscript/tests/function_libs/torch_lib/ops_test.py
titaiwangms: Based on what I found in https://github.com/pytorch/pytorch/pull/100265, indeed, if the type constraints are from onnx spec, most likely it's xfail-lable.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: I see. There probably is but we still need a mapping from torch types to onnx types. Any suggestions?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: I think this is resolved below.

onnxscript/tests/function_libs/torch_lib/ops_test.py
justinchuby: Sometimes it will xpass if we do xfail because ort _can_ choose to implement the op for some types intentionally or unintentionally. So that's a little tricky

titaiwangms: How about put the logic after runs? Just a thought, if it fails, then we check the dtype is supported or not.

justinchuby: Hmm but that’s still going to skip the tests where the annotation is incorrect?

justinchuby: I thought about this again. Maybe that’s ok for now since if we declare that something is not supported via the annotations, then the exporter will just need to bridge the types via type promotion logic. When we see unnecessary type promotion in the graph we can then come back and make changes to the annotations.

justinchuby: Merging for now. Feel free to share more thoughts!

PR #711, opened 2023-05-03
Indexing and slicing support

gramalingam:
Part 1 of extending the support for indexing and slicing (expressions of the form `e[i_1, i_2, i_3]`).

This updates only the runtime (eager) component. The updates to the converter will be part 2 (once this is accepted/finalized).

Primarily extends support to the use of non-scalar tensors as an index (which is effectively a `Gather` operation). Also implements a specific heuristic to choose between the use of `Slice` and `Gather` in some cases where both can be used.

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -162,8 +162,8 @@ code to work for different types of `X`.
 
 {{onnxscript}} supports the use of Python's indexing and slicing operations on
 tensors, which are translated into ONNX's `Slice` and `Gather` operations.
-The semantics of this operation is similar to that of Numpy's
-.
+The semantics of this operation is similar to that of Numpy's.
+
 In the expression `e[i_1, i_2, ..., i_n]`, `n` is either the rank of the
 input tensor or any value less than that. Each index-value `i_j` may be
 a scalar value (a tensor of rank zero) or a higher-dimensional tensor or
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -176,8 +176,9 @@ However, the translator maps indexing using slice-expressions to ONNX's
 operation. The more general case (where `i_j` is an arbitrary tensor) is
 translated using the `Gather` operation.
 
-A limitation of the current implementation is that it does not support the use
-of ellipsis or newaxis in the index.
+Note: The current implementation does not yet support the use of arbitrary
+tensors in the index-expressions. It does not support the use of ellipsis or
+newaxis in the index.
 
 ## Control-Flow
 
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -158,6 +158,25 @@ same type as some other (non-constant) operand. For example, the expression
 `2 * X` is expanded to `op.CastLike(2, X) * X`, which allows the same
 code to work for different types of `X`.
 
+## Indexing and Slicing
+
+{{onnxscript}} supports the use of Python's indexing and slicing operations on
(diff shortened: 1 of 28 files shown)


docs/tutorial/index.md
```diff
@@ -158,6 +158,28 @@ same type as some other (non-constant) operand. For example, the expression
 `2 * X` is expanded to `op.CastLike(2, X) * X`, which allows the same
 code to work for different types of `X`.
 
+## Indexing and Slicing
+
+{{onnxscript}} supports the use of Python's indexing and slicing operations on
+tensors, which are translated into ONNX's `Slice` and `Gather` operations.
+The semantics of this operation is similar to that of Numpy's
+.
```
justinchuby, line 166:
```suggestion
The semantics of this operation is similar to that of Numpy's.

```

minor

onnxscript/tensor.py
```diff
@@ -39,6 +39,10 @@ def value(self) -> np.ndarray:
     def rank(self) -> int:
         return len(self.value.shape)
```
justinchuby, line 40: One issue when I was using rank was np.ndarray does not have it, and so calling it in a trace only function will fail because it was not converted to a Tensor. Now that we have a special class for these functions (TracedOnnxFunction) we may be able to wrap ndarrays there? Or maybe constraint users to always provide Tensors instead of ndarrays.

onnxscript/tensor.py
```diff
@@ -73,40 +77,82 @@ def __getitem__(self, index):
...
-            # Treat 1-dimensional slicing A[i:j] as generic n-dimensional case A[i:j,...]
+        if not isinstance(index, tuple):
+            # Normalize representation to a tuple.
+            # A single index-value is equivalent to a tuple with a single element.
             index = (index,)
+        if len(index) > self.rank:
+            raise ValueError(
+                f"Number of indices {len(index)} is greater than rank {self.rank}"
+            )
+        from onnxscript import autocast
```
justinchuby, line 88: Was this the circular import issue?

docs/tutorial/index.md
gramalingam: Weird. Wonder how it happened.

onnxscript/tensor.py
gramalingam: Yes, we could do something like that for tracing-mode. Orthogonal to current PR, I guess ... (at this point) this code is not meant for tracing-mode.

onnxscript/tensor.py
gramalingam: I put it here to best handle the circularity. (My question the other day was about another issue relating to circular dependence.)

justinchuby: In that case I suggest putting it on top unless there is an absolute need. Hiding imports in a function delays any import errors that should happen at load time and makes refactoring harder.

gramalingam: Putting it on top will create a warning about the circular import. How does that help? It clutters up the messages, and becomes a false-positive. Putting it here, in fact, clarifies that it is a false-positive. False-positives also have the effect of making users simply ignore such messages.

What refactoring is impacted by this?

justinchuby (approved):

onnxscript/tensor.py
gramalingam: Following up our offline discussion: the main reason I have it here is to avoid "noise" due to warnings about cyclic-imports. If we can turn that noise off, that would address what bothers me. (This is based on my impression from the github repo settings a while back, not sure if it is still the same.)

PR #715, opened 2023-05-03
Disable `py/cyclic-import` in codeql

justinchuby:
Module level cyclic import, although should be avoided if possible, is not a problem from the perspective of functionality. It is more preferable than hiding imports in functions from a readability and maintainability point of view.


gramalingam (approved):
Thanks!

jcwchen (approved):

PR #719, opened 2023-05-04
Refactor Overloads | feat(torchlib)

titaiwangms:
After this PR, overloads is deleted, and when we register a **aten overload**, we use it's specific name: `aten::mean.dim`. If it's an **onnx overload** created regarding ONNXFunction limitations, such as `aten_bitwise_not_bool`, we can simply register it under the same aten name. The dispatcher in converter side is expected to find the proper one with op schema.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -44,7 +44,7 @@ def aten_abs(self: TReal) -> TReal:
     return op.Abs(self)
 
 
-@torch_op("aten::abs", overload=True)
+@torch_op("aten::abs")
 def aten_abs_complex(self: TReal) -> TReal:
     """abs(Tensor self) -> Tensor"""
     # self_real = self[..., 0]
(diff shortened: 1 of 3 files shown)


justinchuby (approved):
Thank you!

justinchuby:
> Is it correct to understand that torch **aten overload**s are viewed as different aten ops in onnxscript registry, and each **aten op** can map to multiple **onnx overload**s?

That's right!

> How to decide which **onnx overload** to dispatch to?

Ti-Tai is working on the dispatching logic. @titaiwangms feel free to share more

titaiwangms:
Yes, the main logic would be leveraging on OpSchema. I will see if the PR can be ready this week.

BowenBao:
> Yes, the main logic would be leveraging on OpSchema. I will see if the PR can be ready this week.

Awesome! Is the main logic living in exporter or here?

titaiwangms:
> > Yes, the main logic would be leveraging on OpSchema. I will see if the PR can be ready this week.
> 
> Awesome! Is the main logic living in exporter or here?

In exporter with a Dispatcher and an OpschemaWrapper

PR #721, opened 2023-05-05
Fix ops type/name and adapt API to support torch dispatcher/registry | feat(torchlib)

titaiwangms:
~~Improve logic and fix bugs in Registry to provide completed torchlib to converter.~~

To enable dispatcher in converter: https://github.com/pytorch/pytorch/pull/100660

1. Expose TracedOnnxFunction
2. Add items() into Registry
3. Fix ops type/name (caught by converter side op test)
4. seq(tensor) remains unsupported in TorchScriptTensor

To enable dispatcher in torch converter with utilizing OpSchema, I am annotating all of input arguments from FX graph except seq(tensor), which requires us to address the non-annotated type (Sym value and ONNX Sequence):

https://github.com/pytorch/pytorch/blob/9a811d1df23acd8285e72f938c3919a4993bca69/torch/onnx/_internal/fx/passes/fx_to_onnxscript.py#L197-L208

NOTE:
For Sym value, as we are treating them as normal TorchScriptTensor with aten_add , aten_sub, ... etc, they would be annotated as int64 (SymInt), and float32 (SymFloat).

For ONNX sequence, these are not defined in torch, nor in torchscript. We return NoneType and dispatcher see NoneType/None/empty list as AllType to relax the match.


onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -40,6 +40,13 @@ def register(self, func: Any, name: str, *, private: bool = False) -> None:
     def __getitem__(self, name):
         return self._registry[name]
 
+    def get_functions(self, name: str):
+        return self._registry.get(name, None)
+
+    def is_registered(self, name: str):
```
justinchuby, line 46: I think this is implemented in __contains__ below already. One can use

`name in registry` to check

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -40,6 +40,13 @@ def register(self, func: Any, name: str, *, private: bool = False) -> None:
     def __getitem__(self, name):
         return self._registry[name]
 
+    def get_functions(self, name: str):
```
justinchuby, line 43:
```suggestion
    def get(self, name: str):
```

i would suggest a more pythonic method name, or are you thinking to include methods that get_other_things?

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -40,6 +40,13 @@ def register(self, func: Any, name: str, *, private: bool = False) -> None:
...
+        return self._registry.get(name, None)
```
justinchuby, line 44:
```suggestion
        return self._registry.get(name)
```

onnxscript/function_libs/torch_lib/registration.py
titaiwangms: No worry. After talk to @BowenBao, I will have these registry APIs in converter most likely. This is just a draft that I am debugging while implementing dispatcher. I will have you review on final version if you are available at the time.

titaiwangms: I will probably have to implement registry/dispatcher class in converter. And this one would only be imported as part of registration source.

justinchuby: Sounds like a plan

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -130,13 +134,21 @@ def shape(self, shape: Tuple[int | None, ...]):
     @property
     def dtype(self):
```
BowenBao, line 135: nit: is it possible to annotate return type?

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -49,6 +49,9 @@ def __iter__(self):
     def __repr__(self):
         return repr(self._registry)
 
+    def items(self):
```
BowenBao, line 52: nit: return type annotation?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -75,6 +75,10 @@ def __init__(self, value: torch.Value):
         self._shape: Optional[Tuple[int | None, ...]] = None
         self._name: Optional[str] = None
         self._is_complex: bool = False
+        # NOTE(titaiwang): PyTorch doesn't support ONNX dtype: seq(tensor),
+        # so we use string format type annotation to enable dtype, but it's
+        # not supported in torchscript.
+        self._sequence_dtype_str: Optional[set] = None
```
BowenBao, line 81: why is it a set?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -130,13 +134,21 @@ def shape(self, shape: Tuple[int | None, ...]):
     @property
     def dtype(self):
         # TODO: Return numpy dtype
-        return _type_utils.JitScalarType.from_value(  # type: ignore[attr-defined]
+        torch_dtype = _type_utils.JitScalarType.from_value(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
BowenBao: I see, you don't have a `torch.dtype` for sequence to be passed in from `dtype.setter` to begin with.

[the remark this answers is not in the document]
BowenBao: Frankly feels it's a bit overloaded and doesn't seem to match the design for this class, to be also covering sequence type.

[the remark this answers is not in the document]
titaiwangms: Should we add another attribute?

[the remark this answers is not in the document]
titaiwangms: One of alternative way that I can think of is that we give up this dtype, and try to relax `match_schema` in a way that it finds the best match instead of "the 100% match", as I found some inputs are just None anyway.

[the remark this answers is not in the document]
titaiwangms: Reverted

onnxscript/function_libs/torch_lib/graph_building.py
titaiwangms: Done

gramalingam (approved):

PR #727, opened 2023-05-11
add float16 test base functions and data

xiaowuhu:


onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -71,6 +71,27 @@ def _split_function_and_wrangler(
...
+    torch.float32: (3.7e-5, 1.8e-4),  # default is 1.3e-6, 1e-5
+    torch.float16: (1e-3, 1e-5),
+}
+
+
+def _get_rtol_atol_by_dtype(dtype: torch.dtype) -> tuple(Any, Any):
+    if dtype in OPINFO_PRECISION_TABLE:
+        return OPINFO_PRECISION_TABLE[dtype]
+    else:
+        return (0.0, 0.0)
```
fatcat-z, line 87: Could we apply the default value (default is 1.3e-6, 1e-5) for other possible dtypes?

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -71,6 +71,27 @@ def _split_function_and_wrangler(
...
+
+
+def _is_op_can_run_with_dtype(op_name: str, dtype: torch.dtype) -> bool:
```
fatcat-z, line 90: Suggest a new name: _dtype_is_supported_by_op()

xiaowuhu: I found the previous implementation is return (None, None), so I followed it.

onnxscript/tests/function_libs/torch_lib/ops_test.py
xiaowuhu: I found the previous implementation is return (None, None), so I followed it.

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -32,9 +32,13 @@
 from onnxscript._internal import version_utils
 from onnxscript.tests.function_libs.torch_lib import ops_test_common, ops_test_data
 
-# Test only float32 inputs. All dtypes will be tested on the generated symbolic functions.
+# All dtypes will be tested on the generated symbolic functions.
```
titaiwangms, line 35: How is float32 and float16 all dtypes?

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -71,6 +75,27 @@ def _split_function_and_wrangler(
...
+    # The current most relaxed values are for aten::matmul
+    torch.float32: (3.7e-5, 1.8e-4),  # default is 1.3e-6, 1e-5
+    torch.float16: (1e-3, 1e-5),
+}
+
+
+def _get_rtol_atol_by_dtype(dtype: torch.dtype) -> tuple(Any, Any):
+    if dtype in OPINFO_PRECISION_TABLE:
+        return OPINFO_PRECISION_TABLE[dtype]
+    else:
```
titaiwangms, line 90: nit: unnecessary else

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1316,3 +1316,292 @@ def _where_input_wrangler(
 assert NONDETERMINISTIC_OPS.issubset(
     TESTED_OPS
 ), f"{NONDETERMINISTIC_OPS - TESTED_OPS} not in TESTED_OPS"
+
+# List for different input dtype testing flag
+OPINFO_FUNCTION_TARGET_DTYPE: dict[
```
titaiwangms, line 1321: Why not use skip() on dtypes?

titaiwangms (approved):

titaiwangms (commented):
Oops! I accidentally hit the approve button. Please don't merge yet!

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -303,7 +321,10 @@ def setUp(self) -> None:
     def test_output_match_opinfo_(
         self, device: str, dtype: torch.dtype, op: opinfo_core.OpInfo
     ):
-        """Base test method for testing each op with the eager executor, used by instantiate_device_type_tests."""
+        if not _dtype_is_supported_by_op(op.name, dtype):
+            return
```
justinchuby, line 325: Prefer skipping or else we mistake it as a pass?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1316,3 +1316,292 @@ def _where_input_wrangler(
...
+    str,
+    tuple[Any, ...],
+] = {
```
justinchuby, line 1324: What is our plan to ensure type annotations are correct with this new table, since what is tested is now independent of the annotations ?

onnxscript/tests/function_libs/torch_lib/ops_test.py
xiaowuhu: INT64, INT32, INT8, etc. will be gradually add in, so it is for all dtypes.

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
xiaowuhu: skip itself is a ‘best of a bad bunch' (无奈之举), you can imagine: 

- When we enable torch.float16 testing, we will get hundreds of failed cases, we must skip all the failed case then check in. That means only one dev can work on this, you cannot check in code, cannot parallelly work for multiple devs.
- skip is heavy, it need step into run_test_output_match() function then check skip condition, then quit.
- If we use OPINFO_FUNCTOIN_TARGET_DTYPE, we can gradually add the target functions into the list (one by one), the extreme case is one PR for one aten function testing.

onnxscript/tests/function_libs/torch_lib/ops_test.py
xiaowuhu: changed to use pytest.skip()

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
titaiwangms: I see

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1316,3 +1316,292 @@ def _where_input_wrangler(
...
+        torch.float32,
+        torch.float16,
+    ),
+    "acosh": (
+        torch.float32,
+        torch.float16,
+    ),
+    "add": (
+        torch.float32,
+        # FIXME: float16 failed, tensor-likes are not close for FullGraph mode
```
titaiwangms, line 1351: Interesting

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
xiaowuhu: sorry, not understand your question. @fatcat-z any idea?

titaiwangms:
This is breaking CI, any idea?

gramalingam:
Yes, looks like at least some commas are missing for tuples: eg., [here](https://github.com/microsoft/onnxscript/blob/1681c242f60895e02970c9a47ba71ea547e36200/onnxscript/tests/function_libs/torch_lib/ops_test_data.py#L1478)

xiaowuhu:
fix is here: https://github.com/microsoft/onnxscript/pull/734

PR #735, opened 2023-05-16
test functions in fp16

xiaowuhu:


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1415,193 +1415,728 @@ def _where_input_wrangler(
...
+        torch.float32,
+        torch.float16,
+    ),
+    "ceil": (
+        torch.float32,
+        torch.float16,
+    ),
+    "chunk": (
+        torch.float32,
+        # torch.float16,  # FIXME: SplitToSequence op inference failed
```
fatcat-z, line 1440: nit: Could we link the issue url here? If no, could you please open one?

xiaowuhu: This is just for remembering this issue, will try to fix it later. Otherwise, will file a bug to ORT.

PR #736, opened 2023-05-17
Refine op(full_like) for float16

xiaowuhu:
1. Split full_like into 2 functions: full_like and full_like_dtype. Otherwise cannot pass the FullGraph testing because they return different dtype.
2. File 3 issues for ORT
 - https://github.com/microsoft/onnxruntime/issues/15977
 - https://github.com/microsoft/onnxruntime/issues/16006
 - https://github.com/microsoft/onnxruntime/issues/16007


fatcat-z (approved):

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1556,7 +1572,11 @@ def _where_input_wrangler(
     ),
     "full_like": (
         torch.float32,
-        # torch.float16,  # FIXME: dtype don't match.
+        torch.float16,  # FIXME: dtype don't match.
```
titaiwangms, line 1575: Why is FIXME still here?

PR #738, opened 2023-05-19
RefineOpForFP16(2)

xiaowuhu:
Refine aten_isfinite by adding cast at beginning, so it can support all float dtypes.
The original implementation 'self: Union[FLOAT, DOUBLE]' is only for skip the testcase. When used with FP16, it will failed.

I wrote some Notes in ops_test_data.py, just like:
```python
# List for different input dtype testing flag
# Added Cast inside below functions so they can support all real dtypes naturally
# -- isfinite, isinf, isneginf, isposinf
# Note:
# Converter fp32 to fp16 model is a significant feature for end users,
# another approach is to add cast before/after the function call,
# that way we also need a list to remember which function need cast
OPINFO_FUNCTION_TARGET_DTYPE: dict[
    str,
    tuple[Any, ...],
] = {
...
```
We have a list, so we can know the functions those we made this kind of change.
If we have some other better solutions that can support this scenario, we can also revert this change.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3171,9 +3171,10 @@ def aten_isfinite(self: TFloatOrBFloat16) -> BOOL:
 
 
 @torch_op("aten::isinf")
-def aten_isinf(self: Union[FLOAT, DOUBLE]) -> BOOL:
+def aten_isinf(self: TFloatOrBFloat16) -> BOOL:
     """isinf(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
     return op.IsInf(self)
 
 
@@ -3181,20 +3182,23 @@ def aten_isinf(self: Union[FLOAT, DOUBLE]) -> BOOL:
 def aten_isnan(self: TFloatOrBFloat16) -> BOOL:
     """isnan(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
     return op.IsNaN(self)
 
 
 @torch_op("aten::isneginf")
-def aten_isneginf(self: TReal) -> BOOL:
+def aten_isneginf(self: TFloatOrBFloat16) -> BOOL:
     """isneginf(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
     return op.And(op.Less(self, 0), op.IsInf(self))
 
 
 @torch_op("aten::isposinf")
-def aten_isposinf(self: TReal) -> BOOL:
+def aten_isposinf(self: TFloatOrBFloat16) -> BOOL:
     """isposinf(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
     return op.And(op.Greater(self, 0), op.IsInf(self))
 
 
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3164,7 +3164,8 @@ def aten_isclose(
 def aten_isfinite(self: TFloatOrBFloat16) -> BOOL:
     """isfinite(Tensor self) -> Tensor"""
 
-    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all real types
+    # Added Cast inside the function so it can support all real dtypes naturally
+    self = op.Cast(self, to=FLOAT.dtype)
     not_inf = op.Not(op.IsInf(self))  # op.IsInf() only support FLOAT and DOUBLE
     not_nan = op.Not(op.IsNaN(self))  # TODO: The test case doesnt cover this condition
     return op.And(not_inf, not_nan)
@@ -3189,7 +3190,8 @@ def aten_isnan(self: TFloatOrBFloat16) -> BOOL:
 def aten_isneginf(self: TFloatOrBFloat16) -> BOOL:
     """isneginf(Tensor self) -> Tensor"""
 
-    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
+    # Added Cast inside the function so it can support all real dtypes naturally
+    self = op.Cast(self, to=FLOAT.dtype)
     return op.And(op.Less(self, 0), op.IsInf(self))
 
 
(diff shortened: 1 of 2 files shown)


fatcat-z (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3164,36 +3164,41 @@ def aten_isclose(
 def aten_isfinite(self: TFloatOrBFloat16) -> BOOL:
     """isfinite(Tensor self) -> Tensor"""
 
-    not_inf = op.Not(op.IsInf(self))
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all real types
```
titaiwangms, line 3167: This op.Cast makes no sense to me. op.IsInf simply doesn't support float16. I would change input annotation to TFloatUnlessFloat16

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3164,36 +3164,41 @@ def aten_isclose(
...
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
     return op.IsInf(self)
 
 
 @torch_op("aten::isnan")
 def aten_isnan(self: TFloatOrBFloat16) -> BOOL:
     """isnan(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
```
titaiwangms, line 3185: I don't have error without this op.Cast

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3164,36 +3164,41 @@ def aten_isclose(
...
     not_nan = op.Not(op.IsNaN(self))  # TODO: The test case doesnt cover this condition
     return op.And(not_inf, not_nan)
 
 
 @torch_op("aten::isinf")
-def aten_isinf(self: Union[FLOAT, DOUBLE]) -> BOOL:
+def aten_isinf(self: TFloatOrBFloat16) -> BOOL:
     """isinf(Tensor self) -> Tensor"""
 
+    self = op.Cast(self, to=FLOAT.dtype)  # Make this function support all float types
```
titaiwangms, line 3177: Op simply doesn't support float16.

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -811,12 +811,12 @@ def _where_input_wrangler(
         matcher=lambda sample: sample.args[1] == 2,
         reason="fixme: 'bicubic' mode in ORT implemented differently with Torch",
     ),
-    xfail(
+    skip(
```
titaiwangms, line 814: Some float16 test cases passed?

titaiwangms (changes requested):
Sorry I have to request changes. I don't think adding op.Cast is a good idea, specfically when the op spec states it doesn't support certain dtype. Even this is some upcast/downcast strategy you want to apply on torchlib, I still think we should keep the function inside cleaner, and apply upcast/downcast strategy externally.

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Yes, your consideration is also reasonable. Although I think I'm right, while I cannot say you are wrong. :)
If the annotation TFloaUnlessFloat16 can help when programmatically convert the model to fp16, I'm OK.

titaiwangms: Although OnnxFunction doesn't have to have the same spec as ONNX, our annotation rule rn is to annotate them by the intersection of type constraints of used Ops. Adding Cast here to expand the function support is really weird. Even if we wanted to expand the dtype support  by upcast/downcast, we should do it in converter side IMO.

titaiwangms: @justinchuby and I discussed about adding upcast/downcast into the dispatcher to support more dtype that are out of ONNX spec. Maybe it's similar to what you wanted to do here?

titaiwangms: I wasn't in the discussion, could you elaborate more detail about why we definitely need to support fp16 on a function using non-support fp16 op?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
xiaowuhu: No relation with float16 here. I just think it should be skip instead of xfail. But I don't have strong reason.

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Similar thought but different way. I think FP16 is special, because almost all the users will convert fp32 model to fp16 (because I maintain the convert_fp32_to_fp16 tool in onnxmltools, so I know). For other dtypes such as int64 or double, I think the priority is low, or even don't need to care about.

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Yes, should be by mistake.

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: I see. Could you find somewhere good to put a NOTE something like "FP16 is special case that we support with inserting op.Cast inside function."?

titaiwangms (approved):
I can't say I fully understand the issue we would like to address with special support on FP16, but I am also far from the owner of torchlib. Please put a NOTE with these special considerations we have in somewhere of torchlib, so people would know we have special treatment for FP16. THanks!

xiaowuhu:
> I can't say I fully understand the issue we would like to address with special support on FP16, but I am also far from the owner of torchlib. Please put a NOTE with these special considerations we have in somewhere of torchlib, so people would know we have special treatment for FP16. THanks!

Sure, let me think about it.

justinchuby:
Thanks for the discussion! Personally I would look at this from a correctness point of view: when we insert the cast, is the function still correct for all supported input types? E.g. when the input is a big float64 number (that is out of float32’s range), would the cast make a finite number infinite?

xiaowuhu:
A too big number is meaningless for our scenario (We are working in DL domain, in most time, which the number represent the probability 0~1).
The fact is ORT might also not support fp64 mode. With your assumption, we can do nothing for this case.

PR #744, opened 2023-05-23
Refactor op(scatter add) | feat(torchlib)

titaiwangms:
It wasn't following aten signature, so it's unusable on converter side.
Based on the agreement: https://github.com/microsoft/onnxscript/pull/582#discussion_r1154491445


titaiwangms:
@xiaowuhu @fatcat-z PTAL

xiaowuhu (approved):
A long time ago, int type cannot locate before Tensor type in onnxscript. So it is good now.

justinchuby:
> A long time ago, int type cannot locate before Tensor type in onnxscript. So it is good now.

That’s still true for scripted functions I think?

Issue #703, opened 2023-05-01
[torchlib] Implement mean


BowenBao:
https://github.com/pytorch/pytorch/blob/3e3cf02c44f33d050e12eca4b28cabc1c2ee1aa3/test/onnx/test_fx_to_onnx_with_onnxruntime.py#L306

justinchuby:
@xiaowuhu could you help take a look? Thanks!

titaiwangms:
@xiaowuhu Do you have time to work on this? [TorchBench](https://github.com/microsoft/onnx-converters-private/issues/166) shows that this op basically blocks lots of model, so we figure if we can support it, then we can rocket-increase model coverage.

xiaowuhu:
@justinchuby @titaiwangms Just notice this message, sorry for the late. I will take a look soon.

titaiwangms:
Thank you very much!

PR #748, opened 2023-05-26
Add op(mean) | feat(torchlib)

xiaowuhu:
Revolves https://github.com/microsoft/onnxscript/issues/703


titaiwangms (commented):
Hmm.. we have op.Mean actually. It's not good for aten_mean and aten_mean_dim?

xiaowuhu:
> Hmm.. we have op.Mean actually. It's not good for aten_mean and aten_mean_dim?

op.Mean is element wise computation, its input is list of tensor.
https://onnx.ai/onnx/operators/onnx__Mean.html

titaiwangms (approved):
Thanks!

PR #754, opened 2023-05-31
Add op (AvgPool) | feat (torchlib)

titaiwangms:
1D and 3D

[Fix](https://github.com/microsoft/onnxruntime/pull/16752) can save the op after opset 19, but not before it. So an xfail is created when ceil_mode=True


titaiwangms:
I am blocked by `ceil_mode`. It seems that the `ceil_mode` in ORT is not following ONNNX spec:

```python
import numpy as np
from onnxscript.function_libs.torch_lib.ops.nn import aten_avg_pool2d

x = np.array(
    [
        [
            [
                [1, 2, 3, 4],
                [5, 6, 7, 8],
                [9, 10, 11, 12],
                [13, 14, 15, 16],
            ]
        ]
    ]
).astype(np.float32)
kernal = [3, 3]
stride = [2, 2]

print(aten_avg_pool2d(x, kernel_size=kernal, stride=stride, ceil_mode=True))
# Tensor(array([[[[6., 5.],
#                         [8., 6.]]]], dtype=float32))

# https://github.com/onnx/onnx/blob/main/docs/Operators.md#averagepool
# The answer should be:
# np.array([[[[6, 7.5], [12, 13.5]]]]).astype(np.float32)
```

titaiwangms:
> I am blocked by `ceil_mode`. It seems that the `ceil_mode` in ORT is not following ONNX spec:
> 
> ```python
> import numpy as np
> from onnxscript.function_libs.torch_lib.ops.nn import aten_avg_pool2d
> 
> x = np.array(
>     [
>         [
>             [
>                 [1, 2, 3, 4],
>                 [5, 6, 7, 8],
>                 [9, 10, 11, 12],
>                 [13, 14, 15, 16],
>             ]
>         ]
>     ]
> ).astype(np.float32)
> kernal = [3, 3]
> stride = [2, 2]
> 
> print(aten_avg_pool2d(x, kernel_size=kernal, stride=stride, ceil_mode=True))
> # Tensor(array([[[[6., 5.], [8., 6.]]]], dtype=float32))
> 
> # https://github.com/onnx/onnx/blob/main/docs/Operators.md#averagepool
> # The answer should be:
> # np.array([[[[6, 7.5], [12, 13.5]]]]).astype(np.float32)
> ```
> 
> So the difference is that in `ceil_mode`, ORT tends to divide the value by kernal_size no matter the right padding exists or not. But I don't think they would take this as repro..

Just found this issue is caused by "count_include_pad=True will pad the right side even if it's not padded". So the following code would be able to solve this:

```python
x = np.array(
    [
        [
            [
                [1, 2, 3, 4],
                [5, 6, 7, 8],
                [9, 10, 11, 12],
                [13, 14, 15, 16],
            ]
        ]
    ]
).astype(np.float32)

import onnxscript
from onnxscript.onnx_opset import opset18 as op

@onnxscript.script(default_opset=op)
def avg_pool(x):
    result = op.AveragePool(x, kernel_shape=[3,3], strides=[2,2], ceil_mode=True, count_include_pad=False)
    return result

print(avg_pool(x))
```

Although I still think this doesn't make sense.

titaiwangms:
However, the above doesn't solve the root cause which is that in ORT, the last slide of window tends to dividing the value with kernal size regardless the remain pixel+pads seems like a bug.

titaiwangms:
https://github.com/onnx/onnx/issues/5276

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -126,6 +126,83 @@ def aten_adaptive_max_pool3d_backward(
...
+    result = op.AveragePool(
+        self,
+        ceil_mode=ceil_mode,
+        count_include_pad=count_include_pad,
+        kernel_shape=kernel_shape,
+        pads=pads,
+        strides=strides,
+    )
+
+    # TODO: if want to support divisor_override argument, need to op.Mul(result, mask)
```
fatcat-z, line 190: Since avg_1d doesn't support divisor_override, let's remove this from this function to avoid confusion.

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -946,9 +949,21 @@ def _where_input_wrangler(
         matcher=lambda sample: sample.args[0] != (1, 1, 1),
         reason="only global pooling is supported; only batched inputs are supported",
     ),
+    # xfail(
+    #     "nn.functional.avg_pool1d",
+    #     matcher=lambda sample: sample.kwargs.get("ceil_mode") is True,
+    #     reason="ONNX has different ciel mode strategy to PyTorch",
```
fatcat-z, line 955:
```suggestion
    #     reason="ONNX has different ceil mode strategy to PyTorch",
```

xiaowuhu (approved):

titaiwangms:
https://github.com/microsoft/onnxruntime/issues/16203

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1399,10 +1399,15 @@ def _where_input_wrangler(
...
-    ).xfail(
+    )
+    .xfail(
         matcher=lambda sample: (len(sample.args) > 5 and sample.args[5] is not None)
         or (sample.kwargs.get("divisor_override") is not None),
         reason="ONNX doesn't support divisor_override argument",
+    )
+    .xfail(
+        matcher=lambda sample: sample.kwargs.get("ceil_mode", True),
+        reason="ONNXRUNTIME doesn't match PyTorch when ceil_mode=True until opset 19",
```
justinchuby, line 1410:
```suggestion
        reason="fixme: ORT doesn't match PyTorch when ceil_mode=True until opset 19",
```

justinchuby: as well as the line below

justinchuby (approved):

PR #757, opened 2023-06-01
Fix op (MaxPool) | feat(torchlib)

titaiwangms:
Found this bug when I try to apply this op in a model (https://github.com/pytorch/pytorch/pull/102773). Basically, if the input padding is 3d, we still need to double it to fit the ONNX requirement that we need begin and end padding. I am not sure why we didn't catch this in ops_test.py though, it seems should be caught...

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -711,16 +711,26 @@ def _adjust_attributes_of_max_pool(
     else:
         kernel_shape = kernel_size
 
-    # expand_size is the dimension of pooling kernel,
+    # NOTE: expand_size is the dimension of pooling kernel,
     # ONNX needs begin and end padding so we need to double the padding
+
+    # NOTE: expand size prevents padding from being a single int in
+    # multiple dimension cases
     if isinstance(padding, int):
         pads = [padding] * expand_size * 2
     elif len(padding) == 1:
         pads = padding * expand_size * 2
     elif len(padding) == 2:
-        pads = padding * expand_size
-    elif len(padding) == 3:
+        # 2D padding
         pads = padding * 2
+    elif len(padding) == 3:
+        # 3D padding
+        pads = padding * 2
+    else:
+        # When padding is already done for all dimensions,
+        # we don't need to double it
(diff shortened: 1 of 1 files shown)


BowenBao:
There are some `maxpool3d` related tests that are marked as 'xfail' or 'skip'. Not sure if https://github.com/microsoft/onnxruntime/issues/15446 is related.

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -711,14 +711,16 @@ def _adjust_attributes_of_max_pool(
...
+    # expand_size is the dimension of pooling kernel,
+    # ONNX needs begin and end padding so we need to double the padding
     if isinstance(padding, int):
         pads = [padding] * expand_size * 2
     elif len(padding) == 1:
         pads = padding * expand_size * 2
     elif len(padding) == 2:
         pads = padding * expand_size
-    else:
-        pads = padding
```
titaiwangms, line 721: Not even sure if this is a scenario that we should keep.

titaiwangms:
> There are some `maxpool3d` related tests that are marked as 'xfail' or 'skip'. Not sure if [microsoft/onnxruntime#15446](https://github.com/microsoft/onnxruntime/issues/15446) is related.

Thanks, but I also disabled the very same test, and get the same issue..

titaiwangms:
@fatcat-z PTAL

fatcat-z:
> There are some `maxpool3d` related tests that are marked as 'xfail' or 'skip'. Not sure if [microsoft/onnxruntime#15446](https://github.com/microsoft/onnxruntime/issues/15446) is related.

As you commented in that issue, they are related to the behavior of 'ceil_mode' parameter, so has no relation to this PR.

onnxscript/function_libs/torch_lib/ops/nn.py
fatcat-z: This branch is used to handle the case that user has provided the correct value of padding, in this example it might be (1, 1, 1, 1, 1, 1), so we don't need to do the expand work.

titaiwangms: Done

fatcat-z (approved):
LGTM, thanks!

PR #758, opened 2023-06-01
Preserve correct params order in FunctionIR

### LINKED ISSUE #277: Attribute before inputs for matching signatures
See https://github.com/microsoft/onnx-script/pull/274

### LINKED ISSUE #305: Supply attributes by position and inputs by keyword

BowenBao:
* #759
* __->__ #758

Preserves original arguments order for param_schemas, including when OnnxFunction 
is defined with intersecting inputs and attrs. E.g. `def function(self, input1, attr1, input2, attr2)`.

`eval_function` for `TorchScriptTracingEvaluator` is left unchanged, i.e., splits arguments into inputs
and attributes. Because that is the format expected by TorchScript graph building.

Fixes #277 Fixes #305

### DIFF
--- a/onnxscript/_internal/param_manipulation.py
+++ b/onnxscript/_internal/param_manipulation.py
@@ -67,9 +67,65 @@ def separate_input_attributes_from_arguments(
             # User did not provide the attribute
             if fill_defaults:
                 onnx_attributes[param.name] = param.default
-            else:
-                continue
         elif param.required:
             raise TypeError(f"Required input '{param}' was not provided")
 
     return onnx_inputs, onnx_attributes
+
+
+def tag_arguments_with_param_schemas(
+    param_schemas: Sequence[values.ParamSchema],
+    args,
+    kwargs,
+    fill_defaults: bool = True,
+    allow_extra_kwargs: bool = False,
+) -> tuple[list[tuple[Any, values.ParamSchema]], dict[str, tuple[Any, values.ParamSchema]]]:
+    """Tag Python args and kwargs with matching ONNX ParamSchema.
+
+    Args:
+        param_schemas: The parameter schemas of an Op or a OnnxFunction.
+        args: The Python positional arguments supplied by the caller.
+        kwargs: The Python keyword arguments supplied by the caller.
+        fill_defaults: Whether to fill the default values for attributes.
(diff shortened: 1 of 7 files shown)


titaiwangms:
So we can mix inputs and attributes now with OnnxFunction?

titaiwangms (approved):
I think you would need to change the base before merge.

titaiwangms:
I wonder how many TracedOnnxFunctions are freed by this.

fatcat-z:
> @titaiwangms could you find the link? I think there are still some eager evaluation issues with this PR. I need to revisit.

I think #305 should be the issue which has several correlated issues.

onnxscript/_internal/param_manipulation.py
```diff
@@ -73,3 +73,63 @@ def separate_input_attributes_from_arguments(
...
+            continue
+        if i < len(args):
+            tagged_args.append((args[i], param))
+        elif param.name in kwargs:
+            tagged_kwargs[param.name] = (kwargs[param.name], param)
+        elif param.default is not values._EmptyDefault:  # pylint: disable=protected-access
+            # User did not provide the input/attribute
+            if fill_defaults:
+                tagged_kwargs[param.name] = (param.default, param)
+            else:
```
fatcat-z, line 130: Seems like this 'else' block is not necessary.

justinchuby (approved):

onnxscript/values.py
```diff
@@ -205,37 +205,44 @@ def _is_optional(typeinfo) -> bool:
...
-                    onnx.defs.OpSchema.AttrType(attr_parameter.type)  # type: ignore[call-arg]
-                ),
-                default=_EmptyDefault
-                if attr_parameter.default_value is None
-                else attr_parameter.default_value,
-                is_input=False,
-                required=not attr_parameter.has_default,
-            )
-        )
+    # OnnxFunction supports intersected inputs and attributes.
```
justinchuby, line 234:
```suggestion
    # OnnxFunction supports interleaving inputs and attributes.
```

nit: maybe

justinchuby (approved):
❤️

PR #762, opened 2023-06-01
Add op (hsatck & vstack) | feat(torchlib)

titaiwangms:
Missing operator: https://github.com/microsoft/onnx-converters-private/issues/168
Follow up issue track: https://github.com/microsoft/onnxruntime/issues/16296

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -666,8 +666,8 @@ def aten_atanh(self: TFloat) -> TFloat:
     return op.Atanh(self)
 
 
-@torch_op("aten::atleast_1d")
-def aten_atleast_1d(self: Sequence[TTensor]) -> TTensor:
+@torch_op("aten::atleast_1d", private=True)
+def _aten_atleast_1d_onnx(self: Sequence[TTensor]) -> TTensor:
     """atleast_1d(Tensor self) -> Tensor"""
 
     @graph()
@@ -681,6 +681,11 @@ def aten_atleast_1d(self: Sequence[TTensor]) -> TTensor:
     return op.SequenceMap(self, body=reshape_to_1d)
 
 
+@torch_op("aten::atleast_1d", trace_only=True)
+def aten_atleast_1d(self: Sequence[TTensor]) -> TTensor:
+    return _aten_atleast_1d_onnx(self)
+
+
 @torch_op("aten::atleast_1d")
 def aten_atleast_1d_single_tensor(self: TTensor) -> TTensor:
     """atleast_1d(Tensor self) -> Tensor"""
@@ -692,8 +697,8 @@ def aten_atleast_1d_single_tensor(self: TTensor) -> TTensor:
     return self
 
 
-@torch_op("aten::atleast_2d")
-def aten_atleast_2d(self: Sequence[TTensor]) -> TTensor:
+@torch_op("aten::atleast_2d", private=True)
+def _aten_atleast_2d_onnx(self: Sequence[TTensor]) -> TTensor:
     """atleast_2d(Tensor self) -> Tensor"""
 
     @graph()
@@ -705,6 +710,11 @@ def aten_atleast_2d(self: Sequence[TTensor]) -> TTensor:
         return tensor
 
     return op.SequenceMap(self, body=reshape_to_2d)
+
+
+@torch_op("aten::atleast_2d", trace_only=True)
+def aten_atleast_2d(self: Sequence[TTensor]) -> TTensor:
+    return _aten_atleast_2d_onnx(self)
 
 
 @torch_op("aten::atleast_2d")
@@ -2880,7 +2890,7 @@ def aten_hstack(tensors: Sequence[TTensor]) -> TTensor:
     """hstack(Tensor[] tensors) -> Tensor"""
 
     # Use another onnx function
-    tensors = aten_atleast_1d(tensors)
+    tensors = _aten_atleast_1d_onnx(tensors)
 
     # NOTE: The if/else graph has different shape/type which breaks the
     #       graph matching. We need to use trace_only.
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2811,10 +2811,25 @@ def aten_hspmm(mat1: TensorType, mat2: TensorType) -> TensorType:
...
+    #   TORCH_CHECK(!tensors.empty(),
+    #            "hstack expects a non-empty TensorList");
+    #   auto rep = at::atleast_1d(tensors);
+    #   if (rep[0].dim() == 1) {
+    #     return at::cat(rep, 0);
+    #   }
+    #   return at::cat(rep, 1);
+    # }
+
+    return op.ConcatFromSequence(tensors, axis=-1, new_axis=0)
```
BowenBao, line 2832: I think `axis=-1` works only for tensor of rank 1 and 2, but not others.

Is it possible to check rank of the first element in `tensors`?

titaiwangms:
The op works in test_fx_op_consistency.py, so I figure there might be a bug inside graph_executor on Sequence inputs

fatcat-z:
Currently, for the issue that if/else block may return different shape/type, the alternative is to create 2 functions with different inputs for the same aten op, just like aten_bitwise_not. Could you please change this function as well?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6532,10 +6551,20 @@ def aten_view_copy(self: TensorType, size: INT64) -> TensorType:
     raise NotImplementedError()
 
 
-def aten_vstack(tensors: Sequence[TensorType]) -> TensorType:
+@torch_op("aten::vstack")
+def aten_vstack(tensors: Sequence[TTensor]) -> TTensor:
     """vstack(Tensor[] tensors) -> Tensor"""
 
-    raise NotImplementedError()
+    # TODO: Support at_least2d
```
titaiwangms, line 6558: If I added the code from #767 to here for atleast_2d. It pops error saying "onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : This is an invalid model. Error: Duplicate definition of name (_inline_SequenceMaprank)." in eager mode, and "onnxruntime.capi.onnxruntime_pybind11_state.RuntimeException: [ONNXRuntimeError] : 6 : RUNTIME_EXCEPTION : Exception during initialization: Invalid tensor data type 0." in graph test.

cc @gramalingam @justinchuby any idea?

titaiwangms:
```python
@torch_op("aten::vstack")
def aten_vstack(tensors: Sequence[TTensor]) -> TTensor:
    """vstack(Tensor[] tensors) -> Tensor"""

    # TODO: This is exactly from aten_atleast_2d
    @graph()
    def reshape_to_2dd(tensor):
        shape = op.Shape(tensor)
        rank = op.Size(shape)
        if rank <= 1:
            tensor = op.Reshape(tensor, op.Constant(value_ints=[1, -1]))
        return tensor
    
    new_tensors = op.SequenceMap(tensors, body=reshape_to_2dd)

    return op.ConcatFromSequence(new_tensors, axis=0)
```

fatcat-z: In eager mode, each op will be executed independently as an ONNX graph. Could you start from checking which op caused the "invalid model" error?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2879,20 +2879,11 @@ def aten_hspmm(mat1: TensorType, mat2: TensorType) -> TensorType:
...
-    #   TORCH_CHECK(!tensors.empty(),
-    #            "hstack expects a non-empty TensorList");
-    #   auto rep = at::atleast_1d(tensors);
-    #   if (rep[0].dim() == 1) {
-    #     return at::cat(rep, 0);
-    #   }
-    #   return at::cat(rep, 1);
-    # }
+    # Use another onnx function
+    tensors = aten_atleast_1d(tensors)
```
justinchuby, line 2883: 👀

fatcat-z: Looks like it works.

titaiwangms: Yes, it works. The error is something else. For this particular case, if we need to move aten::atleast_1d to a private share node, it would be a whole function...

justinchuby: I would then lift the whole function. Just don't call one aten in another

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -681,6 +681,11 @@ def reshape_to_1d(tensor):
     return op.SequenceMap(self, body=reshape_to_1d)
 
 
+@torch_op("aten::atleast_1d", trace_only=True)
```
justinchuby, line 684: Possible to make it not trace only, such that the function name is retained in the graph?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -762,6 +766,10 @@ def _where_input_wrangler(
         reason="fixme: ORT fails with invalid model: 'INVALID_ARGUMENT : Failed to load model with error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)'",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
+        "vstack",
+        reason="fixme: A bug of constant-propagation optimization within the subgraph, we can avoid it by turning off graph-optimizations in session options",
```
justinchuby, line 771: Should we turn it off in test and file a bug to ORT?

titaiwangms: Would nested affect performance?

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Would nested affect performance?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
titaiwangms: xfail actually works. I think CI breaks because of something else. Actually @gramalingam spot the bug, but I can still file an issue to track.

justinchuby (approved):

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
titaiwangms: https://github.com/microsoft/onnxruntime/issues/16296

PR #759, opened 2023-06-01
Fix 'aten.gather' input order | fix(torchlib)

BowenBao:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #759
* #758

Fixes #756

PR #764, opened 2023-06-02
Properly check for optional function ir input, fix cross_entropy_loss and test

BowenBao:
* #759
* #758
* __->__ #764

Summary of changes:
- Correct annotation for `target` in `aten_cross_entropy_loss`, set as input instead of attribute.
- Properly check for optional function ir input to correctly set its `required` attribute.
- Correct `xfail` condition for `nn.functional.cross_entropy`. Previously `xfail` was triggered all the time.

Take aways:
- It is easy to set wrong type annotation for OnnxFunction, but hard to notice/debug. Needs careful validation since input vs attribute and the order shuffling internally all depends on type annotation. Perhaps more unittest coverage of explicit specification over opschema/paramschema would help.
- Positional access to `sample.args` is tricky as it excludes the first argument and stores it separately as `sample.input`. I knew this but forgot about this when debugging and it took me a while to realize.


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -754,6 +754,12 @@ def _where_input_wrangler(
         reason="fixme: ORT fails with invalid model: 'INVALID_ARGUMENT : Failed to load model with error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)'",
         test_class_name="TestOutputConsistencyFullGraph",
     ),
+    xfail(
+        "nn.functional.cross_entropy",
+        reason="ORT < 1.15 fails on a few subtests with error 'index < data_.size() was false'. Resolved in ORT 1.15+",
```
justinchuby, line 759: Should we bump ort version in test?

onnxscript/values.py
```diff
@@ -195,6 +195,16 @@ def param_schemas_from_op_schema(
     return tuple(schemas)
 
 
+def _is_optional(typeinfo) -> bool:
+    if typing.get_origin(typeinfo) is Union and type(None) in typing.get_args(typeinfo):
+        # Python < 3.10
+        return True
+    if typing.get_origin(typeinfo) is Optional:
+        # Python >= 3.10
```
justinchuby, line 203: I think there’s a similar function in onnx types or somewhere else. Let me double check

onnxscript/values.py
```diff
@@ -204,7 +214,7 @@ def param_schemas_from_function_ir(
 
     schemas = []
     for arg in function_ir.inputs:
-        if isinstance(arg.typeinfo, onnx.TypeProto.Optional):
+        if _is_optional(arg.typeinfo):
```
justinchuby, line 217: Is this a different type? The previous was checking for typeproto

justinchuby:
Could you make Take away (1) an issue so we can explore further? Thanks!

onnxscript/values.py
BowenBao: Very likely. Perhaps the previous one was typeinfo parsed from OpSchema?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
BowenBao: To do that in a separate PR and clean up all similar decorators w.r.t ORT version?

BowenBao:
> Could you make Take away (1) an issue so we can explore further? Thanks!

#774

onnxscript/values.py
justinchuby: Maybe update https://github.com/microsoft/onnxscript/blob/8521b27c9a549614c7d5250e508ece23c27e3297/onnxscript/type_annotation.py#L141?

onnxscript/values.py
justinchuby: I see. So the previous check did nothing?

justinchuby (approved):
Minor comment about `is_optional`. LGTM!

onnxscript/values.py
BowenBao: thanks for pointer

onnxscript/values.py
BowenBao: This could need more fundamental fix around typeinfo. At least in this case it does not match `IRTypeLike` protocol. I'll open an issue.

BowenBao: #775

onnxscript/type_annotation.py
```diff
@@ -179,7 +185,7 @@ def pytype_to_type_strings(pytype: TypeAnnotationValue) -> list[str]:
     if isinstance(pytype, typing.TypeVar):
         constraints = pytype.__constraints__
         if constraints:
-            return pytype_to_type_strings(Union.__getitem__(constraints))
+            return pytype_to_type_strings(Union[constraints])
```
BowenBao, line 188: Suggested by linter, looks .. strange

justinchuby: Not sure if this will fail in some python versions?

justinchuby: Curious which linter suggested this?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: #770

onnxscript/type_annotation.py
BowenBao: PYLINT, I think I'll revert and skip this

BowenBao:
```
  Advice (PYLINT) C2801
    Unnecessarily calls dunder method __getitem__. Access item via subscript.
    (unnecessary-dunder-call)
    See [unnecessary-dunder-call](https://pylint.pycqa.org/en/latest/
    user_guide/messages/convention/unnecessary-dunder-call.html).

    To disable, use `  # pylint: disable=unnecessary-dunder-call`

        185  |    if isinstance(pytype, typing.TypeVar):
        186  |        constraints = pytype.__constraints__
        187  |        if constraints:
    >>> 188  |            return pytype_to_type_strings(Union.__getitem__(constraints))
        189  |        bound = pytype.__bound__
        190  |        if bound is None:
        191  |            return list(ALL_TENSOR_TYPE_STRINGS)
```

BowenBao: Hmm. The CI test failure with ort-nightly is unrelated right? Saw that on main too. I thought it was due to this change.

justinchuby: Yes it’s being fixed in the version bump PR

BowenBao:
@justinchuby  should I wait for ort-nightly fix to merge this stack?

justinchuby:
Merged

PR #765, opened 2023-06-04
onnx_script_test_case to support multuple outputs models

liqunfu:


onnxscript/tests/models/pnp.py
```diff
@@ -0,0 +1,262 @@
...
+    UINT64,
+)
+from onnxscript.values import Op, Opset
+from typing import Callable, Optional, Sequence, Tuple, Union
+
+import numpy as np
+import onnxruntime as rt
+import onnx
+
+class Opset18Ext(Opset18):
```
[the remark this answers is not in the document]
liqunfu: I need a custom operator so that it can be programmatically edited to replace the op with the core model graph. I should not has used onnx domain though.

onnxscript/tests/models/pnp.py
```diff
@@ -0,0 +1,262 @@
...
+@script()
+def sliding_window_inference(inputs: FLOAT["N", 1, "D", "H", "W"], roi_size: INT64[3]) -> FLOAT["N", "Seg_C", "D", "H", "W"]:
+    """
+    for simplicity, we assume that the step size is the same as the roi size. D/H/W are multiple of roi_size in 3 dimensions,
+    no overlay, no padding. weight is 1.
+    TODOs: sw_batch_size > 1, roi_size is not multiple of D/H/W, overlay, padding, weight is not 1.
+    """
+
+    inputs_shape = op.Shape(inputs)
+    inputs_spatial_shape = op.Slice(inputs_shape, op.Constant(value_ints=[2]), op.Constant(value_ints=[5]), op.Constant(value_ints=[0]))
```
gramalingam, line 239: nit: may be this will be more readable if we move it after the line below, and rewrite it to:
```py
   inputs_spatial_shape = op.Concat(D, H, W, axis=0)
```
Alternatively, we can also say:
```py
   inputs_spatial_shape = op.Shape(inputs, from=2)
```

onnxscript/tests/models/pnp.py
```diff
@@ -0,0 +1,262 @@
...
+@script()
+def extend_roi_indices(roi_indices: INT64["roi_D", "roi_H", "roi_W", 3], seg_C: INT64) -> INT64[1, "seg_C", "roi_D", "roi_H", "roi_W", 5]:
+    """
+    This function is used to extend the roi indices to the segment dimension.
+    """
+    zero = op.Constant(value_int=0)
+    one = op.Constant(value_int=1)
+    one_ = op.Constant(value_ints=[1])
+    roi_D, roi_H, roi_W, _ = op.Split(op.Shape(roi_indices), num_outputs=4)
+    shape_1_seg_C_roi_D_H_W_1 = op.Concat(one_, seg_C, roi_D, roi_H, roi_W, one_, axis=0)
```
gramalingam, line 82: nit: you can directly use `op.Shape(roi_indices)` instead of `roi_D, roi_H, roi_W` ... no need to split it and concat it back.

onnxscript/tests/models/pnp.py
```diff
@@ -0,0 +1,262 @@
...
+    zeros_1_seg_C_D_H_W_1 = op.Cast(op.ConstantOfShape(shape_1_seg_C_roi_D_H_W_1), to=INT64.dtype) # seg_C, roi_D, roi_H, roi_W, 1
```
gramalingam, line 83: nit: you can directly specify the value/type as attribute of ConstantOfShape, instead of casting by adding `value=onnx.helper.make_tensor(...)`

onnxscript/tests/models/pnp.py
```diff
@@ -0,0 +1,262 @@
...
+    grid_d = op.Transpose(grid_d_1, perm=[2, 0, 1])
+
+    zeros_D_W_H = op.Transpose(zeros_D_H_W, perm=[0, 2, 1])
+    grid_h_1 = zeros_D_W_H + grid_h_0
+    grid_h = op.Transpose(grid_h_1, perm=[0, 2, 1])
+
+    grid_w = grid_w_0 + zeros_D_H_W
+
+    indices_seq = op.SequenceConstruct(grid_d, grid_h, grid_w)
+    indices = op.ConcatFromSequence(indices_seq, axis=-1, new_axis=1)  # [D, H, W, 3]
```
gramalingam, line 125: I guess you need this because Concat doesn't support the new_axis attributes? I wonder if we should extend Concat to support it.

We could explicitly unsqueeze grid_d etc. to add the extra dimension at end and then use Concat. It is a bit more cumbersome, but onnxruntime doesn't handle sequence op that well (it may create extra copies of the tensor). Anyway, may not be worth bothering about for now.

PR #766, opened 2023-06-05
Migrate diagnostics to onnxscript | feat(diagnostics)

fatcat-z:
Add diagnostics/infra files in PyTorch into onnxscript project, so it could also benefit this project. This is only the infrastructure part, will add more ONNXSCRIPT dedicated classes to make it work.

To remove the dependency of Torch, use beartype which is required by ONNXScript to replace _beartype.


justinchuby:
I recall we didn’t want to take beartype as a hard dependency into onnxscript. Do we need to reconsider the preference? @gramalingam @BowenBao

fatcat-z:
> I recall we didn’t want to take beartype as a hard dependency into onnxscript. Do we need to reconsider the preference? @gramalingam @BowenBao

There is an implementation in PyTorch exporter to define an internal beartype in case it is not installed. To avoid adding torch as a dependency to onnxscript, I replace it with beartype which is imported in onnxscript already.

justinchuby:
How many of these files are generated? Should we include the thing that will generate them somewhere?

fatcat-z:
> How many of these files are generated? Should we include the thing that will generate them somewhere?

So far, 53 files related to Sarif were generated. And yes, we should migrate the tool as well.

justinchuby:
Beartype is only used in torchlib not onnxscript for now

onnxscript/_internal/_beartype.py
```diff
@@ -0,0 +1,125 @@
...
+    _TORCH_ONNX_EXPERIMENTAL_RUNTIME_TYPE_CHECK = os.getenv(
+        "TORCH_ONNX_EXPERIMENTAL_RUNTIME_TYPE_CHECK"
+    )
+    if _TORCH_ONNX_EXPERIMENTAL_RUNTIME_TYPE_CHECK == "WARNINGS":
+        _runtime_type_check_state = RuntimeTypeCheckState.WARNINGS
+    elif _TORCH_ONNX_EXPERIMENTAL_RUNTIME_TYPE_CHECK == "DISABLED":
+        _runtime_type_check_state = RuntimeTypeCheckState.DISABLED
+    else:
+        _runtime_type_check_state = RuntimeTypeCheckState.ERRORS
+    beartype = _create_beartype_decorator(_runtime_type_check_state)
```
justinchuby, line 123: We can make this super simple by removing the warning option. Happy to chat more

justinchuby (approved):

xiaowuhu (approved):

PR #768, opened 2023-06-06
opgen: translate type constraints in an OpSchema to typing.TypeVars in the static API

abock:
Given a hypothetical `OpSchema` _Thing_ that has two identically type-constrained inputs and a type constrained output:

```
inputs:
  T1: tensor(int32), tensor(int64)
  T2: tensor(int32), tensor(int64)
outputs:
  T1
```

We were previously ignoring the type constraint itself, which is problematic especially when the return type should be the same type as one of the constrained arguments, since this is unsolvable.

### Before

```python
def Thing(a: Union[INT32, INT64], b: Union[INT32, INT64) -> Union[INT32, INT64]:
   ...

# error: we don't know that Thing is constrained to return INT64 tensor:
Thing(some_INT64_tensor, some_INT32_tensor)
```

This PR addresses the issue by translating type constraints on the schema into proper `typing.TypeVar` usage:

### After

```python
T1 = TypeVar("T1", INT32, INT64)
T2 = TypeVar("T2", INT32, INT64)

def Thing(a: T1, b: T2) -> T1:
   ...

# because 'a: T1' is given an INT64, we know we will return 'INT64'
Thing(some_INT64_tensor, some_INT32_tensor)
```

### DIFF
--- a/onnxscript/onnx_opset/_impl/opset1.py
+++ b/onnxscript/onnx_opset/_impl/opset1.py
@@ -10,7 +10,7 @@
 # pylint: disable=W0221,W0222,W0237,W0246,R0901,W0611
 # --------------------------------------------------------------------------
 
-from typing import Callable, Optional, Sequence, Tuple, Union
+from typing import Optional, Sequence, Tuple, TypeVar
 
 from onnx import GraphProto, TensorProto
 from onnx.defs import get_schema
@@ -42,9 +42,9 @@ class Opset1(Opset):
     def __init__(self):
         super().__init__()
 
-    def Abs(
-        self, X: Union[DOUBLE, FLOAT, FLOAT16], consumed_inputs: Optional[Sequence[int]] = None
-    ) -> Union[DOUBLE, FLOAT, FLOAT16]:
+    T = TypeVar("T", DOUBLE, FLOAT, FLOAT16)
+
+    def Abs(self, X: T, consumed_inputs: Optional[Sequence[int]] = None) -> T:
         r"""[🌐 Abs(1)](https://onnx.ai/onnx/operators/onnx__Abs.html#abs-1 "Online Documentation")
 
 
@@ -60,17 +60,19 @@ class Opset1(Opset):
         """
 
         schema = get_schema("Abs", 1, "")
-        op: Callable[..., Union[DOUBLE, FLOAT, FLOAT16]] = Op(self, "Abs", schema)
+        op = Op(self, "Abs", schema)
         return op(*self._prepare_inputs(schema, X), consumed_inputs=consumed_inputs)
+
+    T = TypeVar("T", DOUBLE, FLOAT, FLOAT16)
 
     def Add(
         self,
-        A: Union[DOUBLE, FLOAT, FLOAT16],
-        B: Union[DOUBLE, FLOAT, FLOAT16],
+        A: T,
+        B: T,
         axis: Optional[int] = None,
         broadcast: int = 0,
         consumed_inputs: Optional[Sequence[int]] = None,
-    ) -> Union[DOUBLE, FLOAT, FLOAT16]:
+    ) -> T:
         r"""[🌐 Add(1)](https://onnx.ai/onnx/operators/onnx__Add.html#add-1 "Online Documentation")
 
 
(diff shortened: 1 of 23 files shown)


abock:
Note: I need to constrain the return as well, but want CI to run on just input constraints for now to see test coverage.

thiagocrepaldi (commented):
Looking good. That will make type promotion a breeze

onnxscript/onnx_opset/_impl/opset18.py
```diff
@@ -43,10 +43,10 @@ def __new__(cls):
     def __init__(self):
         super().__init__()
 
+    T = TypeVar("T", INT16, INT32, INT64, INT8, UINT16, UINT32, UINT64, UINT8)
+
     def BitwiseAnd(
-        self,
-        A: Union[INT16, INT32, INT64, INT8, UINT16, UINT32, UINT64, UINT8],
-        B: Union[INT16, INT32, INT64, INT8, UINT16, UINT32, UINT64, UINT8],
+        self, A: T, B: T
```
gramalingam, line 49: The return type should also be `T`, I think.

onnxscript/onnx_opset/_impl/opset1.py
```diff
@@ -118,7 +122,9 @@ def Add(
             consumed_inputs=consumed_inputs,
         )
 
-    def And(self, A: BOOL, B: BOOL, axis: Optional[int] = None, broadcast: int = 0) -> BOOL:
+    T = TypeVar("T", bound=BOOL)
+
+    def And(self, A: T, B: T, axis: Optional[int] = None, broadcast: int = 0) -> BOOL:
```
justinchuby, line 127: When there’s a single type, T and BOOL are equivalent and BOOL maybe simpler?

gramalingam:
This is great! I remember that long back you mentioned that this hinders intellisense ... just wondering if intellisense is able to handle typevars now, or do we still have that limitation? Thanks!

onnxscript/onnx_opset/_impl/opset18.py
gramalingam: Ah, just noticed your comment about this

abock:
@gramalingam unfortunately the IntelliSense is still not great, but given that we actually need to make decisions based on constraints, we need the correctness now.

I also played around with `typing.Annotated` to wrap the types, but that's also a dead-end regarding IntelliSense (the annotation expression itself is not surfaced in UI).

At this point, pylance just needs improving and I'm fine with the approach in this PR since it is the most idiomatic Pythonic way to express the constraints. If pylance gets proper support for TypeVar types in the future, there won't be anything further we need to do.

Aside: I almost think surfacing the constraint names is an improvement simply because seeing potentially dozens of types in the unions in the IntelliSense UI is quite hard to visually parse. Users can go-to-definition on functions to see the constraints for now. I think that's a happy medium until tooling catches up.

gramalingam:
> unfortunately the IntelliSense is still not great

Ok. I guess another possibility is to converge on some small set of well-known type-variables (commonly used sets of types), which can hopefully cover a majority of cases while having well-recognized names.

onnxscript/onnx_opset/_impl/opset1.py
abock: I was on the fence about this case since allowing constraints with a single type to fall through means the constraint itself can't be resolved at eval time. It may not matter at all, but I'm inclined to keep single types as TypeVar for consistency.

abock:
@gramalingam yeah, we can definitely think about creating some common constraints that are defined in an outer scope. What's nice is that doing that normalization would require this PR as a foundation anyway.

As it stands with this PR, the API is much clearer even though the constraint-bound types are no longer directly surfaced in IntelliSense:

image

justinchuby (approved):

abock:
I'm good to merge this as-is. There is followup work to support `TypeVar` in the `converter` and `type_annotation` modules. One thing we will need to think about is that because the `TypeVar` `ast.Assign` statements happen before the script function `ast.FunctionDef` is defined, we need to parse the entire `ast.Module`.

gramalingam (approved):

titaiwangms (approved):

PR #770, opened 2023-06-08
Bump ORT to 1.15 | chore(deps)

justinchuby:
This change updates ORT and ort-nightly to the latest versions and cleans up the skipped tests. It also generates opset19 so that the backend tests can run.


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -712,11 +663,6 @@ def _where_input_wrangler(
         "nn.functional.scaled_dot_product_attention_bool_mask",
         reason="fixme: ORT crashes on Windows, segfaults randomly on Linux",
     ),
-    xfail(
-        "nn.functional.upsample_bilinear2d",
-        reason="fixme: ORT fails with invalid model: 'INVALID_ARGUMENT : Failed to load model with error: vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)'",
-        test_class_name="TestOutputConsistencyFullGraph",
-    ),
     xfail(
         "nn.functional.upsample_nearest2d",
```
justinchuby, line 667: upsample_bilinear2d is fixed but upsample_nearest2d is not. Weird

BowenBao (commented):
LG overall, why do you need opset19?

justinchuby:
> LG overall, why do you need opset19?
> 
> Bumping ORT introduced opset19 tests into CI?

Yes it looks like they were discovered. I didn’t look too closely at what happened

BowenBao (approved):

BowenBao:
Strange, but won't block on this. Could you follow up?

BowenBao:
Actually wait... do we want to release with Opset19? cc @abock

justinchuby:
Sure https://github.com/microsoft/onnxscript/issues/776

abock:
We should probably not include Opset 19 in this PR? I think while you were away @justinchuby I added support for skipping opsets in `opgen`:

```
python -m opgen -x 19 -x 20
```

abock:
In fact, yes, let's split Opset 19 effort out of this PR since I really need ORT 1.15 for #773.

abock (changes requested):
Re-run opgen to drop Opset 19 and open a separate PR for that.

```
python -m opgen -x 19 -x 20
```

justinchuby:
Without opset19 the backend tests fail. I can dig a little deeper

abock (approved):
After speaking with Justin, I had misunderstood why Opset 19 was coming in. We should take it as part of the bump.

PR #773, opened 2023-06-08
opgen: make all OpSchema attributes keyword-only args in Python

abock:
> ⚠️🚨🔥 ***Note:*** this is an API break since all attributes are now keyword-only, so positional usage is no longer allowed. Our tests uncovered a handful of places where we were specifing attributes as positional arguments.

By explicitly projecting all OpSchema attributes to keyword-only arguments in Python, we do not need to reorder attributes based on whether or not they have default values. This allows the Python API to express the same ordering as the op itself for inputs and attributes.

This also makes attribute specification consistent with variadic input ops which already have that implicit requirement.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -4433,8 +4433,8 @@ def _aten_native_batch_norm_inference_onnx(
         training_mode=training,
     )
     # Cannot return 2 dup output, so have to do twice with different variable name
-    empty_mean = op.Cast(op.Shape(input, 0, 0), to=FLOAT.dtype)
-    empty_var = op.Cast(op.Shape(input, 0, 0), to=FLOAT.dtype)
+    empty_mean = op.Cast(op.Shape(input, start=0, end=0), to=FLOAT.dtype)
+    empty_var = op.Cast(op.Shape(input, start=0, end=0), to=FLOAT.dtype)
     return norm, empty_mean, empty_var
 
 
@@ -4663,7 +4663,7 @@ def aten_new_empty_dtype(
 
     # using zero to simulate empty array
     result = op.ConstantOfShape(size)
-    return op.Cast(result, dtype)
+    return op.Cast(result, to=dtype)
 
 
 @torch_op("aten::new_empty_strided")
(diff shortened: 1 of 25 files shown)


justinchuby:
fyi I am bumping ort and nightly in https://github.com/microsoft/onnxscript/pull/770

abock:
Yeah, just trying to see if the problem I am seeing on the ort-nightly CI was addressed already. Seems not.

abock:
Actually I think I'll rebase this for now on your branch so we're on the same page.

abock:
Nice. Thanks @justinchuby, your branch makes this PR go green.

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4433,8 +4433,8 @@ def _aten_native_batch_norm_inference_onnx(
         training_mode=training,
     )
     # Cannot return 2 dup output, so have to do twice with different variable name
-    empty_mean = op.Cast(op.Shape(input, 0, 0), to=FLOAT.dtype)
-    empty_var = op.Cast(op.Shape(input, 0, 0), to=FLOAT.dtype)
+    empty_mean = op.Cast(op.Shape(input, start=0, end=0), to=FLOAT.dtype)
```
gramalingam, line 4436: This line of code looks odd, wonder what it is trying to do! (I realize this has nothing to do with current PR.)

opgen/onnx_opset_builder.py
```diff
@@ -494,14 +509,12 @@ def _make_function_input_args(
         return args
 
     def _make_function_attr_args(self, schema: OpSchema) -> Iterable[cg.Arg]:
-        attr_args = []
+        generate_kwonly_sentinal = True
```
gramalingam, line 512: nit: "sentinel" spelling

gramalingam (approved):

opgen/onnx_opset_builder.py
abock: Fixed, thanks

onnxscript/function_libs/torch_lib/ops/core.py
abock: Filed: https://github.com/microsoft/onnxscript/issues/786

PR #778, opened 2023-06-09
opgen: unique names for TypeVars; TypeAlias for some outputs

abock:
___Depends on #773.___

This PR addresses two somewhat related issues introduced by #768. These issues were not caught locally or in CI because mypy has been disabled for generated code in `.lintrunner.toml`.

### Issue 1: Re-enable mypy for generated code

I believe that there was a pathological issue in an older version of mypy that caused us to disable running it for generated code because it would cause CI to timeout. Simply enable it again and we'll keep an eye on it.

### Issue 2: type constraint names must be unique

We cannot generate multiple `TypeVar`s with the same name because `from __future__ import annotations` defers the evaluation of type annotation expressions. For example, this is not valid for statically reasoning about code:

```python
T = TypeVar("T", int, float)
def func1(thing: T) -> T: ...

T = TypeVar("T", str, bool)
def func2(thing: T) -> T: ...
```

Under deferred type checking, the last assignment to `T` will propagate to both `func1` and `func2` (IOW, `T` is `(str, bool)` across all usages).

This PR suffixes constraint names with the name of the function in which they're used:

```python
T_func1 = TypeVar("T", int, float)
def func1(thing: T) -> T: ...

T_func2 = TypeVar("T", str, bool)
def func2(thing: T) -> T: ...
```

### Issue 3: output constraints must be used as input constraints

The initial support for projecting Python `TypeVar`s from ONNX type constraints is too aggressive and creates invalid Python for some ops when an ONNX output parameter has a type constraint that is not used for any of its input parameters.

This scenario will be fully supported in Python 3.12 with the new subscript syntax for typevars bound to the function itself and not just the function's arguments, e.g.:

```python
def func[T]() -> T: ...
```

The solution for now to retain the name of output constraints that are not used for inputs is to generate a `TypeAlias` instead. This at least keeps the Python signature consistent with its `OpSchema` with respect to constraints, and these aliases can become proper constraints when Python adoption has caught up.


justinchuby (approved):

justinchuby (approved):

justinchuby:
Great, codecov is back

gramalingam (approved):

abock:
Rebasing after resolving conflicts from misspelling in #773.

PR #779, opened 2023-06-12
Remove trace_only for scatter_add | chore(torchlib)

justinchuby:
Now that we support mixed input/attribute ordering from #758


BowenBao (approved):
LGTM

PR #789, opened 2023-06-16
Add documentation about tensor constants

gramalingam:
Clarify in documentation about usage of constant literals in onnxscript as short-hand for tensor-constants.

Also update eager-mode implementation, which doesn't currently support list constants.


justinchuby (approved):

docs/tutorial/index.md
```diff
@@ -79,9 +80,20 @@ the one above.
...
 
-This works for scalar constants. However, if the user wants to
-create an 1-dimensional tensor containing a single value, instead of a 0-dimensional
-tensor, they need to do so more explicitly (as in the previous example).
+The direct usage of literals can be used to create scalars or one-dimensional tensors
+of type `FLOAT` or `INT64` or `STRING`, as shown in the table below.
+
+| Python source  | Generated ONNX constant               |
+| -------------- | ------------------------------------- |
+| 0              | Scalar value 0 of type INT64          |
```
justinchuby, line 88:
```suggestion
| 0              | Scalar value 0 of type `INT64`          |
```

nit: this and below for readability?

onnxscript/autocast.py
```diff
@@ -8,23 +8,42 @@
...
-        return tensor.Tensor(np.array(x, dtype=dtype))
-    return x
+    if isinstance(pyvalue, bool):
+        return np.bool_
+    elif isinstance(pyvalue, int):
+        return np.int64
+    elif isinstance(pyvalue, float):
+        return np.float32
+    elif isinstance(pyvalue, list):
+        if list:
```
justinchuby, line 23:
```suggestion
        if pyvalue:
```

gramalingam: :-)

onnxscript/autocast.py
```diff
@@ -1,30 +1,113 @@
...
+    if onnx_type is onnx.TensorProto.STRING:
+        return helper.make_tensor(tensor_name, onnx_type, [], vals=[pyvalue.encode("utf-8")])
+
+    return helper.make_tensor(tensor_name, onnx_type, [], [pyvalue])
+
+
+# Utilities to convert python values into onnxscript tensors.
+
+
+def promotable(x):
```
justinchuby, line 67:
```suggestion
def _promotable(x) -> bool:
```

nit. Also should it be private for now?

gramalingam:
@justinchuby : do you know why torch_lib/ops_test.py is failing large number of tests? Is that expected?

justinchuby:
A large number of tests failing is not expected. But it seems like there's only one flaky test failing and a pypi pytorch installation hiccup right now so we should be good.

onnxscript/autocast.py
```diff
@@ -1,30 +1,113 @@
...
+        return onnx.TensorProto.FLOAT
+    if pytype is str:
+        return onnx.TensorProto.STRING
+    raise ValueError(f"Tensor element of type {pytype} not supported")
+
+
+def pyvalue_to_onnx_tensor(tensor_name: str, pyvalue):
+    if isinstance(pyvalue, np.ndarray):
+        return numpy_helper.from_array(pyvalue, tensor_name)
+    if isinstance(pyvalue, list):
```
abock, line 41: Check for `isinstance(pyvalue, Sequence)`?

[typing.Sequence](https://docs.python.org/3/library/typing.html#typing.Sequence)

abock (approved):

onnxscript/autocast.py
justinchuby: I think tuples are treated differently as attributes so

gramalingam: May be ok to start by being more restrictive for now, IMO. It's a bit tricky overall since we want to handle things consistently in the static (converter) mode and the eager mode, and in the eager mode we (currently) lack some of the static information. Still, I think tuples should be doable, but perhaps we can do that in a separate PR.

PR #799, opened 2023-06-20
Improve xfails by allowing specifying expected errors | test(torchlib)

### LINKED ISSUE #794: [torchlib] Improve `xfail` decorator to be more robust and accurate
`xfail` now is powered by `pytest.xfail(reason=reason)`. While it is great and better than `skip`, it does not offer check on the actual failure reason. Hence through time the descriptive `reason` may become outdated and misleading.

An option is to rewrite decorator with `self.assertRaiseRegex`.

cc @justinchuby, @titaiwangms

justinchuby:
- Allow specifying expected errors in xfails
- Raise `RuntimeError` instead of `AssertionError` when ORT raises an error in test for more accurate xfails.
- Specify the expected errors in some xfails (as a first pass)

Fixes #794


BowenBao:
Thanks @justinchuby this is great! Can we guard on the error message too? Exception type feels a bit broad, for example I may want to flag change in behavior when ORT starts to complain about mismatching type instead of type not implemented.

justinchuby:
The error message is harder because there isn't a pytest decorator for it. I could implement one but that will take longer

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -581,11 +596,15 @@ def normal_xfail_skip_test_behaviors(
...
     # to go over all test cases to find the right exception type.
-    except Exception as e:  # pylint: disable=broad-exception-caught
+    except Exception:  # pylint: disable=broad-exception-caught
         if test_behavior is None:
-            raise e
+            raise
         if test_behavior == "xfail":
-            pytest.xfail(reason=reason)
+            if raises is not None:
+                with pytest.raises(raises):
```
BowenBao, line 604: Passing in the unittest instance, can we do

```python
        if test_behavior == "xfail":
            if raises is not None:
                with test_case.assertRaisesRegex(raises, expected_error_message_regex):
                    raise
            pytest.xfail(reason=reason)
```

BowenBao: nvm, I think we need a context manager similar to this but wrapped over subtest calling.

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -100,17 +103,26 @@ def xfail(
...
+        raises: The expected error type.
     """
+    if raises is None:
+        decorator = pytest.mark.xfail(reason=reason)
+    else:
+        decorator = pytest.mark.xfail(reason=reason, raises=raises)
     return DecorateMeta(
         op_name=op_name,
         variant_name=variant_name,
-        decorator=unittest.expectedFailure,
```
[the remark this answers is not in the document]
justinchuby: There is the decorator that will decorate the test method, and there are context managers that handle the subtests. I haven't found a good way to unify the two. I may be possible for us to create our own decorators but I am not sure how it is supported by pytest?

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: It should work for any subtests, but I am not sure how to support the test methods?

titaiwangms:
> The error message is harder because there isn't a pytest decorator for it. I could implement one but that will take longer. Do you have any suggestions? Thanks!

A naive way would be we do string match on reason and exception message in `normal_xfail_skip_test_behaviors`, and it would raise xfail if only they match at some points.

justinchuby:
> > The error message is harder because there isn't a pytest decorator for it. I could implement one but that will take longer. Do you have any suggestions? Thanks!
> 
> A naive way would be we do string match on reason and exception message in `normal_xfail_skip_test_behaviors`, and it would raise xfail if only they match at some points.
> 
> I guess it doesn't have to be strict. Just point out the error occurred location and type.

That's for the subtest I think? I am thinking if there's a way to use the decorators

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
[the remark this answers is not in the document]
titaiwangms: It's not supported..

titaiwangms:
> > > The error message is harder because there isn't a pytest decorator for it. I could implement one but that will take longer. Do you have any suggestions? Thanks!
> > 
> > 
> > A naive way would be we do string match on reason and exception message in `normal_xfail_skip_test_behaviors`, and it would raise xfail if only they match at some points.
> > I guess it doesn't have to be strict. Just point out the error occurred location and type.
> 
> That's for the subtest I think? I am thinking if there's a way to use the decorators

Do we have more cases needing xfail on whole test? The other dtypes?

justinchuby:
Yes. Sometimes ORT fails for the op so we xfail the whole test

PR #802, opened 2023-06-22
Add support for serialization without initializers

thiagocrepaldi:
Needed by https://github.com/pytorch/pytorch/pull/103865/

When a PyTorch model is traced using `FakeTensor` weights[1], the resulting IR will not actually contain Tensors that can be serialized into protobuf.

In that scenario, the model must be serialized without any initializer to prevent invalid data access on the underlying PyTorch tensor. This PR proposes an optional parameter `include_initializers: bool = True` to `to_model_proto` API for allowing the serializer to ignore any initializer during serialization.

[1] `FakeTensor` is a subclass of `Tensor`, but without the actual data. See below for an example:

```python
{'_param_constant0': Parameter(FakeTensor(..., size=(2, 2), requires_grad=True)),
 '_param_constant1': Parameter(FakeTensor(..., size=(2,), requires_grad=True))}
```

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -649,7 +649,9 @@ class TorchScriptGraph:
         return onnx_function
 
     @beartype
-    def to_model_proto(self, opset_version: int) -> onnx.ModelProto:
+    def to_model_proto(
+        self, opset_version: int, include_initializers: bool = True
+    ) -> onnx.ModelProto:
         function_proto_dict: Mapping[
             Tuple[str, str], onnx.FunctionProto
         ] = self.fetch_function_proto_dict(opset_version)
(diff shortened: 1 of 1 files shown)


thiagocrepaldi:
> lgtm thanks!

Thanks! Does this test look flaky? only for windows. The API change should be BC

```python
FAILED onnxscript/tests/function_libs/torch_lib/ops_test.py::TestOutputConsistencyFullGraphCPU::test_output_match_opinfo__maximum_cpu_float16
```

justinchuby:
It’s a flaky test due to ORT. It should be safe to merge

BowenBao:
Ah already merged. Should we add a test case for this?

BowenBao:
@thiagocrepaldi for above

thiagocrepaldi:
> @thiagocrepaldi for above

sure, can add one

justinchuby:
https://github.com/microsoft/onnxscript/blob/main/onnxscript/function_libs/torch_lib/graph_building_test.py should be good

PR #792, opened 2023-06-19
Add ops (addr) | feat(torchlib)

justinchuby:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #792

-----

- Mark nan==nan in assert_close
- Fix aten_isfinite by allowing the exporter to do type promotion according to #563 and introduce `TFloatHighPrecision` for it
- Fix typo in `TRealOrUInt8`

PR #791, opened 2023-06-19
Implement add* ops | feat(torchlib) (#785)

justinchuby:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #792
* __->__ #791

- addbmm
- addcdiv
- addcmul
- addmv

Signed-off-by: Justin Chu <justinchu@microsoft.com>

PR #803, opened 2023-06-23
Leverage TorchLibOpInfo in more tests and clean up variables | test(torchlib)

justinchuby:
* #792
* #804
* __->__ #803
* #791


onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -102,52 +102,49 @@ def _get_rtol_atol_by_dtype(dtype: torch.dtype) -> tuple[Any, Any]:
...
-            if isinstance(func, onnxscript.OnnxFunction):
-                raise AssertionError(
-                    f"'{func.name}' is an OnnxFunction. "
-                    "If the function is not trace_only, please remove trace_only=True "
-                    "in the TorchLibOpInfo entry."
-                )
+        for info in ops_test_data.TESTED_TORCHLIB_OPS:
+            if not info.trace_only:
+                continue
+            with self.subTest(name=info.op_info_name):
```
fatcat-z, line 121: What's the purpose of using subTest here?

onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -102,52 +102,49 @@ def _get_rtol_atol_by_dtype(dtype: torch.dtype) -> tuple[Any, Any]:
...
+                func = info.op
+                if not isinstance(func, onnxscript.TracedOnnxFunction):
```
titaiwangms, line 123: If we don't have to explicit trace_only attribute, we don't need this second guard to make sure it's TracedOnnxFunction.

titaiwangms (approved):

onnxscript/tests/function_libs/torch_lib/ops_test.py
justinchuby: I wanted to make it harder to use trace_only (because we eventually want to script all functions, as far as we know now) so that authors know they are making a conscious choice.

onnxscript/tests/function_libs/torch_lib/ops_test.py
justinchuby: Using subTest to show a separate test error for each violation. I didn't use parameterized because I didn't want the tests to show separate lines when they succeed.

PR #810, opened 2023-06-27
Add ops (tensor.int, tensor.float, tensor.bool) | feat(torchlib)

BowenBao:
`aten.tensor` is emitted from TypePromotionPass in dynamo onnx exporter.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6238,8 +6238,8 @@ def aten_tanh(self: TFloat) -> TFloat:
     return op.Tanh(self)
 
 
-@torch_op("aten::tensor.int")
-def aten_tensor_int(self: int, dtype: int) -> TensorType:
+@torch_op("aten::tensor.bool")
+def aten_tensor_bool(self: bool, dtype: int) -> TensorType:
     tensor = op.Constant(value_int=self)
     return op.Cast(tensor, to=dtype)
 
@@ -6250,8 +6250,8 @@ def aten_tensor_float(self: float, dtype: int) -> TensorType:
     return op.Cast(tensor, to=dtype)
 
 
-@torch_op("aten::tensor.bool")
-def aten_tensor_bool(self: bool, dtype: int) -> TensorType:
+@torch_op("aten::tensor.int")
+def aten_tensor_int(self: int, dtype: int) -> TensorType:
     tensor = op.Constant(value_int=self)
     return op.Cast(tensor, to=dtype)
 
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6238,6 +6238,24 @@ def aten_tanh(self: TFloat) -> TFloat:
     return op.Tanh(self)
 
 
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6238,6 +6238,24 @@ def aten_tanh(self: TFloat) -> TFloat:
...
+@torch_op("aten::tensor.float")
+def aten_tensor_float(self: float, dtype: int) -> TensorType:
+    tensor = op.Constant(value_float=self)
+    return op.Cast(tensor, to=dtype)
+
+
+@torch_op("aten::tensor.bool")
+def aten_tensor_bool(self: bool, dtype: int) -> TensorType:
+    tensor = op.Constant(value_int=self)
+    return op.Cast(tensor, to=dtype)
```
justinchuby, line 6256: I would sort by name. Thanks!

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1464,6 +1464,11 @@ def _where_input_wrangler(
         input_wrangler=_sum_input_wrangler,
         trace_only=True,
     ),
+    TorchLibOpInfo("aten.tensor.int", core_ops.aten_tensor_int),  # Custom from extra_opinfo
+    TorchLibOpInfo(
+        "aten.tensor.float", core_ops.aten_tensor_float  # Custom from extra_opinfo
+    ),
+    TorchLibOpInfo("aten.tensor.bool", core_ops.aten_tensor_bool),  # Custom from extra_opinfo
```
justinchuby, line 1471: I would sort this as well

justinchuby (approved):
Thanks!

BowenBao:
Windows CI crash has become more frequent :(

justinchuby:
So far they look like ops related to ReduceMin / ReduceMax

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6257,6 +6257,24 @@ def aten_tanh(self: TFloat) -> TFloat:
     return op.Tanh(self)
 
 
+@torch_op("aten::tensor.bool")
+def aten_tensor_bool(self: bool, dtype: int) -> TensorType:
```
titaiwangms, line 6261: Should we modify the TensorType to one of TypeVar we have?

BowenBao: it's `dtype` dependent and is equivalent to the scope of `onnx::Cast`, what should we use? I mimic-ed what `prims_convert_element_type`, aka `cast` did.


```def prims_convert_element_type(a: TensorType, dtype: int) -> TensorType:```

justinchuby: We can probably use the underlying type for `TTensor`. (not the typevar since the input and output are not constraint to the same type) Right now it is `_TensorType`. We can give it a better name to avoid collision.

justinchuby: Since `TensorType` may accept String as well which is not we want, but it is not a big deal

PR #814, opened 2023-06-27
Add ops (view*) | feat(torchlib)

justinchuby:
We always operate on the real representation of a complex number in torchlib. So view_as_real/view_as_complex are no-op. This PR additionally implements other view* ops.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -28,6 +28,7 @@ from onnxscript.function_libs.torch_lib.tensor_typing import (
     TRealUnlessFloat16OrInt8,
     TRealUnlessInt16OrInt8,
     TTensor,
+    TTensor2,
     TTensorOrString,
 )
 from onnxscript.onnx_opset import opset18 as op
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6758,40 +6758,56 @@ def aten_view(self: TTensor, size: IntType) -> TTensor:
...
 def aten_view_as(self: TensorType, other: TensorType) -> TensorType:
     """view_as(Tensor(a) self, Tensor other) -> Tensor(a)"""
 
-    raise NotImplementedError()
+    size = op.Shape(other)
+    return op.Reshape(self, size)
 
 
+@torch_op("aten::view_as_complex")
 def aten_view_as_complex(self: TensorType) -> TensorType:
```
titaiwangms, line 6770: Why not update TensorType to something in tensor_typing.py? I thought TensorType should always be changed?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Good point. Thanks!

justinchuby: Done. Updated to TTensor

onnxscript/function_libs/torch_lib/tensor_typing.py
```diff
@@ -56,6 +56,7 @@
 ]
 
 TTensor = TypeVar("TTensor", bound=_TensorType)
+TTensor2 = TypeVar("TTensor2", bound=_TensorType)
```
fatcat-z, line 59: What kind of problem will be caused if we don't have such duplicates? If there is a reason, it'd better to leave here as comments for a reminder.

justinchuby: Done. Added a comment

fatcat-z (approved):

PR #827, opened 2023-07-03
Indexing and slicing support (part 2)

gramalingam:
Fixes to the translation of expressions of the form `exp1[exp2]`. This is a continuation of https://github.com/microsoft/onnxscript/pull/711 which fixed the eager-mode handling of such expressions. This PR implements the same for the script translator. 

In the expression `e[i_1, i_2, ..., i_n]`, `n` is either the rank of the input tensor or any value less than that. Each index-value `i_j` may be a scalar value (a tensor of rank zero) or a higher-dimensional tensor or a slice-expression of the form `start:end:step`. Semantically, a slice-expression `start:end:step` is equivalent to a 1-dimensional tensor containing the corresponding sequence of values.

Main changes:
* Previously the translation of `X[Y]` assumed `Y` was a scalar. This PR handles the situation correctly when the rank of Y is unknown. 
* Handle default start/stop values using minint/maxint.

TODO:
* Doesn't yet handle default-values for start/stop when the step-direction is unknown. This can be handled by emitting code to choose the default value based on a runtime check on the sign of the step-value.


onnxscript/_internal/ast_utils.py
```diff
@@ -3,9 +3,12 @@
 
 import ast
 import inspect
+import sys
 import textwrap
 import types
 
+py_version_ge_39 = sys.version_info[:2] >= (3, 9)
```
justinchuby, line 10:
```suggestion
PY_VERSION_GE_39 = sys.version_info >= (3, 9)
```

Upper case, also this seems to work: https://github.com/asottile/flake8-2020#rationale

onnxscript/_internal/ast_utils.py
```diff
@@ -22,3 +25,22 @@ def get_src_and_ast(f: types.FunctionType) -> tuple[str, ast.FunctionDef]:
     f_ast = top_level_ast.body[0]
     assert isinstance(f_ast, ast.FunctionDef)
     return src, f_ast
+
+
+def normalize_subscript_expr(expr: ast.Subscript):
+    # Normalizes the representation of a subscripted expression, handling python version
+    # differences as well as variations between A[x] (single-index) and A[x, y] (multiple indices)
+    # Returns a list of expressions, denoting the indices, after stripping the extraneous "Index"
+    # wrapper present in python versions before 3.9
```
justinchuby, line 34: Convert to docstring?

onnxscript/_internal/ast_utils.py
```diff
@@ -22,3 +25,22 @@ def get_src_and_ast(f: types.FunctionType) -> tuple[str, ast.FunctionDef]:
...
+    if py_version_ge_39:
+        if isinstance(index_expr, ast.Tuple):
+            return index_expr.elts  # multiple indices
+        else:
+            return [index_expr]  # single index
+    else:
+        if isinstance(index_expr, ast.ExtSlice):
+            indices = index_expr.dims  # multiple indices
+        else:
+            indices = [index_expr]  # single slice-index
```
justinchuby, line 45: Looks like we are missing some coverage

onnxscript/_internal/ast_utils.py
justinchuby: Also move this to version_utils?

onnxscript/tests/models/getitem.py
```diff
@@ -3,115 +3,183 @@
...
+def get_last_row(A: INT32[...]) -> INT32[...]:
+    return A[-1]
 
 
-@script()
-def getitem_index_int0(A: FLOAT[...]) -> FLOAT[...]:
-    zero = op.Constant(value=make_tensor("zero", TensorProto.INT64, [1], [0]))
-    r = A[zero]
-    return r
+get_last_row_test = test(get_last_row, input=x, output=[9, 10, 11])
```
justinchuby, line 54: Possible to parameterize with `parameterized` instead?

justinchuby (commented):
Just some nits - I understand this is WIP so just fyi!

onnxscript/_internal/ast_utils.py
gramalingam: Done

onnxscript/_internal/ast_utils.py
gramalingam: This is meant to handle differences between python 3.8 and 3.9. I assume that the code-coverage is running on one of the two versions. (Wonder if there is a way of combining code-coverage from multiple python versions!)

onnxscript/tests/models/getitem.py
gramalingam: How? Why? I would like to keep the script-function being tested, and the input/output values together in one place (makes it easier for the reader). The existing structure achieves this.

onnxscript/_internal/ast_utils.py
justinchuby: hmm. I think codecov should combine coverage info from different runs

onnxscript/tests/models/getitem.py
justinchuby: I think the follow can achieve this (keeping things together) while making the test case look like the standard tests:

```python
import parameterized


class GetItemTest(unittest.TestCase):
    @parameterized.parameterized.expand([
        ("get_first_row", get_first_row, x, [0, 1, 2]),
        ("get_last_row", get_last_row, x, [9, 10, 11]),
    ])
    def test_function(self, _: str, function, input, output):
        FunctionTestParams(function, wrap_input_output(input), wrap_input_output(output))

```

reference: https://pypi.org/project/parameterized/

justinchuby (approved):

onnxscript/tests/models/getitem.py
gramalingam: I understand the use of parameterized. But it still will split the definition of `get_first_row` from the test-case that tests it ... e.g., when you have 15 such functions. Your suggestion would apply to the existing test-case definition [here](https://github.com/microsoft/onnxscript/blob/bcee0ecf57506b8e6cae5dc7c93ae5195ee7071b/onnxscript/converter_test.py#L435) ... but I found it tedious to correlate these test-cases to the definition of the function being tested. (Of course, the situation was worsened since the definitions were in a different file.)

Maybe we can define a function that combines `@script` and `@parametrized` into one decorator. Something for a future PR.

PR #835, opened 2023-07-06
Make aten::contiguous and device_put no-op | fix(torchlib)

justinchuby:
Make contiguous & device_put  an no-op because memory format or device is not a notion in ONNX.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1367,17 +1367,14 @@ def aten_constant_pad_nd(self: TTensor, pad: INT64, value: float = 0.0) -> TTens
...
 
-    if memory_format in ["contiguous_format", "preserve_format"]:
-        return op.Identity(self)
-    else:
-        # TODO: Find out a way to annotate constraints for argument, as part of the function meta data structure.
-        raise NotImplementedError(
-            "memory_format value supports 'contiguous_format' or 'preserve_format' only."
-        )
+    # ONNX does not have the notion of memory_format. It is always treated as a no-op.
+    return op.Identity(self)
```
fatcat-z, line 1377: I'm wondering if we do so, users calling this function will have a wrong misunderstanding that all of formats were processed successfully, which is not right.
If we can handle this op in an earlier phase by exporter, this should be fine, and we'd better leave a comment here.

justinchuby: I understand the format is an inherent representation and does not affect computation in terms of the result? I gathered that from https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html

justinchuby: @BowenBao do you have more info on this op? Do you think we should filter it out in a fx pass?

BowenBao (approved):

PR #846, opened 2023-07-10
Ensure no duplicate output names

gramalingam:
Fix the translation of `return Y, Y`. Since ONNX does not allow duplicate output names, we need to insert an extra copy via Identity in such cases.

Add test-case for the same.

### DIFF
--- a/onnxscript/converter_test.py
+++ b/onnxscript/converter_test.py
@@ -559,6 +559,18 @@ class TestConverter(testutils.TestBase):
         node = none_as_input.to_function_proto().node[0]
         self.assertEqual(node.input[1], "")
 
+    def test_no_duplicate_output_name(self):
+        """Test that the converter does not generate duplicate output names."""
+
+        @script()
+        def duplicate_output(X):
+            Y = op.Neg(X)
+            return Y, Y
+
+        # The converter should generate distinct names for the two outputs
+        outputs = duplicate_output.to_function_proto().output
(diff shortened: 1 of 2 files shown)


justinchuby (approved):
❤️

onnxscript/converter_test.py
```diff
@@ -559,6 +559,18 @@ def none_as_input(X):
         node = none_as_input.to_function_proto().node[0]
         self.assertEqual(node.input[1], "")
 
+    def test_no_duplicate_output_name(self):
+        """Test that the converter does not generate duplicate output names."""
+
+        @script()
+        def duplicate_output(X):
+            Y = op.Neg(X)
+            return Y, Y
```
justinchuby, line 568: if there were four Ys (hypothetically even though unrealistic), would the last one be Y_copy_copy_copy?

gramalingam: No. The names are determined from the root "Y_copy" by name-uniquification strategy. It produces the below:
```
DuplicateOutputName (X) => (Y, Y_copy, Y_copy_0, Y_copy_1)
{
   Y = Add (X, X)
   Y_copy = Identity (Y)
   Y_copy_0 = Identity (Y)
   Y_copy_1 = Identity (Y)
}
```

justinchuby: I see. Thanks!

PR #847, opened 2023-07-11
Implement aten._local_scalar_dense | feat(torchlib) 

wschin:
This op is used in llama when capturing graph using dynamo.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -41,7 +41,7 @@ _MATH_PI = math.pi
 
 @torch_op("aten::_local_scalar_dense")
 def aten__local_scalar_dense(self: TTensor) -> TTensor:
-    """aten::_local_scalar_dense(Tensor self) -> Scalar"""
+    """_local_scalar_dense(Tensor self) -> Scalar"""
 
     # Return the first element in tensor as a scalar.
     return op.Gather(op.Reshape(self, [-1]), 0)
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -39,6 +39,14 @@ _INT64_MIN = -9223372036854775808
 _MATH_PI = math.pi
 
 
+@torch_op("aten::_local_scalar_dense")
+def aten__local_scalar_dense(self: TTensor) -> TTensor:
+
+    flattened = op.Reshape(self, op.Constant(value_ints=[-1]))
+    first_item_as_scalar = op.Gather(flattened, 0)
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -39,6 +39,14 @@
 _MATH_PI = math.pi
 
 
+@torch_op("aten::_local_scalar_dense")
+def aten__local_scalar_dense(self: TTensor) -> TTensor:
```
justinchuby, line 43: Could you add a docstring with its aten signature from https://aka.ms/native-functions  (looks like `_local_scalar_dense(Tensor self) -> Scalar`. Would be nice to have a reference to its implementation or documentation too.

justinchuby (approved):
Thanks!

justinchuby (commented):
I suggest moving this function to the `internal.py` files because internal (`_`) aten ops seems to be a different category. What do people think? @xiaowuhu @fatcat-z @titaiwangms @BowenBao ?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -39,6 +39,14 @@
...
+    """aten::_local_scalar_dense(Tensor self) -> Scalar"""
```
justinchuby, line 44:
```suggestion
    """_local_scalar_dense(Tensor self) -> Scalar"""
```

nit: consistency

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -556,6 +556,7 @@ def sample_inputs_bernoulli_p_deterministic(op_info, device, dtype, requires_gra
 OP_DB: List[opinfo_core.OpInfo] = [
     opinfo_core.OpInfo(
         "aten._local_scalar_dense",
+        # pylint: disable=protected-access
         op=torch.ops.aten._local_scalar_dense,
```
justinchuby:
```suggestion
        op=torch.ops.aten._local_scalar_dense,  # pylint: disable=protected-access
```

Otherwise it will disable the rule for everything below this line

justinchuby: Also possible:

`# pylint: disable-next=protected-access`

https://stackoverflow.com/questions/28829236/is-it-possible-to-ignore-one-single-specific-line-with-pylint

thiagocrepaldi (approved):

justinchuby:
> I suggest moving this function to the `internal.py` files because internal (`_`) aten ops seems to be a different category. What do people think? @xiaowuhu @fatcat-z @titaiwangms @BowenBao ?

I will move later

Issue #706, opened 2023-05-02
Auto computed type constraints for functions


justinchuby:
Currently the type constraints for onnx functions in the generated OpSchemas are hand specified (e.g. `TReal` etc.). Since the graph in a function is fully defined, we may be able to generate the type constraints as well (using some graph algorithms) and automatically update/correct the type annotations in the function definitions.

cc @xiaowuhu @fatcat-z @titaiwangms @BowenBao @gramalingam

titaiwangms:
@abock Is the next step of https://github.com/microsoft/onnxscript/pull/768 this issue?

PR #849, opened 2023-07-11
Type constraint deducer for OnnxFunction | feat(torch_lib)

BowenBao:
First step towards #706

High level goals:

* Produces the least strict type constraints allowed, by deducing from OpSchema in FunctionProto.
* Produces explanation for type constraint tightening.
* Expects Tensors to be pre-annotated as `TensorType`.
* Expects Attributes to be pre-annotated correctly.
* trace_only=True is not planned.

What's not there yet:

* Nested functions.
* Loop & Scan op.
* Organized explanation.
* Tool makes suggestions to existing type annotation.
* Tool makes ast codegen rewrites to existing type annotation.

What's not planned:

* trace_only=True type deduction.
* Attribute type deduction.

Example on an incorrectly type annotated function:

```python
import inspect
from onnxscript.function_libs.torch_lib.ops.core import _aten_argmax_dim
from onnxscript.function_libs.tools.torch_lib import deduce_type_constraints

onnx_function = _aten_argmax_dim
print("Original signature: ", inspect.signature(onnx_function.function))
print(deduce_type_constraints.deduce_type_constraints(onnx_function))
```
Outputs:
```
Original signature:  (self: 'TRealOrUInt8', dim: 'int', keepdim: 'bool' = False) -> 'TRealOrUInt8'
Type Constraints:
  Inputs: 
    self: T0
  Outputs: 
    result_7: T1
  Type Constraints: 
    T0: {'tensor(uint8)', 'tensor(int64)', 'tensor(float16)', 'tensor(int32)', 'tensor(float)', 'tensor(bfloat16)', 'tensor(uint32)', 'tensor(int8)', 'tensor(uint64)', 'tensor(int16)', 'tensor(uint16)', 'tensor(double)'}
    T1: {'tensor(int64)'}
```
The output type constraint produced by deducer is much tighter. Taking a closer look
```python
import onnxscript
print(onnxscript.proto2text(onnx_function.to_function_proto()))
print(deduce_type_constraints.deduce_type_constraints(onnx_function, signature_only=False))
```
```
_aten_argmax_dim <dim>(self) => (result_7)
{
   tmp = Shape (self)
   tmp_0 = Size (tmp)
   int64_0 = Constant <value = int64 int64_0 {0}> ()
   int64_0_cast = CastLike (int64_0, tmp_0)
   self_is_scaler = Equal (tmp_0, int64_0_cast)
   self_4 = If (self_is_scaler) <then_branch = thenGraph_6 () => ( self_2) {
      tmp_1 = Constant <value_ints = [-1]> ()
      self_2 = Reshape (self, tmp_1)
   }, else_branch = elseGraph_6 () => ( self_3) {
      self_3 = Identity (self)
   }>
   result = ArgMax <axis: int = @dim, keepdims: int = @keepdim> (self_4)
   result_7 = If (self_is_scaler) <then_branch = thenGraph_10 () => ( result_5) {
      result_5 = Squeeze (result)
   }, else_branch = elseGraph_10 () => ( result_6) {
      result_6 = Identity (result)
   }>
}
Type Constraints:
  Inputs: 
    self: T0
  Outputs: 
    result_7: T1
  Type Constraints: 
    T0: {'tensor(int32)', 'tensor(uint64)', 'tensor(int16)', 'tensor(float16)', 'tensor(double)', 'tensor(float)', 'tensor(uint16)', 'tensor(uint8)', 'tensor(int8)', 'tensor(uint32)', 'tensor(int64)', 'tensor(bfloat16)'}
    T1: {'tensor(int64)'}
  Intermediate Values: 
    self: T0
    tmp: T2
    tmp_0: T3
    int64_0: T4
    int64_0_cast: T3
    self_is_scaler: T5
    self_4: T0
    tmp_1: T6
    self_2: T0
    self_3: T0
    result: T1
    result_7: T1
    result_5: T1
    result_6: T1
  Intermediate Type Constraints: 
    T0: {'tensor(int32)', 'tensor(int64)', 'tensor(double)', 'tensor(uint8)', 'tensor(uint32)', 'tensor(float)', 'tensor(int16)', 'tensor(int8)', 'tensor(uint64)', 'tensor(float16)', 'tensor(uint16)', 'tensor(bfloat16)'}
    T2: {'tensor(int64)'}
    T3: {'tensor(int64)'}
    T4: {'tensor(int64)'}
    T5: {'tensor(bool)'}
    T6: {'tensor(int64)'}
    T1: {'tensor(int64)'}
```
We can find `T1` first appeared as output from `result = ArgMax <axis: int = @dim, keepdims: int = @keepdim> (self_4)`. Notice that `ArgMax` constraints output type to `INT64`.


onnxscript/function_libs/tools/torch_lib/deduce_type_constraints.py
```diff
@@ -0,0 +1,403 @@
...
+        param_schemas = list(param_schemas)
+        # If the last parameter is variadic, duplicate it to match the number of parameters.
+        if (
+            len(param_schemas) < len(param_names)
+            and param_schemas[-1].option == onnx.defs.OpSchema.FormalParameterOption.Variadic
+        ):
+            param_schemas += [param_schemas[-1]] * (len(param_names) - len(param_schemas))
+        for name, schema in zip(param_names, param_schemas):
+            if is_output:
+                if name in self.values:
```
gramalingam, line 265: Skip optional outputs too.

gramalingam:
This is great! (This is what's called a unification-based type-inference.) One of the challenges, though, is that while this can handle type-constraints explicitly captured in the opschema, it cannot take into account the logic expressed in the type-and-shape-inference-method, which requires the duplicating that logic in `perform_extra_type_constraint_tightening` (e.g., for Constant/Cast etc.,). For these ops specifically, we may be better off invoking ONNX's shape-inference method to do it, but not sure how one might do it for arbitrary ops.

It may be worth following this initial inference, with a final validation (exhaustive checking of all inferred type-combinations by invoking the ONNX checker, as implemented [here](https://github.com/onnx/onnx/blob/86f7744b866a253a4f1286a108f29128509777c0/onnx/test/cpp/function_verify_test.cc#L266) ... that has the advantage of applying extra constraints specified in the operator's type-inference-method. What is hard there is refining the result: e.g., if we know (X: int64, Y:float) is invalid, we can't determine if X's type should be refined or Y's type should be refined.

onnxscript/function_libs/tools/torch_lib/deduce_type_constraints.py
```diff
@@ -0,0 +1,403 @@
...
+        self.name: str = name
+
+    def merge_type_constraint(self, other: Value):
+        if other.type_constraint is not None:
+            other.type_constraint.bind_value(self)
+        elif self.type_constraint is not None:
+            self.type_constraint.bind_value(other)
+        else:
+            raise ValueError(
+                f"Cannot merge two values without type constraints. {self} and {other}"
```
gramalingam, line 117: It could be done if None is treated as the set of all types. This may be useful to eliminate the restriction that inputs must have the "TensorType" annotation. Of course, the set of all types is infinite, but in practice, ONNX too restricts to types of depth 2 or 3 (like sequences of tensors, etc.). So, two choices to do this would be to use some depth-bounded set of all typestrs, or use None as a shorthand for that in class TypeConstraint also.

onnxscript/function_libs/tools/torch_lib/deduce_type_constraints.py
```diff
@@ -0,0 +1,403 @@
+from __future__ import annotations
```
justinchuby, line 1: As discussed it would be helpful to include an explanation of the binding and updating process as documentation.

onnxscript/function_libs/tools/torch_lib/deduce_type_constraints.py
```diff
@@ -0,0 +1,403 @@
...
+
+class TypeConstraintDeducer:
+    def __init__(self, onnx_function: onnxscript.OnnxFunction):
+        self.onnx_function = onnx_function
+        self.values: Dict[str, Value] = {}
+
+    def type_constraints(self, signature_only: bool = True) -> OnnxFunctionTypeConstraints:
+        """Retrieve deduced type constraints for the ONNX function."""
+        if not self.values:
+            raise ValueError("Must call deduce() first")
```
justinchuby, line 198: Potentially reorganize the logic to make type_constraints() and deduce() more separate with clear responsibilities.

justinchuby (approved):
Awesome tool. LGTM with comments, thanks!

justinchuby:
> This is what's called a unification-based type-inference.

Very cool knowledge, thanks! A search led me to https://www.cs.cmu.edu/~fp/courses/15816-f16/schedule.html , https://papl.cs.brown.edu/2016/Type_Inference.html, and https://www.cs.cmu.edu/~janh/courses/ra20/assets/pdf/lect03.pdf. Should have taken PL I guess

justinchuby:
Is there a principled way to implement the algorithm? @gramalingam

justinchuby:
This is so far a manual tool because we can have other considerations to what types we want to use. I think Bowen has run it

BowenBao:
No work was done for linter. There are some good comments I didn't have time to address too. The description explains the current state of the tool.

Issue #848, opened 2023-07-11
[torchlib] `_attention_scale` generates function_proto that breaks SSA


BowenBao:
```python
from onnxscript.function_libs.torch_lib.ops.nn import _attention_scale
# Look for node `Gather`, input and output are both 'tmp'.
"""
node {
  input: "tmp"
  input: "int64_m1"
  output: "tmp"
  name: "n2"
  op_type: "Gather"
  attribute {
    name: "axis"
    i: 0
    type: INT
  }
  domain: ""
}
"""
print(_attention_scale.to_function_proto())
```

gramalingam:
I suspect it must be from the extension I merged in for translation of `X[Y]`. I will take a look.

I wonder if we should have test-cases that check the validity of function-translation in the CI.

BowenBao:
Thanks @gramalingam. In fact I'm surprised it is not caught. I think the ort fullgraph test under ops_test.py should run with model_proto generated from it. Maybe the branch uses `_attention_scale` is not covered? @justinchuby

justinchuby:
Not seeing crashes in my latest tests. Need to verify.

BowenBao:
I believe OP is resolved by #851. We can close this and track enabling `nn.functional.scaled_dot_product_attention` in CI separately.

PR #851, opened 2023-07-11
Ensure names are unique

gramalingam:
`generate_unique_name` was not being called along one path in the translation of index-expressions. (See https://github.com/microsoft/onnxscript/issues/848 ). Fix the bug.

### DIFF
--- a/onnxscript/converter_test.py
+++ b/onnxscript/converter_test.py
@@ -559,6 +559,15 @@ class TestConverter(testutils.TestBase):
         node = none_as_input.to_function_proto().node[0]
         self.assertEqual(node.input[1], "")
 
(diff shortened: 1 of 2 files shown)


onnxscript/converter_test.py
```diff
@@ -559,6 +559,15 @@ def none_as_input(X):
         node = none_as_input.to_function_proto().node[0]
         self.assertEqual(node.input[1], "")
 
+    def test_unique_names_in_subscript_expr(self):
```
justinchuby, line 562: `..._for_nested_index_expression`? since it looks like its testing this specific case?

justinchuby (approved):

Issue #761, opened 2023-06-01
Flaky CI: worker 'gw0' crashed while running xxx


BowenBao:
Some example runs:
https://github.com/microsoft/onnxscript/actions/runs/5138839927/jobs/9248599814?pr=755
https://github.com/microsoft/onnxscript/actions/runs/5149591642/jobs/9272793720?pr=759
https://github.com/microsoft/onnxscript/actions/runs/5159201503/jobs/9293826467?pr=758

justinchuby:
Some flaky ort again:

https://github.com/microsoft/onnxscript/actions/runs/5225635296/jobs/9435363354#step:5:14132

`[gw1] win32 -- Python 3.10.11 D:\a\onnxscript\onnxscript\.nox\test_torch_nightly\Scripts\python.EXE
worker 'gw1' crashed while running 'onnxscript/tests/function_libs/torch_lib/ops_test.py::TestOutputConsistencyFullGraphCPU::test_output_match_opinfo__min_other_binary_cpu_float16'`

justinchuby:
So far they look like ops related to ReduceMin / ReduceMax

BowenBao:
And so far windows only?

justinchuby:
I've seen linux once or twice

jcwchen:
Are these 2 tests (`test_output_match_opinfo__maximum_cpu_float16` and `test_output_match_opinfo__minimum_cpu`) also flaky? I saw them failed on Windows in a few PRs which seem irrelevant. Since it happened quite often, perhaps we can skip these tests on Windows for now?

justinchuby:
Sure, I will add a skip

justinchuby:
https://github.com/microsoft/onnxruntime/issues/16492

PR #852, opened 2023-07-11
Enable int64 tests for torchlib | test(torchlib)

justinchuby:
Enable int64 tests and marked all failures as xfail with information to follow up.

- Fix annotation for `div` and `logsumexp`
- Fix `bernoulli` for int64 inputs
- Use the ONNX native `Bernoulli` to implement bernoulli

cc @wschin


justinchuby:
> There are testing errors about min/minimum in WIN.

Those are known and will be addressed by #761

titaiwangms (approved):

justinchuby:
Merging now. Future comments will be addressed in a follow PR. Thanks!

PR #861, opened 2023-07-11
Simplify bernoulli implementation | fix(torchlib)

justinchuby:
I tested with 

```python
@torch_op("aten::bernoulli.p")
def aten_bernoulli_p(self: TTensor, p: float) -> TTensor:
    """Proximal implementation of aten::bernoulli.p(Tensor self, float p, *, Generator? generator=None)

    Ignore `generator` due to the limit on ONNX expressiveness.
    """
    self_shape = op.Shape(self)
    p = op.Cast(p, to=FLOAT.dtype)
    probs = op.Expand(p, self_shape)
    sampled = op.Bernoulli(probs)
    return op.CastLike(sampled, self)
```

But the results don't match with torch

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -970,6 +970,7 @@ def aten_bernoulli_p(self: TTensor, p: float) -> TTensor:
 
     Ignore `generator` due to the limit on ONNX expressiveness.
     """
+    # NOTE: We will lose some precision when input is float64 but that's considered insignificant
     self_float = op.Cast(self, to=FLOAT.dtype)
     rands = op.RandomUniformLike(
         self_float,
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -973,7 +970,6 @@ def aten_bernoulli_p(self: TTensor, p: float) -> TTensor:
 
     Ignore `generator` due to the limit on ONNX expressiveness.
     """
-    # NOTE: We will lose some precision when input is float64 but that's considered insignificant
```
BowenBao, line 976: is this deletion intended?

BowenBao (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Yes, because we don't rely on self's value to create the result

justinchuby: I will add it back.

PR #862, opened 2023-07-12
Implement aten::index | feat(torchlib)

justinchuby:
WIP: need more tests

Gather op:
- https://github.com/openxla/xla/blob/main/docs/operation_semantics.md?rgh-link-date=2023-07-13T01%3A09%3A16Z#gather
- https://www.pathpartnertech.com/gather-scatter-operation-in-deep-learning-framework/

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -14,8 +14,6 @@ from __future__ import annotations
 import math
 from typing import Any, Optional, Sequence, Tuple, Union
 
-import numpy as np
-
 from onnxscript import BOOL, DOUBLE, FLOAT, INT8, INT16, INT32, INT64, graph
 from onnxscript.function_libs.torch_lib.registration import torch_op
 from onnxscript.function_libs.torch_lib.tensor_typing import (
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3046,12 +3046,23 @@ def _has_none_in_middle(indices) -> bool:
 def aten_index(self: TensorType, indices: Sequence[Optional[INT64]]) -> TensorType:
     """index.Tensor(Tensor self, Tensor?[] indices) -> Tensor"""
 
-    # Move the None indices to the end
-    ordered_indices = sorted(range(len(indices)), key=lambda i: (indices[i] is None, i))
+    # NOTE: Understanding aten::index
+    # The indexing operation x[0, :, 1:2, tensor([[4,5]])] will be translated to
+    # A bunch of Slice operations, followed by aten::index with
+    # self = rank ? tensor
+    # indices = [None, None, None, tensor([[4,5]])]
+    # TODO(justinchuby): Clarify what happens with 0
+
+    # reordered_positions is the permutation of the index positions where
+    # positions with None are move to the end
+    # For example, if indices = [None, 1, None, 2], then reordered_positions = [1, 3, 0, 2]
+    reordered_positions = sorted(range(len(indices)), key=lambda i: (indices[i] is None, i))
     # Fill the list with the remaining indices up to the rank of the tensor
-    ordered_indices = [*ordered_indices, *range(len(ordered_indices), len(self.shape))]
-    # Transpose the tensor to the axis in the order of [provided, None, not provided]
-    self = op.Transpose(self, perm=ordered_indices)
+    # For example, if indices = [None, 1, None, 2], and the rank of self is 6,
+    # then reordered_positions = [1, 3, 0, 2, 4, 5]
+    reordered_positions = [*reordered_positions, *range(len(reordered_positions), len(self.shape))]
+    # Transpose self according to the reordered positions
+    self = op.Transpose(self, perm=reordered_positions)
 
     # Broadcast the indices to the same shape then concatenate
     not_none_indices = [idx for idx in indices if idx is not None]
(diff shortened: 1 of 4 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3026,10 +3028,54 @@ def aten_imag(self: TensorType) -> TensorType:
...
+        return sorted(lst) == list(range(min(lst), max(lst) + 1))
+
+
+def _is_none_in_middle(indices):
+    not_none_indices = [i for i, idx in enumerate(indices) if idx is not None]
+    return not are_consecutive(not_none_indices)
+
+
+@torch_op("aten::index", trace_only=True)
+def aten_index(self: TensorType, indices: Sequence[Optional[INT64]]) -> TensorType:
```
BowenBao, line 3044: This should work for all cases except for bool mask index, if possible.

Let me know if you can find a bug! @justinchuby

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3026,10 +3028,54 @@ def aten_imag(self: TensorType) -> TensorType:
...
 
-    raise NotImplementedError()
+    ordered_indices = sorted(range(len(indices)), key=lambda i: (indices[i] is None, i))
+    ordered_indices += list(range(len(ordered_indices), len(self.shape)))
+
+    self = op.Transpose(self, perm=ordered_indices)
+
+    # Need broadcast concat lol.
+    not_none_indices = [idx for idx in indices if idx is not None]
+    broadcast_shape = np.broadcast_shapes(*[idx.shape for idx in not_none_indices])
```
justinchuby, line 3054: We may not have shapes here: need to assert or use alternatives like implicit broadcasting?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3026,10 +3028,54 @@ def aten_imag(self: TensorType) -> TensorType:
...
+
+    self = op.GatherND(self, final_index, batch_dims=0)
+    if _is_none_in_middle(indices):
+        return self
+
+    # Need to transpose back.
+    adv_res_rank = len(broadcast_shape)
+    # [adv1, adv2, ..., adv_res_rank, x1, x2, ..., xk, ..., xn] -> [x1, x2, ..., xk, adv1, ..., adv_res_rank, ..., xn]
+    perm = (
+        list(range(adv_res_rank, adv_res_rank + ordered_indices[0]))
```
justinchuby, line 3068: I would unpack instead

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -446,6 +446,22 @@ def sample_inputs_col2im(op_info, device, dtype, requires_grad, **kwargs):
...
+def sample_inputs_index(op_info, device, dtype, requires_grad, **kwargs):
+    del op_info  # Unused
+    del kwargs  # Unused
+    make_arg = functools.partial(
+        torch_testing.make_tensor, dtype=dtype, device=device, requires_grad=requires_grad
+    )
+    s = 5
+    test_args = [
+        ([common_methods_invocations.index_variable(2, s, device=device)],),
+        # ([torch.tensor()],)
```
justinchuby, line 458: @justinchuby will add more tests

onnxscript/function_libs/torch_lib/ops/core.py
BowenBao: yea that's kinda rough. If we don't, or if there are symbols, we need to dynamically broadcast, probably via `add`. However at some point later we need the rank of broadcasted tensor, which is needed to construct the `perm` for transpose. If we cannot get it statically, then the transpose becomes a headache.

onnxscript/function_libs/torch_lib/ops/core.py
BowenBao: feel free to make changes. the current version is quick hack style

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: I think we can assume the rank. This is more relaxed than assuming shape

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
BowenBao: Sharing snippet I used to test. It poses complication here as indexing, i.e., `data[]`, lowers to multiple operators, usually couple of `aten.slice` + `aten.index` in the end. So you might need to figure out a way to construct the proper input state if you are testing `aten.index.Tensor` alone.

```python
import torch
import onnxruntime
import onnxscript


class IndexModel(torch.nn.Module):
    def forward(self, data, index, index2):
        # return data[..., index, index2]
        # return data[index, :, index2, ...]
        # return data[..., index, index2, :]
        # return data[index, :, index2]
        return data[:, :, index, index2]


data = torch.arange(0, 7 * 3 * 4 * 5 * 6).view(7, 3, 4, 5, 6)
index = torch.tensor([2, 1])
index2 = torch.tensor([[[0]], [[2]]])

model = IndexModel()
pt_output = model(data, index, index2)
print("PyTorch output shape:", pt_output.shape)

export_output = torch.onnx.dynamo_export(model, data, index, index2)
export_output.diagnostic_context.dump("log.sarif")

print(onnxscript.proto2text(export_output.model_proto))
sess = onnxruntime.InferenceSession(
    export_output.model_proto.SerializeToString(), providers=["CPUExecutionProvider"]
)
output = sess.run(None, {"arg0": data.numpy(), "arg1": index.numpy(), "arg2": index2.numpy()})

torch.testing.assert_close(pt_output.numpy(), output[0])

print("PASS")
```

justinchuby:
Pictures: https://github.com/openxla/xla/blob/main/docs/operation_semantics.md#informal-description-and-examples

gramalingam:
Is this migrating the logic used by the previous exporter (for this op)? Or, is this different or new?

BowenBao:
> Is this migrating the logic used by the previous exporter (for this op)? Or, is this different or new?

It is new and hopefully more robust, while the logic is similar.

The old implementation was using opset9.GatherElements, at that point GatherND does not exist yet.

gramalingam:
> The old implementation was using opset9.GatherElements, at that point GatherND does not exist yet.

Oh ... I thought Spandan introduced the new Gather op variants to support the pytorch converter.

BowenBao:
> Oh ... I thought Spandan introduced the new Gather op variants to support the pytorch converter.

Yes it was used in some other ops, as well as extended edge case support for `aten.index` after opset11, but there wasn't a full rewrite.

justinchuby:
Squashing to another branch

Issue #828, opened 2023-07-03
Use explicit ATen overload name in torchlib for all ops


titaiwangms:
In torchlib, we don't register all ATen overloads. FOr example, `aten::add.Tensor` and `aten::add.Int` both match to `aten::add`. The converter is currently using default overload fall back to find ATen overload if it's not in the registry. Should we manage to match the exact overload name? Or we are good at fall back and warning in SARIF?
```[tasklist]
### Tasks
- [ ] https://github.com/microsoft/onnxscript/pull/863
- [ ] https://github.com/microsoft/onnxscript/issues/865
- [ ] Update the dispatching logic in the exporter to account for the new information
```

BowenBao:
We are omitting overload names in most cases in torchlib decorators. Following torch semantics omitting means "default" overload.

The caveat is most of the cases the implementation is for "Tensor" overload.

titaiwangms:
@BowenBao  I see. I guess this is also the reason that when we generate `decomposition table`, we relax the constraints of checking whether the op is supported (fall back to default). Do you think it's risky that we just fall back to default instead of saying we don't support the overload (I put SARIF warning on this though). At some level, it looks like nearest match, and I wonder if we could say we are safe as long as all OpInfo tests are passed both in torchlib and fx exporter.

BowenBao:
> Do you think it's risky that we just fall back to default instead of saying we don't support the overload

titaiwangms:
@justinchuby How about 

```python
@torch_op("aten::add", overloads=["Tensor", "Int"])
def aten_add(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
    """add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor"""
    # FIXME(titaiwang): get rid of this when we have type_promotion
    # TODO(microsoft/onnxruntime#15977): Improve fp16 precision
    other = op.CastLike(other, self)
    alpha = op.CastLike(alpha, other)
    other = op.Mul(other, alpha)
    return op.Add(self, other)
```

justinchuby:
We can do `@torch_op(("aten::add.Tensor", "aten::addInt"))` if desired

justinchuby:
Next steps:

1. Annotate overload information on torchlib functions
2. Update the dispatching logic in the exporter to account for the new information
3. Update the torchlib decorator to allow https://github.com/microsoft/onnxscript/issues/828#issuecomment-1622693049

[...]

PR #863, opened 2023-07-12
Allow functions to be registered to multiple names | feat(torchlib)

justinchuby:
* #864
* __->__ #863

To support registering aliases and overloads of an op, `torch_op` can now take a tuple. #828

### DIFF
--- a/onnxscript/function_libs/torch_lib/registration.py
+++ b/onnxscript/function_libs/torch_lib/registration.py
@@ -102,7 +102,7 @@ def torch_op(
         name: Qualified ATen name of the function. E.g. "aten::relu", "aten::add.Tensor".
             Or a tuple of names e.g. ("aten::add.Scalar", "aten::add.Tensor").
             Default overloads should be specified by omitting the overload part,
-            i.e. ""aten::relu" instead of "aten::relu.default".
+            i.e. "aten::relu" instead of "aten::relu.default".
         registry: Registry to register the function to. If None, the default registry is used.
         trace_only: Whether the function should only be traced and not compiled.
         private: Whether the function is private (not directly exposed). It should
--- a/onnxscript/function_libs/torch_lib/registration.py
+++ b/onnxscript/function_libs/torch_lib/registration.py
@@ -2,10 +2,16 @@
 
 from __future__ import annotations
 
+import re
 from types import FunctionType
 from typing import Any, Callable, Generator, Optional
 
 import onnxscript
+
+# Regex that will match "aten::add" and "aten::add.Tensor"
+_QUALIFIED_OPERATOR_NAME_REGEX = re.compile(
+    r"^(?P<namespace>[a-zA-Z0-9_]+)::(?P<name>[a-zA-Z0-9_]+)(?P<overload>\.[a-zA-Z0-9._]+)?$"
+)
 
 
 class OverloadedFunction:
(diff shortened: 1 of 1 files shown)


titaiwangms (commented):
Do we need a name format checking to make sure it's "namespace::op_name.overload"? Or it's covered in tests or something.

justinchuby:
> Do we need a name format checking to make sure it's "namespace::op_name.overload"? Or it's covered in tests or something.

I can do that

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -74,7 +74,8 @@ def torch_op(
     """Register a torch op.
 
     Args:
-        name: ATen name of the function. E.g. "aten::add".
+        name: Qualified ATen name of the function. E.g. "aten::relu", "aten::add.Tensor".
```
wschin, line 77: `name Union[str, Tuple[str, ...]]` to annotate this function?

justinchuby: Thanks, just added

justinchuby:
> > Do we need a name format checking to make sure it's "namespace::op_name.overload"? Or it's covered in tests or something.
> 
> I can do that

Done.

titaiwangms (approved):

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -2,11 +2,17 @@
 
 from __future__ import annotations
 
+import re
 from types import FunctionType
 from typing import Any, Callable, Generator, Optional
 
 import onnxscript
 
+# Regex that will match "aten::add" and "aten::add.Tensor"
```
BowenBao, line 11: nit: would aten::<op_name>[.<overload>] be better? comment gives feeling the regex is specifically for add.

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -63,8 +69,27 @@ def items(self) -> Generator[tuple[str, OverloadedFunction], None, None]:
...
+    names: tuple[str, ...]
+
+    if isinstance(name, str):
+        names = (name,)
+    else:
+        names = name
+    if not isinstance(names, tuple):
+        raise TypeError(f"Name must be a string or a tuple of strings, got {name}")
+    for name_ in names:
+        if name_.endswith(".default") or not _QUALIFIED_OPERATOR_NAME_REGEX.fullmatch(name_):
```
BowenBao, line 82: so we don't allow registering for default? should we call out in `torch_op` docstring

onnxscript/function_libs/torch_lib/registration.py
justinchuby: Done

BowenBao (approved):

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -63,8 +69,27 @@ def items(self) -> Generator[tuple[str, OverloadedFunction], None, None]:
...
+            raise ValueError(
+                f"Invalid name '{name_}'. Must be in the form 'namespace::name.overload' "
```
thiagocrepaldi, line 84: maybe highlight the `.default` is not accepted here too?

onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -74,7 +99,10 @@ def torch_op(
     """Register a torch op.
 
     Args:
-        name: ATen name of the function. E.g. "aten::add".
+        name: Qualified ATen name of the function. E.g. "aten::relu", "aten::add.Tensor".
+            Or a tuple of names e.g. ("aten::add.Scalar", "aten::add.Tensor").
+            Default overloads should be specified by omitting the overload part,
+            i.e. ""aten::relu" instead of "aten::relu.default".
```
thiagocrepaldi, line 105:
```suggestion
            i.e. "aten::relu" instead of "aten::relu.default".
```

onnxscript/function_libs/torch_lib/registration.py
justinchuby: It’s on the next line?

onnxscript/function_libs/torch_lib/registration.py
justinchuby: Thanks

onnxscript/function_libs/torch_lib/registration.py
thiagocrepaldi: indeed minor, but the message doesn't say `torch.add.default` is not allowed, only that `torch.add` is the default overload. that might mean that both `torch.add` (implicit form) and `torch.add.default` (explicit form) are acceptable as default overload

justinchuby: I rephrased the message. Hopefully better now?

PR #867, opened 2023-07-12
Properly check and guard overloading initializer name | fix(graph_building)

BowenBao:
Previously the code path is never triggered due to another bug in exporter that initializers are always added with unique names.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -404,7 +404,7 @@ class TorchScriptGraph:
             # NOTE: Previously it raises when `name` is already set. This is relaxed
             # because this will be invoked multiple times when submodule is called
             # multiple times.
-            if name in self._initializers and self._initializers[name] != value:
+            if name in self._initializers and self._initializers[name] is not value:
                 raise ValueError(
                     f"Initializer '{name}' exists already with a different value."
                 )
--- a/onnxscript/function_libs/torch_lib/graph_building_test.py
+++ b/onnxscript/function_libs/torch_lib/graph_building_test.py
@@ -118,6 +118,7 @@ class TestTorchScriptTracingEvaluator(unittest.TestCase):
         expected = outer.to_model_proto()
         onnxscript.testing.assert_isomorphic(traced, expected)
 
+
 class TestTorchScriptGraph(unittest.TestCase):
     def test_add_initializer_raises_when_the_same_name_used_for_different_tensors(self):
         graph = graph_building.TorchScriptGraph()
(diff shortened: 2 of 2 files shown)


BowenBao:
Not sure which topic label to pick

onnxscript/function_libs/torch_lib/graph_building_test.py
```diff
@@ -118,6 +118,18 @@ def outer(x: FLOAT[1, 2, 3]):
         expected = outer.to_model_proto()
         onnxscript.testing.assert_isomorphic(traced, expected)
 
+    def test_adding_different_tensors_using_same_initializer_name_raises(self):
```
justinchuby, line 121:
```suggestion
class TestTorchScriptGraph(unittest.TestCase):
    def test_add_initializer_raises_when_the_same_name_used_for_different_tensors(self):
    
```

Create a test class for the graph class and update the same to something along this line?

titaiwangms (approved):
I guess there should be another fix in PyTorch. Would you link this PR to the fixed in PyTorch?

onnxscript/function_libs/torch_lib/graph_building_test.py
```diff
@@ -118,6 +118,18 @@ def outer(x: FLOAT[1, 2, 3]):
...
+        graph = graph_building.TorchScriptGraph()
+        graph.add_initializer("x", torch.ones((1, 2, 3), dtype=torch.float32))
+        with self.assertRaises(ValueError):
+            graph.add_initializer("x", torch.ones((1, 2, 3), dtype=torch.float32))
+
+    def test_adding_same_tensors_using_same_initializer_name_passes(self):
```
justinchuby, line 127:
```suggestion
    def test_add_initializer_allows_adding_the_same_tensor_twice_using_same_name(self):
```

etc.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -404,7 +404,7 @@ def add_initializer(self, name: str, value: torch.Tensor) -> TorchScriptTensor:
             # NOTE: Previously it raises when `name` is already set. This is relaxed
             # because this will be invoked multiple times when submodule is called
             # multiple times.
-            if name in self._initializers and self._initializers[name] != value:
```
justinchuby, line 407: So this comparison was invalid right?

BowenBao: It does tensor value comparison, i.e. torch.neq(a, b). It returns a bool tensor.

The check should instead see if they are same tensor object.

BowenBao:
> I guess there should be another fix in PyTorch. Would you link this PR to the fixed in PyTorch?

No code changes on torch side. We'll bump commit in CI in next related PR.

PR #869, opened 2023-07-13
Fix `train` dtype for native_dropout | fix(torchlib)

justinchuby:
Previously the `train` attribute in native_dropout was not casted to BOOL. We need to do that because the underlying attribute type for python bool is INT64 when promoted to input. This change also added a test that will catch the old error.


justinchuby:
> Is it like the difference between using it as an attribute or an input? Looks like it's going to be used as input of op.Dropout, so we need to promote it to BOOL.

That's correct.

PR #871, opened 2023-07-13
Rename extra opinfo names and simplify the OpInfos | test(torchlib)

### LINKED ISSUE #870: Rename all with the op.aten. prefix
This way OpInfo will find the op automatically without us having to specify the op.

justinchuby:
Rename all with the `op.aten.` prefix in extra_opinfos. This way OpInfo will find the op automatically without us having to specify the op. Also simplify the custom OpInfo to remove redundant information.

Fixes #870

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
+++ b/onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
@@ -603,7 +603,7 @@ OP_DB: List[opinfo_core.OpInfo] = [
         supports_out=False,
     ),
     opinfo_core.OpInfo(
-        "ops.aten.conv3d",
+        "nn.functional.conv3d",
         aten_name="conv3d",
         dtypes=common_dtype.floating_and_complex_types_and(torch.int64, torch.bfloat16),
         sample_inputs_func=sample_inputs_conv3d,
@@ -662,21 +662,21 @@ OP_DB: List[opinfo_core.OpInfo] = [
         supports_out=False,
     ),
     opinfo_core.OpInfo(
-        "ops.aten.max_pool1d_with_indices",
+        "nn.functional.max_pool1d_with_indices",
         aten_name="max_pool1d_with_indices",
         dtypes=common_dtype.floating_types_and(torch.bfloat16),
         sample_inputs_func=sample_inputs_max_pool1d_with_indices,
         supports_out=False,
     ),
     opinfo_core.OpInfo(
-        "ops.aten.max_pool2d_with_indices",
+        "nn.functional.max_pool2d_with_indices",
         aten_name="max_pool2d_with_indices",
         dtypes=common_dtype.floating_types_and(torch.bfloat16),
         sample_inputs_func=sample_inputs_max_pool2d_with_indices,
         supports_out=False,
     ),
     opinfo_core.OpInfo(
-        "ops.aten.max_pool3d_with_indices",
+        "nn.functional.max_pool3d_with_indices",
         aten_name="max_pool3d_with_indices",
         dtypes=common_dtype.floating_types_and(torch.bfloat16),
         sample_inputs_func=sample_inputs_max_pool3d_with_indices,
--- a/onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
+++ b/onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
@@ -553,173 +553,133 @@ def sample_inputs_bernoulli_p_deterministic(op_info, device, dtype, requires_gra
             yield opinfo_core.SampleInput(t, kwargs={"p": p})
 
 
+# NOTE: How to create an OpInfo:
+# 1. Create a function that generates sample inputs for the op.
+#    This function should yield SampleInputs.
+#    Use `sample_inputs_col2im` as an example.
+# 2. Specify dtypes that the op supports.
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -553,173 +552,134 @@ def sample_inputs_bernoulli_p_deterministic(op_info, device, dtype, requires_gra
...
         # dtypes can be a tuple of (torch.float, torch.double).
         dtypes=common_dtype.all_types(),
         sample_inputs_func=sample_inputs_bernoulli_p,
     ),
     opinfo_core.OpInfo(
         # Deterministic bernoulli sampling where p is either 0 or 1
-        "aten.bernoulli.p_deterministic",
-        aten_name="bernoulli.p",
+        "ops.aten.bernoulli.p_deterministic",
         op=torch.ops.aten.bernoulli.p,
```
BowenBao, line 681: rm?

justinchuby: Name is `ops.aten.bernoulli.p_deterministic` so we need to specify

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -577,183 +573,160 @@ def sample_inputs_bernoulli_p_deterministic(op_info, device, dtype, requires_gra
...
 
 
+# NOTE: How to create an OpInfo:
+# 1. Create a function that generates sample inputs for the op.
+#    This function should yield SampleInputs.
+#    Use `sample_inputs_col2im` as an example.
+# 2. Specify dtypes that the op supports.
+# 3. Use how you would call the op in PyTorch as the name of the OpInfo.
+#    For example, `torch.ops.aten.col2im` should be named "ops.aten.col2im".
+#    This way OpInfo knows to use `torch.ops.aten.col2im` as the op.
```
BowenBao, line 583: For my understanding, is this implemented by torch's opinfo? or by us?

justinchuby: It's by opinfo

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1361,7 +1361,7 @@ def _where_input_wrangler(
         reason="String padding is not accepted by aten::conv2d",
     ),
     TorchLibOpInfo(
-        "nn.functional.conv3d",
```
BowenBao, line 1364: This changes the tested torch function. Is it expected? what's the guideline for picking which torch function?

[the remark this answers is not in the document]
justinchuby: They only need to match the opinfo names. If the name does not exist in op_db we cannot use it here

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1376,7 +1376,7 @@ def _where_input_wrangler(
     ),
     TorchLibOpInfo("nn.functional.linear", nn_ops.aten_linear, trace_only=True),
     TorchLibOpInfo(
-        "nn.functional.max_pool1d",
```
BowenBao, line 1379: a few other impacted `nn.functional` ones

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
justinchuby: Because I created two cases `ops.aten.bernoulli.p` and `ops.aten.bernoulli.p_deterministic` so we can at least test some numbers. `ops.aten.bernoulli.p` tests only dtype and shape

justinchuby: `ops.aten.bernoulli.p` is the op being tested.

BowenBao: I see. So if op is unspecified, by default OpInfo searches for function `torch.<op_info_name>`.

Feels like we can extend existing op info w/ extra sample data? Hmm I recall you have PR in torch so if that's merged we don't need the special case here.

BowenBao: But if we just want to do extensions locally, that should be feasible as well right?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: Fixed

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
justinchuby: > But if we just want to do extensions locally, that should be feasible as well right?

It should be. `ops.aten.bernoulli.p` was not in op_db so I created one. It contains more valid inputs than `p_deterministic`, (p = 0.5) which are not deterministic.

We could use the `variant` field to create two opinfos of the same name and this issue should go away, but right now in our TorchLibOpInfo there is no way to pick specific variants. We are only able to skip/xfail on variants. So this can be improved in another pass.

PR #873, opened 2023-07-13
Implement a ONNX to ONNX Script code generator based on libcst

abock:
## [WIP] Proposal to use [libcst](https://github.com/instagram/libcst) for code generation

This PR is somewhat experimental, but goes far enough to illustrate using libcst for code generation of ONNX to ONNX Script. It is nearly as complete as the existing template-based `onnxscript.proto2python`, and lays the groundwork for implementing more "raising" passes to produce idiomatic Python code.

I implemented this to learn more about libcst and currently I am heavily leaning on wanting to take a full dependency on libcst. I see adoption happening in three phases:

1. Continue with the work in this PR by implementing more transformers to produce idiomatic Python ONNX Script from ONNX. This keeps the libcst dependency scoped to just this feature.

2. Port the existing `opgen` generator to use libcst. Again, the libcst dependency would be scoped to just this code generator.

3. Finally, let's port our existing Python AST support that uses the builtin `ast` module to `libcst`. I suspect if we do this, lots of utilities that are necessary for (1) and (2) will be highly useful here. More importantly, libcst provides [scope analysis](https://libcst.readthedocs.io/en/latest/scope_tutorial.html) and qualified name resolving, which is the basis for doing real semantic analysis work with CSTTransformer.

Also note, that if we implement various passes as CSTTransformers, we can begin to define a very nice UX in IDEs for "lightbulb fixes". For example any pass we may apply as part of builtin code generation for ONNX to ONNX Script (phase 1), we can surface to the IDEs to upgrade user code to more idiomatic Python. PyTorch is also beginning to use libcst for providing fixes to PyTorch code as well.

### Depending on libcst

libcst itself does not carry many dependencies and is relatively small (about 10MB). Comparatively, onnx itself is rather large (112MB). Thus, I am not too concerned about the size of the libcst dependency, considering it would become central to ONNX Script across the three identified phases.

#### Dependency Cost

| Command | Installed site-packages | Disk Usage | Delta |
| - | - | - | - |
| `python3.10 -m venv venv --upgrade-deps` | setuptools, pip | 19MB | _Baseline_ |
| `pip install onnx` | onnx, google, protobuf, numpy, typing_extensions | 131MB | +112MB |
| `pip install libcst` | libcst, pyyaml, mypy_extensions, typing_inspect | 142MB | +11MB |

### Current capability of this PR:

- Adds some general codegen utilities based on libcst

- Implements an ONNX to ONNX Script generator: the base converter produces ONNX Script that is very 1:1 with the structure of ONNX, and transformers are implemented to raise the generated code to more idiomatic Python that ONNX Script supports; this commit provides support for raising to Python binary operators and raising Constant/make_tensor to supported Python constants; more transformers need to be implemented, but this commit can be used as a guide.

- Adds a new top-level command line interface, allowing the code generator to be invoked:

  ```bash
  python -m onnxscript convert model.onnx
  ```

### DIFF
--- a/onnxscript/codeanalysis/onnx_to_onnxscript.py
+++ b/onnxscript/codeanalysis/onnx_to_onnxscript.py
@@ -1,0 +1,533 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import BinaryIO, Final, Literal, Sequence, cast, overload
+
+import libcst as cst
+import libcst.matchers as cstm
+import libcst.metadata as cstmeta
+import onnx
+
+from onnxscript.codeanalysis import (
+    CstCodeGenerator,
+    RemoveUnusedImportsTransformer,
+    format_code,
+    make_const_expr,
+)
+
+__all__ = [
+    "OnnxScriptCodeGenerator",
+    "OnnxScriptTransformer",
+    "OnnxToPythonOperatorTransformer",
+    "OnnxConstantOpToPythonConstantTransformer",
+    "Driver",
+]
+
+DEFAULT_OPSET_VERSION: Final = 18
+
+
+@dataclass
+class QualifiedOnnxOp:
+    domain: str
+    name: str
+    version: int = 0
+
+    def __eq__(self, __value: object) -> bool:
+        if isinstance(__value, str):
+            return self.domain == "" and self.name == __value
+        elif isinstance(__value, QualifiedOnnxOp):
+            return (
+                self.domain == __value.domain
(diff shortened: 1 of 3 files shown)


onnxscript/codeanalysis/onnx_to_onnxscript.py
```diff
@@ -0,0 +1,533 @@
...
+
+    def translate_tensor_proto(self, tensor_proto: onnx.TensorProto) -> cst.BaseExpression:
+        if onnx.external_data_helper.uses_external_data(tensor_proto):
+            raise NotImplementedError("tensors with external data are not supported")
+
+        self.add_import("onnx")
+
+        numpy_tensor = onnx.numpy_helper.to_array(tensor_proto)
+
+        return cst.Call(
```
gramalingam, line 97: This is one example where verbosity makes this harder to read than the existing template-based mechanism. Do you think it might make sense to judiciously mix parsed strings, with formatted strings, with cst constructors to get the best of both worlds?

PR #874, opened 2023-07-13
Skip reduce min/max ops on Windows | test(torchlib)

justinchuby:
Because ORT randomly aborts on these tests on Windows. ORT is also flaky on Linux/MacOS, but it is much rarer.

https://github.com/microsoft/onnxscript/issues/761, https://github.com/microsoft/onnxruntime/issues/16492


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -708,6 +711,11 @@ def _where_input_wrangler(
     TorchLibOpInfo("mT", core_ops.aten_mT),
     TorchLibOpInfo("mT", core_ops.aten_mT_complex, complex=True),
     TorchLibOpInfo("min_dim", core_ops.aten_min_dim)
```
justinchuby, line 713: We haven't implemented max_dim etc.

jcwchen (approved):
Thanks for skipping them!

PR #875, opened 2023-07-14
Implement rand* ops | feat(torchilb)

justinchuby:
Implement rand* ops. I split all ops that can take a dtype into two overloads to avoid trace_only logic. In pytorch `low` and `high` are symints that can be dynamic, but since Random* ops in ONNX take them as attributes, I annotated the arguments as float for compatibility. SymInts are unsupported until we update the ONNX ops.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5361,47 +5361,132 @@ def aten_rad2deg(self: TensorType) -> TensorType:
...
+    """randint(SymInt high, SymInt[] size, *, ScalarType? dtype=long, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
 
-def aten_randint_like(
-    self: TensorType, high: int, memory_format: Optional[str] = None
+    shaper = op.ConstantOfShape(size)
+    rand = op.RandomUniformLike(shaper)
+    # Round to ints first
+    rand = op.Cast(rand, to=INT64.dtype)
+    # Scale to [0, high]
+    rand = rand * high
```
BowenBao, line 5394: scale first then round

justinchuby: Just came to my mind: why not use `Round` but Cast instead?

justinchuby: > scale first then round

Good catch thanks

justinchuby: Would cast introduce a bias?

BowenBao: > Just came to my mind: why not use `Round` but Cast instead?

That's a good catch, we should use Floor. Because randint is inclusive on low and exclusive on high.

BowenBao: I thought cast behaves like floor, but it seems not the case for negative numbers.

justinchuby:
Merging to the main repo first

PR #876, opened 2023-07-14
Break aten::max into overloads | fix(torchlib)

justinchuby:
`aten::max` was implemented as a big trace_only function. This change breaks it into multiple overloads like we did for `aten::min`.

- Also merged `aten::max/min.other` with `aten::minimum/maximum` as an alias.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3929,8 +3929,8 @@ def aten_mean(self: TReal) -> TReal:
 
 
 @torch_op("aten::mean.dim")
-def aten_mean_dim(self: TReal, dim: INT64, keepdim: bool = False) -> TReal:
-    """mean(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
+def aten_mean_dim(self: TReal, dim: int, keepdim: bool = False) -> TReal:
+    """mean.dim(Tensor self, int[1]? dim, bool keepdim=False, *, ScalarType? dtype=None) -> Tensor"""
 
     if op.Size(op.Shape(self)) == 0:
         result = self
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3883,62 +3883,37 @@ def aten_matrix_power(self: TensorType, n: int) -> TensorType:
...
     return result
 
 
-@torch_op("aten::max", private=True)
-def _aten_max_with_other(self: TReal, other: TReal) -> TReal:
-    result = op.Max(self, other)
-    return result
-
+@torch_op("aten::max.dim")
+def aten_max_dim(self: TReal, dim: int, keepdim: bool = False) -> Tuple[TReal, INT64]:
```
titaiwangms, line 3903: Downstairs on line 3932, there is a mean.dim, I think these two have a pretty similar implementation, but that one seems to be wrongly annotated with dim: INT64. Wonder why is that one passed?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Thanks for catching! I updated the function. Let me check the test

justinchuby: OK so it was actually correct. That is because dim in `mean_dim` is `int[1]? dim`, which is a sequence that is passed as an input. We treat it as INT64 to feed it into the ReduceMean op.

titaiwangms: Did test fail? I thought every parameters went to attribute of an Op, should be primitive type. If it's list of ints, shouldn't it be Sequence[int]?

titaiwangms: Ah I see there is a 

```python
 op.Size(op.Shape(dim)) == 0:
```

justinchuby: I am not sure if there's an easy way to process `Sequence[int]` from an attribute to make it an input. ConcatFromSequence apparently doesn't work plus it is going to be slower. So we make it an input right from the beginning instead.

PR #882, opened 2023-07-17
Simplify model error message in test | test(torchlib)

justinchuby:
* (to be filled)

Remove the duplicate check_model message when displaying

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/ops_test.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test.py
@@ -16,7 +16,7 @@ Usage:
 from __future__ import annotations
 
 import unittest
-from typing import Any, Callable, Optional, Sequence, Tuple
+from typing import Callable, Optional, Sequence, Tuple
 
 import numpy as np
 import onnx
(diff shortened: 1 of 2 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test.py
```diff
@@ -72,18 +72,6 @@ def _should_skip_xfail_test_sample(
     return None, None
 
 
-def _split_function_and_wrangler(
```
BowenBao, line 75: intended change?

BowenBao (approved):

onnxscript/tests/function_libs/torch_lib/ops_test.py
justinchuby: Yes because it was unused

justinchuby:
It’s https://github.com/microsoft/onnxscript/pull/882/files#diff-1b8336d32142b9f922d0006b0a59b39dd3cbf21f36415d98b6a17d8bc8bda34dR520

looks like the diff included previous changes as well

PR #890, opened 2023-07-17
Support OptionalType to get shape | feat(graph_building)

titaiwangms:
Due to the need of supporting None args, we need to support None (torch.OptionalType) to get shape/dtype.
To support https://github.com/pytorch/pytorch/pull/105263

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building_test.py
+++ b/onnxscript/function_libs/torch_lib/graph_building_test.py
@@ -118,6 +118,12 @@ class TestTorchScriptTracingEvaluator(unittest.TestCase):
         expected = outer.to_model_proto()
         onnxscript.testing.assert_isomorphic(traced, expected)
 
+    def test_add_input_works_with_optionaltype(self):
+        graph = graph_building.TorchScriptGraph()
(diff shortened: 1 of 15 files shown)


justinchuby (approved):
Possible to add a test?

onnxscript/function_libs/torch_lib/graph_building_test.py
```diff
@@ -118,6 +118,12 @@ def outer(x: FLOAT[1, 2, 3]):
         expected = outer.to_model_proto()
         onnxscript.testing.assert_isomorphic(traced, expected)
 
+    def test_add_input_works_with_optionaltype(self):
```
justinchuby, line 121: Would be nice to clarify what "works with" mean? Does it mean None can now be an input to the graph, or something else?

titaiwangms: How about test_add_input_with_optionaltype_does_not_raise_torch_internal_error

justinchuby: SG!

PR #894, opened 2023-07-18
Return converter expression uniformly

gramalingam:
A few converter methods were returning either a ConverterExpression or a string. Ensure that they uniformly return a ConverterExpression.

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -1266,7 +1266,8 @@ class Converter:
             )
         body = self.exit_scope()
         inputs = [o_loop_bound, o_true] + [
-            self.py_var_to_onnx_var(pv, self.source_of(loop_stmt)).name for pv in loop_state_vars
+            self.py_var_to_onnx_var(pv, self.source_of(loop_stmt)).name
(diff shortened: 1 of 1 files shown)


onnxscript/converter.py
```diff
@@ -1253,7 +1251,7 @@ def translate_loop_stmt(self, loop_stmt: Union[ast.For, ast.While]) -> None:
             self.source_of(loop_stmt),
         )
         for pv in loop_state_vars:
-            ov = self.py_var_to_onnx_var(pv, self.source_of(loop_stmt))
+            ov = self.py_var_to_onnx_var(pv, self.source_of(loop_stmt)).name
```
justinchuby, line 1254: nit: What are ov and pv? Possible to have better names for them?

justinchuby (approved):

onnxscript/converter.py
gramalingam: pv stands for python-variable and ov for the corresponding onnx-variable. Let me postpone this fix to another PR (there are more occurrences like this, and the CI takes too long to run again, so merged it for now).

PR #895, opened 2023-07-18
Temporarily rename package to onnxscript-preview during release | chore(release)

justinchuby:
Because onnxscript is owned. Also updated to use trusted publisher: https://docs.pypi.org/trusted-publishers/using-a-publisher/

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,6 +29,9 @@ jobs:
       - name: Install Python build dependencies
         run: |
           python -m pip install --upgrade pip build wheel
+      - name: Temporarily rename package to onnx-preview
+        run: |
+          sed -i 's/name = "onnxscript"/name = "onnxscript-preview"/' 'pyproject.toml'
       - name: Build ONNX Script wheel dev version
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
justinchuby, line 1: release ci did not run for some reason

jcwchen: It seems only working in the main branch and was skipped in other PRs as well. I am taking a closer look.

jcwchen: Added run release CIs label and update this PR (new commit or reopen this PR) should run this release pipeline.

justinchuby: Nice! Running now

jcwchen (approved):

jcwchen:
Do we need to cover more environments going forward? Now it only has Python 3.10 + ubuntu

liqunfu (approved):

justinchuby:
> Do we need to cover more environments going forward? Now it only has Python 3.10 + ubuntu

I think that's ok because we test in normal ci already?

PR #896, opened 2023-07-19
Cleanup translator

gramalingam:
Cleanup following aspects of the translator:

* `ConverterExpression` was being misused to cover 0, 1, or more ONNX variables created during translation. The resulting type-annotations were unclear and the code was fragile in a few places. It now is used to represent a single ONNX variable.
* The translation of multiple-variable assignments, like `x, y, z = op.SomeOp(...)` has been refactored.
* `translate_bool_op_expr` was not using the suggested target name (the original python variable name) for generated onnx variable. Fix this.
* Fix `translate_subscript_expr` so that it returns a `ConverterExpression`, rather than a string, for uniformity.

### DIFF
--- a/onnxscript/converter.py
+++ b/onnxscript/converter.py
@@ -93,8 +93,9 @@ class ConverterExpressionKind(IntEnum):
 
 
 class ConverterExpression:
-    def __init__(self, name: str, kind: ConverterExpressionKind):
-        assert isinstance(name, str)
+    def __init__(
+        self, name: str, kind: ConverterExpressionKind = ConverterExpressionKind.CONST
+    ):
         self.name = name
         self.kind = kind
 
@@ -102,7 +103,6 @@ class ConverterExpression:
         return self.kind == ConverterExpressionKind.CONST
 
     def __str__(self) -> str:
-        assert isinstance(self.name, str), "`name` is not a string. This is likely a bug."
         return self.name
 
 
@@ -529,7 +529,7 @@ class Converter:
         elif isinstance(node, (ast.BinOp, ast.BitAnd, ast.BitOr)):
             r = self.translate_bin_op_expr(node)
         elif isinstance(node, ast.BoolOp):
-            r = self.translate_bool_op_expr(node)
+            r = self.translate_bool_op_expr(node, target)
         elif isinstance(node, ast.UnaryOp):
             r = self.translate_unary_op_expr(node)
         elif isinstance(node, ast.Compare):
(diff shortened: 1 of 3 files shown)


onnxscript/converter.py
```diff
@@ -980,45 +983,60 @@ def translate_stmt(self, node: ast.stmt, index_of_stmt=None) -> None:
...
-
-                ids = [id(x) for x in lhs.elts]
-                onnxids = self.translate_expr(rhs, ids).name
-                for x, y in zip(ids, onnxids):
-                    self.bind(x, values.Dynamic(y, values.DynamicKind.Intermediate, info))
+                outputs = [generate_onnx_name(x) for x in lhs.elts]
+                self.emit(outputs, callee, inputs, attrs)
             else:
-                fail("Unsupported construct in LHS of assignment.")
+                self.fail(lhs, f"Unsupported construct in LHS of assignment: {type(lhs)!r}")
```
justinchuby, line 1022:
```suggestion
                self.fail(lhs, f"Unsupported construct in LHS of assignment: '{type(lhs)!r}'")
```

Perhaps quote this for clarity

onnxscript/converter_test.py
```diff
@@ -581,6 +581,19 @@ def duplicate_output(X):
...
         self.assertNotEqual(outputs[0], outputs[1])
 
+    def test_bool_op_name_generation(self):
+        """Verify that python variable name is used as onnx variable name (when possible)
+        for boolean operation translation.
+        """
+
+        @script(default_opset=op)
+        def bool_op(X, Y):
+            T = X and Y
```
justinchuby, line 591: python requires `and` return a bool I think. Does this work in eager mode?

justinchuby (approved):

onnxscript/converter_test.py
gramalingam: Good point. It doesn't. Perhaps this whole support should be removed, since it is not that useful. The only way to support it (in eager mode) would be to transform the AST itself, and recompiling. Doesn't seem worth it.

gramalingam: I will go ahead and merge the PR, so that this fixed version is in the git history. I will separately delete the support for AND operator (unless someone has any good reason to have it).

gramalingam: Just a couple more thoughts: The eager-mode does support it in the scalar case (i.e., a tensor of rank 0). Which is what is expected in if-conditions and while-conditions. But it won't handle the element-wise version for tensors (such as the condition in a Where op). So, there are some situations where it makes sense and can be supported. But not sure it is worth having the complications by supporting it. The other difference is the short-circuit evaluation, which seems to be the reason that Python doesn't support user-classes to override these boolean operators via some magic __method__. ONNX doesn't have any short-circuit evaluation (though a backend can choose to implement it that way).

justinchuby: I agree. Supporting special cases may end up causing more confusion.

Issue #786, opened 2023-06-15
Questions around _aten_native_batch_norm_inference_onnx


abock:
[As commented](https://github.com/microsoft/onnxscript/pull/773#pullrequestreview-1482443336) in #773, @gramalingam is curious what the intent of these lines are:

https://github.com/microsoft/onnxscript/blob/f106c1e906cb05ffc37491c9111402779e79de2d/onnxscript/function_libs/torch_lib/ops/core.py#L4435-L4438

justinchuby:
I think it is saying onnxscript wouldn't allow

```python
    ...
    empty = op.Cast(op.Shape(input, start=0, end=0), to=FLOAT.dtype) 
    return norm, empty, empty
```

Perhaps

```python
    ...
    empty_mean = op.Cast(op.Shape(input, start=0, end=0), to=FLOAT.dtype)
    empty_var = empty_mean
    return norm, empty_mean, empty_var
```

works? Or we can update the translator to account for this.

@xiaowuhu feel free to add or correct

xiaowuhu:
```
a = op.Cast(xxxx)
b = a
```
This lines cannot work in onnxscript, because they are actually one var.   ```b = op.Identity(a)``` can work.

gramalingam:
I understand the comment and the duplicated lines. What I am puzzled by is: what is the actual value it is trying to return? Why is the first dimension value being returned as a "mean" and "variance"? Something seems off. Why is it called "empty_*" ? Is it a case where the returned value doesn't matter? Then, why invoke "Shape" at all?

gramalingam:
May be it is trying to create a don't-care value of the given dimension? Then, the code is wrong, and needs an Expand perhaps.

xiaowuhu:
For the test case of aten_native_batch_norm(), one case's expected result should be:
```
[tensor([[[ -1.7358, ...0.4818]]]), tensor([]), tensor([])]
```
So we need make sure the first element ```tensor([[[ -1.7358, ...0.4818]]])``` is correct, it is correct.
But for the second and third element, they must be empty tensor. In onnxscript, I didn't find any other way to create a real empty tensor.

gramalingam:
By the way, the explicit call to `make_tensor` also specifies the exact output shape required. An empty tensor can have many different shapes, so may be good to be explicit about that too.

justinchuby:
[...]

Issue #860, opened 2023-07-11
Allow `ConstantOfShape` to take python constants


justinchuby:
Instead of writing 

```python
op.ConstantOfShape(
    new_shape, value=make_tensor("zerof", TensorProto.FLOAT, [1], [0])
)
```

it would be great to be able to write

```python
op.ConstantOfShape(
    new_shape, value=0.0)
)
```

cc @gramalingam

gramalingam:
Yes, that would be nice. It is best done by utilizing type information for attributes from the opschema, since attributes can be ints or floats or tensorsL: we don't want to promote a float-typed attribute into a tensor-attribute.

gramalingam:
I think this works now? Closing it, please feel free to reopen in case of problems.

PR #902, opened 2023-07-19
RefineOp(native-batch-norm) | feat(torchlib)

xiaowuhu:
Explain issue: https://github.com/microsoft/onnxscript/issues/786
If anybody have better way to create a real empty tensor in OnnxScript, please let me know.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4644,9 +4644,13 @@ def _aten_native_batch_norm_inference_onnx(
         momentum=momentum,
         training_mode=training,
     )
+
+    # For this function's test case, the returned mean and var must be empty tensors
+    # Seems we don't have any other way to create a real empty tensor, cannot use zero to simulate here
+    empty_int = op.Shape(input, start=0, end=0)
```
gramalingam, line 4650: I suggest
```py
   empty = op.Constant(value = onnx.helper.make_tensor("empty", onnx.TensorProto.FLOAT, [0], []))
```
That will also eliminate the need for the cast.

Unfortunately, there is no shortcut: currently, we can't use something like `[]` since onnxscript doesn't know what type to use. May be we can find some way to support it later.

justinchuby: Does constant(value_floats=[]) work?

justinchuby: Or do we need 0d

gramalingam: A 0d-tensor has size=1, can't be empty.

gramalingam: Right, `value_floats=[]` should work, but seems to not work ... seems to produce an int64 tensor, for some reason.

gramalingam: I think this may be due to a bug-feature in onnxruntime ... I remember something from a while back. Need to track it down.

xiaowuhu: > Does constant(value_floats=[]) work?

works in eager mode, but failed on check_model() in full graph mode.

xiaowuhu: Changed to:
```
op.Constant(value = onnx.helper.make_tensor("empty", onnx.TensorProto.FLOAT, [0], []))
```

justinchuby: I propose that we don’t introduce the onnx import and just wait for the ORT fix to use value_floats, if others don’t think differently.

gramalingam: I was mistaken. The issue is in onnxscript (I should have known). It is similar to the reason why plain [] doesn't work, the onnxscript translator doesn't know what (attribute) type to use (in the current implementation). If we use the context information (that is, the opschema info for the op being called, namely "Constant" here), we will be able to translate it correctly.

Anyway, feel free to not make any changes (as Justin suggests) for now, if you don't want to introduce the onnx import.

gramalingam: The issue is the same (more or less) as the one discussed here: https://github.com/microsoft/onnxscript/issues/860

xiaowuhu: OK. I will close this PR.

PR #903, opened 2023-07-19
Remove support for boolean operators 'and' and 'or'

gramalingam:
Removing support for boolean operators, as discussed in https://github.com/microsoft/onnxscript/pull/896 

This only removes the support for expressions like `X and Y`, which must instead be specified as `op.And(X, Y)`. The reason is that the shorthand notation cannot be supported in eager-mode easily. (It works for the scalar-case, but not for element-wise boolean ops on non-scalar tensors. Further, python uses short-circuit evaluation for these expressions, which will also cause a divergence in behavior with respect to the translated code.


justinchuby (approved):
Thanks!

PR #906, opened 2023-07-20
More annotations, cleanup, and comments

gramalingam:
Updates to autocast.py and converter.py.

### DIFF
--- a/onnxscript/autocast.py
+++ b/onnxscript/autocast.py
@@ -5,14 +5,17 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Optional, Sequence
 
 import numpy as np
 import onnx
 from onnx import helper, numpy_helper
 from onnx.defs import OpSchema
 
-from onnxscript import tensor, values
+from onnxscript import tensor
+
+if TYPE_CHECKING:
+    from onnxscript import converter
 
 # Conversions from python values to ONNX are used by both the script converter as well
 # as the eager-mode runtime and both need to be consistent. The script converter converts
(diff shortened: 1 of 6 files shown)


onnxscript/autocast.py
```diff
@@ -173,30 +176,33 @@ def get_type_info(x):
...
-            tmp = converter.generate_unique_name(f"{x.name}_cast")
-            converter.emit(
-                [tmp],
-                values.Op(converter.default_opset, "CastLike"),
-                [x.name, typeinfo],
-                [],
-            )
+        if x.is_castable and y is not None:
+            # Polymorphic constant x is cast to the type of y:
+            tmp = converter_.generate_unique_name(f"{x.name}_cast")
```
justinchuby, line 203: Consider a more meaningful name than tmp?

onnxscript/converter.py
```diff
@@ -180,22 +185,22 @@ def __init__(
...
 
         # States initialized by `init_function_translation`
-        self._outer = []
-        self._current_fn = None
-        self._nextvar = 0
-        self._used_vars = set()
+        self._outer: List[irbuilder.IRFunction] = []
+        self._current_fn: irbuilder.IRFunction = None
+        self._nextvar: int = 0
+        self._used_vars: Set[str] = set()
```
justinchuby, line 191:
```suggestion
        self._used_vars: set[str] = set()
```

justinchuby (approved):

onnxscript/converter.py
```diff
@@ -542,18 +545,18 @@ def translate_expr(
...
         if isinstance(r, tuple):
             callee, args, attrs = r
             target = "tmp" if target is None else target
             assert isinstance(target, str)
             result = self.generate_unique_name(target)
             self.emit([result], callee, args, attrs)
-            return ConverterExpression(result, ConverterExpressionKind.ANY)
-        return ConverterExpression(r, ConverterExpressionKind.ANY)
+            return Variable(result)
+        return Variable(r)
```
justinchuby, line 557: Looks like this line needs a test

gramalingam: Good catch, it is not needed any longer, since the translate_ method(s) that were returning a string have been updated to return a Variable.

onnxscript/converter.py
```diff
@@ -696,7 +696,7 @@ def translate_slice(slice_expr: ast.Slice) -> tuple[str, str, str]:
         if not (sliced_indices or scalar_indices or non_scalar_indices):
             # Edge case: no index specified. Eg. A[:, :]
             self.emit([target], "Identity", [var_name])
-            return ConverterExpression(target)
+            return Variable(target)
```
justinchuby, line 699: This line need a test too?

onnxscript/converter.py
```diff
@@ -880,13 +880,13 @@ def translate_compare_expr(self, node):
...
             not_op = values.Op(self.default_opset, "Not")
             return not_op, [tmp], []
 
         return op, [left, right], []
 
-    def translate_name_expr(self, node: ast.Name) -> ConverterExpression:
+    def translate_name_expr(self, node: ast.Name) -> Variable:
         return self.py_var_to_onnx_var(node.id, self.source_of(node))
 
     # pylint: disable=inconsistent-return-statements
```
justinchuby, line 892: not this pr but this needs to be re-enabled after this method block

PR #907, opened 2023-07-21
Add BOOL variants to bitwise ops | fix(torchlib)

justinchuby:
In torchbench we see errors like `[ONNXRuntimeError] : 10 : INVALID_GRAPH : Load model from bench_dynamo_onnx_model/attention_is_all_you_need_pytorch/model.onnx failed:This is an invalid model. Type Error: Type 'tensor(bool)' of input parameter (unsqueeze_1) of operator (BitwiseAnd) in node (n0__9) is invalid.`

This change fixes them by registering the Boolean variants from the exiting logical* implementations.

The reason for not creating new bitwise* variants is that the implementation is exactly the same as the logical* ones, which are covered by test cases. The bool bitwise* BOOL variants, if implemented separately, besides completely identical logic, would require test case duplication, which imo is not worth it.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -1010,15 +1010,9 @@ def aten_binomial(
 @torch_op(("aten::bitwise_and", "aten::bitwise_and.Tensor"))
 def aten_bitwise_and(self: TInt, other: TInt) -> TInt:
     """bitwise_and.Tensor(Tensor self, Tensor other) -> Tensor"""
+    # logical_and implements the BOOL variant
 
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1034,27 +1041,41 @@ def aten_bitwise_not_bool(self: BOOL) -> BOOL:
     return op.Not(self)
```
justinchuby, line 1041: bitwise not has an implementation already

titaiwangms (approved):

BowenBao:
Thank you for spotting and fixing this!

PR #908, opened 2023-07-23
AddOp(linalg_vector_norm) | feat(torchlib)

xiaowuhu:
Also updated the should_skip_xfail logic in test to account for data types.

## Notes

Have to skip one kind of test case: ord=6, dtype=float16.
In Pytorch:
```
>>> b
tensor([[2.3730, 0.9316, 0.6240, 6.1523, 0.1758],
        [5.3984, 7.9375, 4.9062, 1.8809, 6.1016],
        [4.0234, 8.2344, 3.2695, 0.8701, 1.3447]], dtype=torch.float16)
>>> la.vector_norm(a, dim=0, ord=6)
tensor([5.5483, 9.0838, 4.9754, 6.1532, 6.1017])
>>> la.vector_norm(b, dim=0, ord=6)
tensor([5.5469,    inf, 4.9727, 6.1523, 6.0977], dtype=torch.float16)
>>>
```

But in ORT, the result is:
```
tensor([5.5469,  9.08, 4.9727, 6.1523, 6.0977], dtype=loat16)
```

the second element ```9.08``` should be ```inf```. It works in eager mode, failed in FullGraph mode only.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/linalg.py
+++ b/onnxscript/function_libs/torch_lib/ops/linalg.py
@@ -13,16 +13,11 @@ from __future__ import annotations
 
 from typing import Optional, Sequence
 
-from onnxscript.onnx_types import TensorType
-
 from onnxscript import BOOL, FLOAT, INT64
 from onnxscript.function_libs.torch_lib.registration import torch_op
-from onnxscript.function_libs.torch_lib.tensor_typing import (
-    TFloat,
-)
+from onnxscript.function_libs.torch_lib.tensor_typing import TFloat
 from onnxscript.onnx_opset import opset18 as op
-
-import math
+from onnxscript.onnx_types import TensorType
 
 
 def aten_linalg_cholesky(self: TensorType, upper: bool = False) -> TensorType:
@@ -362,7 +357,9 @@ def _aten_linalg_vector_norm_no_dim_onnx(self: TFloat, ord: float, keepdim: bool
 
 
 @torch_op("aten::linalg_vector_norm", private=True)
-def _aten_linalg_vector_norm_onnx(self: TFloat, ord: float, dim: INT64, keepdim: bool) -> TFloat:
+def _aten_linalg_vector_norm_onnx(
+    self: TFloat, ord: float, dim: INT64, keepdim: bool
+) -> TFloat:
     self_rank = op.Size(op.Shape(self))
     if self_rank == 0:
         self = op.Unsqueeze(self, axes=[0])
--- a/onnxscript/function_libs/torch_lib/ops/linalg.py
+++ b/onnxscript/function_libs/torch_lib/ops/linalg.py
@@ -15,6 +15,11 @@ from typing import Optional, Sequence
 
 from onnxscript.onnx_types import TensorType
 
+from onnxscript.function_libs.torch_lib.registration import torch_op
+from onnxscript.onnx_opset import opset18 as op
+
+import math
+
 
 def aten_linalg_cholesky(self: TensorType, upper: bool = False) -> TensorType:
     """linalg_cholesky(Tensor self, *, bool upper=False) -> Tensor"""
@@ -305,6 +310,7 @@ def aten_linalg_vecdot(x: TensorType, y: TensorType, dim: int = -1) -> TensorTyp
     raise NotImplementedError()
 
 
+@torch_op("aten::linalg_vector_norm", trace_only=True)
 def aten_linalg_vector_norm(
     self: TensorType,
     ord: float = 2,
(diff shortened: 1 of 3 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -630,6 +640,17 @@ def _where_input_wrangler(
...
+    TorchLibOpInfo(
+        "linalg.vector_norm",
+        linalg_ops.aten_linalg_vector_norm,
+        trace_only=True,
+        tolerance={torch.float16: (2e-3, 2e-3)},
+        input_wrangler=_linalg_vector_norm_input_wrangler,
+    ).skip(
+        matcher=lambda sample: sample.kwargs.get("ord") == 6
+        and sample.input.dtype == torch.float16,
+        reason="ORT return wrong value for float16 with ord=6 (expected=Inf, actual=9.48).",
```
justinchuby, line 652: Looks like the number we get is actually more accurate. Is that true?

justinchuby: We can make it an xfail here potentially, so we know it when things get fixed

onnxscript/function_libs/torch_lib/ops/linalg.py
```diff
@@ -305,13 +309,78 @@ def aten_linalg_vecdot(x: TensorType, y: TensorType, dim: int = -1) -> TensorTyp
     raise NotImplementedError()
 
 
+@torch_op("aten::linalg_vector_norm", trace_only=True)
 def aten_linalg_vector_norm(
-    self: TensorType,
+    self: TFloat,
     ord: float = 2,
```
justinchuby, line 315:
```suggestion
    ord: float = 2.0,
```

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/linalg.py
justinchuby: This should fix the full graph error

xiaowuhu: Will use XFail.
try INT64 as input.

onnxscript/function_libs/torch_lib/ops/linalg.py
```diff
@@ -305,13 +309,78 @@ def aten_linalg_vecdot(x: TensorType, y: TensorType, dim: int = -1) -> TensorTyp
...
     dim: Optional[int] = None,
     keepdim: bool = False,
-    dtype: Optional[int] = None,
-) -> TensorType:
+    dtype: int = -1,
+) -> TFloat:
     """linalg_vector_norm(Tensor self, Scalar ord=2, int[1]? dim=None, bool keepdim=False, *, ScalarType? dtype=None) -> Tensor"""
 
-    raise NotImplementedError()
+    if dtype != -1:
```
titaiwangms, line 322: @justinchuby I am confused. Do we want the diaptcher to take care of dtype overload?

The dtype scenario should follow aten_argmax or not?

justinchuby: We do. But there's also the `dim: Optional[int] = None,` which can be split, in which case we will have 4 variants of the same function. Since the function is currently trace_only not because of the dtype situation here, I thought it's not worth it having 4 copies of this thing

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
titaiwangms: If ORT is more accurate, we can change the word "wrong value".

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -630,6 +640,17 @@ def _where_input_wrangler(
...
     TorchLibOpInfo("isposinf", core_ops.aten_isposinf),
+    TorchLibOpInfo(
+        "linalg.vector_norm",
+        linalg_ops.aten_linalg_vector_norm,
+        trace_only=True,
+        tolerance={torch.float16: (2e-3, 2e-3)},
+        input_wrangler=_linalg_vector_norm_input_wrangler,
+    ).skip(
+        matcher=lambda sample: sample.kwargs.get("ord") == 6
+        and sample.input.dtype == torch.float16,
```
justinchuby, line 651:
```suggestion
        matcher=lambda sample: sample.kwargs.get("ord") == 6,
        dtypes=[torch.float16],
```

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -630,6 +640,17 @@ def _where_input_wrangler(
     TorchLibOpInfo("isnan", core_ops.aten_isnan),
     TorchLibOpInfo("isneginf", core_ops.aten_isneginf),
     TorchLibOpInfo("isposinf", core_ops.aten_isposinf),
+    TorchLibOpInfo(
+        "linalg.vector_norm",
+        linalg_ops.aten_linalg_vector_norm,
+        trace_only=True,
+        tolerance={torch.float16: (2e-3, 2e-3)},
+        input_wrangler=_linalg_vector_norm_input_wrangler,
+    ).skip(
```
justinchuby, line 649:
```suggestion
    ).xfail(
```

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby, line 652:
```suggestion
        reason="ORT returns a more accurate value for float16 with ord=6 (expected=Inf, actual=9.48).",
```

justinchuby: Done

onnxscript/function_libs/torch_lib/ops/linalg.py
justinchuby: Never mind xfailing, because some cases will pass and thus fail the test

justinchuby:
I took the liberty to merge. Happy to follow up if there's more discuss desired

Issue #879, opened 2023-07-17
op.While cannot be inside op.If


xiaowuhu:
If the script looks like:
```
cond = ...
while cond:
   ...
```
It works well. But if the script looks like:
```
cond1 = ...
if cond1:
    cond2 = ...
    while cond2:
        ....
```
It cannot pass the stmt checking for the op.While().

cc @gramalingam

gramalingam:
I think this should work now? I will close it. Please reopen if there is any issue, thanks!

PR #909, opened 2023-07-24
AddOp(embedding bag) | feat(torchlib)

xiaowuhu:
- This PR is only for aten_embedding_bag function.
- There have 4 outputs for this function, we only care about the first one, for other 3, we just make the shape correct, with all zero values filled.
- aten_embedding_bag_padding_idx will be another PR.
- max_norm, I think this is rare case but not sure. If given, each embedding vector with norm larger than max_norm is renormalized to have norm max_norm. Note: this will modify weight in-place. Not sure if we need implement ```embedding_renorm``` function.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
     raise NotImplementedError()
 
 
+@torch_op("aten::embedding_bag", trace_only=True)
```
justinchuby, line 2263:
```suggestion
@torch_op(("aten::embedding_bag", "aten::_embedding_bag"), trace_only=True)
```

We can register the aten-private version as well because we happen to see them in actual models. It is essentially the same logic with contiguous input assumptions: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L2219

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
...
+    # Parse offsets=[0,1,3,3,4]
+    # When include_last_offset=False, means: [0:1],[1:3],[3:3],[3:4],[4:end]
+    # When include_last_offset=True, means: [0:1],[1:3],[3:3],[3:4]
+    len_tensor = op.Reshape(op.Size(offsets), neg_1)
+    if include_last_offset == True:  # pylint: disable=singleton-comparison
+        len_tensor = len_tensor - 1
+    else:
+        offsets = op.Concat(offsets, op.Shape(indices), axis=0)  # Replace end with number
+
+    # The element in sequence must be FLOAT32 dtype due to ORT bug
```
justinchuby, line 2314: Should we file a bug to ORT and link it here?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -883,6 +894,16 @@ def _where_input_wrangler(
...
     ),
+    TorchLibOpInfo(
+        "nn.functional.embedding_bag",
+        core_ops.aten_embedding_bag,
+        input_wrangler=_embedding_bag_input_wrangler,
+        trace_only=True,
+        tolerance={torch.float16: (1e-3, 1e-2)},
+    ).skip(
+        matcher=lambda sample: "padding_idx" in sample.kwargs or "max_norm" in sample.kwargs,
+        reason="padding_idx is used for training mode, and max_norm is rare case.",
```
justinchuby, line 905: Looks like max_norm is not part of the aten signature. I assume we can safely ignore them?

```suggestion
        reason="padding_idx or max_norm is not part of the aten signature.",
```

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
...
+    # Get weight out according to indices,
+    # e.g. indices=[0,1,2,4] means get weight[0],weight[1],weight[2],weight[4]
+    new_weight = op.Gather(weight, indices)
+    new_weight = op.Mul(new_weight, op.Unsqueeze(per_sample_weights, axes=1))
+    dim_1_size = op.Reshape(op.Gather(op.Shape(weight), 1), neg_1)
+    # Parse offsets=[0,1,3,3,4]
+    # When include_last_offset=False, means: [0:1],[1:3],[3:3],[3:4],[4:end]
+    # When include_last_offset=True, means: [0:1],[1:3],[3:3],[3:4]
+    len_tensor = op.Reshape(op.Size(offsets), neg_1)
+    if include_last_offset == True:  # pylint: disable=singleton-comparison
```
justinchuby, line 2309:
```suggestion
    if op.Equal(include_last_offset, True):
```

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: I wonder if we can skip trae_only and let dispatcher do the work.

BowenBao: And `aten::_embedding_bag_forward_only` as well.

BowenBao: Both aten private versions have `padding_idx=-1`. 

`aten::embedding_bag.padding_idx` also has this argument, but the default might be slightly different `int? padding_idx`.

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L2247

justinchuby: Thanks for the info! I’m inclined to suggest we just use -1 and ignore the None case for ease of implementation

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
BowenBao: https://github.com/pytorch/pytorch/pull/105862 shows very likely `max_norm` triggers an extra `aten.embedding_renorm.default` operator. 

Based on latest discussion, we want to exclude only cases when `padding_idx` is None or `max_norm` is not None.

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Already have this bug in ORT.

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: Do you mean padding_idx is not None?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
...
-    sparse: bool = False,
-    per_sample_weights: Optional[TensorType] = None,
+    weight: TFloat,
+    indices: INT64,
+    offsets: INT64 = None,  # Could be None accotding to the doc, go 2d branch
+    scale_grad_by_freq: bool = False,  # pylint: disable=unused-argument
+    mode: int = 0,  # [0,1,2] indicate ["mean", "sum", "max"]
+    sparse: bool = False,  # pylint: disable=unused-argument
+    per_sample_weights: Optional[TFloat] = None,
     include_last_offset: bool = False,
```
justinchuby, line 2272:
```suggestion
    include_last_offset: bool = False,
    padding_idx: int = -1
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
...
-) -> tuple[TensorType, TensorType, TensorType, TensorType]:
+) -> TFloat:
     """embedding_bag(Tensor weight, Tensor indices, Tensor offsets, bool scale_grad_by_freq=False, int mode=0, bool sparse=False, Tensor? per_sample_weights=None, bool include_last_offset=False) -> (Tensor, Tensor, Tensor, Tensor)"""
 
-    raise NotImplementedError()
+    if per_sample_weights is None:
+        per_sample_weights = op.Expand(op.Constant(value_floats=[1.0]), op.Shape(indices))
+        per_sample_weights = op.CastLike(per_sample_weights, weight)
+
+    if len(indices.shape) == 1:  # 1d
```
justinchuby, line 2280: We can change this to the normal way of computing rank then combine both functions into a script() one.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,107 @@ def aten_embedding_backward(
...
+    indices: INT64,
+    offsets: INT64 = None,  # Could be None accotding to the doc, go 2d branch
+    scale_grad_by_freq: bool = False,  # pylint: disable=unused-argument
+    mode: int = 0,  # [0,1,2] indicate ["mean", "sum", "max"]
+    sparse: bool = False,  # pylint: disable=unused-argument
+    per_sample_weights: Optional[TFloat] = None,
     include_last_offset: bool = False,
-) -> tuple[TensorType, TensorType, TensorType, TensorType]:
+) -> TFloat:
     """embedding_bag(Tensor weight, Tensor indices, Tensor offsets, bool scale_grad_by_freq=False, int mode=0, bool sparse=False, Tensor? per_sample_weights=None, bool include_last_offset=False) -> (Tensor, Tensor, Tensor, Tensor)"""
```
justinchuby, line 2274: @BowenBao The function returns four outputs. Do you know if they are all used?

BowenBao: Looks like unused. Would be best to e2e test it out w/ exporter.

https://github.com/pytorch/pytorch/blob/26e3b4020f01d4fc2b7f63e1de4c94d2c8b362b5/torch/nn/functional.py#L2422

https://github.com/pytorch/pytorch/blob/26e3b4020f01d4fc2b7f63e1de4c94d2c8b362b5/aten/src/ATen/native/EmbeddingBag.cpp#L1233

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
BowenBao: No, I thought after discussion the torchlib op signature expects `padding_idx` as int with default, so we only skip tests when `padding_idx` is explicitly None (this case may not exist)?

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Tried but failed, because ```op.While()``` cannot be inside ```op.If()```.

justinchuby: We should fix onnxscript for this

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
BowenBao: yes, `padding_idx` could be wrangled around. Semantically it doesn't affect forward so yea all cases should work.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: @gramalingam do you have an idea on what would be needed to make this work? Thanks!

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2260,19 +2260,117 @@ def aten_embedding_backward(
...
+    indices: INT64,
+    offsets: INT64 = None,  # Could be None accotding to the doc, go 2d branch
+    scale_grad_by_freq: bool = False,  # pylint: disable=unused-argument
+    mode: int = 1,  # [0,1,2] indicate ["sum", "mean", "max"], default is "mean"
+    sparse: bool = False,  # pylint: disable=unused-argument
+    per_sample_weights: Optional[TFloat] = None,
     include_last_offset: bool = False,
-) -> tuple[TensorType, TensorType, TensorType, TensorType]:
+    padding_idx: int = -1,  # pylint: disable=unused-argument
+) -> TFloat:
```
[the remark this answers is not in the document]
xiaowuhu: @justinchuby Do we still need this?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
xiaowuhu: - When ```padding_idx``` is None, the logic just be what I have implemented.
- When ```padding_idx``` has value, (such as  -1, 0, 1, or 2), it will affect the final result. So the ```-1``` cannot be default value.

onnxscript/function_libs/torch_lib/ops/core.py
gramalingam: Yes, I guess this is the same as the issue mentioned in https://github.com/microsoft/onnxscript/issues/879 ... will fix it, thanks

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
BowenBao: The analysis makes sense to me. I'm not certain. My impression on `-1` is only that it is the default value for `padding_idx` in `native_functions.yaml`. It depends on torch implementation regarding `padding_idx`. 

If we are not implementing `padding_idx` case, do we raise error when `padding_idx` is provided?

justinchuby: We tend to not raise in torchlib. This is a trace only function so it may be more relaxed, but I wouldn't rely on it. 

Let me check the -1 case.

justinchuby: OK so it looks like -1 doesn't have a special meaning (from https://github.com/pytorch/pytorch/blob/487ebcac3bc10b4b4b0631dafe2a12ddb0852f2a/torch/csrc/api/src/nn/modules/embedding.cpp#L105). In this case I suggest we do nothing to support `-1`, and add a `ValueError` for the time being, with a `TODO` next to it.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: #925 is merged. Could you merge from main and see if it works now?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
xiaowuhu: So far, from my try:
```
when padding_idx is not None(e.g. padding_idx=2 in test case 6):
    when in eager mode: actual == expected
    when in full graph mode: actual != expected
```
So it might be an ORT bug.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2266,19 +2266,144 @@ def aten_embedding_backward(
...
+    weight: TFloat,
+    indices: INT64,
+    offsets: INT64 = None,  # Could be None accotding to the doc, go 2d branch
+    scale_grad_by_freq: bool = False,  # pylint: disable=unused-argument
+    mode: int = 1,  # [0,1,2] indicate ["sum", "mean", "max"], default is "mean"
+    sparse: bool = False,  # pylint: disable=unused-argument
+    per_sample_weights: Optional[TFloat] = None,
     include_last_offset: bool = False,
-) -> tuple[TensorType, TensorType, TensorType, TensorType]:
+) -> TFloat:
```
justinchuby, line 2286:
```suggestion
) -> Tuple[TFloat, TFloat, TFloat, TFloat]:
```

PR #911, opened 2023-07-24
Mark tests as flaky and enable reruns | test(torchlib)

justinchuby:
We may still want to run tests even if they are flaky. This PR uses pytest-rerunfailures to rerun flaky tests so they are tested.

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/ops_test_common.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test_common.py
@@ -56,6 +56,8 @@ FLOAT_TYPES = (
 
 TEST_OPSET_VERSION = 18
 IS_WINDOWS = os.name == "nt"
+# The number of reruns for flaky tests
+FLAKY_RERUNS = 5
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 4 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -56,6 +56,8 @@
 
 TEST_OPSET_VERSION = 18
 IS_WINDOWS = os.name == "nt"
+# The number of reruns for flaky tests
+FLAKY_RERUNS = 5
```
titaiwangms, line 60: Would 5 be too much?

titaiwangms (approved):

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
justinchuby: Potentially? But we should be able to handle them with 2-3 reruns; I set to 5 just in case (following the example in the pytest-rerunfailures library). We can adjust this if the tests don't run as we want them to

BowenBao (approved):
This looks great! How do we know this works on flaky segfaults, i.e., segfaults don't crash entire test suite?

justinchuby:
> This looks great! How do we know this works on flaky segfaults, i.e., segfaults don't crash entire test suite?

This should handle cases when Python doesn't have a fatal error - i.e. when pytest-xdist is still able to recover. If the main Python process fails I don't think this will be able to handle it.

justinchuby:
Hmm, disappointing https://github.com/microsoft/onnxscript/actions/runs/5649751839/job/15304883084#step:5:16355

PR #915, opened 2023-07-24
Revert "Mark tests as flaky and enable reruns | test(torchlib)"

justinchuby:
Reverts microsoft/onnxscript#911

No rerun was triggered. Need more testing.


BowenBao (approved):
segfaults too strong

Issue #865, opened 2023-07-12
Annotate overload information on torchlib functions


justinchuby:

BowenBao:
List of unannotated ops found in diagnostics from onnx bench. @justinchuby @titaiwangms 

```
{'aten.add.Tensor': 30596,
 'aten.clamp.Tensor': 2330,
 'aten.div.Tensor': 9528,
 'aten.eq.Scalar': 72,
 'aten.eq.Tensor': 112,
 'aten.fill.Tensor': 28,
 'aten.ge.Scalar': 4,
 'aten.ge.Tensor': 4,
 'aten.gt.Scalar': 45,
 'aten.le.Tensor': 32,
 'aten.lt.Scalar': 80,
 'aten.mul.Tensor': 20459,
 'aten.ne.Scalar': 612,
 'aten.roll.default': 22,
 'aten.rsub.Scalar': 358,
 'aten.select.int': 4532,
 'aten.slice.Tensor': 15708,
 'aten.split.Tensor': 2331,
 'aten.sub.Tensor': 6945,
 'aten.transpose.int': 12068,
 'aten.unbind.int': 832,
 'aten.where.self': 578}
```

justinchuby:
aha! do you have a list you are working on? I didn't do anything yet

BowenBao:
https://github.com/microsoft/onnxscript/pull/919

BowenBao:
There is `aten::div.Tensor_Mode` left, which should be unsupported, but somehow worked with fallback.

BowenBao:
> There is `aten::div.Tensor_Mode` left, which should be unsupported, but somehow worked with fallback.

Tracked by https://github.com/pytorch/pytorch/issues/106057, along with other exporter dispatcher nearest match warnings.

titaiwangms:
> There is `aten::div.Tensor_Mode` left, which should be unsupported, but somehow worked with fallback.

Fallback seems reasonable, given the current implementation. However, there is one perfect match for this op in th sarif you provided in https://github.com/pytorch/pytorch/issues/106057. That seems like a big bug lol.

PR #919, opened 2023-07-25
Add overlooked overload information on torchlib functions

BowenBao:
* #920
* __->__ #919

Resolves most reported missing overloads from https://github.com/microsoft/onnxscript/issues/865

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -81,7 +81,7 @@ def aten_acosh(self: TFloat) -> TFloat:
     return op.Acosh(self)
 
 
-@torch_op("aten::add")
+@torch_op(("aten::add", "aten::add.Tensor"))
 def aten_add(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
     """add.Tensor(Tensor self, Tensor other, *, Scalar alpha=1) -> Tensor"""
     # TODO(microsoft/onnxruntime#15977): Improve fp16 precision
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4462,15 +4462,15 @@ def aten_msort(self: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::mul")
+@torch_op(("aten::mul", "aten::mul.Tensor"))
```
BowenBao, line 4465: note: tool didn't pick this up due to multiple impl.

titaiwangms:
Are these all of them? I can start working on refactor `get_aten_name` in onnx dispatcher.

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4883,7 +4883,7 @@ def aten_native_norm(self: TensorType, p: float = 2.0) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::ne")
+@torch_op(("aten::ne", "aten::ne.Scalar"))
```
justinchuby, line 4886: Missing Tensor?

BowenBao:
> Are these all of them? I can start working on refactor `get_aten_name` in onnx dispatcher.

Updating with latest bench log.
There is `aten::div.Tensor_Mode` left, which should be unsupported, but somehow worked with fallback.

onnxscript/function_libs/torch_lib/ops/core.py
BowenBao: Tensor wasn't observed by benchmark. I'll add here manually.

justinchuby:
Maybe the rounding mode didn't affect things much?

BowenBao:
> Maybe the rounding mode didn't affect things much?

Maybe. Maybe type promotion does the right thing for it. Maybe `rounding_mode` was not provided so the signature matched, and behavior also matched for default `true_divide`.

I think if `rounding_mode` is explicitly set, export will fail, as it should.

BowenBao:
@justinchuby  do you know if `CI / build_docs (windows-latest) (pull_request) ` failure is related?

justinchuby:
No. It started to fail these two days

PR #920, opened 2023-07-25
Tool to modify torchlib overload names via libcst

BowenBao:
* __->__ #920
* #919


justinchuby (dismissed):

onnxscript/function_libs/tools/torch_lib/modify_overload_names.py
```diff
@@ -0,0 +1,181 @@
...
+    source_tree.visit(op_overload_collector)
+    transformer = _TorchlibOpOverloadAdder(op_overload_collector._op_overloads, overload_names)
+    modified_tree = source_tree.visit(transformer)
+    module_path.write_text(modified_tree.code)
+    return transformer._results
+
+
+def main():
+    new_overload_names = {
+        "aten::add.Tensor",
```
justinchuby, line 148: Possible to add all overloads for these ops? I imagine `x for x in torch.ops.aten.add.overloads() if "out" not in x` etc.

BowenBao: Yes, the tool allows such usage. 

Frankly this PR is libcst exercise lol. I think the only useful scenario for this tool is when we have a large batch of missing overloads that we know would work, typically from model benchmark.

onnxscript/function_libs/tools/torch_lib/modify_overload_names.py
```diff
@@ -0,0 +1,214 @@
...
+        self._op_overloads: Dict[str, List[Tuple[str, List[str]]]] = {}
+        self._stack: List[str] = []
+
+    def visit_FunctionDef(self, node: FunctionDef) -> bool | None:
+        self._stack.append(node.name.value)
+
+    def leave_FunctionDef(self, node: FunctionDef) -> None:
+        self._stack.pop()
+
+    def visit_Call(self, node: cst.Call) -> None:
```
justinchuby, line 60: So the function def is visited before the decorator call?

BowenBao: I think so, from cst tree level, decorators are a part of the function.

PR #923, opened 2023-07-26
Use type-directed translation of attribute values

gramalingam:
Use the attribute-type from opschema, when available, to handle attribute-values during translation, as well as in eager-mode.

This enables constructs like `Constant(value_ints=[])`, `Constant(value=0)`, `Constant(value=[1,-1])`. 

See https://github.com/microsoft/onnxscript/issues/860 and https://github.com/microsoft/onnxscript/issues/786


onnxscript/converter.py
```diff
@@ -481,10 +496,17 @@ def translate_attr(
...
+
         if isinstance(expr, ast.Name):
             val = self.lookup(expr.id, self.source_of(expr))
             if isinstance(val, values.AttrRef):
-                return self.ir_builder.make_attr_ref(attr_name, val.value, val.typeinfo)
+                attr_ref = self.ir_builder.make_attr_ref(attr_name, val.value, val.typeinfo)
+                if attr_meta and (attr_ref.type != attr_meta.type):
+                    self.fail(
+                        expr,
+                        f"Attribute type {attr_ref.type} does not match expected type {attr_meta.type}",
```
justinchuby, line 507: Quote user inputs for clarity?

```suggestion
                        f"Attribute type '{attr_ref.type} 'does not match expected type '{attr_meta.type}'",
```

justinchuby, line 507: Potentially also specify where the “expected type” came from? E.g. “specified by the function definition”

At some point we may be able to produce errors that also knows the source location?

onnxscript/converter.py
```diff
@@ -508,8 +530,17 @@ def translate_attr(
         # The caller is responsible for omitting such attribute-values from the list of attributes
         # in a NodeProto.
         if val is None:
+            if attr_meta and attr_meta.required:
+                self.fail(expr, "Attribute {attr_name} is required.")
```
justinchuby, line 534: Quote user inputs like the above

onnxscript/converter.py
```diff
@@ -902,6 +936,7 @@ def translate_opset_expr(self, node: ast.Attribute) -> values.Opset:
         else:
             self.fail(node, "Invalid opset expression.")
 
+    # pylint: enable=inconsistent-return-statements
```
justinchuby, line 939: Be sure to enable it afterwards? Or move to the file top to disable it for the whole file

onnxscript/converter_test.py
```diff
@@ -607,6 +607,54 @@ def if_then_else_expanded(flag: typing.List[bool], Y, Z):
...
+
+        expected = np.array([], dtype=np.float32)
+        self.check_run(empty_floats, [], expected)
+
+    def test_int_as_tensor_attribute(self):
+        @script()
+        def int_as_tensor():
+            return op.Constant(value=17)
+
+        expected = np.array([17], dtype=np.int64).reshape(())
```
justinchuby, line 631:
```suggestion
        expected = np.array(17, dtype=np.int64)
```

is this possible?

onnxscript/converter.py
gramalingam: It does show source-location (from the source-location of first-argument `expr` of `self.fail`).

onnxscript/converter.py
justinchuby: My bad :)

justinchuby:
It started failing earlier this week. Looks like a dependency just updated and broke it

onnxscript/converter.py
```diff
@@ -508,8 +530,17 @@ def translate_attr(
         # The caller is responsible for omitting such attribute-values from the list of attributes
         # in a NodeProto.
         if val is None:
+            if attr_meta and attr_meta.required:
```
justinchuby, line 533: Should we test the error cases too?

onnxscript/evaluator.py
```diff
@@ -432,13 +432,25 @@ def _call_ort(
     args = [_os_to_ort_value(x) for x in args]
     implicit_args = {k: _os_to_ort_value(v) for k, v in implicit_args.items()}
 
+    # Utility to convert kwarg to ONNX AttributeProto:
+    def make_attr(key: str, value: Any) -> onnx.AttributeProto:
```
justinchuby, line 436:
```suggestion
    def make_attr(key: str, value: Any) -> onnx.AttributeProto:
    """Utility to convert kwarg to ONNX AttributeProto."""
```

justinchuby, line 436: Also lift the function to module and mark private to reduce nested functions?

justinchuby (approved):

onnxscript/_internal/autocast.py
```diff
@@ -64,6 +64,49 @@ def pyvalue_to_onnx_tensor(tensor_name: str, pyvalue):
...
+        onnx.AttributeProto.TYPE_PROTOS,
+    ]
+)
+
+
+def pyvalue_to_onnx_attribute(
+    key: str,
+    value: Any,
+    name_generator: Callable[[], str],
+    attr_type: Optional[onnx.AttributeProto.AttributeType] = None,
```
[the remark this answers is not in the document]
gramalingam: Not yet. Currently, we don't stop translation if we encounter an unknown op, eg: `op.Unknown(foo=5)`. If we decide to give that up, we can make the attr_type required (if desired). I am not sure we should give that up at this point.

PR #925, opened 2023-07-26
Handle loops in analysis of assigned vars

gramalingam:
The analysis function "assigned_vars" was not handling loops. Fix this.

Should fix issue https://github.com/microsoft/onnxscript/issues/879

See also PR https://github.com/microsoft/onnxscript/pull/909

### DIFF
--- a/onnxscript/analysis.py
+++ b/onnxscript/analysis.py
@@ -51,7 +51,7 @@ def local_defs(lhs: ast.expr) -> Set[str]:
     return {get_id(lhs)}
 
 
-def defs(stmt: ast.stmt) -> Set[str]:
+def defs(stmt: ast.stmt | list[ast.stmt], formatter: sourceinfo.Formatter) -> Set[str]:
     """Return the set of all variables that may be defined (assigned to) in an
     execution of input stmt.
     """
@@ -59,7 +59,7 @@ def defs(stmt: ast.stmt) -> Set[str]:
     def block_defs(block: Sequence[ast.stmt]) -> Set[str]:
         result: set[Any] = set()
         for s in block:
-            result = result | defs(s)
+            result = result | defs(s, formatter)
         return result
 
     if isinstance(stmt, ast.Assign):
@@ -70,12 +70,18 @@ def defs(stmt: ast.stmt) -> Set[str]:
         return set()
     if isinstance(stmt, ast.If):
         return block_defs(stmt.body) | block_defs(stmt.orelse)
+    if isinstance(stmt, ast.For):
+        return block_defs(stmt.body) | {get_loop_var(stmt, formatter)}
+    if isinstance(stmt, ast.While):
+        return block_defs(stmt.body)
     if isinstance(stmt, list):
         return block_defs(stmt)
     if isinstance(stmt, ast.Break):
         return set()
     if ast_utils.is_print_call(stmt):
         return set()
+    if isinstance(stmt, list):
+        return block_defs(stmt)
     raise ValueError(f"Unsupported statement type {type(stmt)!r}.")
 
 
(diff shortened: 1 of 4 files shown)


onnxscript/analysis.py
```diff
@@ -51,31 +49,39 @@ def get_id(e):
...
+        return assigned_in_block(stmt.body)
     if isinstance(stmt, list):
-        return block_defs(stmt)
+        return assigned_in_block(stmt)
     if isinstance(stmt, ast.Break):
         return set()
     if ast_utils.is_print_call(stmt):
         return set()
+    if isinstance(stmt, list):
+        return assigned_in_block(stmt)
```
justinchuby, line 84: Should we test this too?

onnxscript/analysis_test.py
```diff
@@ -163,5 +163,57 @@ def nested():  # pylint: disable=unused-variable
         self.assertUses(f, {"x"})
 
 
+class TestAssignedVarAnalysis(unittest.TestCase):
+    def assertDefs(self, f, expected: set[str]):
+        source, parse_tree = ast_utils.get_src_and_ast(f)
+        result = analysis.assigned_vars(parse_tree.body, formatter(source))
+        self.assertEqual(result, expected)
+
```
justinchuby, line 171: I would lift this out and make sure what is being asserted is clear from its name

```suggestion
def assert_defined_variables(self, f, expected: set[str]):
    source, parse_tree = ast_utils.get_src_and_ast(f)
    result = analysis.assigned_vars(parse_tree.body, formatter(source))
    self.assertEqual(result, expected)


class TestAssignedVarAnalysis(unittest.TestCase):

```

onnxscript/converter.py
```diff
@@ -1102,9 +1102,11 @@ def ret(exp, i, suffix):
 
     def translate_if_stmt(self, stmt: ast.If) -> None:
         if hasattr(stmt, "live_out"):
-            live_defs = list(stmt.live_out.intersection(analysis.defs(stmt)))
+            live_defs = list(
+                stmt.live_out.intersection(analysis.assigned_vars(stmt, self.message))
+            )
         else:
-            live_defs = list(analysis.defs(stmt))
+            live_defs = list(analysis.assigned_vars(stmt, self.message))
```
justinchuby, line 1109: Would be good to explain this logic with a comment?

onnxscript/analysis.py
gramalingam: It is tested by every test for this function, since `parse_tree.body` is a list of statements. In fact, this extension exists primarily for enabling the unit tests.

onnxscript/analysis_test.py
gramalingam: Renamed it. But leaving it inside the class ... that seems preferable to me.

onnxscript/converter.py
gramalingam: Good question. I didn't add it, and it seems questionable to me. I need to check it out, I suspect it was added in some case where the liveness analysis had not run for some reason. but the fix should probably me something else.

onnxscript/analysis.py
justinchuby: hmm strange. 
image

coverage shows only this line is not tested

justinchuby: Let me print some things here to see

onnxscript/analysis_test.py
justinchuby: ok

onnxscript/analysis.py
justinchuby: Duplicate of line 77?

justinchuby (approved):
LGTM w/ duplicated logic removed

onnxscript/analysis.py
gramalingam: Good catch, thanks!

PR #927, opened 2023-07-27
Remove outdated skips | test

justinchuby:
Remove outdated skips in function tests which were skipping ONNX 1.13 or torch 1.13.


justinchuby:
I cannot reproduce the errors locally. Strange.

jcwchen (commented):
> I cannot reproduce the errors locally. Strange.

The error happens due to output mismatch. Would it be related to NumPy verison or different platform?

justinchuby:
Bumping some versions to check

gramalingam (approved):

Issue #493, opened 2023-03-02
TorchScript Graph: Support model > 2GB


BowenBao:
Dependent on `torch.Graph._export_onnx`, large model export requires specifying a target path to store the model and weights proto. Right now this path is fixed as "".

https://github.com/microsoft/onnx-script/blob/4986a4be9f8ff1b0e2b7bdde738836574f90e900/onnxscript/function_libs/torch_aten/graph_building.py#L475

justinchuby:
cc @shubhambhokare1

BowenBao:
we do but we need a str path provided. A tedious restriction that we want to eliminate.

justinchuby:
Will we be able to provide an option in `to_model_proto` to leverage what we already have in torch for this?

justinchuby:
Would you like to share some of the options?

BowenBao:
I'm thinking of model / weight separation by default.

wschin:
Can someone please assign this to @wschin? I have a version for demo.

justinchuby:
One thing I think we can do is to save the model out to disk with torchscript, then load it back and pass the model proto onto the exporter for it to decide how to serialize. When we move away from using torchscript this will not be a problem anymore

PR #929, opened 2023-07-27
Pass folder path to store initializers as files

wschin:
One solution for https://github.com/microsoft/onnxscript/issues/493. It requires small PyTorch exporter change to passing the newly added parameter from user code -> pytorch-to-onnx exporter -> onnxscript.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import typing
 import warnings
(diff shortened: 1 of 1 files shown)


BowenBao:
I'm thinking about making the api to return model_proto and initializers separately. It is similar to what you did with `symbolic_trace`. The model_proto and initializers are only written to disk on the `save`/`serialize` call.

```python
# model_proto frame and initializers are still in memory, inside export_output.
export_output = torch.onnx.dynamo_export(model, ...)
# Now we have the path, serialize model to there.
export_output.save("path_to_model.onnx")
```

justinchuby:
Added comments in #493.

PR #930, opened 2023-07-27
Temporary solution for demo | chore(torchlib)

### LINKED ISSUE #493: TorchScript Graph: Support model > 2GB

wschin:
Hack for #493 for demo. Will need a more proper solution.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -66,6 +66,12 @@ jobs:
           CATCH_ORT_SEGFAULT: "${{ matrix.os == 'ubuntu-latest' && '1' || '0' }}"
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+      - name: Upload Test Results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: dynamo-report (${{ matrix.name }}-${{ matrix.os }})
+          path: report_dynamo_export.sarif
 
   build_docs:
     strategy:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -66,12 +66,6 @@ jobs:
           CATCH_ORT_SEGFAULT: "${{ matrix.os == 'ubuntu-latest' && '1' || '0' }}"
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
-      - name: Upload Test Results
-        if: always()
-        uses: actions/upload-artifact@v3
-        with:
-          name: dynamo-report (${{ matrix.name }}-${{ matrix.os }})
-          path: report_dynamo_export.sarif
 
   build_docs:
     strategy:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -60,12 +60,19 @@ jobs:
       - name: Install nox
         run: python -m pip install nox
       - name: Run tests
-        run: nox -t ${{ matrix.nox-tag }} --forcecolor -- -v --cov=onnxscript --cov-report=xml --cov-append --cov-branch -n=auto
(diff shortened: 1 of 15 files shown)


justinchuby (approved):

justinchuby:
Merging for demo

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -698,7 +699,11 @@ def to_model_proto(
             keep_initializers_as_inputs=False,
             custom_opsets={},
             add_node_names=True,
-            onnx_file_path="",
+            # TODO(wechi): Passing in this instead of reading from env.
```
justinchuby, line 702:
```suggestion
            # TODO(#493): Passing in this instead of reading from env.
```

onnxscript/function_libs/torch_lib/graph_building_test.py
```diff
@@ -139,5 +141,44 @@ def test_add_initializer_allows_adding_the_same_tensor_twice_using_same_name(sel
...
+        # # of model parameters:
+        #  input_size x hidden_size + hidden_size +
+        #  hidden_size x output_size + output_size
+        # Thus, we have
+        #  (4 x 50000000 + 50000000 = 250000000) +
+        #  (50000000 x 10 + 10 = 500000010)
+        #  = 750000010 parameters
+        #  = 750000010 * 4 bytes
+        #  = 3 GB
+        batch_size, input_size, hidden_size, output_size = 1, 4, 50000000, 10
```
justinchuby, line 168: Would this consume a lot of memory? We don't have much in the CI machines. Maybe skip this in CI too?

onnxscript/function_libs/torch_lib/graph_building.py
wschin: Ok.

onnxscript/function_libs/torch_lib/graph_building_test.py
wschin: I will change the size of this test to `small`. Although it won't test the actual functionality, it can still test the code to fetch env variable.

.github/workflows/main.yaml
```diff
@@ -66,6 +66,12 @@ jobs:
           CATCH_ORT_SEGFAULT: "${{ matrix.os == 'ubuntu-latest' && '1' || '0' }}"
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+      - name: Upload Test Results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: dynamo-report (${{ matrix.name }}-${{ matrix.os }})
+          path: report_dynamo_export.sarif
```
BowenBao, line 74: Is this intended change? looks unrelated to this pr

noxfile.py
```diff
@@ -14,6 +14,7 @@
     "numpy==1.23.5",
     "typing_extensions",
     "beartype",
+    "tabulate",
```
BowenBao, line 17: same here

.github/workflows/main.yaml
wschin: Reverted. Basically, dynamo_export call triggers several exceptions in CI but no call stack or details were provided. I will revert them and create another issue.

wschin: Tracked by #934.

wschin:
@justinchuby, I disable that test in CI because CI is not ready. That new line is always run by other to_model_proto calls, so it should be fine.

PR #936, opened 2023-07-28
Handle docstring in analyses

gramalingam:
Handle doc-string in a couple of analyses where it wasn't being handled.


onnxscript/analysis_test.py
```diff
@@ -162,6 +175,16 @@ def nested():  # pylint: disable=unused-variable
 
         self.assertUses(f, {"x"})
 
+    def test_doc_string(self):
+        def f(x):
```
justinchuby, line 179: I remember

```
@script()
def foo():

    @graph()
    def foo_graph():
        """Docstring"""
        ...
```

would cause an error. Does this fix handles it too?

gramalingam: I hope so. At least, I tried an example like above, and it works.

justinchuby: Possible to add a nested function def and see?

justinchuby: So just

```
def foo():
    def foo_graph():
        """Docstring"""
        pass
```

justinchuby: nested python function def I meant

PR #942, opened 2023-07-31
Create publish pipelines in Azure | chore(release)

justinchuby:
I created two pipelines. They should be triggered by the individual github workflows as a second step.

### DIFF
--- a/.azure-pipelines/release-dev.yml
+++ b/.azure-pipelines/release-dev.yml
@@ -1,6 +1,8 @@
 # Build the dev version of the package and publish to artifacts
 
-trigger: none
+trigger:
+  - main
+  - releases/*
 
 pool:
   vmImage: ubuntu-latest
--- a/.azure-pipelines/release-dev.yml
+++ b/.azure-pipelines/release-dev.yml
@@ -1,0 +1,40 @@
+trigger: none
+
+pool:
+  vmImage: ubuntu-latest
+steps:
+  - displayName: 'Use Python 3.10'
+    task: UsePythonVersion@0
+    inputs:
+      versionSpec: '3.10'
+  - displayName: Install Python build dependencies
+    script: |
+      python -m pip install --upgrade pip build wheel
+  - displayName: Temporarily rename package to onnx-preview
+    script: |
+      sed -i 's/name = "onnxscript"/name = "onnxscript-preview"/' 'pyproject.toml'
+  - displayName: Build ONNX Script wheel dev version
+    script: |
+      python -m build
+  - task: CopyFiles@2
+    displayName: 'Copy Python Wheel to: $(Build.ArtifactStagingDirectory)'
+    inputs:
+      SourceFolder: 'dist'
+      Contents: '*.whl'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)'
+  # - task: PublishBuildArtifacts@1
+  #   displayName: 'Publish onnxscript'
+  #   inputs:
+  #     ArtifactName: onnxscript
+
+  # Test the wheels. This needs to happen after PublishBuildArtifacts
+  # to avoid interference with the artifacts
+  - name: Install Python dependencies
+    script: |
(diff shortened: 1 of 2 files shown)


.azure-pipelines/release-dev.yml
```diff
@@ -0,0 +1,42 @@
...
+  - displayName: Build ONNX Script wheel dev version
+    script: |
+      python -m build
+  - task: CopyFiles@2
+    displayName: 'Copy Python Wheel to: $(Build.ArtifactStagingDirectory)'
+    inputs:
+      SourceFolder: 'dist'
+      Contents: '*.whl'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)'
+  # - task: PublishBuildArtifacts@1
```
justinchuby, line 27: To uncomment when things are set up

.azure-pipelines/release-dev.yml
```diff
@@ -0,0 +1,42 @@
...
+trigger: none
+
+pool:
+  vmImage: ubuntu-latest
+steps:
+  - displayName: 'Use Python 3.10'
+    task: UsePythonVersion@0
+    inputs:
+      versionSpec: '3.10'
+  - displayName: Install Python build dependencies
```
justinchuby, line 12: Should we move the reusable bits into a common template? I don't know how to do that.

jcwchen: Sounds good. That's doable in AzurePipelines: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops&pivots=templates-includes.

A while ago it's not doable in GitHub Action, but it may now.

justinchuby: I just want them to work before refining. Maybe merge the current thing and create the pipelines first?

.azure-pipelines/release-dev.yml
```diff
@@ -0,0 +1,42 @@
...
+
+trigger: none
+
+pool:
+  vmImage: ubuntu-latest
+steps:
+  - displayName: 'Use Python 3.10'
+    task: UsePythonVersion@0
+    inputs:
+      versionSpec: '3.10'
```
jcwchen, line 11: Not relevant to this PR: When shall we start to use Python 3.11 in onnxscript if there is no obvious blocker?

.azure-pipelines/release.yml
```diff
@@ -0,0 +1,45 @@
...
+      SourceFolder: 'dist'
+      Contents: '*.whl'
+      TargetFolder: '$(Build.ArtifactStagingDirectory)'
+  # - task: PublishBuildArtifacts@1
+  #   displayName: 'Publish onnxscript'
+  #   inputs:
+  #     ArtifactName: onnxscript
+
+  # Test the wheels. This needs to happen after PublishBuildArtifacts
+  # to avoid interference with the artifacts
```
jcwchen, line 36: That might make sense for dev wheels, but for released wheel, it seems safer that we require the built wheel to pass all tests and then do upload?

.azure-pipelines/release-dev.yml
jcwchen: Yes. Let's forward this PR first

jcwchen (approved):

.azure-pipelines/release.yml
justinchuby: Sure. I will make the change in a separate PR.

.azure-pipelines/release-dev.yml
justinchuby: I guess now?

.azure-pipelines/release.yml
justinchuby: We can trigger the release if the pipeline succeeds

.azure-pipelines/release-dev.yml
justinchuby: Done

jcwchen (approved):

PR #954, opened 2023-08-01
Disable release triggers on Github | chore(release)

justinchuby:
We will use Azure Dev Ops for compliant releases


jcwchen (approved):
Why not remove it entirely? For backup purpose?

Issue #934, opened 2023-07-28
dynamo_export fails without call stack and details in CI


wschin:
When developing #930, `dynamo_export` hit some errors but CI didn't provide details to debug or locate the cause. We need to at least print the call stack or upload the diagnostic to artifact when exception occurs.

cc @BowenBao , @justinchuby

BowenBao:
Could you share an example job link? thx

BowenBao:
```
2023-07-28T00:48:46.9368021Z [gw0][36m [ 98%] [0m[32mPASSED[0m onnxscript/tests/function_libs/torch_lib/ops_test.py::TestOutputConsistencyFullGraphCPU::test_output_match_opinfo__full_like_dtype_cpu_float16 
2023-07-28T00:49:30.1877854Z ##[error]The operation was canceled.
2023-07-28T00:49:30.2008818Z Post job cleanup.
```

It shows the job was canceled. Do we know this is a test error?

justinchuby:
When there's a new push the old jobs will be canceled

wschin:
Check this one: https://[pipelines.actions.githubusercontent.com/serviceHosts/18101f61-82cd-4f0f-b124-60b829998504/_apis/pipelines/1/runs/13947/signedlogcontent/24?urlExpires=2023-07-28T20%3A02%3A38.4940021Z&urlSigningMethod=HMACV1&urlSignature=WVhqf%2F2dbYkY5eeHUO8iHgBv8lvuupTiXeb%2FWSzJMso%3D](https://pipelines.actions.githubusercontent.com/serviceHosts/18101f61-82cd-4f0f-b124-60b829998504/_apis/pipelines/1/runs/13947/signedlogcontent/24?urlExpires=2023-07-28T20%3A02%3A38.4940021Z&urlSigningMethod=HMACV1&urlSignature=WVhqf%2F2dbYkY5eeHUO8iHgBv8lvuupTiXeb%2FWSzJMso%3D)

Should see this error:

2023-07-28T00:31:01.9080429Z [31mFAILED[0m onnxscript/function_libs/torch_lib/graph_building_test.py::[1mTestLargeModelSave::test_save_initializer_to_files_for_large_model[0m - torch.onnx.OnnxExporterError: Failed to export the model to ONNX. Generating SARIF report at report_dynamo_export.sarif. Please report a bug on PyTorch Github: https://github.com/pytorch/pytorch/issues
[...]

PR #963, opened 2023-08-04
Reorder pytest results | test

justinchuby:
`fEsX` -> `sfEX` to move failures to the bottom of the pytest error summary. I guess the order matters?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ onnx = ["py.typed"]
 
 [tool.pytest.ini_options]
 filterwarnings = ["ignore::UserWarning", "ignore::DeprecationWarning"]
-addopts = "-rfEsX --tb=short --color=yes"
+addopts = "-rsfEX --tb=short --color=yes"
 
 [tool.mypy]
 follow_imports = "silent"   # TODO: Remove when we fix all the mypy errors


jcwchen (approved):

BowenBao (approved):
```
This change alters the pytest configuration in terms of how pytest behaves during testing.

In the pytest command line arguments:

- `-r` is used to display short test summary info at the end of test session for specified info.

- The character following `-r` specifies the type of summary information to be included:

    - `s`: skipped tests
    - `f`: failed tests
    - `E`: tests with error
    - `X`: xfailed tests (expected failure)
   
- `-rfEsX` would therefore report on failed, error, and expected failure tests.

- `-rsfEX` will report on skipped, failed, error, and expected failure tests.

- `--tb=short` specifies that traceback output should be in the "short" format.

- `--color=yes` enables colored output.

So, the difference between `-rfEsX` and `-rsfEX` is that the former does not show reports for skipped tests, while the latter does.
```

In chatgpt we trust. (Ignoring the last line which is clearly a mistake)

I'm not sure how much it helps with https://github.com/microsoft/onnxscript/issues/934, iiuc this changes the ordering inside 'summary', the detailed failure is still way up there.

PR #967, opened 2023-08-04
Include the textual model representation in error reports | test(torchlib)

justinchuby:
For better readability, we include a section in the error report that is the textual representation of the ONNX model. We cannot use it programmatically instead of textproto because it does not have initializers and would be invalid sometimes. Textproto is always the exact ONNX proto.

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/error_reproduction.py
+++ b/onnxscript/tests/function_libs/torch_lib/error_reproduction.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import logging
 
 import pathlib
 import platform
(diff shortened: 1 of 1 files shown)


BowenBao (commented):
Wouldn't it be repetitive to include both repro code with model proto encoded and now textproto?

justinchuby:
The text repr is in the markdown. It is easier to read than the proto. The proto is more correct as explained. The text repr is not always identical as the actual model.

BowenBao (approved):
Ok updated my understanding of terminology.

- textproto: `onnx_model_text = str(onnx_model)` used in repro code.
- text repr: `onnx.printer.to_text(onnx_model)` added in this PR.

Hence my previous question should be rephrased as

> Wouldn't it be repetitive to include both repro code with textproto and text repr in the markdown?

Guess answer is no.

justinchuby:
Right it wouldn't be repetitive (as explained in the pr description).

PR #971, opened 2023-08-04
Enable tests for `upsample_nearest2d` | test(torchlib)

justinchuby:
* #979
* #978
* #977
* #976
* #973
* #972
* __->__ #971


BowenBao (approved):
but how is it possible to remove an xfail without other code changes ... is it overshadowed by another skip/xfail?

justinchuby:
That is a very good question. Let me run more tests

justinchuby:
OK. So the xfail guards _any_ subtests failing. Here we have another xfail that checks for specific subtests. Thus we aways have some failures and this big xfail was not triggered.

PR #992, opened 2023-08-08
make shape inference optional due to segfault

wschin:
Although `onnx.shape_inference.infer_shapes` is put into try-except block, segfault still causes the entire program crashes. Making shape inference optional will be helpful if user hits segfault problem.

A segfault from PyTorch CI of my [PR](https://github.com/pytorch/pytorch/pull/106589) after upgrading onnxruntime=1.15.0 to onnxruntime-training=1.15.1,
image

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -1,10 +1,8 @@
 """Graph building functions for torchscript graph backend."""
 from __future__ import annotations
 
-import logging
 import os
 import typing
-import warnings
 from typing import Any, Dict, Final, List, Mapping, Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -672,7 +670,7 @@ class TorchScriptGraph:
 
     @runtime_typing.checked
     def to_model_proto(
-        self, opset_version: int, include_initializers: bool = True
+        self, opset_version: int, include_initializers: bool = True, infer_shapes: bool = True
     ) -> onnx.ModelProto:
         function_proto_dict: Mapping[
             Tuple[str, str], onnx.FunctionProto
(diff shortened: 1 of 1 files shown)


justinchuby:
If onnx segfaults i think we should fix onnx. I am a little worried that we are starting to grow the to_model_proto method. Ideally it mirrors onnxscript’s to_model_proto method which does not provide many options. 

Which version of ONNX is segfaulting? Could you share environment info? Thanks!

justinchuby (changes requested):
Blocking for now. Thanks

wschin:
This is one example --- https://github.com/pytorch/pytorch/actions/runs/5801748185/job/15727346375

I installed onnx-weekly first, then installed onnxruntime-training. Then my test hits segfault in to_model_proto. I guess segfault happens because onnxruntime-training overrides onnx-weekly with an earlier onnx.

justinchuby:
I see. That could be possible. cc @jcwchen just in case this is a new segfault.

@wschin could you share the ONNX files so we can test that to make sure?

wschin:
I guess it's related to onnx function. Maybe exporting this model with `dynamo_export` can just repro
```
        example_args_collection = tuple(
            (torch.randn(batch, 2, dtype=torch.float32),) for batch in (1, 2, 4, 6, 8)
        )

        class MLP(nn.Module):
            def __init__(self):
                super().__init__()
                self.fc1 = nn.Linear(2, 4, bias=True)
                self.fc2 = nn.Linear(4, 2, bias=True)

            def forward(self, tensor_x: torch.Tensor):
                tensor_x = self.fc1(tensor_x)
                tensor_x = torch.sigmoid(tensor_x)
                tensor_x = self.fc2(tensor_x)
                tensor_x = torch.sigmoid(tensor_x)
                return tensor_x
```

justinchuby:
I tested with the latest onnx. Looks like no segfaults

justinchuby:
with:

```python
for arg in example_args_collection: torch.onnx.dynamo_export(MLP(), *arg)
```

ONNX version 1.15.0.dev20230807

justinchuby:
Also tested with ONNX 1.14.1

justinchuby:
1.14 segfaults

wschin:
Let's close it and open an issue to onnxruntime-training instead, because it sounds onnxruntime-training should fetch onnx-1.14.1 in its dep file. Many thanks.

Issue #945, opened 2023-07-31
[torchlib] Bug with `aten::slice_scatter`


BowenBao:
Found in https://github.com/microsoft/onnx-converters-private/issues/166, affecting below models

```
        pyhpc_turbulent_kinetic_energy
        pyhpc_isoneutral_mixing
        MBartForConditionalGeneration
        ghostnet_100
        BartForConditionalGeneration
        eca_halonext26ts
        PLBartForConditionalGeneration
```

Minimized repro seems to be pointing to a bug with `aten::slice_scatter` torchlib implementation.

```python
import torch
import onnxruntime

def func(x, a):
    return torch.ops.aten.slice_scatter.default(x, a, 3, 0, 9223372036854775807)

x = torch.randn(8, 16, 112, 122)
a = torch.randn(8, 16, 112, 122)


func(x, a)
print("pt eager passed")
model = torch.onnx.dynamo_export(func, x, a).model_proto
session = onnxruntime.InferenceSession(model.SerializeToString(), providers=['CPUExecutionProvider'])
print(session.run(None, {'arg0': x.numpy(), 'arg1': a.numpy()}))
```
```
pt eager passed
2023-07-31 14:40:42.980386186 [E:onnxruntime:, sequential_executor.cc:514 ExecuteKernel] Non-zero status code returned while running ScatterElements node. Name:'_inline__aten_slice_scatter_onnxn17' Status Message: Indices vs updates dimensions differs at position=3 0 vs 122
Traceback (most recent call last):
  File "repro_scatter_slice.py", line 15, in <module>
    print(session.run(None, {'arg0': x.numpy(), 'arg1': a.numpy()}))
  File "/home/bowbao/anaconda3/envs/torch38/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py", line 219, in run
    return self._sess.run(output_names, input_feed, run_options)
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Non-zero status code returned while running ScatterElements node. Name:'_inline__aten_slice_scatter_onnxn17' Status Message: Indices vs updates dimensions differs at position=3 0 vs 122
```

justinchuby:
Assigning to [er3x3](https://github.com/er3x3). Thanks for creating the PR!

PR #995, opened 2023-08-09
Fix slice_scatter

er3x3:
Current logic for slice_scatter doesn't handle the case when param start or end is out of range for [0, dim_value), while this case is valid. Most of the case (for transformer models) it will use MAX_INT for end. Current test data doesn't have such case so it doesn't cover such case. Using (torch.zeros(8, 8), torch.ones(2, 8), 0, 6, 100, 1) can fail current implementation.

This PR is to fix this, and also adjust the logic to make it simpler. The Transpose logic in current implementation is not necessary.

The new implementation passes all current test data as well as above one. I don't know how to add new test data to current UT code, someone please help me on this.

Fixes https://github.com/microsoft/onnxscript/issues/945

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6251,7 +6251,7 @@ def aten_slice_copy(
     raise NotImplementedError()
 
 
-@torch_op("aten::slice_scatter", trace_only=True)
+@torch_op("aten::slice_scatter")
 def aten_slice_scatter(
     self: TTensor,
     src: TTensor,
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6247,40 +6247,21 @@ def aten_slice_scatter(
...
+        op.Unsqueeze(step, zero),
     )
-    # Generate index but not finalized, need to do transpose later
-    # e.g. [[0,1,2],[0,1,2],[0,1,2]...,[0,1,2]], total count = 2x3x7
-    index_base = op.Range(start, end, step)  # e.g. [0,1,2]
-    index_expand = op.Expand(index_base, shape_expand)
-    indices = op.Transpose(index_expand, perm=perm)
-
+    index_base = op.Unsqueeze(index_base, op.Range(1, op.Size(self_shape) - dim, 1))
+    shape_expand = op.ScatterElements(self_shape, op.Unsqueeze(op.Constant(value_int=dim), zero), one, axis=0)
```
justinchuby, line 6263: Could you add a few lines to explain the logic for future readers? Thanks!

justinchuby (commented):
Thanks for the clean implementation!

justinchuby:
To add new tests, you may follow this: https://github.com/microsoft/onnxscript/blob/4d6dc45edce3c00c53acd8060b41c41501cde7d8/onnxscript/tests/function_libs/torch_lib/extra_opinfo.py#L754-L767

justinchuby:
Does the PR fix this issue? https://github.com/microsoft/onnxscript/issues/945

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6247,40 +6247,21 @@ def aten_slice_scatter(
...
-    dim_tensor = op.Reshape(op.Constant(value_int=dim), neg_1)
-    shape_before_dim = op.Slice(src_shape, op.Constant(value_ints=[0]), dim_tensor)
-    shape_after_dim = op.Slice(src_shape, op.Add(dim_tensor, 1), last_dim)
-    shape_expand = op.Concat(
-        shape_before_dim, shape_after_dim, op.Constant(value_ints=[1]), axis=0
+    zero = op.Constant(value_ints=[0])
+    one = op.Constant(value_ints=[1])
+    self_shape = op.Shape(self)
+    dim_shape = op.Gather(self_shape, dim, axis=0)
+    index_base = op.Range(0, dim_shape, 1)
```
justinchuby, line 6254: Would be great to have a comment explaining how the case described in the PR is handled here

fatcat-z:
> Does the PR fix this issue? #945

Looks like it does.

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -751,6 +751,131 @@ def sample_inputs_unfold(op_info, device, dtype, requires_grad, **kwargs):
...
+    )
+
+    L = 20
+    cases = (
+        (
+            (L, L, L),
+            (
+                L,
+                L,
+                L,
```
justinchuby, line 768:
```suggestion
                L
```

nit: Remove the last comma to let black format the tuple into a compact line

justinchuby: Below as well. Thanks!

justinchuby (approved):
Lgtm with comments

onnxscript/tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -751,6 +751,35 @@ def sample_inputs_unfold(op_info, device, dtype, requires_grad, **kwargs):
...
+        ((L, L, L), (L // 4, L, L), (0, L // 2, L, 2)),
+        ((L, L, L), (L, L, L), (1, 0, L, 1)),
+        ((L, L, L), (L, L // 2, L), (1, L // 2, L, 1)),
+        ((L, L, L), (L, L // 4, L), (1, L // 2, L, 2)),
+        ((L, L, L), (L, L, L), (2, 0, L, 1)),
+        ((L, L, L), (L, L, L // 2), (2, L // 2, L, 1)),
+        ((L, L, L), (L, L, L // 4), (2, L // 2, L, 2)),
+        ((L, L, L), (L, L // 2, L), (1, L // 2, L * 2, 1)),  # end > L
+        ((L, L, L), (L, L, L), (-2, 0, L, 1)),  # negative dim
+        ((L, L, L), (L, L, L // 4), (-1, L // 2, L * 2, 2)),  # end > L and negative dim
```
justinchuby, line 774: Nice comments, Thanks! Would you be willing to add these cases to pytorch as well?

er3x3: > Nice comments, Thanks! Would you be willing to add these cases to pytorch as well?

I would rather not do this as I am not quite familiar with the process. You guys can do this. :-)

justinchuby: The process is simple. It is just a matter of adding these three lines in a PR.

PR #996, opened 2023-08-09
Always upload SARIF in PR | chore

justinchuby:
I realized the SARIF file wasn’t uploaded without always()

Use `i` to toggle comment display if needed.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -73,6 +73,8 @@ jobs:
         run: |
           python -m lintrunner_adapters to-sarif lint.json lintrunner.sarif
       - name: Upload SARIF file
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -73,8 +73,7 @@ jobs:
         run: |
           python -m lintrunner_adapters to-sarif lint.json lintrunner.sarif
       - name: Upload SARIF file
-        # Do not display on drafts to make reviewing easier
-        if: github.event.pull_request.draft == false
+        if: always()
```
jcwchen, line 76: nit: perhaps comment why always() here for future reference?

jcwchen (approved):

PR #1003, opened 2023-08-10
Support >2G model export | torchlib(feat)

### LINKED ISSUE #493: TorchScript Graph: Support model > 2GB

justinchuby:
Support >2G model export by caching the model to disk when necessary.

Tested locally with `test_save_initializer_to_files_for_large_model`

Fixes #493

cc @wschin

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -346,7 +346,7 @@ def _create_op_call_in_torch_graph(
     return node_ouputs
 
 
-def _estimate_tensor_size(tensor: torch.Tensor) -> int:
+def _tensor_rawdata_size(tensor: torch.Tensor) -> int:
     """Estimate the size of a tensor in bytes.
 
     Args:
@@ -700,7 +700,7 @@ class TorchScriptGraph:
             unique_custom_domains[function_proto.domain] = 1
 
         initializers_size = sum(
-            _estimate_tensor_size(tensor) for tensor in self.initializers.values()
+            _tensor_rawdata_size(tensor) for tensor in self.initializers.values()
         )
 
         large_model = initializers_size > _LARGE_MODEL_SIZE_THRESHOLD
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -732,7 +732,8 @@ class TorchScriptGraph:
                 ) = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
                     **export_kwargs
                 )
-                onnx_model = onnx.load_model(proto, load_external_data=True)
+                onnx_model = onnx.load_from_string(proto)
+                onnx.load_external_data_for_model(onnx_model, temp_dir)
         else:
             (
                 proto,
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -727,7 +727,7 @@ class TorchScriptGraph:
                 _ = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
                     **export_kwargs
                 )
-            onnx_model = onnx.load_model(onnx_file_path, load_external_data=True)
+                onnx_model = onnx.load_model(onnx_file_path, load_external_data=True)
         else:
             (
                 proto,
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -724,10 +724,15 @@ class TorchScriptGraph:
             with tempfile.TemporaryDirectory() as temp_dir:
                 onnx_file_path = os.path.join(temp_dir, "exported_model.onnx")
                 export_kwargs["onnx_file_path"] = onnx_file_path
-                _ = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
+                (
+                    proto,
+                    _,
+                    _,
+                    _,
+                ) = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
                     **export_kwargs
                 )
-                onnx_model = onnx.load_model(onnx_file_path, load_external_data=True)
+                onnx_model = onnx.load_model(proto, load_external_data=True)
         else:
             (
                 proto,
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -64,6 +64,9 @@ ValidTorchValueType: TypeAlias = Union[
     None,
 ]
 
+# Be sure to leave ample room for the rest of the proto fields.
+_LARGE_MODEL_SIZE_THRESHOLD = int(2**30 * 1.8)  # 1.8GB
+
 # TODO(justinchuby): Build a context manager to handle source information.
 
 
@@ -700,14 +703,11 @@ class TorchScriptGraph:
             _estimate_tensor_size(tensor) for tensor in self.initializers.values()
         )
 
-        # Treat models > 1GB as large models so that we have ample room
-        # for the rest of the proto fields.
-        large_model = initializers_size > (2**30)
+        large_model = initializers_size > _LARGE_MODEL_SIZE_THRESHOLD
 
         export_kwargs: dict[str, Any] = dict(
             initializers=self.initializers if include_initializers else {},
             onnx_opset_version=opset_version,
-            # TODO(justinchuby): Figure out how to get the dynamic axes from the inputs
             dynamic_axes={},
             defer_weight_export=False,
             operator_export_type=torch.onnx.OperatorExportTypes.ONNX,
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -725,10 +756,14 @@ def to_model_proto(
         )
 
         try:
-            onnx_model = onnx.shape_inference.infer_shapes(
-                onnx_model, check_type=True, strict_mode=False, data_prop=True
-            )
-            onnx.checker.check_model(onnx_model, full_check=True)
+            if not cache_model_to_disk:
```
justinchuby, line 759: In the follow up PR we will remove the checks altogether from a discussion with Aaron: we should not check it here.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -342,6 +343,18 @@ def _create_op_call_in_torch_graph(
...
+def _estimate_tensor_size(tensor: torch.Tensor) -> int:
+    """Estimate the size of a tensor in bytes.
+
+    Args:
+        tensor: The tensor to estimate the size of.
+
+    Returns:
+        The estimated size of the tensor in bytes.
+    """
+    return tensor.numel() * tensor.element_size()
```
justinchuby, line 355: Thanks GPT

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -725,10 +762,14 @@ def to_model_proto(
         )
 
         try:
-            onnx_model = onnx.shape_inference.infer_shapes(
-                onnx_model, check_type=True, strict_mode=False, data_prop=True
-            )
-            onnx.checker.check_model(onnx_model, full_check=True)
+            if not cache_model_to_disk:
+                # Only check the model if it is in memory.
+                # Otherwise the checker and shape_inference will fail because
```
wschin, line 767: For shape inference, can we still load shape and element type from model file (not initializer files) and then run infer_shape?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -683,12 +696,15 @@ def to_model_proto(
...
-            proto,
-            _,
-            _,
-            _,
-        ) = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
+        initializers_size = sum(
+            _estimate_tensor_size(tensor) for tensor in self.initializers.values()
+        )
+
+        # Treat models > 1GB as large models so that we have ample room
```
wschin, line 703: Humm, maybe increase to 1.8 GB? I never see a model > 100MB without initializers.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -699,15 +715,36 @@ def to_model_proto(
...
-            onnx_file_path=os.path.join(
-                os.getenv("EXTERNAL_ONNX_INITIALIZER_FOLDER", ""), "dummy_model_path.onnx"
-            ),
             node_attr_to_name={},
         )
 
-        onnx_model = onnx.load_from_string(proto)
+        cache_model_to_disk = include_initializers and large_model
+
+        if cache_model_to_disk:
```
wschin, line 723: Whether or not storing initializers should be controlled by a user flag. Assume that I export a 1GB model on remote machine. I want to visualize it locally. I really don't want to download its initializers with home internet. If this flag can be turned on, I will be able to just download the structure of model and debug faster.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -699,15 +715,36 @@ def to_model_proto(
...
+                (
+                    proto,
+                    _,
+                    _,
+                    _,
+                ) = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
+                    **export_kwargs
+                )
+                onnx_model = onnx.load_from_string(proto)
+                onnx.load_external_data_for_model(onnx_model, temp_dir)
```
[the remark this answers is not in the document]
justinchuby: Do we return a list of initializers?

Taking this further, we don’t even need to pass in initializers to _export_onnx. We can serialize them ourselves outside with onnx. This way we don’t need to change the PyTorch c++ implementation.

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: We could, but we also don’t need to because PyTorch supplies all the shape info.

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: I think once the user get the model proto, they can do whatever they want? A user has full control when they get the dynamo export output as an object. 

Further yet include_initializers is already an argument

onnxscript/function_libs/torch_lib/graph_building.py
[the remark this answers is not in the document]
justinchuby: Tested with 

```
def _add_initializers(model_proto: onnx.ModelProto, initializers: Mapping[str, torch.Tensor]):
    tensor_protos = []
    for name, tensor in initializers.items():
        print(name, "0")
        tensor_numpy = tensor.detach().numpy()
        print(name)
        tensor_proto = onnx.helper.make_tensor(
            name=name,
            data_type=onnx.helper.np_dtype_to_tensor_dtype(tensor_numpy.dtype),
            dims=tensor_numpy.shape,
            vals=tensor_numpy,
        )
        print(name, "done1")
        tensor_protos.append(tensor_proto)
    model_proto.graph.initializer.extend(tensor_protos)
```

But it is very slow.

I think returning a list of TensorProtos can work. But for now it seems to me allowing the compatibility may be nice.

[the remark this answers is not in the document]
justinchuby: *compatibility with torch2.0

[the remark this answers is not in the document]
justinchuby: Because we will move away from torchscript eventually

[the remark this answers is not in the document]
BowenBao: Let's try `make_tensor(...torch_tensor.data_ptr().to_bytes(torch_tensor.element_size() * torch_tensor.numel(), byteorder=sys.byteorder), raw=True)` and if that doesn't work or is still slow then we go back to initial solution.

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: A drawback due to https://github.com/onnx/onnx/issues/5487, we don't have much inner node shape info left now that modules are functions.

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: Agree with @wschin 's goal and @justinchuby 's explanation. A thing to consider for `ExportOutput.save` or `ExportOutputSerializer`.

onnxscript/function_libs/torch_lib/graph_building.py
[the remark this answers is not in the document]
BowenBao: Should be 
```python
import ctypes
torch_tensor = torch.tensor([2, 3])
raw_data = bytes(ctypes.c_ubyte*torch_tensor.element_size()*torch_tensor.numel()).from_address(torch_tensor.data_ptr())
tensor_proto = make_tensor(..., vals=raw_data, raw=True)
```
I was misled by someone doing `data_ptr().to_bytes()`, but what that really does was converting the pointer integer itself to bytes... lol

I'm not sure if it is worth it, looks hacky, but this should resemble what `_export_onnx` is doing on c++ side. If that is still slow then there is nothing more we can do.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -342,6 +343,18 @@ def _create_op_call_in_torch_graph(
     return node_ouputs
 
 
+def _estimate_tensor_size(tensor: torch.Tensor) -> int:
```
BowenBao, line 346:
```suggestion
def _tensor_rawdata_size(tensor: torch.Tensor) -> int:
```
would this be more accurate?

BowenBao (approved):

BowenBao:
Would be nice to mention the perf impact / comparison too.

justinchuby:
> Would be nice to mention the perf impact / comparison too.

Done

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Done

onnxscript/function_libs/torch_lib/graph_building.py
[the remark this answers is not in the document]
justinchuby: https://github.com/microsoft/onnxscript/pull/1004

PR #1004, opened 2023-08-11
Support >2G model export - alternative implementation | torchlib(feat)

justinchuby:
Support >2G model export - alternative implementation where we build initializers ourselves. This is a follow up of #1003. We now build the TensorProto ourselves directly from PyTorch tensors. This circumvents torchscript _export_onnx's limitation of 2G protobuf serialization and additional serialization, because we are now keeping everything in memory.

2G model is now no longer a special case because we add initializers in a separate step.

### TODO

- [ ] Test to_model_proto with initlizers for all torch data types.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -740,20 +740,4 @@ class TorchScriptGraph:
             ]
         )
 
-        try:
-            if not cache_model_to_disk:
-                # Only check the model if it is in memory.
-                # Otherwise the checker and shape_inference will fail because
-                # we cannot serialize the model.
-                onnx_model = onnx.shape_inference.infer_shapes(
-                    onnx_model, check_type=True, strict_mode=False, data_prop=True
-                )
-                onnx.checker.check_model(onnx_model, full_check=True)
-        except (onnx.checker.ValidationError, onnx.shape_inference.InferenceError) as e:
-            warnings.warn(f"ONNX model is invalid: {e}", stacklevel=1)
-            logging.debug(
-                "ONNX model:\n%s\n\nTorchScript graph:\n%s",
-                onnxscript.proto2text(onnx_model),
-                self.torch_graph,
-            )
         return onnx_model
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -732,7 +732,8 @@ class TorchScriptGraph:
                 ) = self._torch_graph._export_onnx(  # type: ignore[attr-defined] # pylint: disable=protected-access
                     **export_kwargs
                 )
-                onnx_model = onnx.load_model(proto, load_external_data=True)
+                onnx_model = onnx.load_from_string(proto)
+                onnx.load_external_data_for_model(onnx_model, temp_dir)
         else:
             (
                 proto,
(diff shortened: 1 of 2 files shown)


justinchuby:
@BowenBao Speed seems to be reasonably good. There's errors: `Unsuported type proto value case.`

BowenBao:
Adding back infer_shapes and check_model made tests pass locally.

justinchuby:
Let me test again

justinchuby:
> Adding back infer_shapes and check_model made tests pass locally.

Nice, thanks! I updated this PR.

BowenBao (approved):
🚀

justinchuby:
Needs validation with real models

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -281,12 +277,38 @@ def eval_function(  # type: ignore[override]
...
+    for name, tensor in initializers.items():
+        tensor = tensor.detach().cpu().contiguous()
+        # Take the raw data directly from the tensor to avoid the overhead of
+        # data manipulation in onnx.helper.make_tensor
+        raw_data = bytes(
+            (ctypes.c_ubyte * tensor.element_size() * tensor.numel()).from_address(
+                tensor.data_ptr()
+            )
+        )
+        tensor_proto = onnx.helper.make_tensor(
```
wschin, line 295: consider optionally making it `external_data` so you never keep the model in memory.

justinchuby: I think that's what we wanted (return a single model proto, so the interface is uniform)? If we didn't need the initializers we can set include_initializers to False. The exporter is responsible for deciding what to do with the model (save to external data, serialize etc.)

justinchuby: This way to_model_proto does not have side effects.

justinchuby: Are you concerned about the case when we don't have enough memory so we don't want to duplicate the data? I wonder how we want to design the export api for this case?

wschin: Yes. I meant. TensorProto can be stored as schema (name, shape, element_type) plus a path to a file containing its content -- it's a valid in-memory TensorProto. If we optionally enable it, we can make this part of code scalable to huge scale models. I haven't gone through all parts end-to-end, so this idea might be local instead global necessary.

wschin: There are multiple fields you can use under TensorProto to store tensor content; `float_data`, `raw_data`, and `external_data`. I feel we should let use to optionally use `external_data`.

Ref: https://github.com/onnx/onnx/blob/ec38b7c4cf3656b9115a68f525017a6e69aa8773/onnx/onnx-ml.proto#L608

justinchuby: I see! That makes sense. Let me think more on this

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -378,12 +374,38 @@ def eval_function(  # type: ignore[override]
         return self._graph.add_function_call(function, inputs, attributes)
 
 
+def _add_initializers(
```
BowenBao, line 377: fyi @ kunal-vaishnavi creating onnx tensorproto from torch tensor dataptr

Issue #965, opened 2023-08-04
[torchlib] Fix permute


justinchuby:
_Possible just fix the function signature_

### Summary

ONNX Runtime raises `[ONNXRuntimeError] : 10 : INVALID_GRAPH : This is an invalid model. In Node, ("_0x7ef6c40_n0", Transpose, "", -1) : ("input_0": tensor(float16),) -> ("_val_1": tensor(float16),) , Error Attribute 'perm' is expected to have field 'ints'` when executing test `ops_test.TestOutputConsistencyFullGraphCPU.test_output_match_opinfo__permute_cpu_float16` in ONNX Script `TorchLib`.

To recreate this report, use

```bash
CREATE_REPRODUCTION_REPORT=1 python -m pytest onnxscript/tests/function_libs/torch_lib/ops_test.py -k test_output_match_opinfo__permute_cpu_float16
```

### To reproduce

```python
import google.protobuf.text_format
import numpy as np
from numpy import array, float16, float32, float64, int32, int64
import onnx
import onnxruntime as ort

# Run n times
N = 1

onnx_model_text = """
ir_version: 8
producer_name: "pytorch"
producer_version: "2.1.0"
graph {
  node {
    input: "input_0"
    output: "_val_1"
    name: "aten_permute_0"
    op_type: "aten_permute"
    attribute {
      name: "dims"
      type: INTS
    }
    doc_string: ""
    domain: "pkg.onnxscript.torch_lib"
  }
  name: "torch_jit"
  input {
    name: "input_0"
    type {
      tensor_type {
        elem_type: 10
        shape {
        }
      }
    }
  }
  output {
    name: "_val_1"
    type {
      tensor_type {
        elem_type: 10
        shape {
        }
      }
    }
  }
}
opset_import {
  domain: "pkg.onnxscript.torch_lib"
  version: 1
}
opset_import {
  domain: ""
  version: 18
}
functions {
  name: "aten_permute"
  input: "self"
  output: "return_val"
  attribute: "dims"
  node {
    input: "self"
    output: "return_val"
    name: "n0"
    op_type: "Transpose"
    attribute {
      name: "perm"
      type: INTS
      ref_attr_name: "dims"
    }
    domain: ""
  }
  doc_string: "permute(Tensor(a) self, int[] dims) -> Tensor(a)"
  opset_import {
    domain: ""
    version: 18
  }
  domain: "pkg.onnxscript.torch_lib"
}

"""

[...]

PR #1015, opened 2023-08-21
Fix permute for all tested inputs | fix(torchlib)

justinchuby:
Add negative axis and empty axis support to `permute`. We have to mark it as `trace_only` because we need to manipulate axis. However this should be acceptable because the result is still a single op in graph.

Fixes https://github.com/microsoft/onnxscript/issues/965

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -5342,9 +5342,14 @@ def aten_pdist(self: TensorType, p: float = 2.0) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::permute")
+@torch_op("aten::permute", trace_only=True)
 def aten_permute(self: TTensor, dims: Sequence[int]) -> TTensor:
     """permute(Tensor(a) self, int[] dims) -> Tensor(a)"""
+
+    if not dims:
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5342,10 +5342,16 @@ def aten_pdist(self: TensorType, p: float = 2.0) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::permute")
+@torch_op("aten::permute", trace_only=True)
 def aten_permute(self: TTensor, dims: Sequence[int]) -> TTensor:
     """permute(Tensor(a) self, int[] dims) -> Tensor(a)"""
 
+    if not dims:
```
titaiwangms, line 5349: `aten_permute` with no dims or dims=None will always be nearest match in SARIF. If we want a perfect match on this, we will need a separate function like `aten_permute_non_dim` to match the no dims situation. But I am not blocking this.

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: dims won't be `None`, it may be empty.

PR #1017, opened 2023-08-21
Fix circular import with torch exporter

BowenBao:
* (to be filled)

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -6,7 +6,7 @@ import os
 import tempfile
 import typing
 import warnings
-from typing import Any, Dict, Final, List, Mapping, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


justinchuby (approved):
Thanks a lot!

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -6,7 +6,7 @@
 import tempfile
 import typing
 import warnings
-from typing import Any, Dict, Final, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Union
```
justinchuby, line 9:
```suggestion
from typing import Any, Dict, Final, List, Mapping, Optional, Sequence, Tuple, Union
```

BowenBao: strange, not sure why lintrunner removed this. thanks for fix.

PR #1023, opened 2023-08-24
Add initializer setter to support constant tensors in fake mode export | torchlib(feat)

titaiwangms:
`include_initializer` is created and controled by fake_mode to detach initializers with their default values, preventing fake tensor becomes initializers. However, initializer could be constant tensors which should not be detached from its default value, as it's always real tensor. This PR removes the parameter, because after https://github.com/pytorch/pytorch/pull/107836, the filtering logic is added into converter.

NOTE: `include_initializers` is kept for potentially usage of debugging/experimenting on large scale models.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -402,13 +402,6 @@ class TorchScriptGraph:
     def initializers(self) -> Mapping[str, torch.Tensor]:
         return self._initializers
 
-    # NOTE: This setter is used in torch converter when we activate fake mode,
-    #       we need to filter out the initializers that has fake tensor. This
-    #       is because we don't want to introduce fake tensor in onnxscript.
-    @initializers.setter
-    def initializers(self, initializers: Dict[str, torch.Tensor]):
-        self._initializers = initializers
-
     @property
     def initializers_inputs(self) -> Mapping[str, TorchScriptTensor]:
         return self._initializers_inputs
@@ -454,7 +447,9 @@ class TorchScriptGraph:
         return tensor_value  # type: ignore[return-value]
 
     @runtime_typing.checked
-    def add_initializer(self, name: str, value: torch.Tensor) -> TorchScriptTensor:
+    def add_initializer(
+        self, name: str, value: torch.Tensor, save_value_to_default: bool = True
+    ) -> TorchScriptTensor:
         if name in self._initializers_inputs:
             # NOTE: Previously it raises when `name` is already set. This is relaxed
             # because this will be invoked multiple times when submodule is called
@@ -473,14 +468,17 @@ class TorchScriptGraph:
             # to root graph, and add as input to current graph.
             self._initializers_inputs_from_parent[
                 name
-            ] = self._parent_torch_script_graph.add_initializer(name, value)
+            ] = self._parent_torch_script_graph.add_initializer(
+                name, value, save_value_to_default
+            )
             torch_value = self._torch_graph.addInput(name)
             torch_value.setType(torch.TensorType.create_from_tensor(value))
             tensor_value = _wrap_torch_value_to_tensor(torch_value)
             self._initializers_inputs[name] = tensor_value  # type: ignore[assignment]
             return tensor_value  # type: ignore[return-value]
-
-        self._initializers[name] = value
+        # NOTE: If an initializer relies on external input, its default value is detached.
+        if save_value_to_default:
+            self._initializers[name] = value
         torch_value = self._torch_graph.addInput(name)
         torch_value.setType(torch.TensorType.create_from_tensor(value))
         tensor_value = _wrap_torch_value_to_tensor(torch_value)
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -402,6 +402,13 @@ def torch_graph(self):
     def initializers(self) -> Mapping[str, torch.Tensor]:
         return self._initializers
 
+    # NOTE: This setter is used in torch converter when we activate fake mode,
+    #       we need to filter out the initializers that has fake tensor. This
+    #       is because we don't want to introduce fake tensor in onnxscript.
+    @initializers.setter
```
justinchuby, line 408: Is add_initializer not enough?

titaiwangms: We could filter fake tensor INSIDE `add_initializer`, but do you think we should bring fake tensor into it?

justinchuby: I suggests filtering outside of the call if possible. The exporter should be responsible for deciding which initializer to add

titaiwangms: We want it to be initializers, but we don't want it to have default values.

justinchuby: So a tensor proto w/o data?

titaiwangms: Constant tensor is not an input and we want to keep its value. The reason that it's not going through `_wrap_fx_args_as_onnxscript_args` is because ORT prefers constant tensors to be initializers. As for others, buffers and weights under fake mode, those values would become fake tensor which further crashes during `to_model_proto` if we don't detach the fake tensor from value.

justinchuby: I see. I am actually a little confused on how initializers are handled. Happy to chat more tomorrow.

justinchuby: Specifically 

        # Mapping from intializer name to data(torch.Tensor).
        self._initializers: Dict[str, torch.Tensor] = {}
        # Mapping from intializer name to input(TorchScriptTensor).
        self._initializers_inputs: Dict[str, TorchScriptTensor] = {}

@bowenbao

titaiwangms: Should we have this by 8/25 to catch 2.1?

BowenBao: Changes in pytorch need to catch the 8/25 cut.

BowenBao: @justinchuby  here is an example of what these three initializer name mapped values are and used.


        initializer_data: value in self._initializers
        initializer_input: value in self._initializers_inputs
        initializer_input_from_parent: value in self._initializers_inputs_from_parent

```
MainGraph(...)
  %initializer_input_from_parent = onnx::Constant(value_t=initializer_data)
  %sub_graph_output = onnx::SubGraph(%initializer_input_from_parent)

SubGraph(%initializer_input)
  %output = onnx::Op(%initializer_input, ...)
```

BowenBao: All the complexity is for building the symbolic connection of subgraph input and maingraph initializer. We can probably refactor this away and move connection logic to exporter, such that the logic for dismissing fake tensor initializer data look more natural, but the deadline is too tight.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -447,7 +447,9 @@ def add_input(
         return tensor_value  # type: ignore[return-value]
 
     @runtime_typing.checked
-    def add_initializer(self, name: str, value: torch.Tensor) -> TorchScriptTensor:
+    def add_initializer(
+        self, name: str, value: torch.Tensor, save_value_to_default: bool = True
```
BowenBao, line 451: I feel the flag `save_value_to_default` is same as making `value` an `Optional[torch.Tensor]` and store to map when not None.

Both feels very weird, and are doing things like "add input but also help me construct the symbolic connection as if this is an initializer, except that there is no initializer data". It really doesn't make sense until at the very end fake mode serialization comes in and brings back all the initializer data and reveals "hey that's why". The whole process also assumes all missing gap will be precisely filled by fake mode serialization for it to work.

We need to be careful with API changes as it will introduce backward compatibility dependency: we cannot remove this arg even if later exporter refactored since people using torch 2.1 with newer onnxscript would crash otherwise. 

Maybe after all introducing fake tensor concept into graph building is the best solution for now ... We do a filter in `to_model_proto` and only serialize initializer if tensor is concrete.

BowenBao (approved):
Discussed various options offline. The current being most non-intrusive and bc friendly.

titaiwangms:
Added back `include_initializers` for potentially debugging usage.

justinchuby (approved):

PR #1024, opened 2023-08-25
Implement `aten::_softmax` | feat(torchlib)

justinchuby:
Move `aten_special_softmax` -> `aten_softmax`, support alias and when dtype input is None. Implement `_softmax` using `softmax`.

An alternative would be to create a decomp in pytorch from _softmax to softmax. But since pytorch optimization tends to create decomp to private ops (and not the other way around) in different optimizations steps as @BowenBao pointed out, we still need to support _softmax here.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -67,6 +67,29 @@ def aten__local_scalar_dense_int(self: IntType) -> INT64:
     return op.Cast(op.Gather(op.Reshape(self, [-1]), 0), to=INT64.dtype)
 
 
+@torch_op("aten::_softmax", trace_only=True)
+def aten__softmax(self: Union[FLOAT16, BFLOAT16], dim: int, half_to_float: bool) -> FLOAT:
+    """_softmax(Tensor self, int dim, bool half_to_float) -> Tensor"""
+
+    # trace_only because we need to cast conditionally based on half_to_float
+    if half_to_float:
+        self = op.Cast(self, to=FLOAT.dtype)
+
+    return aten_softmax_no_dtype(self, dim)
+
+
+@torch_op("aten::_softmax", trace_only=True)
+def aten__softmax(
+    self: TFloatHighPrecision, dim: int, half_to_float: bool
+) -> TFloatHighPrecision:
+    """_softmax(Tensor self, int dim, bool half_to_float) -> Tensor"""
+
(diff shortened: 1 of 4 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -67,6 +67,29 @@ def aten__local_scalar_dense_int(self: IntType) -> INT64:
...
+    # trace_only because we need to cast conditionally based on half_to_float
+    if half_to_float:
+        self = op.Cast(self, to=FLOAT.dtype)
+
+    return aten_softmax_no_dtype(self, dim)
+
+
+@torch_op("aten::_softmax", trace_only=True)
+def aten__softmax(
+    self: TFloatHighPrecision, dim: int, half_to_float: bool
```
fatcat-z, line 83: What does these 2 _softmax() functions has different dtypes for self?

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1197,7 +1197,7 @@ def _where_input_wrangler(
     TorchLibOpInfo("sinh", core_ops.aten_sinh),
     TorchLibOpInfo(
         "softmax",
-        special_ops.aten_special_softmax,
+        core_ops.aten_softmax,
```
fatcat-z, line 1200: Should we add aten__softmax into test as well?

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: This is because we want to cast float16 to float32 when half_to_float is true. Since we have no idea what the input dtype will be within the function body, we rely on the dispatcher to pick the function that already limits the input dtype it can accept to know we are dealing with float16 types.

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
justinchuby: We should

fatcat-z: > We should

Will it be done in this PR or next one in future?

justinchuby: I will update the PR. Sorry for the confusion.

fatcat-z (approved):

PR #1029, opened 2023-08-27
Fix typo mpdel --> model

chipmunk-pixel:
Fix typo in README


justinchuby:
Thanks! Could you also fix https://github.com/microsoft/onnxscript/blob/main/docs/tutorial/examples/hardmax_end_to_end.py? This way the files can be synced

chipmunk-pixel:

chipmunk-pixel:
> Thanks! Could you also fix https://github.com/microsoft/onnxscript/blob/main/docs/tutorial/examples/hardmax_end_to_end.py? This way the files can be synced

Updated!

justinchuby (approved):

PR #1032, opened 2023-08-29
Add Op(multinomial) | torchlib(feat)

titaiwangms:
~~1. xfail reason: ONNX spec expects the first dim is batch size, and the whole input should be 2D.~~
~~2. extra opinfo reason: We would like to have a pure 2D datasets to make sure the op is functional.~~

ATen supports 1D and 2D, but ONNX only supports 2D, so Unsqueeze would be used when it's 1D input.

### DIFF
--- a/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
@@ -1117,11 +1117,7 @@ TESTED_TORCHLIB_OPS: tuple[TorchLibOpInfo, ...] = (
         matcher=lambda sample: len(sample.input.shape) == 0,
         reason="fixme: output 'shape' do not match: torch.Size([0, 1]) != torch.Size([0, 0]).",
     ),
-    TorchLibOpInfo(
-        "normal",
-        core_ops.aten_normal,
-        nondeterministic=True,
-    )
+    TorchLibOpInfo("normal", core_ops.aten_normal, nondeterministic=True)
     .skip(
         matcher=lambda sample: len(sample.args) > 0 and not isinstance(sample.args[0], float),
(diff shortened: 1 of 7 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1529,6 +1529,19 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.args) > 0,
         reason="this ATen overload only supports one tensor as input by design",
     ),
+    TorchLibOpInfo(
+        "multinomial",
+        core_ops.aten_multinomial,
+        nondeterministic=True,
+    ).xfail(
+        matcher=lambda sample: len(sample.input.shape) < 2,
+        reason="onnx spec requires input tensor to have at least 2 dimensions (the first one is batch size)",
```
justinchuby, line 1538: Can we use unqueeze and squeeze to handle this?

titaiwangms: Done

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1545,6 +1545,11 @@ def _where_input_wrangler(
         matcher=lambda sample: len(sample.args) > 0,
         reason="this ATen overload only supports one tensor as input by design",
     ),
+    TorchLibOpInfo(
+        "multinomial",
+        core_ops.aten_multinomial,
+        nondeterministic=True,
```
justinchuby, line 1551:
```suggestion
        nondeterministic=True
```

so the lines can be formatted compactly

justinchuby (approved):
Lgtm!

PR #1034, opened 2023-08-29
Create aten_normal_float_float | feat(torchlib)

justinchuby:
Fixes the new test error caused by torch-nightly


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5393,6 +5393,18 @@ def aten_normal(
     return result
 
 
+@torch_op("aten::normal.float_float")
+def aten_normal_float_float(
+    mean: float, std: float, size: INT64, dtype: int = FLOAT.dtype
+) -> TensorType:
```
titaiwangms, line 5399: Output dtype?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Since the type is determined by the dtype argument, it can be anything specified. Tensortype is the best I could come up with. Any suggestions?

justinchuby:
Merging for now. Will follow up if needed.

PR #1036, opened 2023-08-29
Fix aten_var_mean for float16 inputs

xiaowuhu:
So far, no idea about the shape inference differ error. If we comment the check_model(), the testing will pass.


justinchuby (approved):

justinchuby:
Should we create an issue in the onnx repo for the shape error?

PR #1038, opened 2023-08-29
Implement maximum/minimum for Boolean inputs | feat(torchlib)

justinchuby:
Use And and Or to create variants because ONNX Min/Max does not take BOOL input.


titaiwangms (approved):
Why we don't need to skip non-BOOL cases in the duplicated dataset?

justinchuby:
> Why we don't need to skip non-BOOL cases in the duplicated dataset?

That's because we use the function signature to skip unsupported dtypes automatically

PR #1039, opened 2023-08-30
Fix `attribute_proto` generation of function in protobuf 4.X

take-cheeze:
In protobuf 4.x fields aren't defined as class attribute so it should check instance attribute instead
```
>>> import google.protobuf
>>> google.protobuf.__version__
'4.21.0'
>>> import onnx
>>> hasattr(onnx.FunctionProto, "attribute_proto")
False
>>> hasattr(onnx.FunctionProto(), "attribute_proto")
True
```

### DIFF
--- a/onnxscript/irbuilder.py
+++ b/onnxscript/irbuilder.py
@@ -468,7 +468,7 @@ class IRFunction:
             attributes=attribute_names,
             doc_string=self.docstring,
         )
-        if hasattr(onnx.FunctionProto, "attribute_proto"):
+        if hasattr(f, "attribute_proto"):
             f.attribute_proto.extend(
                 [attr.attr_proto for attr in self.attrs if attr.has_default]
             )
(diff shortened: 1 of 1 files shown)


onnxscript/irbuilder.py
```diff
@@ -468,7 +468,7 @@ def to_function_proto(self) -> onnx.FunctionProto:
             attributes=attribute_names,
             doc_string=self.docstring,
         )
-        if hasattr(onnx.FunctionProto, "attribute_proto"):
+        if hasattr(f, "attribute_proto"):
```
justinchuby, line 471: Could you add the same note in the Pr description as a comment here? Thanks!

justinchuby, line 471:
```suggestion
        # In protobuf 4.x fields aren't defined as class attribute so it should check instance attribute instead
        if hasattr(f, "attribute_proto"):
```

justinchuby: I did it

justinchuby (approved):

onnxscript/irbuilder.py
take-cheeze: Thank you!

PR #1041, opened 2023-08-31
try to fix TopK

xiaowuhu:


xiaowuhu:
It only happened on Linux/MacOS, impact the second output. On Windows, it works well. So it might be due to the sort order.

PR #1044, opened 2023-09-01
Register `_embedding_bag` correctly | fix(torchlib)

justinchuby:
Previously `_embedding_bag` was registered to `aten_embedding_bag` when it should have been registered to `aten_embedding_bag_padding_idx` because of its signature.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -2337,9 +2337,9 @@ def aten_embedding_backward(
 def aten_embedding_bag(
     weight: TFloat,
     indices: INT64,
-    offsets: Optional[INT64] = None,  # Could be None accotding to the doc, go 2d branch
+    offsets: INT64,
     scale_grad_by_freq: bool = False,  # pylint: disable=unused-argument
-    mode: int = 1,  # [0,1,2] indicate ["sum", "mean", "max"], default is "mean"
+    mode: int = 0,  # [0,1,2] indicate ["sum", "mean", "max"]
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2466,7 +2459,14 @@ def _aten_embedding_bag_onnx(
...
 
-@torch_op("aten::embedding_bag.padding_idx", trace_only=True)
+@torch_op(
+    (
+        "aten::embedding_bag.padding_idx",
+        "aten::_embedding_bag",
+        "aten::_embedding_bag_forward_only",
+    ),
+    trace_only=True,
+)
```
titaiwangms, line 2469: Should we change `padding_idx` to non-Optional value as it's an attribute?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Good call. @xiaowuhu do you know what happens when idx is -1?

xiaowuhu: ```padding_idx``` must exist in embedding_bag.padding_idx function. 
If it is none, it should be use embedding_bag function.

justinchuby: > Should we change `padding_idx` to non-Optional value as it's an attribute?

Tried but couldn't because arguments before it are kwargs. Would this be an issue for the dispatcher?

PR #1045, opened 2023-09-04
Fix arange issue for int8/int16/int32 dtype

### LINKED ISSUE #974: [torchlib] Fix arange for int32
Output shape mismatch in edge cases.

fatcat-z:
ONNX op 'Range' behaves differently with Torch op 'arange' when the expected output dtype is INT32.

Totally 2 differences are found in two conditions below:

1. 'start' is negative, or 
2. 'step' is negative. 

This PR will make necessary adjustments to make these 2 ops behave same.

Fix #974 .

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -507,7 +507,7 @@ def _adjust_args_for_arange_int_dtype(
     step = op.Cast(step, to=FLOAT.dtype)
 
     if start < zero:
-        start = op.Cast(op.Cast(start, to=dtype), to=FLOAT.dtype)
+        start = op.Ceil(start)
 
     if step < zero:
         start = op.Floor(start)
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -531,7 +531,7 @@ def aten_arange_start_step(
         # PyTorch arange op handles these integral types differently from INT64,
         # so we have to adjust these arguments accordingly.
         # https://github.com/pytorch/pytorch/blob/121cfb60c0817816fcbe2190303b7f6d05c77cf3/torch/_refs/__init__.py#L4794
-        start, end, step = _adjust_args_for_arange_int_dtype(start, end, step, dtype)
+        start, end, step = _adjust_args_for_arange_int_dtype(start, end, step)
         result = op.Cast(op.Range(start, end, step), to=dtype)
     elif dtype == INT64.dtype:
         end = op.Cast(end, to=dtype)
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -521,7 +521,7 @@ def aten_arange_start_step(
     if dtype == -1:
         result = op.Range(start, end, step)
     elif _range_supported(dtype):
-        if dtype in [INT16.dtype, INT32.dtype]:
+        if dtype in (INT16.dtype, INT32.dtype):
             start, end, step = _aten_arange_start_step_onnx(start, end, step, dtype)
             result = op.Cast(op.Range(start, end, step), to=dtype)
         else:
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -488,9 +488,8 @@ def aten_arange_start(
 def _aten_arange_start_step_onnx(
     start: TRealUnlessFloat16OrInt8,
     end: TRealUnlessFloat16OrInt8,
-    step: TRealUnlessFloat16OrInt8
+    step: TRealUnlessFloat16OrInt8,
 ) -> Tuple[FLOAT, FLOAT, FLOAT]:
-
     zero = op.Cast(0.0, to=FLOAT.dtype)
     start = op.Cast(start, to=FLOAT.dtype)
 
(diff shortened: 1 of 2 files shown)


justinchuby:
Thanks! We can work together to simplify the logic on Wednesday. For now - should we make this compatible with int16 and int8 as well, since they should be the same as int32?

fatcat-z:
> Thanks! We can work together to simplify the logic on Wednesday. For now - should we make this compatible with int16 and int8 as well, since they should be the same as int32?

Has updated the code to support them as well. It passed in local testing.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -484,6 +484,28 @@ def aten_arange_start(
...
+    start: TRealUnlessFloat16OrInt8,
+    end: TRealUnlessFloat16OrInt8,
+    step: TRealUnlessFloat16OrInt8,
+    dtype: int,
+) -> Tuple[FLOAT, FLOAT, FLOAT]:
+    zero = op.Cast(0.0, to=FLOAT.dtype)
+    start = op.Cast(start, to=FLOAT.dtype)
+
+    if start < zero:
+        start = op.Cast(op.Cast(start, to=dtype), to=FLOAT.dtype)
```
justinchuby, line 498: Is this effectively taking Floor?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -484,6 +484,28 @@ def aten_arange_start(
...
 
+@torch_op("aten::arange.start_step", private=True)
+def _aten_arange_start_step_onnx(
+    start: TRealUnlessFloat16OrInt8,
+    end: TRealUnlessFloat16OrInt8,
+    step: TRealUnlessFloat16OrInt8,
+    dtype: int,
+) -> Tuple[FLOAT, FLOAT, FLOAT]:
+    zero = op.Cast(0.0, to=FLOAT.dtype)
+    start = op.Cast(start, to=FLOAT.dtype)
```
justinchuby, line 495:
```suggestion
    start = op.Cast(start, to=FLOAT.dtype)
    step = op.Cast(step, to=FLOAT.dtype)
    end = op.Cast(end, to=FLOAT.dtype)
```

We can move all casts to the top for clarity

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -484,6 +484,28 @@ def aten_arange_start(
     return result
 
 
+@torch_op("aten::arange.start_step", private=True)
+def _aten_arange_start_step_onnx(
```
justinchuby, line 488:
```suggestion
def _adjust_endpoints_for_arange_int_dtype(
```

or something more descriptive?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -499,18 +521,26 @@ def aten_arange_start_step(
     if dtype == -1:
         result = op.Range(start, end, step)
     elif _range_supported(dtype):
-        end = op.Cast(end, to=dtype)
-        start = op.Cast(start, to=dtype)
-        step = op.Cast(step, to=dtype)
-        result = op.Range(start, end, step)
+        if dtype in [INT16.dtype, INT32.dtype]:
```
justinchuby, line 524:
```suggestion
        if dtype in (INT8.dtype INT16.dtype, INT32.dtype):
```

I would also use an immutable type (tuple) for immutable uses

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -499,18 +521,26 @@ def aten_arange_start_step(
...
+            result = op.Range(start, end, step)
     else:
         # Cast input to float if dtype is not supported by Range,
         # because the input dtype may be e.g. bfloat16 / int8 etc.
         # which Range does not support. The output type is ensured because the output
         # is casted to the specified dtype.
-        end = op.Cast(end, to=FLOAT.dtype)
-        start = op.Cast(start, to=FLOAT.dtype)
-        step = op.Cast(step, to=FLOAT.dtype)
+        if dtype == INT8.dtype:
```
justinchuby, line 537: Looks like we can merge this with the other int types? Only INT64 is different

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Sorry my bad - I should say Ceil

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -484,6 +484,28 @@ def aten_arange_start(
     return result
 
 
+@torch_op("aten::arange.start_step", private=True)
+def _aten_arange_start_step_onnx(
+    start: TRealUnlessFloat16OrInt8,
+    end: TRealUnlessFloat16OrInt8,
+    step: TRealUnlessFloat16OrInt8,
+    dtype: int,
+) -> Tuple[FLOAT, FLOAT, FLOAT]:
```
justinchuby, line 493: Note: comment that links to the pytorch ref implementation that explains this logic

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -425,6 +425,15 @@ def _range_supported(dtype: int) -> bool:
     }
 
 
+def _integral_to_be_adjusted(dtype: int) -> bool:
```
titaiwangms, line 428: Should we have another file for utils?

titaiwangms (approved):

justinchuby (approved):
Thanks for the great work!

PR #1050, opened 2023-09-06
Register `view_as_real` as complex functions | fix(torchlib)

justinchuby:
Previously I omitted `complex=True`


titaiwangms (approved):
Is there any test guarding it?

justinchuby:
> Is there any test guarding it?

That’s a good question. I haven’t come up with ways of doing it. Any suggestions?

Will merge for now. Happy to follow up!

titaiwangms:
> > Is there any test guarding it?
> 
> That’s a good question. I haven’t come up with ways of doing it. Any suggestions?
> 
> Will merge for now. Happy to follow up!

It can't be added into ops_test.py with extra_opinfo?

justinchuby:
That's true. My bad

PR #1051, opened 2023-09-06
Ensure exporter creates valid python name for opset vars

gramalingam:
The existing exporter creates invalid python identifiers like "onnxscript.atenlib1" when generating variable names for opsets. Change this to create a name like "onnxscript_atenlib1".

### DIFF
--- a/onnxscript/backend/onnx_export.py
+++ b/onnxscript/backend/onnx_export.py
@@ -432,21 +432,22 @@ class Exporter:
         ]
         return "".join(text)
 
+    def translate_opset_import(self, domain: str, version: int) -> str:
+        if (domain == "") or (domain == "ai.onnx"):
+            return f"from onnxscript.onnx_opset import opset{version}\n"
+        else:
+            varname = self.make_opset_name(domain, version)
+            return f"{varname} = Opset('{domain}', {version})\n"
+
     def translate_opset_imports(self, opset_imports: Sequence[onnx.OperatorSetIdProto]) -> str:
-        text = ""
-        for imported in opset_imports:
-            if (imported.domain == "") or (imported.domain == "ai.onnx"):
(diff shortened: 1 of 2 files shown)


onnxscript/backend/onnx_export.py
```diff
@@ -230,6 +212,28 @@ def _rename_variable_s(self, name):
         """Renames all names equal to a python keyword."""
         return str(self._rename_variable(name))
 
+    def _rename_domain(self, domain: str) -> str:
+        if domain == "":
+            return "opset"
+        return domain.replace(".", "_")
```
justinchuby, line 218: Move this method out to the global namespace as it doesn't depend on `self`?

onnxscript/backend/onnx_export.py
```diff
@@ -230,6 +212,28 @@ def _rename_variable_s(self, name):
...
+
+    def make_opset_name(self, domain, version):
+        return f"{self._rename_domain(domain)}{version}"
```
justinchuby, line 221: We can move this as well

onnxscript/backend/onnx_export.py
```diff
@@ -230,6 +212,28 @@ def _rename_variable_s(self, name):
...
+
+    def _python_make_node_name(self, domain, version, name, node=False):
```
justinchuby, line 223: This too

onnxscript/backend/onnx_export.py
gramalingam: It doesn't use self yet. But, logically, it may need other state information (to avoid name clashes). I will leave it in its current place for now. Also, I am making minimal changes for now, as this is likely to undergo more significant changes (like elimination of templates) (re. https://github.com/microsoft/onnxscript/pull/873 )

justinchuby (approved):

PR #1064, opened 2023-09-18
Update beartype version | ci

justinchuby:
Avoid 0.16.0 because of https://github.com/beartype/beartype/issues/282

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -13,7 +13,7 @@ COMMON_TEST_DEPENDENCIES = (
     "jinja2",
     "numpy==1.23.5",
     "typing_extensions",
-    "beartype",
+    "beartype!=0.16.0",
     "types-PyYAML",
     "expecttest",
     "hypothesis",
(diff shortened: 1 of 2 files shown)


jcwchen (approved):

noxfile.py
```diff
@@ -13,7 +13,7 @@
     "jinja2",
     "numpy==1.23.5",
     "typing_extensions",
-    "beartype",
+    "beartype!=0.16.0",
```
jcwchen, line 16:
```suggestion
    "beartype<0.16.0",
```
nit: or just smaller for now? Not sure whether they can fix the issue in the next version and if not, it will break again.

jcwchen (commented):
nit: or

noxfile.py
justinchuby: I expect they will fix the issue in the next version because it is impacting many users. I will send a patch if that doesn't happen

Issue #1056, opened 2023-09-11
Embedding_bag mismatch


justinchuby:
```
_ TestOnnxModelOutputConsistency_opset18CPU.test_output_match_nn_functional_embedding_bag_cpu_float16 (inputs='(tensor([[ 3.4629,  4.7383, -5.2305, -7.2500,  3.7188],\n        [ 0.5361, -6.0039, -1.3975,  0.3867, -1.9512],\n        [ 8.1719,  0.7734, -8.7109, -3.3320, -6.9609],\n        [-2.2578, -8.6875,  6.7500,  4.0352, -6.4531],\n        [-7.2695, -1.1426,  8.3438,  2.2324,  0.7031],\n        [ 3.9199, -7.8047, -7.3398, -7.8672, -7.1641],\n        [ 3.2344, -6.4414,  3.3047,  3.0508, -7.5234],\n        [ 7.7695, -1.9336, -7.4531,  4.7188, -6.1602],\n        [ 7.6562, -4.5352, -1.9424,  1.0020,  7.0156],\n        [ 0.2549,  6.6172,  2.5391, -7.7188,  2.3984]], dtype=torch.float16), tensor([[9, 3, 8, 4, 4],\n        [9, 4, 4, 9, 9],\n        [5, 7, 0, 3, 2],\n        [0, 8, 6, 1, 9],\n        [1, 6, 9, 8, 6]]))', kwargs="{'mode': 'sum', 'per_sample_weights': tensor([[ 4.7383,  8.6016,  1.9072, -3.1719, -1.7842],\n        [-6.3008,  1.5293,  1.3008, -0.6592,  4.7539],\n        [ 5.0898, -4.1914, -1.9072,  3.4531, -8.3594],\n        [-6.3906, -0.9229,  7.9023,  7.1562, -2.5840],\n        [ 8.3125, -1.3008,  6.7773,  7.0234, -2.1719]], dtype=torch.float16)}", opset=18, sample_num=4) _
Traceback (most recent call last):
  File "/home/justinchu/dev/pytorch/test/onnx/test_fx_op_consistency.py", line 679, in _run_test_output_match
    test_suite.run_test_with_fx_to_onnx_exporter_and_onnx_runtime(
[...]

PR #1067, opened 2023-09-20
FixOp(embeddign_bag) change to use nn.functional.embeding_bag function | feat(torchlib)

xiaowuhu:
For issue #1056, no idea what test case it is using and which target function is using ( nn.functional.embedding_bag or ops.aten.embedding_bag?)
This PR changed to use nn.functional.embedding_bag as reference.
The difference are:
- Only return 1 output
- Support when indices is 2d tensor. In this case, the 'offsets' argument can be None.
- The padding_idx argument not support, need another overload to support.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -2646,7 +2646,7 @@ def aten_embedding_backward(
 def aten_embedding_bag(
     weight: TFloat,
     indices: INT64,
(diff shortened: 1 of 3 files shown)


justinchuby:
We should still return four outputs as discussed in teams

justinchuby (changes requested):
Blocking for now. Thanks!

xiaowuhu:
Not this guess, try another way.

PR #1071, opened 2023-09-21
New graph api prototype

titaiwangms:
Where we use PyTorch API: https://github.com/pytorch/pytorch/blob/main/torch/onnx/_internal/fx/fx_onnx_interpreter.py
The old API: https://github.com/microsoft/onnxscript/blob/main/onnxscript/function_libs/torch_lib/graph_building.py

Related issue and discussion: https://github.com/microsoft/onnx-converters-private/issues/158

The PR uses three component to build the graph: Tensor, Node, Graph.

Tensor:
- Add shape and dtype
- Add onnx_type
- Add `to_onnx_value_info()`
- TODO: Should we make `onnx_type` an object/class?

Node:
- inputs and outputs
- operators
- attributes
- onnx_function(optional)
- `to_node_proto()`

Graph:
- inputs and outputs
- initializers
- nodes
- function_store
- `to_onnx_model()`
- TODO: `load_from_onnx()`

TODO:
1. multiple outputs test
2. subgraph test
3. opset_imports
4. doc_string
5. Should we keep evaluator? We could use `add_node` to directly insert node into graph
6. graph sorting

### DIFF
--- a/onnxscript/function_libs/torch_lib/new_graph_prototype.py
+++ b/onnxscript/function_libs/torch_lib/new_graph_prototype.py
@@ -83,8 +83,8 @@ class Tensor:
             )
         self._name = name
         self._nparray = nparray
-        self._shape = shape or nparray.shape
-        self._dtype = dtype or nparray.dtype
+        self._shape = nparray.shape if nparray is not None else shape
+        self._dtype = nparray.dtype if nparray is not None else dtype
         # FIXME(justinhuby): Create a better way to determine the opset version
         self._opset: Any = opset or onnx_opset.opset18
         self._onnx_type = onnx_type
@@ -407,13 +407,15 @@ class Node:
 
 # TODO(titaiwang): How we deal with subgraph?
 class Graph:
-    def __init__(self) -> None:
+    def __init__(self, producer_name="pytorch") -> None:
         # All the functions used, deduplicated by name
         # key: (name, domain)
         self._function_store: Dict[Tuple[str, str], onnxscript.OnnxFunction] = {}
         self._nodes: List[Node] = []
         self._inputs: List[Tensor] = []
         self._outputs: List[Tensor] = []
+        self._initializer: List[Tensor] = []
+        self._producer_name: str = producer_name
         # TODO: what are these two string for?
         self._doc_string: str = ""
         self._name: str = ""
@@ -438,10 +440,39 @@ class Graph:
     def add_output(self, tensor: Tensor):
         self._outputs.append(tensor)
 
+    # TODO(titaiwang): this function is not friendly to users when
+    # they want to build the graph from scratch, as they need to
+    # manually set the value to initializers, so we need another
+    # method to get all initializers from the graph `self.initializers`
+    @runtime_typing.checked
+    def add_initializer(self, tensor: Tensor, name: str):
+        pass
+
+    def _get_constant_tensors(self) -> Dict[str, Tensor]:
+        input_const = [tensor for tensor in self._inputs if not tensor.is_fake]
+        output_const = [tensor for tensor in self._outputs if not tensor.is_fake]
+        node_related_const = [
+            tensor
+            for node in self._nodes
+            for tensor in (node.inputs + node.outputs)
+            if not tensor.is_fake
+        ]
+        return {
+            tensor.name: tensor for tensor in input_const + output_const + node_related_const
+        }
+
+    @property
+    def initializers(self) -> List[Tensor]:
+        return self._initializer or self._get_constant_tensors()
+
     def _to_onnx_graph(self) -> onnx.GraphProto:
         # convert tensor to tensor value info according to the onnx_type
         input_value_infos = [tensor.to_value_info() for tensor in self._inputs]
         output_value_infos = [tensor.to_value_info() for tensor in self._outputs]
+        # convert initializer
+        initializer = [
+            constant_tensor.to_value_int() for constant_tensor in self.initializers.values()
+        ]
         # convert node to node proto
         node_value_infos = [node.to_node_proto() for node in self._nodes]
         # convert graph to graph proto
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+    str,
+    int,
+    float,
+    bool,
+    None,
+]
+
+
+class GraphTensor(onnxscript_tensor.Tensor):
+    """An implementation of ONNX Tensors, based on a wrapper around numpy arrays.
```
gramalingam, line 45: Is this meant to be a wrapper around torch.Tensor?

onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+            self._shape = tuple(
+                dim if isinstance(dim, int) else str(dim) for dim in torch_tensor.size()
+            )
+        # FIXME(justinhuby): Create a better way to determine the opset version
+        self._opset: Any = opset or onnx_opset.opset18
+
+        # TODO: Type class?
+        # NOTE: This is used to determine the type of the tensor
+        # when we convert it to onnx graph
+        if onnx_type not in {"tensor_type", "sequence_type", "optional_type"}:
```
gramalingam, line 73: I am not sure about this. This sounds like an implicit coercion/cast of a tensor into a sequence of tensors or optional tensor. It may be simpler and better to make such cast explicit in the IR.

gramalingam:
I do have a very high-level question: what is the goal/intention of this IR? It would be good to have a list of use-cases and scenarios it is intended to support.

onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+    "GraphTensor",
+]
+
+
+ValidArgumentType: TypeAlias = Union[
+    "GraphTensor",
+    Sequence["GraphTensor"],
+    Sequence[float],
+    Sequence[int],
+    str,
```
gramalingam, line 36: Assuming this corresponds to AttributeProto, `Sequence[str]` is missing

onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+from onnxscript._internal import param_manipulation, runtime_typing
+
+__all__ = [
+    "Graph",
+    "GraphEvaluator",
+    "GraphTensor",
+]
+
+
+ValidArgumentType: TypeAlias = Union[
```
gramalingam, line 31: Maybe something like `Attribute` or `AttributeLike` would be a more meaningful name

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: Ok, on further looking at the code, it seems to me that this is meant to be a static representation of a runtime value. In that case, it would be better to rename the class (to something like `Value` or `Var`), and to _not_ inherit from Tensor.

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: Also needs a way to represent Graphs, for control-flow ops.

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: nit: the type-representation is incomplete. a sequence-type must describe the element-type too.

onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+            n_outputs=n_outputs,
+        )
+        self._function = function
+
+    @property
+    def function(self) -> onnxscript.OnnxFunction:
+        return self._function
+
+
+class ModuleNode(Node):
```
gramalingam, line 314: Not sure I understand this. Seems pytorch-specific.

gramalingam:
> I do have a very high-level question: what is the goal/intention of this IR? It would be good to have a list of use-cases and scenarios it is intended to support.

Just answering my question based on my understanding: looks like this is motivated by the needs of the trace-only onnxscript function construction, is that correct? (So, it's a replacement for what's in there for that purpose?

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: We either need equivalent classes to represent these types (or must use strings like "sequence(tensor(int64))" as used in ONNX, but classes may be better).

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: answer: not really. a "constant value" (represented by a torch.Tensor or numpy.array) is a special case of the more general "value" class that this corresponds to.

onnxscript/function_libs/torch_lib/new_graph_prototype.py
```diff
@@ -0,0 +1,626 @@
...
+
+###################################################################################################
+
+
+class Node:
+    def __init__(
+        self,
+        namespace: str,
+        op_name: str,
+        inputs: List[GraphTensor],
```
BowenBao, line 185: Why is `outputs: List[GraphTensor]` not part of initializer args?

justinchuby:
@shubhambhokare1 talked with me on evolving the PR. We think it would be possible to use a feature flag to control which implementation to use, and hide the new logic under the same apis. @shubhambhokare1 will look into doing this and enabling the torchlib tests. Thanks!

onnxscript/function_libs/torch_lib/new_graph_prototype.py
gramalingam: Specifically, inheriting from onnxscript's Tensor doesn't sound right here.

PR #1088, opened 2023-10-11
Fix randint dtypes | test(torchlib)

justinchuby:
Fix test input types for randint by replacing float input tests to integer inputs.


fatcat-z (approved):
LGTM, thanks!

Issue #1095, opened 2023-10-18
[torchlib] trace_ok


justinchuby:
Potentially allow a mode to evaluate if conditions on ranks to fold away if conditions.

- [x] Create the Rank operator for it to support both tracing and eager mode
- [x] Create a special case for Rank in the evaluator to return python values. Create a new opset for it. https://github.com/microsoft/onnxscript/pull/1178
- [x] Potentially fold away constant nodes with castlike as well https://github.com/microsoft/onnxscript/pull/1178
- [x] Create an option to control the trace behavior. We probably do not want it to be the default and should call it experimental https://github.com/microsoft/onnxscript/pull/1176
- [ ] #1218

https://github.com/microsoft/onnxscript/issues/1089

justinchuby:
Discussed with @BowenBao. With the passes Bowen has created we should be able to fold away the if branches and hit comparable performance.

Given this it may make more sense to follow a profile guided process to avoid premature optimization on the torchlib side. I will instead:

1. Still create the Rank op to simplify the functions and improve readability
2. Drive https://github.com/onnx/onnx/issues/5487
3. Create Gemm overloads to temporarily fix https://github.com/microsoft/onnxscript/issues/1089

thiagocrepaldi:
@gramalingam's work on constant folding as a post-export step might do the trick too, so it is fine to table trace_ok for now.

> Discussed with @BowenBao. With the passes Bowen has created we should be able to fold away the if branches and hit comparable performance.
Which passes are we talking about, specifically? Can you provide links to the code? ORT core mentioned they noticed graphs with several If nodes in it. I can check the code and compare with a graph and see what is up

> Given this it may make more sense to follow a profile guided process to avoid premature optimization on the torchlib side. I will instead:
> 1. Still create the Rank op to simplify the functions and improve readability
Do you foresee any possible issue constant folding this later? Will it create more branching?

> 2. Drive [[Feature request] field to represent shape/type for nodes within function onnx/onnx#5487](https://github.com/onnx/onnx/issues/5487)
Indeed this seems a high-value task to work on

> 3. Create Gemm overloads to temporarily fix [Linear from PyTorch must map to Gemm in ONNX #1089](https://github.com/microsoft/onnxscript/issues/1089)
[...]

Issue #834, opened 2023-07-06
[torchlib] Allow calling shared functions from other onnx functions


justinchuby:
Currently the torchscript evaluator does not support calling OnnxFunctions from within an OnnxFunction. We should enable this to allow shared implementations. To do this, we will need to keep references of the called functions inside an OnnxFunction.

BowenBao:
Done by #725 ?

justinchuby:
Ahh that’s right! I remember this pr but got a test error when I did this. I must have confused myself. Thanks!

justinchuby:
So we need to use `add_module_call` it seems? Looks like the graph builder is not handling cases where a function calls another (when using OnnxFunction `to_function_proto`).

I can think of two options:

1. Create a new API in OnnxFunction to return a list of function protos used by the function (recursively) (deduplication would be an issue)
2. Record the functions in the Graph on the fly
3. Do what to_model_proto does (I haven't looked)

BowenBao:
@justinchuby We don't need `add_module_call` for this, the rest makes sense. There is existing api `IRFunction.to_graph_and_functions` / `IRFunction.called_functions` used by `OnnxFunction.to_model_proto` to grab inner functions. I believe all we need is exposing the api at `OnnxFunction` level and use it in `TorchScriptGraph.add_function_call`.

PR #1105, opened 2023-10-23
Create the Rank and IsScalar shared functions | feat(torchlib)

justinchuby:
This change introduces two shared operators `Rank` and `IsScalar`. They are used to replace the `Size(Shape())` pattern for code reuse and readability.

I used a hack to always include these shared functions in the model proto because without #834 we cannot dynamically add these functions to the model as they are used. I added a TODO for this.

The first usage is in `aten_all`. I will update the rest of the functions in a separate PR.

#1095

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -728,7 +728,6 @@ class TorchScriptGraph:
             opset_imports=onnx_model.opset_import,
             doc_string=onnx_model.doc_string,
         )
-        # TODO: onnx.checker.check_function(onnx_function)?
         return onnx_function
 
     @runtime_typing.checked
@@ -811,6 +810,13 @@ class TorchScriptGraph:
                 onnx.helper.make_opsetid(domain, version)
                 for domain, version in unique_custom_domains.items()
             ]
+        )
+        # Include the library shared opset domain
+        # TODO: Remove after https://github.com/microsoft/onnxscript/issues/834 is fixed
+        onnx_model.opset_import.append(
+            onnx.helper.make_opsetid(
+                common_ops.common_opset.domain, common_ops.common_opset.version
+            )
         )
 
         try:
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -21,6 +21,7 @@ import onnxscript
 from onnxscript import evaluator
 from onnxscript import tensor as onnxscript_tensor
 from onnxscript._internal import param_manipulation, runtime_typing
+from onnxscript.function_libs.torch_lib.ops import common as common_ops
 
 __all__ = [
     "TorchScriptTensor",
@@ -361,6 +362,16 @@ def _tensor_rawdata_size(tensor: torch.Tensor) -> int:
         The estimated size of the tensor in bytes.
     """
     return tensor.numel() * tensor.element_size()
+
+
+def _shared_functions() -> list[onnx.FunctionProto]:
+    """Hack to always include the share ops."""
+
+    # TODO: Remove after https://github.com/microsoft/onnxscript/issues/834 is fixed
+    return [
+        common_ops.Rank.to_function_proto(),
+        common_ops.IsScalar.to_function_proto(),
+    ]
 
 
 class TorchScriptGraph:
@@ -786,6 +797,7 @@ class TorchScriptGraph:
             onnx_model = onnx.load_from_string(proto)
 
         onnx_model.functions.extend(function_proto_dict.values())
+        onnx_model.functions.extend(_shared_functions())
 
         # `_export_onnx` only exports opset_imports that is visible to it. It does not
         # export opset_imports for nested functions, since it does not have access to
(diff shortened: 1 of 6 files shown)


onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,40 @@
...
+        return Rank(x)
+
+    def IsScalar(self, x: tensor_typing.TTensor) -> BOOL:
+        return IsScalar(x)
+
+
+common_opset = CommonOpset()
+
+# TODO(justinchuby): How does onnxscript know what an op is a function?
+# Or do we need onnxscript to support calling functions from a different module?
```
justinchuby, line 26:
```suggestion
```

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,40 @@
...
+
```
justinchuby, line 27:
```suggestion
```

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,37 @@
...
+from onnxscript.function_libs.torch_lib import _constants, tensor_typing
+
+DOMAIN = f"{_constants.DOMAIN}.common"
+
+
+class CommonOpset(onnxscript.values.Opset):
+    def __new__(cls):
+        return onnxscript.values.Opset.__new__(cls, DOMAIN, 1)
+
+    def Rank(self, x: tensor_typing.TTensor) -> INT64:
```
justinchuby, line 16:
```suggestion
    def Rank(self, input: tensor_typing.TTensor) -> INT64:
```

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,37 @@
...
+        return Rank(x)
+
+    def IsScalar(self, x: tensor_typing.TTensor) -> BOOL:
```
justinchuby, line 19:
```suggestion
    def IsScalar(self, input: tensor_typing.TTensor) -> BOOL:
```

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -800,6 +811,13 @@ def to_model_proto(
                 for domain, version in unique_custom_domains.items()
             ]
         )
+        # Include the library shared opset domain
+        # TODO: Remove after https://github.com/microsoft/onnxscript/issues/834 is fixed
+        onnx_model.opset_import.append(
+            onnx.helper.make_opsetid(
```
BowenBao, line 817: Could we be double appending opsetid? (appended first here, then somewhere else where it is appended normally for torchlib functions)

BowenBao: nvm, I think we won't double append unless we register "common" domain ops in pytorch exporter registry.

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,37 @@
+"""Common operators shared in the torchlib library."""
+
+import onnxscript
+import onnxscript.values
+from onnxscript import BOOL, INT64
+from onnxscript import opset18 as op
+from onnxscript.function_libs.torch_lib import _constants, tensor_typing
+
+DOMAIN = f"{_constants.DOMAIN}.common"
```
BowenBao, line 9: Are we going to keep the domain name separate if #834 is fixed?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -320,8 +321,7 @@ def aten_align_to(self: TensorType, names: Sequence[str]) -> TensorType:
 def aten_all(self: TTensor) -> BOOL:
     """all(Tensor self) -> Tensor"""
 
-    self_rank = op.Size(op.Shape(self))
-    if self_rank == 0:
+    if common_opset.IsScalar(self):
```
BowenBao, line 324: nit: name bikeshedding. Should we import as `common_op`, keeping the convention of certain `opset` imported as `op`?

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class CommonOpset(onnxscript.values.Opset):
+    def __new__(cls):
+        return onnxscript.values.Opset.__new__(cls, DOMAIN, 1)
+
+    def Rank(self, input: tensor_typing.TTensor) -> INT64:
```
BowenBao, line 16: Why do these need to be defined explicitly for `CommonOpset`, while for `torchlib` a single decorator is enough?

onnxscript/function_libs/torch_lib/ops/common.py
justinchuby: I expect that because the common functions are not torch functions.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: I simplified the usage and removed common_opset reference.

onnxscript/function_libs/torch_lib/ops/common.py
justinchuby: This was a hack to allow both script and eager to work because `common.Rank` is currently rejected by the `script()` translator. I updated to alias `Rank` in the module level to circumvent this and removed this class.

justinchuby:
@BowenBao PTAL

BowenBao (approved):

PR #1108, opened 2023-10-24
Use Rank and IsScalar | feat(torchlib)

justinchuby:
Use the common Rank and IsScalar functions in torchlib.

Follow up of https://github.com/microsoft/onnxscript/pull/1105


justinchuby:
Blocked by https://github.com/onnx/onnx/issues/5701

titaiwangms (approved):

Issue #1089, opened 2023-10-11
Linear from PyTorch must map to Gemm in ONNX


baijumeswani:
PyTorch Model:

```python
class NeuralNet(torch.nn.Module):
    def __init__(self, input_size, hidden_size, num_classes):
        super().__init__()

        self.fc1 = torch.nn.Linear(input_size, hidden_size)
        self.relu = torch.nn.ReLU()
        self.fc2 = torch.nn.Linear(hidden_size, num_classes)

    def forward(self, input1):
        out = self.fc1(input1)
        out = self.relu(out)
        out = self.fc2(out)
        return out
```

Exporting with torch script based exporter yields:

![image](https://github.com/microsoft/onnxscript/assets/12852605/dd087259-a202-4b7a-ab71-c55d54d68c7f)

which makes sense. It is after all a linear layer followed by a ReLU followed by another Linear layer.

Exporting the same model with torch dynamo based exporter yields:

<img width="263" alt="image" src="https://github.com/microsoft/onnxscript/assets/12852605/f7eb9abb-ba67-4f01-a25a-8e287f83cf72">

Two levels beneath the linear layer, I find:

<img width="380" alt="image" src="https://github.com/microsoft/onnxscript/assets/12852605/03751eb6-3400-4a46-8d04-cb88e0f2012b">

It seems like the Gemm is somehow manifested as a subgraph with matmuls, muls, adds, and castlikes. And digging deeper, I find that this definition comes from https://github.com/microsoft/onnxscript/blob/a981b8add9a4c7e67ad0d28622b23d4c6a55a76e/onnxscript/function_libs/torch_lib/ops/core.py#L220-L229

It seems wasteful that an op as simple as a Gemm needs to be represented as this subgraph. Looking at [this document](https://github.com/microsoft/onnxscript/issues/165), this seems to be a design choice.

> We favor general ops like MatMul than specialized ops like Gemm in the function lib.

But imagine a model having thousands of `Gemm`s. Each `Gemm` is now this subgraph. Which means this optimization/fusion needs to run thousands of times to achieve something that probably can be achieved very easily at the source.

It would benefit ONNX Runtime (inference and training) and the larger ONNX community if this subgraph were represented as a Gemm node after export.

baijumeswani:
cc: @BowenBao @justinchuby

justinchuby:
[...]

PR #1111, opened 2023-10-24
Decompose addmm with Gemm | feat(torchlib)

justinchuby:
Decompose addmm with Gemm by creating a special variant for `FLOAT` and conditionally check for the ranks if the input tensors. The if branch is expected to be folded away by constant folding passes.

I have not found other instances where Gemm is used in the torch.onnx exporter.

Fixes https://github.com/microsoft/onnxscript/issues/1089

cc @baijumeswani

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -282,6 +282,7 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
             param = name_to_schema[name]
             # Cast int to float if needed
             if param.type in {float, "float"}:
+                print(name, param.type)
                 # FIXME(justinchuby): Create invariant on the type of param.type to simplify this
                 attributes[name] = float(value)
         return self._graph.add_function_call(function, inputs, attributes)
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -282,7 +282,6 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
             param = name_to_schema[name]
             # Cast int to float if needed
             if param.type in {float, "float"}:
-                print(name, param.type)
                 # FIXME(justinchuby): Create invariant on the type of param.type to simplify this
                 attributes[name] = float(value)
         return self._graph.add_function_call(function, inputs, attributes)
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -12,7 +12,7 @@
 from __future__ import annotations
 
 import math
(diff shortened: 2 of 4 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -282,6 +282,7 @@ def eval_function(  # type: ignore[override]
             param = name_to_schema[name]
             # Cast int to float if needed
             if param.type in {float, "float"}:
+                print(name, param.type)
```
justinchuby, line 285:
```suggestion
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -222,14 +235,42 @@ def aten_addcmul(
...
+    alpha: float = 1.0,
+) -> TRealUnlessFloat32:
     """addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor"""
 
     mat1_mat2 = op.MatMul(mat1, mat2)
     scaled_mat1_mat2 = op.Mul(mat1_mat2, alpha)
     scaled_self = op.Mul(self, beta)
-    return op.Add(scaled_self, scaled_mat1_mat2)
+    result = op.Add(scaled_self, scaled_mat1_mat2)
+    return result
```
justinchuby, line 250:
```suggestion
    return op.Add(scaled_self, scaled_mat1_mat2)
```

BowenBao:
`aten::mm` is also exported as Gemm by torchscript exporter. Although I'm not sure if it is better to export as Gemm or Matmul in this case.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -232,6 +249,29 @@ def aten_addmm(
     return op.Add(scaled_self, scaled_mat1_mat2)
 
 
+@torch_op("aten::addmm")
+def aten_addmm_gemm(
+    self: FLOAT, mat1: FLOAT, mat2: FLOAT, beta: float = 1.0, alpha: float = 1.0
```
BowenBao, line 254: Why only do 'gemm' variant on float inputs?

justinchuby:
> `aten::mm` is also exported as Gemm by torchscript exporter. Although I'm not sure if it is better to export as Gemm or Matmul in this case.

Yeah Gemm seems to be more specialized for that case. Backends should just do the right thing for matmul imo

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: ORT implements Gemm for float inputs only. So I set it accordingly for practicality

justinchuby: Changed to TFloat because it supports float16 as well

BowenBao (approved):
LGTM, thanks!

PR #1112, opened 2023-10-25
Experimental flag for controlling initializers as inputs | feat(torchlib)

justinchuby:
Create a flag `TORCHLIB_EXPERIMENTAL_INITIALIZERS_AS_INPUTS=1` to mark initializers as inputs to the model.

**This flag is experimental only for ONNX Runtime training and should not be assumed to exist in production.**

Reference: https://github.com/microsoft/onnx-converters-private/issues/182

Tested with the open-llama model with `transformers==4.31.0` with script https://gist.github.com/abock/2115d34d98df15a77516e8a2899b121c

image


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -749,14 +750,17 @@ def to_model_proto(
 
         large_model = initializers_size > _LARGE_MODEL_SIZE_THRESHOLD
 
+        # TODO(justinchuby): If initializers is empty, do we still know which ones are initializers? Should we use the defer_weight_export argument?
```
justinchuby, line 753:
```suggestion
```

We do

BowenBao (commented):
I feel like though this requirement is more naturally satisfied on exporter side by adding an option that allows adjusting the `include_initializer` flag that decides this, wdyt?

onnxscript/function_libs/torch_lib/_flags.py
```diff
@@ -0,0 +1,11 @@
+"""Experimental flags.
+
+NOTE: These flags are experimental only. Any flag here can be removed at any
+time without notice.
+"""
+
+import os
+
+EXPERIMENTAL_INITIALIZERS_AS_INPUTS: bool = (
```
BowenBao, line 9: nit: add a comment explaining the behavior?

onnxscript/function_libs/torch_lib/_flags.py
```diff
@@ -0,0 +1,11 @@
+"""Experimental flags.
+
+NOTE: These flags are experimental only. Any flag here can be removed at any
```
BowenBao, line 3: Should we emit warning whenever flag uses under this file is detected?

justinchuby: Good idea

BowenBao: And raise error when this flag is removed but still used by user.

BowenBao (approved):

onnxscript/function_libs/torch_lib/_flags.py
justinchuby: Done. Added as string. PTAL

onnxscript/function_libs/torch_lib/_flags.py
justinchuby: Done

onnxscript/function_libs/torch_lib/_flags.py
```diff
@@ -0,0 +1,41 @@
...
+    deprecated: bool = False,
+) -> bool:
+    """Load a boolean flag from environment variable.
+
+    Args:
+        name: The name of the environment variable.
+        this_will: A string that describes what this flag will do.
+        deprecated: Whether this flag is deprecated.
+    """
+    state = os.getenv(name) == 1
```
justinchuby, line 26:
```suggestion
    state = os.getenv(name) == "1"
```

BowenBao (approved):

justinchuby:
Thanks for reviewing!

Issue #1110, opened 2023-10-24
[torchlib][tracking] Handle Gemm with a graph rewriting pass


justinchuby:
Handle Gemm with a graph rewriting pass instead of hard coding the branching logic

https://github.com/microsoft/onnxscript/issues/1089

PR #1113, opened 2023-10-25
Use Gemm to implement addmm | fix(torchlib)

justinchuby:
When I looked at the test coverage for `addmm` (below), I realized mat1 and mat2 are always 2d tensors. So the rank check is redundant. `addmm` is now fully mapped to `Gemm`, which should completely resolve https://github.com/microsoft/onnxscript/issues/1089

Closes https://github.com/microsoft/onnxscript/issues/1110

image

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -222,37 +222,12 @@ def aten_addcmul(
 
 @torch_op("aten::addmm")
 def aten_addmm(
-    self: TInt, mat1: TInt, mat2: TInt, beta: float = 1.0, alpha: float = 1.0
-) -> TInt:
-    """addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor"""
-
-    mat1_mat2 = op.MatMul(mat1, mat2)
-    scaled_mat1_mat2 = op.Mul(mat1_mat2, alpha)
-    scaled_self = op.Mul(self, beta)
-    return op.Add(scaled_self, scaled_mat1_mat2)
-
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -222,39 +222,27 @@ def aten_addcmul(
...
+
+
+@torch_op("aten::addmm")
+def aten_addmm_int(
     self: TInt, mat1: TInt, mat2: TInt, beta: float = 1.0, alpha: float = 1.0
 ) -> TInt:
     """addmm(Tensor self, Tensor mat1, Tensor mat2, *, Scalar beta=1, Scalar alpha=1) -> Tensor"""
 
     mat1_mat2 = op.MatMul(mat1, mat2)
+    # NOTE: The following Mul should be optimized away by a pass when alpha and beta are 1.0
```
BowenBao, line 240: Let's comment it's because onnx::gemm not supporting int type?

BowenBao (approved):

onnxscript/function_libs/torch_lib/ops/core.py
BowenBao: That's my preference too, but I don't have strong opinion either way.

gramalingam:
This looks good. But curious, since the PR description mentions test-cases ... are we dependent on test-cases for figuring out spec aspects (like broadcasting)? It isn't documented anywhere?

justinchuby:
It is documented in the torch documentation https://pytorch.org/docs/stable/generated/torch.addmm.html. The old exporter treats this as a special case because we have received reports of rank>2 inputs. I suppose we don’t get these inputs anymore.

PR #1121, opened 2023-11-01
Add Op(aten_assert_async) | feat(torchlib)

titaiwangms:
This is an op invoked whenever assertion is used in models, but with no output.

~~NOTE: Converter side needs extra handles on shape/type filling, as the op doesn't contain meta['val']~~

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -853,7 +853,9 @@ def aten_asinh(self: TFloat) -> TFloat:
 
 
 @torch_op("aten::_assert_async.msg")
-def aten_assert_async(self: TTensor) -> TTensor:
+def aten_assert_async(
+    self: TTensor, assert_msg: str = ""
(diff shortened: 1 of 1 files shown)


titaiwangms:
Actually I am not sure we want this to be an op or we want a pass to delete all of them.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -852,6 +852,15 @@ def aten_asinh(self: TFloat) -> TFloat:
     return op.Asinh(self)
 
 
+@torch_op("aten::_assert_async.msg")
+def aten_assert_async(
+    self: TTensor, assert_msg: str = ""
+) -> TTensor:  # pylint: disable=unused-argument`
+    """_assert_async.msg(Tensor self, str assert_msg) -> ()"""
```
BowenBao, line 859: Signature says this function does not have an output

titaiwangms: hmmm maybe we should delete it.

PR #1127, opened 2023-11-06
Fix var_mean when input dim is a list | fix(torchlib)

justinchuby:
Previously `var_mean.dim` handles dim only when it is a tuple. It can be a list as well in practice.


BowenBao (approved):

justinchuby:
Will follow up with tests

PR #1134, opened 2023-11-07
Implement `transpose` for complex inputs | feat(torchlib)

justinchuby:
Handle negative dimensions when transposing complex inputs.


titaiwangms (approved):
Should the original one also handle negative axis?

justinchuby:
> Should the original one also handle negative axis?

It does already

PR #1135, opened 2023-11-07
Add Op(built-in function ops) | feat(torchlib)

titaiwangms:
Fix https://github.com/microsoft/onnx-converters-private/issues/190

With https://github.com/pytorch/pytorch/pull/112758, this PR moves built-in function ops mapping into torchlib.
NOTE: module name of `operator.add` is _operator

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -124,7 +124,7 @@ def aten__softmax(
     return aten_softmax_no_dtype(self, dim)
 
 
-@torch_op("aten::abs")
+@torch_op(("aten::abs", "_operator::abs"))
 def aten_abs(self: TRealOrUInt8) -> TRealOrUInt8:
     """abs(Tensor self) -> Tensor"""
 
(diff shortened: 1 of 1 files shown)


justinchuby:
LGTM. Cc @BowenBao for namespace naming conventions

justinchuby:
I am thinking something like python.operator and python.math may be more principled

titaiwangms:
> I am thinking something like python.operator and python.math may be more principled

A different domain from DOMAIN = "pkg.onnxscript.torch_lib"? I don't know if we should mark the difference in the graph, as they are the same (eg: operator.add and aten::add) in ONNX perspective.

justinchuby:
I meant `python.operator::add` or `python.math::div`

titaiwangms:
> I meant `python.operator::add` or `python.math::div` to denote they are python built-in values and avoid potential name conflicts (although unlikely)

In that case, string processing would need to live somewhere in torchlib registration or converter registration to match _operator.add (original name).

justinchuby:
SG. That's a good argument to keep the existing names

justinchuby (approved):

PR #1139, opened 2023-11-08
Implement squeeze.dim for complex inputs | feat(torchlib)

justinchuby:
Discovered through https://github.com/pytorch/pytorch/issues/113067


titaiwangms:
hmmm I don't see a missing op request from that issue?

justinchuby:
That happens when there is not a batch dim

titaiwangms (approved):

PR #1147, opened 2023-11-11
Create shared function to handle complex casting | fix(torchlib)

justinchuby:
https://github.com/pytorch/pytorch/issues/113444 Exposed an error where when `dtype=torch.complex64` in `aten_scalar_tensor_sym_number`, we produce a node with it output being complex. This change extracts the logic from `prims::convert_element_type` to a common function to handle complex dtypes.

Adds complex test for `scalar_tensor`.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7042,7 +7042,7 @@ def aten_rsub_complex(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
 
 
 @torch_op("aten::scalar_tensor", trace_only=True)
-def aten_scalar_tensor(s: float, dtype: int = FLOAT.dtype) -> TTensor:  # type: ignore[type-var]
+def aten_scalar_tensor(s: float, dtype: int = FLOAT.dtype) -> RealType:
     """scalar_tensor(Scalar s, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
 
     # Set trace_only=True because different if branches return different dtypes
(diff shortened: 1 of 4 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7041,20 +7041,22 @@ def aten_rsub_complex(self: TReal, other: TReal, alpha: float = 1.0) -> TReal:
...
+    # Set trace_only=True because different if branches return different dtypes
+    # which is not supported in an ONNX function
+    return common_ops.cast_to(s, dtype=dtype)
 
 
-@torch_op("aten::scalar_tensor")
-def aten_scalar_tensor_sym_number(
-    s: Union[FLOAT, INT32, BOOL], dtype: int = FLOAT.dtype
-) -> TTensor:
+@torch_op("aten::scalar_tensor", trace_only=True)
```
titaiwangms, line 7053: Have you tried it with dispatcher yet? I don't think this is supported... hmm.

titaiwangms: ONNX variants part

justinchuby: I did. Which was not supported?

titaiwangms: I thought the dispatcher wouldn't be able to tell the two aten_scalar_tensor variants apart.

justinchuby: It should be able to I think? Because the opschema is not affected in this change. One has attribute only and the other has one input and a single attribute

titaiwangms: I see

PR #1166, opened 2023-11-18
Handle name conflict (attribute parameter) when converting back to python

gramalingam:
Attribute parameters and normal values (like tensors) use the same namespace in onnxscript/python. This can cause a conflict when converting ONNX proto back to onnxscript (since these are different namespaces in ONNX). Examples of where this happens shown in test-case below: eg., when an attribute-parameter "yyy" is used as a value, the onnxscript translator introduces a value called "yyy" which is bound to the attribute-value "yyy".

Fix this in the exporter.

### DIFF
--- a/onnxscript/backend/onnx_export.py
+++ b/onnxscript/backend/onnx_export.py
@@ -25,15 +25,7 @@ from onnxscript.onnx_types import {{ ", ".join(unique_types) }}
 {%- endif %}
 {{translate_opset_imports_of(main_model)}}
 {% for domain, name, fct in functions: %}
-@script({{make_opset_name(domain, 1)}})
-def {{ python_make_node_name(fct['proto'].domain, 1, fct['proto'].name) }}{{
-    translate_function_signature(fct['proto'])}}
-    {% if fct['proto'].doc_string %}"""
-    {{ fct['proto'].doc_string }}
-    """{%- endif %}
-    {%- for node in fct['proto'].node: %}
-{{ python_make_node(node, opsets, indent=1) }}{% endfor %}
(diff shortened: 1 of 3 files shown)


onnxscript/backend/onnx_export.py
```diff
@@ -199,14 +191,70 @@ def _attribute_value(attr: onnx.AttributeProto):
...
+
+def _update_names_used_in_function(names: set[str], fun: FunctionProto) -> None:
+    names.update(fun.input)
+    names.update(fun.output)
+    for node in fun.node:
+        _update_names_used_in_node(names, node)
+
+
+def _names_used_in_function(fun: FunctionProto) -> set[str]:
+    names = set()
```
justinchuby, line 221:
```suggestion
    names: set[str] = set()
```

justinchuby (approved):

onnxscript/backend/onnx_export.py
```diff
@@ -221,7 +270,9 @@ def make_opset_name(self, domain, version):
         return f"{self._rename_domain(domain)}{version}"
 
     def _python_make_node_name(self, domain, version, name, node=False):
-        name = _rename_variable(name)
+        name = _rename_variable(
+            name
+        )  # TODO: Is this a typo? Is it supposed to be self._rename_variable(name)?
```
justinchuby, line 275: Need to fix this?

gramalingam: I added that question ... I am not sure, need to look at it closely. I have another followup PR with other cleanup and will then get to this.

Issue #1167, opened 2023-11-18
Onnx to onnxscript conversion errors


gramalingam:
A few different issues show up when converting the model below back to onnxscript: (TODO: not sure how to attach an ONNX file here.)

The issues relate to:
* name conflicts between attribute-parameters and corresponding values (obtained by promotion from attribute-parameter).
* Some misalignment of loop-carried dependences for the loop operator (looks like they are off-by-one).

justinchuby:
The onnx file can be zipped

gramalingam:
[stft.zip](https://github.com/microsoft/onnxscript/files/13398649/stft.zip)
Model attached

PR #1168, opened 2023-11-20
Eliminate use of template in exporter

gramalingam:
Some cleanup of the exporter before fixing issue https://github.com/microsoft/onnxscript/issues/1167. This eliminates the use of template in the exporter (which was harder to understand, in particular to debug).

### DIFF
--- a/onnxscript/backend/onnx_export.py
+++ b/onnxscript/backend/onnx_export.py
@@ -560,7 +560,7 @@ class Exporter:
         add(f"def {function_name}{_translate_signature(graph.input, graph.output)}")
         if doc:
             add(f'    """{doc}"""')
-        self._python_make_node_graph(graph, self.opsets, indent=1)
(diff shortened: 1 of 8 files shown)


onnxscript/backend/onnx_export.py
```diff
@@ -199,14 +182,76 @@ def _attribute_value(attr: onnx.AttributeProto):
     raise NotImplementedError(f"Unable to return a value for attribute {attr!r}.")
 
 
+def _update_names_used_in_graph(names: set[str], graph: GraphProto) -> None:
+    """Returns the names used in a graph."""
```
justinchuby, line 186: This returns None. Maybe "Adds the named used in a graph to the set names." ?

justinchuby (approved):

onnxscript/backend/onnx_export.py
gramalingam: Done

PR #1172, opened 2023-11-21
[Experimental] Export with symbolic shapes

BowenBao:
Experimental feature to store symbolic shapes produced by torch dynamo inside the exported onnx model.
There is no official ONNX spec to support nodes within FunctionProto to have value info, https://github.com/onnx/onnx/issues/5487. The names for value info are generated uniquely to be retrievable based on the call site and call stack.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -93,8 +93,6 @@ class TorchScriptTensor(onnxscript_tensor.Tensor):
     def __init__(
         self,
         value: torch.Value,
-        shape: Optional[torch.Size] = None,
-        dtype: Optional[torch.dtype] = None,
     ):
         super().__init__(None)
         self._torch_value: torch.Value = value
@@ -161,7 +159,7 @@ class TorchScriptTensor(onnxscript_tensor.Tensor):
         # Normalize torch symbolic dimension size to str.
         torch_sym_types = (torch.SymInt, torch.SymFloat, torch.SymBool)
         self._shape = tuple(
-            str(dim.node) if isinstance(dim, torch_sym_types) else dim for dim in shape
+            str(dim.node) if isinstance(dim, torch_sym_types) else dim for dim in shape  # type: ignore[union-attr]
         )
         # jit api does not support assigning symbolic shapes,
         # hence symbols are replaced as None.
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -692,6 +793,41 @@ def add_module_call(
             n_outputs=sub_torch_script_graph.num_outputs,
         )
 
+    @runtime_typing.checked
+    def generate_function_value_info_proto(
```
justinchuby, line 797: Nice!

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -208,6 +208,14 @@ def symbolic_value(self) -> torch.Value:
         """The symbolic Value in torch.Graph."""
         return self._torch_value
 
+    def value_info(self) -> Optional[onnx.ValueInfoProto]:
+        try:
+            dtype = self.onnx_dtype.value
+        except torch.onnx.errors.OnnxExporterError:
```
justinchuby, line 214: Curious: why do we get OnnxExporterError?

BowenBao: Got it from below in torch/onnx/_type_utils.py
```
        if isinstance(value.type(), torch.ListType):
            try:
                return cls.from_dtype(value.type().getElementType().dtype())
            except RuntimeError:
                return cls._from_name(str(value.type().getElementType()))
```

BowenBao:
@justinchuby are the torch nightly CI failures safe to ignore?

justinchuby:
Yes. I will create a fix. Thanks!

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Any chance we can take the logic out and remove dependence on type utils?

BowenBao: I'm not following, would you prefer a general exception class? In general, we want this method to return None if dtype is not retrievable or undefined.

justinchuby: I am just thinking we should move away from using `_type_utils.JitScalarType` from torch.onnx. But this can be done separately

PR #1176, opened 2023-11-22
Define the EXPERIMENTAL_PREFER_TRACING flag

justinchuby:
* (to be filled)

### DIFF
--- a/onnxscript/function_libs/torch_lib/registration.py
+++ b/onnxscript/function_libs/torch_lib/registration.py
@@ -99,6 +99,7 @@ def torch_op(
     trace_only: bool = False,
     private: bool = False,
     complex: bool = False,
+    traceable: bool = False,
 ) -> Callable[[FunctionType], onnxscript.OnnxFunction | onnxscript.values.TracedOnnxFunction]:
     """Register a torch op.
 
@@ -112,6 +113,7 @@ def torch_op(
         private: Whether the function is private (not directly exposed). It should
             be true for all functions with names starting with "_".
         complex: Whether the function expects complex-valued inputs.
+        traceable: Whether the function can be traced.
     """
     if registry is None:
         registry = default_registry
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/registration.py
```diff
@@ -112,6 +113,7 @@ def torch_op(
         private: Whether the function is private (not directly exposed). It should
             be true for all functions with names starting with "_".
         complex: Whether the function expects complex-valued inputs.
+        traceable: Whether the function can be traced.
```
BowenBao, line 116: I don't understand this comment. Could you elaborate?

justinchuby: A function is traceable if it can both be scripted and traced and produce the same result for a given input. Does that sound reasonable? I will add this to the docstring if it does.

justinchuby: - A function can be tagged with traceable if its if branches (if any) can be statically evaluated. 
- A function should be tagged with traceable if it contains if branches and/or CastLike nodes. 
- A function without if branches or CastLike nodes _should not_ be tagged with traceable because inlining will do the same thing.
- A function with `@graph` defined for a `Scan` op is not traceable yet.

justinchuby: Done

titaiwangms (approved):

onnxscript/function_libs/torch_lib/registration.py
titaiwangms: Is there any chance one could wrongly tag a function traceable, and create a bug?

justinchuby: There is a chance. We should be able to catch them in the op info tests.

PR #1178, opened 2023-11-22
Implement the experimental evaluator for folding branches and castlikes | feat(torchlib)

justinchuby:
* __->__ #1178

As an effort described in https://github.com/microsoft/onnxscript/issues/1095, this PR

- Implements the experimental evaluator for folding branches and castlikes so that they are eagerly evaluated when possible.
- Updates implementation for `addr` for it to be traceable.
- Conditionally enabled previously xfailed tests.

Set `TORCHLIB_EXPERIMENTAL_PREFER_TRACING=1` and tested in CI.

E.g. clamp_min now becomes

```
<
   ir_version: 8,
   opset_import: ["" : 18, "pkg.onnxscript.torch_lib.common" : 1],
   producer_name: "pytorch",
   producer_version: "2.2.0"
>
main_graph (int32[5,10,5] input_0, int32[10,5] input_1) => (int32[5,10,5] _val_9) 
   <int64 _val_2, int64[2] _val_3, int64 _val_4, int64 _val_5, bool _val_6, int64 _val_7, bool _val_8>
{
   _val_2 = Size (input_0)
   _val_3 = Shape <start: int = 0> (input_1)
   _val_4 = Size (_val_3)
   _val_5 = Constant <value: tensor = int64 {0}> ()
   _val_6 = Equal (_val_2, _val_5)
   _val_7 = Constant <value: tensor = int64 {0}> ()
   _val_8 = Equal (_val_4, _val_7)
   _val_9 = Max (input_0, input_1)
}
<
  domain: "pkg.onnxscript.torch_lib.common",
  opset_import: ["" : 18]
>
Rank (input) => (return_val)
{
   tmp = Shape (input)
   return_val = Size (tmp)
}
<
  domain: "pkg.onnxscript.torch_lib.common",
  opset_import: ["" : 18]
>
IsScalar (input) => (return_val)
{
   tmp = Shape (input)
   tmp_0 = Size (tmp)
   tmp_1 = Constant <value_int: int = 0> ()
   return_val = Equal (tmp_0, tmp_1)
}
```

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -265,7 +265,7 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
             if schema.name == "CastLike":
                 assert len(inputs) == 2
                 # Skip CastLike if the input and output types are the same
-                if inputs[0].dtype == inputs[1].dtype is not None:
+                if inputs[0].dtype == inputs[1].dtype and inputs[1].dtype is not None:
                     return inputs[0]
         return self._graph.add_op_call(schema, inputs, attributes)
 
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -90,11 +90,14 @@ def _rename_intermediate_value(name: str) -> str:
 class TorchScriptTensor(onnxscript_tensor.Tensor):
     """A onnxscript tensor that wraps a torchscript Value."""
 
-    def __init__(self, value: torch.Value):
+    def __init__(
+        self,
+        value: torch.Value,
+    ):
         super().__init__(None)
         self._torch_value: torch.Value = value
         self._concrete_value: Optional[np.ndarray] = None
-        self._shape: Optional[Tuple[int | None, ...]] = None
+        self._shape: Optional[Tuple[int | str | None, ...]] = None
         self._torch_dtype: Optional[torch.dtype] = None
         self._name: Optional[str] = None
         self._is_complex: bool = False
@@ -125,6 +128,9 @@ class TorchScriptTensor(onnxscript_tensor.Tensor):
 
     @property  # type: ignore[override]
     def rank(self) -> int | None:
+        if self._shape is not None:
+            return len(self._shape)
+
         value_type = self._torch_value.type()
         if value_type is None:
             return None
@@ -132,7 +138,7 @@ class TorchScriptTensor(onnxscript_tensor.Tensor):
         return value_type.dim()
 
     @property  # type: ignore[override]
-    def shape(self) -> Tuple[int | None, ...] | None:
+    def shape(self) -> Tuple[int | str | None, ...] | None:
         if self._shape is not None:
             return self._shape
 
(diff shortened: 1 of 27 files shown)


justinchuby:
Change cast likes to cast

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -303,6 +314,40 @@ def eval_function(  # type: ignore[override]
...
+                if isinstance(args[0], TorchScriptTensor):
+                    if args[0].rank is not None:
+                        return args[0].rank == 0
+                    else:
+                        # Fall to call the function
+                        pass
+                else:
+                    # Python constants are scalars
+                    return True
+            if function.name == "Rank":
```
titaiwangms, line 333: Because rank is immutable, we can return int?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -303,6 +314,40 @@ def eval_function(  # type: ignore[override]
...
+            if function.name == "IsScalar":
+                if len(args) != 1:
+                    raise TypeError(
+                        f"Expected 1 positional argument for function '{function}', got {len(args)}."
+                    )
+                if isinstance(args[0], TorchScriptTensor):
+                    if args[0].rank is not None:
+                        return args[0].rank == 0
+                    else:
+                        # Fall to call the function
```
titaiwangms, line 328: Is it `Fall to `add_function_call`?

onnxscript/function_libs/torch_lib/graph_building.py
titaiwangms: Tracing is scattered in the codebase. I am a bit worried if it would become a debugging problem of dynamic/static for future?

justinchuby: Do you mean the `traceable` flag? Only the functions that are safe to trace are marked as traceable. Do you foresee any issues?

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Good catch thanks

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Yes. The assumption is that Rank is going to be static. This should be true even if we have dynamic axis. If that's not the case we can remove this experiment or find other ways around it (e.g. using the torch.export constraints).

justinchuby:
`addr_cpu_float16 - AssertionError: Tensor-likes are not close` need to adjust precision.

onnxscript/function_libs/torch_lib/graph_building.py
titaiwangms: I guess it's the complexity of the flag bringing us. I think we are trying to keep "scripted as much as it can". However, the node modifications now could happen here, converter(filter_args_XXX), and torchlib. I wonder if we have a plan to make it simpler, so whenever we have an unexpected static axe model issue, we have clear targets where the code might be.

titaiwangms (approved):

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: Resolved offline. We agreed that this is an experimental feature that will be removed or adjusted in the future.

PR #1189, opened 2023-11-28
ModifyOp(embedding_bag) | feat(torchlib)

xiaowuhu:
Remove registery (_embedding_bag) from aten_embedding_bad_padding_idx() function, because ```embedding_bag``` should not has the parameter ```padding_idx```.


justinchuby:
https://github.com/pytorch/pytorch/blob/0bef97fac35bdffc15a4c142bae6a5516799f862/aten/src/ATen/native/native_functions.yaml#L2307

Issue #1191, opened 2023-11-29
Clean up isinstance checks for tuples in trace only functions


justinchuby:
There are cases where PyTorch accepts an integer or a single element sequence in an op. In ONNX the attribute has to be either an integer or a sequence of integer. We cannot annotate as a union because the op schema will not support that. Right now how we do it is annotate the type as Sequence, then use an isinstance check to convert the int into a tuple as needed. This prevents the function from being scripted. 

Instead, we should just treat the input as a sequence all the way through. When the converter supplied an integer that should have been a sequence, we can simply adapt it in the evaluator like we do for int->float conversion.

PR #1192, opened 2023-11-29
Clean up isinstance checks for tuples | feat(torchlib)

justinchuby:
Clean up isinstance checks for tuples to turn some trace_only functions into script functions. This is done by updating the tracing evaluator to turn int attributes into sequences according to the function op schema.

Remove xfails for var_mean.

Fixes https://github.com/microsoft/onnxscript/issues/1191

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -374,7 +374,6 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
         assert op_schema is not None
         for name, value in attributes.items():
             attribute = op_schema.attributes[name]
-            # Cast int to float if needed
             if attribute.type == onnx.defs.OpSchema.AttrType.FLOAT:
                 # Cast int to float if the attribute is FLOAT
                 attributes[name] = float(value)
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -368,13 +368,22 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
         ) = param_manipulation.separate_input_attributes_from_arguments(
             param_schemas, args, kwargs, fill_defaults=True, allow_extra_kwargs=True
         )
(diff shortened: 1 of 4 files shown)


justinchuby:
Windows: Use the helper to cast to int64

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -368,13 +368,22 @@ def eval_function(  # type: ignore[override]
...
-            param = name_to_schema[name]
+            attribute = op_schema.attributes[name]
             # Cast int to float if needed
-            if param.type in {float, "float"}:
-                # FIXME(justinchuby): Create invariant on the type of param.type to simplify this
+            if attribute.type == onnx.defs.OpSchema.AttrType.FLOAT:
+                # Cast int to float is the attribute is FLOAT
                 attributes[name] = float(value)
+            if attribute.type == onnx.defs.OpSchema.AttrType.INTS and isinstance(value, int):
+                attributes[name] = (value,)
```
BowenBao, line 382: when would this happen?

BowenBao (approved):
LGTM w/ inline questions.

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -368,13 +368,22 @@ def eval_function(  # type: ignore[override]
...
         )
-        name_to_schema = {param.name: param for param in param_schemas}
+
+        # Cast attributes to the correct type based on function signature
+        op_schema = function.op_schema
+        assert op_schema is not None
         for name, value in attributes.items():
-            param = name_to_schema[name]
+            attribute = op_schema.attributes[name]
             # Cast int to float if needed
```
justinchuby, line 377:
```suggestion
```

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: In PyTorch, attributes annotated as `int[1]?` accepts an integer or a sequence. When the attribute is an integer, it is treated as a single element sequence. ONNX requires an attribute to either be an integer or a sequence. So we promote the value to a sequence here.

Added as a comment too.

PR #1193, opened 2023-11-29
Remove proto2text | chore!

justinchuby:
Remove proto2text because it is simply an alias of `onnx.printer.to_text`.


titaiwangms (approved):
There was one CI fail before I updated the branch.

justinchuby:
There's one issue with experimental-torchlib-tracing unrelated to this PR. I will create a fix. Thanks!

PR #1194, opened 2023-11-30
Implement aten_inverse with the microsoft operator | feat(torchlib)

justinchuby:
Implement aten_inverse using the microsoft opset Inverse.


justinchuby:
Closed because we only want to use ONNX domain operators in torchlib. This can be used as an example for supporting custom operators.

PR #1203, opened 2023-12-05
AddOp(aten_replication_pad1d) | feat(torchlib)

xiaowuhu:
- A large change in extra_opinfo.py, is just because re-sort the functions by name.


justinchuby:
Thank you! I will fix the lint errors

justinchuby (approved):

PR #1204, opened 2023-12-05
Fix Op(aten::_scaled_dot_product_efficient_attention) under fp16

BowenBao:
Insert castlike where needed.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -1908,8 +1908,8 @@ def _aten_scaled_dot_product_attention_no_mask_onnx(
 
     # https://github.com/pytorch/pytorch/blob/12da0c70378b5be9135c6fda62a9863bce4a4818/aten/src/ATen/native/transformers/attention.cpp#L653
     # Scale q, k before matmul for stability see https://tinyurl.com/sudb9s96 for math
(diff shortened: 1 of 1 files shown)


BowenBao:
Lint CI failing with `/opt/hostedtoolcache/Python/3.10.13/x64/bin/python: No module named lintrunner_adapters`. cc @justinchuby is there something I'm missing?

BowenBao:
build_docs CI failing with
```
ERROR: Could not find a version that satisfies the requirement onnx-weekly==1.15.0.dev20230807 (from versions: 1.16.0.dev20231106, 1.16.0.dev20231113, 1.16.0.dev20231120, 1.16.0.dev20231127, 1.16.0.dev20231204)
ERROR: No matching distribution found for onnx-weekly==1.15.0.dev20230807
```

justinchuby (approved):

justinchuby:
Fixing in https://github.com/microsoft/onnxscript/pull/1200

titaiwangms (approved):
Do we miss the fp16 tests?

justinchuby:
Should `scale: FLOAT` be annotated differently?

BowenBao:
> Should `scale: FLOAT` be annotated differently?

Maybe castlike earlier and update annotation. Those functions that are tagged as `private=True` won't ever be invoked from exporter op dispatcher directly right? So the scale is guaranteed to have been casted properly at that point.

PR #1207, opened 2023-12-05
Fix loop exporting issue

gramalingam:
Fix issue in converting ONNX loops back to python (onnxscript).

This is part 1. The generated code can be further simplified by merging the SSA-variants of a variable that is created by ONNX requirements (eg., a loop in ONNX will represent a state-vairable by four variables representing the values at entry to loop, at exit from loop, at entry into loop-body, and at exit from loop-body). This, however, requires some complex analysis to ensure correctness, and hence will be done separately.


justinchuby:
Looks like some tests failed 👀

justinchuby:
That was solved. The errors seem legit

justinchuby:
I will go ahead and fix the implementations. There doesn't seem too many: https://github.com/microsoft/onnxscript/runs/19532384407

aten_embedding_bag_padding_idx
aten_unfold
aten_embedding_bag

justinchuby (approved):

PR #1209, opened 2023-12-09
Fix Op(aten::diagonal) | feat(torchlib)

titaiwangms:
dim1 and dim2 could go negative.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2440,6 +2440,11 @@ def aten_diagonal(self: TReal, offset: int = 0, dim1: int = 0, dim2: int = 1) ->
     # [0,1,2] -> [2,0,1] when dim1=0 and dim2=1
     # [0,1,2] -> [1,0,2] when dim1=0 and dim2=2
     # [0,1,2] -> [0,1,2] when dim1=1 and dim2=2
+    if dim1 < 0:
+        dim1 = dim1 + len(self.shape)
```
justinchuby, line 2444: Maybe use self.rank in case the shape is not fully known?

justinchuby (approved):

PR #1217, opened 2023-12-12
Fix fp16 aten_native_batch_norm when bias is None and training is True

BowenBao:
Fixes `nfnet` in https://github.com/microsoft/onnx-converters-private/issues/196

Two changes:

- Cast locally created `weight` and `bias` to `input` dtype.
- Upcast `input` when computing `mean` and `var`.

No idea how this was not covered by unittest. The test case seems to be there.

Minimized repro:

```python
import torch

class Module(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.bn = torch.nn.BatchNorm2d(3, track_running_stats=False)

    def forward(self, x):
        # return self.bn(x)
        return torch.nn.functional.batch_norm(x, None, None, self.bn.weight, training=True)
    
model = Module().cuda().to(dtype=torch.float16).eval()
x = torch.randn(1, 3, 224, 224, dtype=torch.float16).cuda()

op = torch.onnx.dynamo_export(model, x)
op.save("report_bn.onnx")
op.save_diagnostics("report_bn.sarif")


import onnxruntime
sess = onnxruntime.InferenceSession("report_bn.onnx")
sess.run(None, {"l_x_": x.cpu().numpy()})

```

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -5555,10 +5555,14 @@ def aten_native_batch_norm(
     """native_batch_norm(Tensor input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor? running_var, bool training, float momentum, float eps) -> (Tensor, Tensor, Tensor)"""
 
     if weight is None:  # Set to 1.0 as default
-        weight = op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2))
+        weight = op.CastLike(
+            op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2)), input
+        )
 
     if bias is None:  # Set to 0.0 as default
-        bias = op.Expand(op.Constant(value_floats=[0.0]), op.Shape(input, start=1, end=2))
+        bias = op.CastLike(
+            op.Expand(op.Constant(value_floats=[0.0]), op.Shape(input, start=1, end=2)), input
+        )
 
     axes = list(range(len(input.shape)))
     axes.pop(1)
@@ -5609,13 +5613,14 @@ def _aten_native_batch_norm_training_onnx(
         training_mode=training,
     )
     # Compute var and rstd
-    mean = op.ReduceMean(input, axes)
-    input_sub_mean = op.Sub(input, mean)
+    upcast_input = op.Cast(input, to=FLOAT.dtype)
+    mean = op.ReduceMean(upcast_input, axes)
+    input_sub_mean = op.Sub(upcast_input, mean)
     sqr = op.Mul(input_sub_mean, input_sub_mean)
     var = op.ReduceMean(sqr, axes, keepdims=False)
     rstd = op.Div(1.0, op.Sqrt(var + eps))
     # Get mean again with size = [1, C]
-    mean = op.ReduceMean(input, axes, keepdims=False)
+    mean = op.ReduceMean(upcast_input, axes, keepdims=False)
     return norm, mean, rstd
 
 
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5555,10 +5555,14 @@ def aten_native_batch_norm(
     """native_batch_norm(Tensor input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor? running_var, bool training, float momentum, float eps) -> (Tensor, Tensor, Tensor)"""
 
     if weight is None:  # Set to 1.0 as default
-        weight = op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2))
+        weight = op.CastLike(
+            op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2)), input
+        )
```
justinchuby, line 5560: Should we cast first than expand? Hopefully that is casting a smaller and constant foldable tensor.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5609,13 +5613,14 @@ def _aten_native_batch_norm_training_onnx(
         training_mode=training,
     )
     # Compute var and rstd
-    mean = op.ReduceMean(input, axes)
-    input_sub_mean = op.Sub(input, mean)
+    upcast_input = op.Cast(input, to=FLOAT.dtype)
```
justinchuby, line 5616: Could you add a comment to explain why upcasting is needed (I assume it is because of training?) Thanks!

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5609,13 +5613,14 @@ def _aten_native_batch_norm_training_onnx(
...
+    mean = op.ReduceMean(upcast_input, axes)
+    input_sub_mean = op.Sub(upcast_input, mean)
     sqr = op.Mul(input_sub_mean, input_sub_mean)
     var = op.ReduceMean(sqr, axes, keepdims=False)
     rstd = op.Div(1.0, op.Sqrt(var + eps))
     # Get mean again with size = [1, C]
-    mean = op.ReduceMean(input, axes, keepdims=False)
+    mean = op.ReduceMean(upcast_input, axes, keepdims=False)
     return norm, mean, rstd
```
justinchuby, line 5625: Do we need to tighten the test tolerance if there is any set?

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
BowenBao: idk, I have no idea which test case is associated with this particular scenario ...

justinchuby: It is specified here: https://github.com/microsoft/onnxscript/blob/7428f920ebb9713ec78bae7555090f0f55e05ef0/onnxscript/tests/function_libs/torch_lib/ops_test_data.py#L1813

justinchuby: But may be unrelated if training is never true in the test cases

justinchuby: Feel free to merge if removing this line fails the tests

BowenBao:
I saw some batch_norm failure in CI, not sure if they are new. Also why only failing on torch-nightly?

> E   onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] Inference error(s): (op_type:_aten_native_batch_norm_training_onnx, node name: _aten_native_batch_norm_training_onnx_1): [TypeInferenceError] Inferred elem type differs from existing elem type: (1) vs (10)

>  AssertionError: Output 0 mismatch

justinchuby:
Looks like there is now assertion errors

PR #1219, opened 2023-12-12
Fix Op(aten::scaled_dot_product_attention) under fp16

BowenBao:
Fixes nanogpt from https://github.com/microsoft/onnx-converters-private/issues/196

Cont of #1204

Minimal repro

```python
import torch

class SubModule(torch.nn.Module):
    def forward(self, x):
        return torch.ops.aten.scaled_dot_product_attention(x, x, x, is_causal=True)
    
class Module(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.sub = SubModule()

    def forward(self, x):
        return self.sub(x)


func = Module().cuda().to(dtype=torch.float16).eval()
x = torch.randn(1, 1, 1, 32, dtype=torch.float16).cuda()
op = torch.onnx.dynamo_export(func, x)
op.save("attn.onnx")
op.save_diagnostics("attn.sarif")

import onnxruntime
sess = onnxruntime.InferenceSession("attn.onnx", providers=["CUDAExecutionProvider"])
sess.run(None, {"l_x_": x.cpu().numpy()})

```


justinchuby (approved):
Thanks!

PR #1220, opened 2023-12-12
Use correct logger in values.py

### LINKED ISSUE #1211: Use the correct logger in values.py
We should get the module logger first in

https://github.com/microsoft/onnxscript/blob/036b6478c7cdef0b5fdee4f564eef9c8d450002c/onnxscript/values.py#L287

Otherwise the log is from "root" which can be confusing

image

AbhiByte:
Fixes #1211

### DIFF
--- a/onnxscript/values.py
+++ b/onnxscript/values.py
@@ -108,7 +108,6 @@ class Opset:
 
     def add_function_def(self, fun):
         if fun.name in self.function_defs:
-            logger = logging.getLogger("onnxscript")
             logger.warning("%s: Already defined.", fun.name)
         self.function_defs[fun.name] = fun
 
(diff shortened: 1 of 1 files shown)


onnxscript/values.py
```diff
@@ -284,7 +284,8 @@ def __init__(
         self._param_schemas: Optional[tuple[ParamSchema, ...]] = None
 
         if self._op_schema is None:
-            logging.debug(
+            logger = logging.getLogger("onnxscript")
```
justinchuby, line 287: Feel free to move this to the top of the module in the global namespace and use `logger = logging.getLogger(__name__)`. Thanks!

AbhiByte: Thanks, just did that.

justinchuby (approved):

justinchuby:
Thank you for your contribution!

PR #1222, opened 2023-12-13
Fix Op(aten::roll) | feat(torchlib)

titaiwangms:
In converter, pyTorch sets default value of dims as empty list.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6915,7 +6915,7 @@ def aten_roll(self: TTensor, shifts: INT64, dims: Sequence[int] = ()) -> TTensor
     elif self.shape[0] == 0:  # empty tensor
         return self
     else:
-        if isinstance(dims, tuple) and len(dims) == 0:  # Empty list
+        if isinstance(dims, (tuple, list)) and len(dims) == 0:  # Empty list
             # assert isinstance(shifts, int)
             return _aten_roll_shift_no_dim_onnx(self, shifts)
         else:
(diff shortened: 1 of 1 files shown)


justinchuby (approved):
Thanks! Could you add the pr description as a comment in code?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6915,7 +6915,7 @@ def aten_roll(self: TTensor, shifts: INT64, dims: Sequence[int] = ()) -> TTensor
     elif self.shape[0] == 0:  # empty tensor
         return self
     else:
-        if isinstance(dims, tuple) and len(dims) == 0:  # Empty list
+        if isinstance(dims, (tuple, list)) and len(dims) == 0:  # Empty list
```
justinchuby, line 6918:
```suggestion
        if isinstance(dims, Sequence) and len(dims) == 0:  # Empty list
```

would this work?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6915,7 +6915,8 @@ def aten_roll(self: TTensor, shifts: INT64, dims: Sequence[int] = ()) -> TTensor
     elif self.shape[0] == 0:  # empty tensor
         return self
     else:
-        if isinstance(dims, tuple) and len(dims) == 0:  # Empty list
+        # NOTE: In pytorch, default value of dims is an empty list.
+        if isinstance(dims, Sequence) and len(dims) == 0:  # Empty sequence
```
justinchuby, line 6919: Now that I look at this, we shouldn’t even need this isinstance check because dims is always a sequence

PR #1226, opened 2023-12-14
Remove check_model and shape_infer in graph building | feat(torchlib)

justinchuby:
Remove check_model and shape_infer in graph building because serialization adds overhead and impacts runtime when calling to_model_proto.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -1029,15 +1029,6 @@ class TorchScriptGraph:
             )
         )
 
-        try:
-            if not cache_model_to_disk:
-                # Only check the model if it is in memory.
-                # Otherwise the checker and shape_inference will fail because
-                # we cannot serialize the model.
-                onnx_model = onnx.shape_inference.infer_shapes(
-                    onnx_model, check_type=True, strict_mode=False, data_prop=True
-                )
-                onnx.checker.check_model(onnx_model, full_check=True)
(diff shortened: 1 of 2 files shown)


justinchuby:
There are some errors with sequence output types

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -1028,21 +1028,4 @@ def to_model_proto(
                 common_ops.common_opset.domain, common_ops.common_opset.version
             )
         )
-
-        try:
```
thiagocrepaldi, line 1032: maybe add a ` def to_model_proto(..., infer_shape=False)` in which this step is skipped by default instead of deleting it?

thiagocrepaldi (commented):
can you share

onnxscript/function_libs/torch_lib/graph_building.py
xadupre: Maybe we should assume the converter produces correct onnx graphs only. It fails if it cannot. In that case, more checkings seem unnecessary and this code should be removed or called only in the unit tests.

xadupre (approved):

onnxscript/function_libs/torch_lib/graph_building.py
justinchuby: I added the shape inference step in test instead.

thiagocrepaldi (approved):
LGTM assuming it does not break the torch.onnx.dynamo_export CI

Issue #1223, opened 2023-12-13
[torchlib] Fix special.erfcx


titaiwangms:
From https://github.com/microsoft/onnxscript/blob/1ff5388dc790e3e910a5a57fa5be70bcd3f650af/onnxscript/tests/function_libs/torch_lib/ops_test_data.py#L818

Tried to enable as
```python
TorchLibOpInfo("special.erfcx", special_ops.aten_special_erfcx),
```
but got redefine test error.

And based on https://pytorch.org/docs/master/special.html#torch.special.erfcx,
The implementation of special.erfcx should rather be:

```python
@torch_op("aten::special_erfcx")
def aten_special_erfcx(self: TFloatOrBFloat16) -> TFloatOrBFloat16:
    """special_erfcx(Tensor self) -> Tensor"""

    return op.Mul(op.Exp(op.Pow(self, 2)), op.Sub(1, op.Erf(self)))
```

This modification makes op test in converter side closer to result matching but still not enough for float32 precision. May need more investigations..

justinchuby:
### Summary

The output of ONNX Runtime does not match that of PyTorch when executing test
`ops_test.TestOutputConsistencyFullGraphCPU.test_output_match_opinfo__special_erfcx_cpu_float32`, `sample 0` in ONNX Script `TorchLib`.

To recreate this report, use

```bash
CREATE_REPRODUCTION_REPORT=1 python -m pytest onnxscript/tests/function_libs/torch_lib/ops_test.py -k test_output_match_opinfo__special_erfcx_cpu_float32
```

### Inputs

Shapes: `['Tensor<torch.Size([20]), dtype=torch.float32>']`

<details><summary>Details</summary>
<p>

```python
kwargs = {}
inputs = (tensor([ 6.8808,  7.4701, -2.1085,  8.2675, -1.9719,  1.8161, -4.3817,  5.2855,
         7.9339, -6.6027,  7.8228,  1.6844,  6.6493,  1.2189,  4.3397, -1.2707,
         6.9380,  1.3303, -4.2016,  2.2941]),)
```

</p>
</details>

### Expected output

Shape: `torch.Size([20])`

<details><summary>Details</summary>
<p>

```python
expected = tensor([8.1154e-02, 7.4867e-02, 1.7025e+02, 6.7753e-02, 9.7418e+01, 2.7655e-01,
[...]

PR #1227, opened 2023-12-14
Fix the erfcx implementation | fix(torchlib)

justinchuby:
https://github.com/microsoft/onnxscript/issues/1223 but it is still numerically unstable.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/special.py
+++ b/onnxscript/function_libs/torch_lib/ops/special.py
@@ -106,7 +106,7 @@ def aten_special_erfc(self: TFloatOrBFloat16) -> TFloatOrBFloat16:
 def aten_special_erfcx(self: TFloatOrBFloat16) -> TFloatOrBFloat16:
     """special_erfcx(Tensor self) -> Tensor"""
 
-    return op.Mul(op.Exp(op.Pow(self, 2)), op.Erf(self))
+    return op.Mul(op.Exp(op.Pow(self, 2)), op.Sub(1, op.Erf(self)))
 
 
 def aten_special_erfinv(self: TensorType) -> TensorType:
--- a/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
+++ b/onnxscript/tests/function_libs/torch_lib/ops_test_data.py
@@ -59,7 +59,6 @@ OPS_DB = copy.deepcopy(common_methods_invocations.op_db)
 
 # Append extra op_db into the op database for testing
 OPS_DB.extend(opinfo_definitions.signal.op_db)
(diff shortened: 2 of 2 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -815,7 +814,9 @@ def _where_input_wrangler(
     TorchLibOpInfo(
         "erfc", special_ops.aten_special_erfc, tolerance={torch.float16: (1e-2, 2e-4)}
     ),
-    # TorchLibOpInfo("erfcx", special_ops.aten_special_erfcx),  # not in OPS_DB
+    TorchLibOpInfo("special.erfcx", special_ops.aten_special_erfcx).xfail(
+        reason="fixme: The implementation is numerical unstable: https://github.com/microsoft/onnxscript/issues/1223"
```
justinchuby, line 818:
```suggestion
        reason="fixme: The implementation is numerically unstable: https://github.com/microsoft/onnxscript/issues/1223"
```

onnxscript/function_libs/torch_lib/ops/special.py
```diff
@@ -106,7 +106,7 @@ def aten_special_erfc(self: TFloatOrBFloat16) -> TFloatOrBFloat16:
 def aten_special_erfcx(self: TFloatOrBFloat16) -> TFloatOrBFloat16:
     """special_erfcx(Tensor self) -> Tensor"""
 
-    return op.Mul(op.Exp(op.Pow(self, 2)), op.Erf(self))
+    return op.Mul(op.Exp(op.Pow(self, 2)), op.Sub(1, op.Erf(self)))
```
justinchuby, line 109: Maybe https://math.stackexchange.com/questions/4056083/how-to-address-the-numerical-instability-of-a-function-involving-erf-and-an-expo ?

titaiwangms: That is what I thought, but it still doesn't match. These are all basic ops and match with pytorch, but when we combine them, it doesn't have enough precision..

titaiwangms (approved):
Let's merge it with an issue raised in ORT?

onnxscript/function_libs/torch_lib/ops/special.py
justinchuby: Do you mean when you use a more numerically stable implementation from this link the results still do not match?

titaiwangms: Ah, I thought the link was about torch implementation.

PR #1238, opened 2023-12-21
Fully support on complex dtype | feat(graph_building)

titaiwangms:
Previous to this PR, the graph didn't support complex dtype attributes, and dim offset of cat/stack is not considered in complex dtype.

Adds:
(1) Support complex constant (convert to real reprentation)
(2) Add cat/stack complex support

NOTE: `aten_scalar_tensor_complex` only supports complex input, and if input is real to convert to complex, `aten_scalar_tensor` is invoked instead, just like we have in tests.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7106,7 +7106,9 @@ def aten_scalar_tensor_complex(
     s: Union[FLOAT, DOUBLE], dtype: int = COMPLEX64.dtype
 ) -> RealType:
     """scalar_tensor(Scalar s, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
-
+    # NOTE: When the input is originally in complex, this function is invoked,
+    # on the other hand, when the input is originally in real, aten_scalar_tensor
+    # is invoked.
     if dtype == COMPLEX128.dtype:
         casted = op.Cast(s, to=DOUBLE.dtype)
     if dtype == COMPLEX64.dtype:
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7106,8 +7106,8 @@ def aten_scalar_tensor_complex(
     s: Union[FLOAT, DOUBLE], dtype: int = COMPLEX64.dtype
 ) -> RealType:
     """scalar_tensor(Scalar s, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
-    # NOTE: When the input is originally in complex, this function is invoked,
-    # on the other hand, when the input is originally in real, aten_scalar_tensor
+    # NOTE: When the input is originally in complex, this function is invoked.
+    # On the other hand, when the input is originally in real, aten_scalar_tensor is used.
     # is invoked.
     if dtype == COMPLEX128.dtype:
         s = op.Cast(s, to=DOUBLE.dtype)
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7110,10 +7110,14 @@ def aten_scalar_tensor_complex(
     # On the other hand, when the input is originally in real, aten_scalar_tensor is used.
     # is invoked.
     if dtype == COMPLEX128.dtype:
-        s = op.Cast(s, to=DOUBLE.dtype)
+        result = op.Cast(s, to=DOUBLE.dtype)
     elif dtype == COMPLEX64.dtype:
-        s = op.Cast(s, to=FLOAT.dtype)
-    return s
+        result = op.Cast(s, to=FLOAT.dtype)
+    else:
+        # NOTE: No-op for non-complex dtype
+        # It's potentially a bug if it comes here with no-op.
+        result = s
+    return result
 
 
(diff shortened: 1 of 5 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1412,6 +1414,15 @@ def aten_cartesian_prod(tensors: Sequence[TensorType]) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::cat", trace_only=True, complex=True)
+def aten_cat_complex(tensors: Sequence[TTensor], dim: int = 0) -> TTensor:
```
justinchuby, line 1418: We can test this I think?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7090,6 +7101,21 @@ def aten_scalar_tensor(s: float, dtype: int = FLOAT.dtype) -> RealType:
     return common_ops.cast_to(s, dtype=dtype)
 
 
+@torch_op("aten::scalar_tensor", trace_only=True, complex=True)
+def aten_scalar_tensor_complex(
+    s: Union[FLOAT, DOUBLE], dtype: int = COMPLEX64.dtype
+) -> RealType:
+    """scalar_tensor(Scalar s, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
+    # NOTE: When the input is originally in complex, this function is invoked,
+    # on the other hand, when the input is originally in real, aten_scalar_tensor
```
justinchuby, line 7110:
```suggestion
    # NOTE: When the input is originally in complex, this function is invoked.
    # On the other hand, when the input is originally in real, aten_scalar_tensor is use.
```

nit

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7090,6 +7101,21 @@ def aten_scalar_tensor(s: float, dtype: int = FLOAT.dtype) -> RealType:
...
+    # is invoked.
+    if dtype == COMPLEX128.dtype:
+        s = op.Cast(s, to=DOUBLE.dtype)
```
justinchuby, line 7113:
```suggestion
        result = op.Cast(s, to=DOUBLE.dtype)
```

titaiwangms: This will cause "Local variable 'result' may be used before it is initialized.", unless we put an else. Do we want to put an else branch with result = op.Identity(s)?

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Not very meaningful. If we want to test this, the input generated in `extra_opinfo` will be real representation of the s. The crucial part is actually during the conversion.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Sure, we can put result = s

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Simply adding an entry with complex=True should do it:

https://github.com/microsoft/onnxscript/blob/635fb437f8270769b1d19b3e286384d7f80a2741/onnxscript/tests/function_libs/torch_lib/ops_test_data.py#L534-L535

titaiwangms: It doesn't work, because the tests are like `input=(-5), kwargs={'dtype':torch.complex64}`, but `aten_scalar_tensor_complex` only supports `input=(-5j), kwargs={'dtype':torch.complex64}`. When it comes to scalar real input, we need the original one. It's special case.

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Done with a NOTE

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Can we filter those cases out?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7520,6 +7550,15 @@ def aten_sspaddmm(
     raise NotImplementedError()
 
 
+@torch_op("aten::stack", trace_only=True, complex=True)
+def aten_stack_complex(tensors: Sequence[TTensorOrString], dim: int = 0) -> TTensorOrString:
+    """stack(Tensor[] tensors, int dim=0) -> Tensor"""
+    # Real representation unsqueezes the last dimension
+    if dim < 0:
```
justinchuby, line 7557: Test can be added for this too?

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: I can do that, but that's all of them.

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Done

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Add it for remind that is a special case.

justinchuby (approved):
Thank you!

PR #1245, opened 2024-01-02
Add op(upsample linear1d) | feat(torchlib)

xiaowuhu:


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2371,12 +2371,42 @@ def aten_upsample_bilinear2d_backward(
...
 
 
+@torch_op("aten::upsample_linear1d", trace_only=True)
 def aten_upsample_linear1d(
-    self: TensorType, output_size: INT64, align_corners: bool, scales: Optional[float] = None
-) -> TensorType:
+    self: TReal, output_size: INT64, align_corners: bool, scales: Optional[float] = None
+) -> TReal:
     """upsample_linear1d(Tensor self, SymInt[1] output_size, bool align_corners, float? scales=None) -> Tensor"""
+    del scales  # aten function ignore the `scales` argument
```
justinchuby, line 2379: Do we include scales in the tests to verify this behavior?

xiaowuhu: Did ever try, but the FullGraph mode failed due the ```scales``` not in use. So remove that case.

justinchuby: I think we should use scale similar to the 2d case

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2368,12 +2368,20 @@ def aten_upsample_bilinear2d_backward(
...
+@torch_op("aten::upsample_linear1d", trace_only=True)
 def aten_upsample_linear1d(
-    self: TensorType, output_size: INT64, align_corners: bool, scales: Optional[float] = None
-) -> TensorType:
+    self: TReal, output_size: INT64, align_corners: bool, scales: Optional[float] = None
+) -> TReal:
     """upsample_linear1d(Tensor self, SymInt[1] output_size, bool align_corners, float? scales=None) -> Tensor"""
-
-    raise NotImplementedError()
+    # FIXME(justinchuby): Support when scales is provided
```
justinchuby, line 2376:
```suggestion
    # FIXME(justinchuby): Support when scales is provided and align_corners is False
```

Issue #1159, opened 2023-11-16
[torchlib] Implement various upsample functions


justinchuby:
- [x] aten_upsample_bicubic2d https://github.com/microsoft/onnxscript/pull/1208
- [ ] aten__upsample_bicubic2d_aa #1259 
- [ ] aten__upsample_bilinear2d_aa #1259 
- [x] aten_upsample_bilinear2d.vec https://github.com/microsoft/onnxscript/pull/1249
- [x] aten_upsample_linear1d #1245 
- [x] aten_upsample_nearest1d #1246 
- [x] aten_upsample_nearest3d #1246 
- [x] aten_upsample_trilinear3d  #1260 

Reference: https://github.com/microsoft/onnxscript/blob/24d77d96ee343fd6cac4b253a6c8e10c7ac4bd2a/onnxscript/function_libs/torch_lib/ops/nn.py#L2121-L2137

justinchuby:
https://github.com/pytorch/pytorch/issues/116706

BowenBao:
How are we validating this works e2e for pytorch export?

I think there might be an issue inside pytorch that decompositions are forced for these ops, similar to the einsum situation. https://github.com/pytorch/pytorch/issues/116684

BowenBao:
`     align_corners: bool = True,  # pylint: disable=unused-argument `

Why is `align_corners` an unused argument?

BowenBao:
Regarding Resize.nearest_mode. In torchscript exporter it was set to "floor". In torchlib it is left as default "round_prefer_floor"

justinchuby:
We do need to implement them and match the behaviors @xiaowuhu please correct me if I’m missing anything

PR #1254, opened 2024-01-16
Fix upsample_bilinear to respect align_corner argument

BowenBao:
Fixes https://github.com/microsoft/onnxscript/issues/1159#issuecomment-1888089502 which indeed turns out to be a problem uncovered by PyTorch CI https://github.com/pytorch/pytorch/actions/runs/7508784822/job/20445196351?pr=117314.

> Fixes `align_corner` default value. The default value from pytorch signature is `False` https://pytorch.org/docs/stable/generated/torch.nn.Upsample.html#torch.nn.Upsample.
> That said, it shouldn't matter since `align_corner` in aten signature in `native_functions.yaml` is a required argument, so in practice this function will never be invoked w/o `align_corner`.  

Above is outdated. The case is more complicated. https://github.com/microsoft/onnxscript/pull/1254#discussion_r1453948494.

In short this PR fixes the torchlib op signature to match with aten spec, and updates input wrangler for test case to bridge from sample test inputs for function `torch.nn.functional.upsample_bilinear`.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2323,7 +2323,7 @@ def aten_upsample_bilinear2d_vec(
     align_corners: bool,
     scale_factors: Optional[Sequence[float]],
 ) -> TReal:
-    """upsample_bilinear2d(Tensor self, SymInt[2] output_size, bool align_corners, float? scales_h=None, float? scales_w=None) -> Tensor"""
+    """upsample_bilinear2d.vec(Tensor input, SymInt[]? output_size, bool align_corners, float[]? scale_factors) -> Tensor"""
     scales_h = scale_factors[0] if scale_factors is not None else None
     scales_w = scale_factors[1] if scale_factors is not None else None
     return aten_upsample_bilinear2d(self, output_size, align_corners, scales_h, scales_w)
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2298,7 +2298,7 @@ def aten_upsample_bilinear2d(
     output_size: Optional[INT64] = None,
     scales_h: Optional[float] = None,
     scales_w: Optional[float] = None,
-    align_corners: bool = True,  # pylint: disable=unused-argument
+    align_corners: bool = True,
 ) -> TReal:
     """upsample_bilinear2d(Tensor self, SymInt[2] output_size, bool align_corners, float? scales_h=None, float? scales_w=None) -> Tensor"""
 
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2298,7 +2298,7 @@ def aten_upsample_bilinear2d(
     output_size: Optional[INT64] = None,
     scales_h: Optional[float] = None,
     scales_w: Optional[float] = None,
-    align_corners: bool = True,
+    align_corners: bool = False,
 ) -> TReal:
     """upsample_bilinear2d(Tensor self, SymInt[2] output_size, bool align_corners, float? scales_h=None, float? scales_w=None) -> Tensor"""
 
@@ -2320,7 +2320,7 @@ def aten_upsample_bilinear2d(
 def aten_upsample_bilinear2d_vec(
     self: TReal,
     output_size: Optional[INT64] = None,
-    align_corners: bool = True,
+    align_corners: bool = False,
     scale_factors: Optional[Sequence[float]] = None,
 ) -> TReal:
     scales_h = scale_factors[0] if scale_factors is not None else None
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2295,10 +2295,10 @@ def aten_upsample_bicubic2d_backward(
 @torch_op("aten::upsample_bilinear2d", trace_only=True)
 def aten_upsample_bilinear2d(
     self: TReal,
-    output_size: Optional[INT64] = None,
+    output_size: Optional[INT64],
+    align_corners: bool,
     scales_h: Optional[float] = None,
     scales_w: Optional[float] = None,
-    align_corners: bool = False,
 ) -> TReal:
     """upsample_bilinear2d(Tensor self, SymInt[2] output_size, bool align_corners, float? scales_h=None, float? scales_w=None) -> Tensor"""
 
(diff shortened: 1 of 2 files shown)


BowenBao:
@justinchuby The reason why it passes on onnxscript side is that we are not testing the `nn.functional.interpolate` test case from `common_methods_invocations.py`, which fails on pytorch side. I don't know how to properly add the test case for this, it is a large test case with various variants. The `upsample` test case we test on probably didn't have `align_corner` flag variation in its sample inputs.

We need more e2e pytorch export tests when adding ops in torchlib to ensure it works.

justinchuby (approved):

justinchuby:
Thanks!

> We need more e2e pytorch export tests when adding ops in torchlib to ensure it works.

Do you have suggestions? Currently we rely on tests in PyTorch for integration tests.

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2298,16 +2298,21 @@ def aten_upsample_bilinear2d(
...
 
+    coordinate_transformation_mode = "align_corners" if align_corners else "pytorch_half_pixel"
     if output_size is not None:
-        result = _aten_upsample_bilinear2d_output_size(self, output_size)
+        result = _aten_upsample_bilinear2d_output_size(
+            self, output_size, coordinate_transformation_mode
+        )
     else:
         assert scales_h is not None
         assert scales_h == scales_w, f"scale_h({scales_h}) != scale_w({scales_w})"
```
justinchuby, line 2312: Is this still valid?

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2298,7 +2298,7 @@ def aten_upsample_bilinear2d(
     output_size: Optional[INT64] = None,
     scales_h: Optional[float] = None,
     scales_w: Optional[float] = None,
-    align_corners: bool = True,
+    align_corners: bool = False,
```
justinchuby, line 2301: Looks like align_corners should be required in the signature

BowenBao: I don't know why it was written in this way. Maybe something to do with onnxscript signature syntax?

onnxscript/function_libs/torch_lib/ops/nn.py
BowenBao: Unclear to me, looks like different scale values are not covered in tests.

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: I don't think it is constrained by onnxscript. Looks like something to be fixed

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: @xiaowuhu

onnxscript/function_libs/torch_lib/ops/nn.py
BowenBao: Hmm, looks like tests start to fail with the default value swap.

justinchuby: The issue is also we are using the tests for `torch.nn.Upsample`. We should probably take the test cases out to create tests for the aten op like @xiaowuhu suggested.

BowenBao: Hmm ok, so it appears the reference test case is for [torch.nn.functional.upsample_bilinear](https://pytorch.org/docs/stable/generated/torch.nn.functional.upsample_bilinear.html#torch.nn.functional.upsample_bilinear), which is deprecated and defaults `align_corners` to True.

I agree the best fix is to make `align_corners` required, and fix the input wrangler.

BowenBao:
> Do you have suggestions? Currently we rely on tests in PyTorch for integration tests.

Based on current infra, one would need to add both test cases to onnxscript and pytorch ... Otherwise, the wrangler itself and operator signature implementation part is missing coverage.

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2295,38 +2295,45 @@ def aten_upsample_bicubic2d_backward(
...
 def aten_upsample_bilinear2d_vec(
     self: TReal,
-    output_size: Optional[INT64] = None,
-    align_corners: bool = True,
-    scale_factors: Optional[Sequence[float]] = None,
+    output_size: Optional[INT64],
+    align_corners: bool,
+    scale_factors: Optional[Sequence[float]],
 ) -> TReal:
+    """upsample_bilinear2d(Tensor self, SymInt[2] output_size, bool align_corners, float? scales_h=None, float? scales_w=None) -> Tensor"""
```
justinchuby:
```suggestion
    """upsample_bilinear2d.vec(Tensor input, SymInt[]? output_size, bool align_corners, float[]? scale_factors) -> Tensor"""
```

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2295,38 +2295,45 @@ def aten_upsample_bicubic2d_backward(
 @torch_op("aten::upsample_bilinear2d", trace_only=True)
 def aten_upsample_bilinear2d(
     self: TReal,
-    output_size: Optional[INT64] = None,
+    output_size: Optional[INT64],
```
justinchuby, line 2298:
```suggestion
    output_size: INT64,
```

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2295,38 +2295,45 @@ def aten_upsample_bicubic2d_backward(
...
 
+    coordinate_transformation_mode = "align_corners" if align_corners else "pytorch_half_pixel"
     if output_size is not None:
-        result = _aten_upsample_bilinear2d_output_size(self, output_size)
+        result = _aten_upsample_bilinear2d_output_size(
+            self, output_size, coordinate_transformation_mode
+        )
     else:
         assert scales_h is not None
         assert scales_h == scales_w, f"scale_h({scales_h}) != scale_w({scales_w})"
```
justinchuby, line 2312: Should we reverse the checks - omit output_size when the scales are not None?

justinchuby: The reference here shows scale is considered first if present: https://github.com/pytorch/pytorch/blob/f6767244cf292cb5e434303fe62e576b456e92eb/torch/_decomp/decompositions.py#L3311-L3315

justinchuby: But testing suggested otherwise

PR #1259, opened 2024-01-26
Add Op (_upsample_bilinear2d_aa, _upsample_bicubic2d_aa) | feat(torchlib)

xiaowuhu:
It seems that the antialias method is different between ONNX and PyTorch, so we can just compare the shape, instead of the value.

Below is the difference between ONXN and PyTorch:

```python
# ONNX
import numpy as np
self = np.array([[[[2,1,1,1],
                   [1,1,1,1],
                   [1,1,1,1],
                   [1,1,1,1]]]]).astype(np.float32)
print(self.shape)
output_size = np.array([1,1]).astype(np.int64)
align_corners = True
r = aten__upsample_bicubic2d_aa(self, output_size, align_corners)
print(r)
```
ONXN output = [[[[1.390625]]]]
```python
# PyTorch
import torch as t
r = t.ops.aten._upsample_bicubic2d_aa(t.tensor(self), t.tensor(output_size), align_corners)
print(r)
```
Torch output = tensor([[[[2.2656]]]])

I also tried some other parameters combination but none of them can match with torch.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2201,12 +2201,21 @@ def _get_upsample_align_corners_mode(align_corners: bool) -> str:
     return "align_corners" if align_corners else "pytorch_half_pixel"
 
 
-@torch_op(("aten::upsample_bicubic2d", "aten::upsample_bilinear2d"), private=True)
+@torch_op(
+    (
+        "aten::upsample_bicubic2d",
+        "aten::upsample_bicubic2d_aa",
+        "aten::upsample_bilinear2d",
+        "aten::upsample_bilinear2d_aa",
+    ),
+    private=True
+)
 def _aten_upsample_output_size(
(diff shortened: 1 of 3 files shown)


justinchuby:
It would be better if we match values because the values should be deterministic. Do we know how PyTorch does it?

xiaowuhu:
> It would be better if we match values because the values should be deterministic. Do we know how PyTorch does it?

please see the description for this PR. I add comparison between onnx and torch.

justinchuby:
Would it be helpful to consult the PyTorch implementation? I suspect we need additional processing to implement antialiasing.

justinchuby:
From our discussion: understanding the PyTorch implementation proved to be harder than anticipated (https://github.com/pytorch/pytorch/blob/bcf35c6ae62bb6560befa3550e37a8283944e5f4/aten/src/ATen/native/cpu/UpSampleKernel.cpp#L2009). We will seek additional help for this.

Issue #1263, opened 2024-02-03
Illegal "IsScalar" node encountered in exported ONNX model


yuanyao-nv:
An exported model contains an `IsScalar` op which is not in the ONNX spec.
![image](https://github.com/microsoft/onnxscript/assets/99693700/e668c040-151f-4c5f-b978-96d7609e3807)

Seems like it comes from the use of `common_ops.IsScalar` in `aten_mean_dim`.

Env:
torch 2.3.0.dev20240202+cu121
onnxscript 0.1.0.dev20240203
MONAI installed with ` pip3 install git+https://github.com/Project-MONAI/MONAI@0.8.1#egg=MONAI`

justinchuby:
IsScalar is a function which should be part of the model local functions. How was the model exported?

yuanyao-nv:
This is the export script:
```
import torch
from monai.networks.nets import DenseNet

def DenseNetDataloader():
    yield torch.randn(1,1,128,128).to("cuda")

model = lambda : DenseNet(
            init_features=64,
            growth_rate=32,
            block_config=[6, 12, 24, 16],
            spatial_dims=2,
            in_channels=1,
            out_channels=15
        )
data = next(DenseNetDataloader())

model = model().eval().to('cuda')

export_output = torch.onnx.dynamo_export(
    model,
    data,
)
export_output.save('Clara_DenseNet_dynamo.onnx')
```

justinchuby:
Was the error raised from the onnx checker?

justinchuby:
Could you attach the exported ONNX file?

yuanyao-nv:
ONNX checker actually doesn't give any error. The error was raised by TensorRT.

I'm checking with the relevant parties whether I can directly share the model. Did the export script I attached above work for you? I used the NGC pytorch container `docker run --gpus all -it --rm nvcr.io/nvidia/pytorch:xx.xx-py3` and installed torch, onnxscript and MONAI as I noted in the description. You might also need to install some dependencies manually:
```
pip install hydra-core sentencepiece youtokentome transformers inflect pyannote.core pyannote.metrics editdistance jiwer h5py sacremoses pytorch-lightning webdataset ijson wandb
```

justinchuby:
On thing I would check is if IsScalar is defined as one of the model local functions.

yuanyao-nv:
[...]

PR #1272, opened 2024-02-07
Correctly construct nested function opset_imports for TorchScriptGraph serialization

BowenBao:
Fixes https://github.com/microsoft/onnxscript/issues/1263

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -732,6 +732,19 @@ class TorchScriptGraph:
         return tensors
 
     @runtime_typing.checked
+    def fetch_opset_imports(self) -> Dict[str, int]:
+        # TODO: All local function domain versions are hardcoded as 1.
+        opset_imports: Dict[str, int] = {}
+        for subgraph in self._sub_torch_script_graphs.values():
+            opset_imports.update(subgraph.fetch_opset_imports())
+            if subgraph.domain_name is not None:
+                opset_imports[subgraph.domain_name] = 1
+        for function in self._function_store.values():
+            opset_imports.update(function.function_ir.get_opset_import())
+            opset_imports[function.function_ir.domain] = 1
+        return opset_imports
+
+    @runtime_typing.checked
     def fetch_function_proto_dict(
         self, opset_version: int
     ) -> Mapping[Tuple[str, str], onnx.FunctionProto]:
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -731,6 +731,19 @@ def _add_torchscript_op_call(
             tensor.name = _rename_intermediate_value(tensor.name)
         return tensors
 
+    @runtime_typing.checked
+    def fetch_opset_imports(self) -> Dict[str, int]:
+        # TODO: All local function domain versions are hardcoded as 1.
```
justinchuby, line 736:
```suggestion
        # FIXME: All local function domain versions are hardcoded as 1.
```

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -731,6 +731,19 @@ def _add_torchscript_op_call(
...
+        opset_imports: Dict[str, int] = {}
+        for subgraph in self._sub_torch_script_graphs.values():
+            opset_imports.update(subgraph.fetch_opset_imports())
```
justinchuby, line 739: Do we need this to be recursive, or do we just need opsets for call the direct invocations?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -925,6 +938,18 @@ def to_function_proto(self, opset_version: int, function_name: str) -> onnx.Func
 
         onnx_model = onnx.load_from_string(proto)
 
+        unique_custom_domains = self.fetch_opset_imports()
+
+        while len(onnx_model.opset_import) > 0:
```
justinchuby, line 943: Is there a way to use a for loop for this to avoid unintentional infinite loops?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -1026,11 +1047,4 @@ def to_model_proto(
                 for domain, version in unique_custom_domains.items()
             ]
         )
-        # Include the library shared opset domain
```
justinchuby, line 1029: Nice!

justinchuby (approved):
Thanks a lot!

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: Need to be recursive, or more precisely, looking inside nested functions. This is exactly what was missing that led to the IsScalar issue.
Now there is indeed in-efficiency that can be improved. Caching lookup results could be one solution.

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: certainly

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: Thinking about it again, caching is definitely no-go since it is more than likely to create bugs in the future.

I'll merge as is and leave it as follow-up in case people are interested to improve the efficiency.

justinchuby: > This is exactly what was missing

I just realized. Because the `aten_mean_dim` function has imported the `common` domain already. That is a little counter intuitive. Although I don't see this explicitly stated in the spec. @gramalingam do you have thoughts?

PR #1275, opened 2024-02-09
Revert "Correctly construct nested function opset_imports for TorchScriptGraph serialization"

justinchuby:
Reverts microsoft/onnxscript#1272


BowenBao (approved):
Turns out "imports are used to resolve _only_ the local nodes" is the correct behavior.

PR #1276, opened 2024-02-09
Simplify value name in symbolic shape export

BowenBao:
* __->__ #1276

Use `{domain}::{function_name}/{value}` instead of `{function_name_1}/{function_name_2}/.../{value}` for symbolic value_info name. There are several reasons behind this update and why a fully qualified name is not needed.

1. The symbolic shape of the value is uniquely associated with a given `TorchScriptGraph` that eventually gets serialized as `FunctionProto`. Prior to serialization it should be made generally compatible for all callsites, in case there are multiple.
2. In ONNX 1.16, FunctionProto will have its own designated value_info field. This aligns with the first point that the shape itself must be made compatible.
3. In practice, all `TorchScriptGraph` objects are exported as specialized function, which has only 1 callsite.
4. This enables both types of downstream model function visitor, whereas previously only the second type was considered.
    1. Visit via function in `model.functions`.
    2. Visit via function from function node callsite.

Next step:
- Update PyTorch onnxscript CI and its unittest.
- Update onnx-rewriter to consume new shape format.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -804,7 +804,7 @@ class TorchScriptGraph:
 
         # Insert value info for nodes within nested function calls.
         # NOTE: This is an experimental feature, will be replaced by ValueInfo inside FunctionProto
-        # in ONNX 1.16.
+        # in ONNX 1.16. https://github.com/microsoft/onnxscript/issues/1268
         # The naming strategy is subject to change. Since all local functions representing
         # nn.Modules exported by dynamo exporter have unique call sites, their function
         # op_type name can serve to form the unique identifier for value info.
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -88,6 +88,14 @@ def _rename_intermediate_value(name: str) -> str:
     return name
 
 
+def _function_id(domain: str | None, name: str) -> str:
+    """Create a unique function id for a function in a domain.
+
+    Used for generating model level unique ids for values inside a function.
+    """
+    return f"{domain if domain is not None else ''}::{name}"
+
+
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -867,38 +866,44 @@ def add_module_call(
...
-        self, prefix: str = ""
+        self, function_op_type: str
     ) -> Mapping[str, onnx.ValueInfoProto]:
-        """Unique naming strategies
-
-            {function1_op_type}/{function2_op_type}/.../{value_name}
-
-        As long as function op_type has unique call site, this is safe.
+        named_value_info: Dict[str, onnx.ValueInfoProto] = {}
+        function_id = f"{self.domain_name}::{function_op_type}"
```
justinchuby, line 873: Maybe lift string manipulation into a separate function?

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -795,16 +803,15 @@ def _override_with_symbolic_value_info_proto(self, onnx_model: onnx.ModelProto):
         del onnx_model.graph.value_info[:]
 
         # Insert value info for nodes within nested function calls.
-        # NOTE: This is an experimental feature, since in official ONNX spec, nodes
-        # within FunctionProto to have value info. https://github.com/onnx/onnx/issues/5487
-        # The names for value info are generated uniquely to be retrievable based on
-        # the call site and call stack.
+        # NOTE: This is an experimental feature, will be replaced by ValueInfo inside FunctionProto
+        # in ONNX 1.16.
```
titaiwangms, line 807: nit: maybe track with an issue

titaiwangms (approved):

onnxscript/function_libs/torch_lib/graph_building.py
BowenBao: done

Issue #1262, opened 2024-02-02
InstanceNorm always has training attribute set to True


yuanyao-nv:
I'm exporting a small ONNX model consisting of just a InstanceNorm2d op. The exported model contains a BatchNorm node with the training attribute set to True.

![image](https://github.com/microsoft/onnxscript/assets/99693700/b7e3d4b7-1c7b-4404-998c-3cddc1ccbc11)

Here's my script:
```
class Model(nn.Module):
  def __init__(self):
      super().__init__()
      self.instancenorm = torch.nn.InstanceNorm2d(100)

  def forward(self, tensor_x: torch.Tensor):
      output = self.instancenorm(tensor_x)
      return output

def Dataloader():
    yield torch.randn(20, 100, 35, 45).cuda()

model = Model()
data = next(Dataloader())

export_output = torch.onnx.dynamo_export(
    model.eval().to('cuda'),
    data,
)
export_output.save('instancenorm_dynamo.onnx')
```
If I understand it correctly, the origin of this issue should be in the torch repo since onnxscript just takes whatever value is passed for the `training` input of `aten_native_batch_norm()`?

Thanks.

justinchuby:
I think you are right that it should be PyTorch's issue setting the parameter. Could you print the fx graph, and/or try exporting with torch.export.export()?

yuanyao-nv:
If I print with
```
exported_program = torch.export.export(model.eval().to('cuda'), (data,))
print(exported_program)
```
I get
```
ExportedProgram:
    class GraphModule(torch.nn.Module):
        def forward(self, l_tensor_x_: "f32[20, 100, 35, 45]"):
            # File: /ws/dynamo_exporter_workflow/export_instancenorm.py:24, code: output = self.instancenorm(tensor_x)
            view: "f32[1, 2000, 35, 45]" = torch.ops.aten.view.default(l_tensor_x_, [1, 2000, 35, 45]);  l_tensor_x_ = None
            _native_batch_norm_legit = torch.ops.aten._native_batch_norm_legit.no_stats(view, None, None, True, 0.1, 1e-05);  view = None
            getitem: "f32[1, 2000, 35, 45]" = _native_batch_norm_legit[0];  _native_batch_norm_legit = None
            view_1: "f32[20, 100, 35, 45]" = torch.ops.aten.view.default(getitem, [20, 100, 35, 45]);  getitem = None
            return (view_1,)

[...]

PR #1284, opened 2024-02-29
Add Op (instance_norm) | feat(torchlib)

### LINKED ISSUE #1280: [torchlib] aten.instance_norm
Related to this issue: https://github.com/microsoft/onnxscript/issues/1262

We discovered that `aten::instance_norm` is decomposed during export, emitting combination of reshape + reduce + batchnorm.
It is possible to skip decomposition in exporter, and try to convert to onnx::InstanceNormalization instead. This needs torchlib support for this operator.

cc @xiaowuhu, @fatcat-z, @justinchuby, @titaiwangms

BowenBao:
* __->__ #1284

- Fixes #1280, #1262. Avoid exporting as onnx::BatchNormalization with training=True.
- Fixes mismatch in unittest.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -4168,20 +4168,55 @@ def aten_inner(self: TensorType, other: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::instance_norm", trace_only=True)
 def aten_instance_norm(
-    input: TensorType,
-    weight: Optional[TensorType],
-    bias: Optional[TensorType],
-    running_mean: Optional[TensorType],
-    running_var: Optional[TensorType],
-    use_input_stats: bool,
-    momentum: float,
-    eps: float,
-    cudnn_enabled: bool,
-) -> TensorType:
+    input: TFloat,
+    weight: Optional[TFloat] = None,
+    bias: Optional[TFloat] = None,
+    running_mean: Optional[TFloat] = None,
+    running_var: Optional[TFloat] = None,
+    use_input_stats: bool = True,
+    momentum: float = 0.1,
+    eps: float = 1e-05,
+    cudnn_enabled: bool = False,
+) -> TFloat:
     """instance_norm(Tensor input, Tensor? weight, Tensor? bias, Tensor? running_mean, Tensor? running_var, bool use_input_stats, float momentum, float eps, bool cudnn_enabled) -> Tensor"""
+    del cudnn_enabled  # unused
+    if weight is None:  # Set to 1.0 as default
+        weight = op.CastLike(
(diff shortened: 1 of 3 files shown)


onnxscript/tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -292,7 +292,7 @@ def convert_kwargs_for_onnx(kwargs: dict[str, Any]) -> dict[str, Any]:
         if key == "dtype":
             value = TORCH_TYPE_TO_ONNX[value]
         if isinstance(value, torch.Tensor):
-            value = np.array(value)
+            value = np.array(value.cpu())
```
titaiwangms, line 295: Did this break anything, or it's just safer to have?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4168,20 +4168,55 @@ def aten_inner(self: TensorType, other: TensorType) -> TensorType:
...
+            op.Expand(op.Constant(value_floats=[1.0]), op.Shape(input, start=1, end=2)), input
+        )
 
-    raise NotImplementedError()
+    if bias is None:  # Set to 0.0 as default
+        bias = op.CastLike(
+            op.Expand(op.Constant(value_floats=[0.0]), op.Shape(input, start=1, end=2)), input
+        )
+
+    if use_input_stats:
```
titaiwangms, line 4195: Better to have a comment on what this is about.

titaiwangms (approved):
LG w/ non-blockings

onnxscript/tests/function_libs/torch_lib/ops_test_common.py
BowenBao: I think safer since its noop on cpu. Without it I'm observing test failure on local env with cuda.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4168,20 +4168,58 @@ def aten_inner(self: TensorType, other: TensorType) -> TensorType:
...
+        return op.InstanceNormalization(input, weight, bias, epsilon=eps)
+
+    assert (
+        running_mean is not None and running_var is not None
+    ), "running_mean and running_var must be provided when use_input_stats is False"
+
+    batch_size = op.Shape(input, start=0, end=1)
+    bn_input = op.Reshape(input, op.Concat([1, -1], op.Shape(input, start=2), axis=0))
+    weight = op.Tile(weight, batch_size)
```
justinchuby, line 4207: Curious: When should we use Tile vs Expand? Is there a difference here?

BowenBao: Expand only works when source dimension size is either 1 or equal to target dimension size. 

Tile on the other hand is like repeat. Tile and Expand are only equivalent when source dimension size is 1.

PR #1292, opened 2024-03-08
Fix native_batch_norm_legit on cuda | fix(torchlib)

justinchuby:
Adjusted output dtypes for the op to match the graph exported with GPU.


justinchuby:
Thanks @titaiwangms for helping with the environemnt!

titaiwangms (approved):

PR #1303, opened 2024-03-14
Fix eager evaluation of the `Rank` op | fix(torchlib)

justinchuby:
I mistakenly copied the logic for `IsScalar` to `Rank`. `Rank` needs its own logic.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building.py
+++ b/onnxscript/function_libs/torch_lib/graph_building.py
@@ -365,8 +365,8 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
                     else:
                         # Fall to call add_function_call
                         pass
-                elif isinstance(args[0], Sequence):
-                    return False
+                elif isinstance(args[0], Sequence) and all(isinstance(arg, (int, float)) for arg in args[0]):
+                    return 1
                 else:
                     # Python constants are scalars
                     return 0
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -365,8 +365,8 @@ def eval_function(  # type: ignore[override]
                     else:
                         # Fall to call add_function_call
                         pass
-                elif isinstance(args[0], Sequence):
-                    return False
+                elif isinstance(args[0], Sequence) and all(isinstance(arg, (int, float)) for arg in args[0]):
+                    return 1
```
justinchuby, line 369:
```suggestion
                elif isinstance(args[0], Sequence)
                    if all(isinstance(arg, (int, float)) for arg in args[0]):
                        return 1
                    else:
                        # Fall to call add_function_call
                        pass
```

onnxscript/function_libs/torch_lib/graph_building.py
```diff
@@ -365,8 +365,12 @@ def eval_function(  # type: ignore[override]
                     else:
                         # Fall to call add_function_call
                         pass
-                elif isinstance(args[0], Sequence):
-                    return False
+                elif isinstance(args[0], Sequence)
```
justinchuby, line 368:
```suggestion
                elif isinstance(args[0], Sequence):
```

BowenBao (approved):

PR #1308, opened 2024-03-19
More memory efficient impl of 'aten_slice_scatter'

BowenBao:
The previous implementation constructs an expanded indices tensor, which
can be quite large. This step is unnecessary if the operator export is
implemented with `ScatterND` instead of `ScatterElements`.

I **have not verified** the performance impact of this change.

For memory efficiency, using the example provided by @xadupre, for 1 layer
llama we are reducing 134MB or 18.6% of one of the backward model size.

    Before
    720398178 dort-b-ort_1.opt.onnx
    After
    586184069 dort-new-ort_1.opt.onnx

The rest of the large model size consists almost solely by ORT constant
folding a `ConstantOfShape` node.

    ('new_zeros_4', 512000KB)

This is a legit usage as observed in source FX IR, as well as inductor IR, so it should be unavoidable.
Maybe avoiding allocating this upfront can reduce peak memory usage.

FX IR

    new_zeros_4: "f32[32000, 4096]" = torch.ops.aten.new_zeros.default(add_23, [32000, 4096], pin_memory = False);

Inductor IR

    buf48 = empty_strided_cuda((32000, 4096), (4096, 1), torch.float32)

cc @xadupre, @wschin, @thiagocrepaldi


xadupre:
Nice. It did the same. That's why i created a pr in onnxruntime to implement scatternd on cuda for opset >= 13.

justinchuby (approved):
Thanks! Please see codeql messages and also update the test info entry with trace_only

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7440,20 +7441,26 @@ def aten_slice_scatter(
...
+            dims = list(range(src_rank))
+            dims[0], dims[dim] = dims[dim], dims[0]
+            # Python code ends
+
+            src = op.Transpose(src, perm=dims)
+            self = op.Transpose(self, perm=dims)
+
+    output = op.ScatterND(self, index_base, src)
+    if dims is not None:
+        output = op.Transpose(output, perm=dims)
```
justinchuby, line 7462: Maybe revert the condition so dims do not have to be None?

```python
    if dim == 0:
        return output
    elif (src_rank := src.rank) != 0:
        dims = list(range(src_rank))
        dims[0], dims[dim] = dims[dim], dims[0]
        # Python code ends

        src = op.Transpose(src, perm=dims)
        self = op.Transpose(self, perm=dims)

        output = op.ScatterND(self, index_base, src)
        output = op.Transpose(output, perm=dims)
    else:
        output = op.ScatterND(self, index_base, src)
```

PR #1312, opened 2024-03-24
Implement sym_float | feat(torchlib)

justinchuby:


BowenBao:
Could you add more context?

BowenBao (approved):

justinchuby:
Wei-Sheng discovered that we are missing these two functions for flash attention. But these functions are even aten ops so I am a little confused. We may need to update the dispatcher to handle them?

PR #1314, opened 2024-03-25
Add support for _unsafe_index_put

xadupre:


onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -863,6 +863,14 @@ def _where_input_wrangler(
         matcher=lambda sample: not (sample.args[0][0].dtype == torch.int64),
         reason="this Aten overload only supports tensor(int) as indices",
     )
```
justinchuby, line 865:
```suggestion
    ),
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4076,6 +4076,16 @@ def aten_index_put(
     return result
 
 
+@torch_op("aten::_unsafe_index_put")
```
justinchuby, line 4079:
```suggestion
@torch_op("aten::_unsafe_index_put", trace_only=True)
```

Because it is reusing aten_index_put

onnxscript/tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -863,6 +863,14 @@ def _where_input_wrangler(
...
+    TorchLibOpInfo(
+        "_unsafe_index_put",
```
justinchuby, line 867: Looks like _unsafe_index_put does not exist in  https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_methods_invocations.py so we may need to create our own test cases.

Since it looks like it is the same as index_put, we can actually make it the same function

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4076,6 +4076,16 @@ def aten_index_put(
...
+def aten__unsafe_index_put(
+    self: TReal,
+    indices: Sequence[INT64],
+    values: TReal,
+    accumulate: bool = False,
+) -> TReal:
+    return aten_index_put(self, indices, values, accumulate)
```
justinchuby, line 4086:
```suggestion
```

On line 4054 above, we can simply register the function to this name since they are the same:

```python
@torch_op(("aten::index_put", "aten::_unsafe_index_put"))
```

justinchuby:
Curious what is unsafe about this op itself?

xadupre:
Dropped for #1315.

PR #1344, opened 2024-04-04
[IR] Mutable IR implementation

justinchuby:
Implement mutation methods on Graph and Function and a DoublyLinkedList to support safe mutation during iteration. Nodes in different graphs can be moved to other graphs safely during iteration.

### Methods implemented

```
- __getitem__
- __len__
- __iter__
- __reversed__
- append
- extend
- remove
- insert_after
- insert_before
- sort (declared, not implemented)
```

The mutation methods are inspired by the pytorch FX graph methods.

### Safe Iterators

The behavior of the iterators is:

- If new elements are inserted after the current node, the iterator will iterate over them as well.
- If new elements are inserted before the current node, they will not be iterated over in this iteration.
- If the current node is lifted and inserted in a different location, iteration will start from the "next" node at the _original_ location.

A node cannot be added to a graph if it belongs to another graph. It needs to be removed first. The user is responsible for removing nodes from other graphs before adding the same node to a new graph.

### Linked list implementation
The doubly linked list implementation is inspired by the pytorch FX graph: They are both doubly linked lists with a dummy root node. Whereas in FX `graph` contains the root node and `Node`s are the links; we create the `DoublyLinkedOrderedSet` class to decouple the list implementation form the graph. Additionally, we created the `LinkedBox` class as a data struct to hold the prev/next pointers for the Nodes to allow nodes to move across graphs. This is in contrast to FX restriction of all nodes must belong to the same graph in their lifecycle. By allowing the nodes to move to different graphs, we are able to flatten/unflatten graphs without copying.

DoublyLinkedOrderedSet is unit tested.

### Next steps

1. Naming authority
2. Graph input/output/initializers mutation


onnxscript/ir/_core.py
```diff
@@ -633,18 +645,57 @@ def inputs(self, _: Any) -> None:
             "Directly mutating the input sequence is unsupported. Please use Node.replace_input_with() instead."
         )
 
-    def replace_input_with(self, index: int, new_input: Value | None) -> None:
+    def set_input(self, index: int, value: Value | None) -> None:
```
justinchuby, line 648: Make it a setter instead

onnxscript/ir/_core.py
```diff
@@ -633,18 +645,57 @@ def inputs(self, _: Any) -> None:
             "Directly mutating the input sequence is unsupported. Please use Node.replace_input_with() instead."
```
justinchuby, line 645: Update

onnxscript/ir/_core.py
```diff
@@ -952,6 +1003,193 @@ def __init__(
...
+        prev._next = next
+    if next is not None:
+        next._prev = prev
+
+
+def _connect_node_sequence(nodes: Sequence[Node]) -> None:
+    """Connect a sequence of nodes in a graph."""
+    if not nodes:
+        return
+    for i in range(len(nodes) - 1):
```
justinchuby, line 1018: Handle single element case

onnxscript/ir/_core.py
```diff
@@ -952,6 +1003,193 @@ def __init__(
...
+                if node._next is not None:
+                    node._next._prev = node._prev
+                if self._head is node:
+                    self._head = node._next
+                if self._tail is node:
+                    self._tail = node._prev
+                node._graph = None
+            node = node._next
+
+    def append(self, node: Node) -> None:
```
justinchuby, line 1088: Remove

onnxscript/ir/_core.py
```diff
@@ -952,6 +1003,193 @@ def __init__(
...
+                yield node
+            node = node._next
+        # TODO: Find the right time to call _remove_erased_nodes
+
+    def __len__(self) -> int:
+        return self._length
+
+    def __getitem__(self, index: int) -> Node:
+        if index < 0:
+            raise IndexError("Index must be non-negative")
```
justinchuby, line 1061: Update to support

onnxscript/ir/_linked_list.py
BowenBao, line 1: need a way to somehow disable the auto messages from codecov and scanning

justinchuby: Why?

justinchuby: Oh, to hide the annotations, you can press `a`

justinchuby:
Modification in double iteration on the same graph is safe, but moving nodes from one to another is not. How should we handle/guard against it? Use boxes?

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,217 @@
...
+# Inspired by https://github.com/pytorch/pytorch/blob/064a650b635e6fdaa8cf1a0dbc7dbbd23a37265d/torch/fx/graph.py
+
+from __future__ import annotations
+
+import warnings
+from typing import Callable, Generic, Iterable, Iterator, Protocol, TypeVar
+
+
+class Linkable(Protocol):
+    """A link in a doubly linked list that has a reference to the actual object in the link.
```
justinchuby, line 14:
```suggestion
    """A link in a doubly linked list.
```

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,217 @@
...
+
+    A fields are private and are managed by DoublyLinkedList
```
justinchuby, line 16:
```suggestion
    All fields are private and are managed by DoublyLinkedList
```

onnxscript/ir/_core.py
```diff
@@ -995,11 +1053,11 @@ def __init__(
...
         self._metadata_props: dict[str, str] | None = None
-
-        # Assign this graph as the owning_graph of all nodes
-        self._nodes = list(nodes)
-        for node in self._nodes:
-            node.graph = self
+        self._nodes: _linked_list.DoublyLinkedList[Node] = _linked_list.DoublyLinkedList(
+            root=_create_root_node_for_linked_list
+        )
+        # Call self.extend not self._nodes so the graph reference is added to the nodes
```
justinchuby, line 1059:
```suggestion
        # Call self.extend not self._nodes.extend so the graph reference is added to the nodes
```

justinchuby:
```
Program: /home/justinchu/dev/onnx-script/testtest.py

0.048 <module>  testtest.py:1
├─ 0.031 deserialize_model  onnxscript/ir/serde.py:279
│  ├─ 0.028 _deserialize_graph  onnxscript/ir/serde.py:360
│  │  ├─ 0.026 <listcomp>  onnxscript/ir/serde.py:403
│  │  │  ├─ 0.025 _deserialize_node  onnxscript/ir/serde.py:624
│  │  │  │  ├─ 0.010 [self]  onnxscript/ir/serde.py
│  │  │  │  ├─ 0.006 deserialize_value_info_proto  onnxscript/ir/serde.py:453
│  │  │  │  │  ├─ 0.004 deserialize_type_proto_for_shape  onnxscript/ir/serde.py:464
│  │  │  │  │  │  ├─ 0.002 <listcomp>  onnxscript/ir/serde.py:470
│  │  │  │  │  │  │  ├─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  │  │  │  └─ 0.001 deserialize_dimension  onnxscript/ir/serde.py:527
│  │  │  │  │  │  │     └─ 0.001 Dimension.__init__  onnxscript/ir/_core.py:368
│  │  │  │  │  │  └─ 0.002 Shape.__init__  onnxscript/ir/_core.py:394
│  │  │  │  │  │     └─ 0.002 isinstance  <built-in>
│  │  │  │  │  └─ 0.002 deserialize_type_proto_for_type  onnxscript/ir/serde.py:492
│  │  │  │  │     ├─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  │     └─ 0.001 DataType.__call__  enum.py:359
│  │  │  │  │           [2 frames hidden]  enum
│  │  │  │  ├─ 0.004 <listcomp>  onnxscript/ir/serde.py:651
│  │  │  │  │  ├─ 0.003 _deserialize_attribute  onnxscript/ir/serde.py:574
│  │  │  │  │  │  ├─ 0.002 [self]  onnxscript/ir/serde.py
│  │  │  │  │  │  └─ 0.001 _get_field  onnxscript/ir/serde.py:235
│  │  │  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  ├─ 0.003 Node.__init__  onnxscript/ir/_core.py:487
│  │  │  │  │  ├─ 0.002 Value.add_user  onnxscript/ir/_core.py:901
│  │  │  │  │  │  ├─ 0.001 [self]  onnxscript/ir/_core.py
│  │  │  │  │  │  └─ 0.001 set.add  <built-in>
│  │  │  │  │  └─ 0.001 [self]  onnxscript/ir/_core.py
│  │  │  │  └─ 0.002 _get_field  onnxscript/ir/serde.py:235
│  │  │  │     └─ 0.002 NodeProto.HasField  <built-in>
│  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  └─ 0.002 Graph.__init__  onnxscript/ir/_core.py:1026
│  │     └─ 0.002 Graph.extend  onnxscript/ir/_core.py:1113
│  │        └─ 0.002 DoublyLinkedList.extend  onnxscript/ir/_linked_list.py:227
│  │           └─ 0.002 DoublyLinkedList.append  onnxscript/ir/_linked_list.py:221
│  │              ├─ 0.001 DoublyLinkedList._insert_one_after  onnxscript/ir/_linked_list.py:153
│  │              └─ 0.001 [self]  onnxscript/ir/_linked_list.py
│  ├─ 0.002 deserialize_function  onnxscript/ir/serde.py:418
│  │  └─ 0.002 <listcomp>  onnxscript/ir/serde.py:424
│  │     └─ 0.002 _deserialize_node  onnxscript/ir/serde.py:624
│  │        ├─ 0.001 Node.__init__  onnxscript/ir/_core.py:487
│  │        └─ 0.001 _get_field  onnxscript/ir/serde.py:235
│  │           └─ 0.001 getattr  <built-in>
│  └─ 0.001 _deserialized_experimental_value_info_for_function_ir9  onnxscript/ir/serde.py:307
│     └─ 0.001 _parse_experimental_function_value_info_name  onnxscript/ir/serde.py:250
└─ 0.017 serialize_model  onnxscript/ir/serde.py:677
   └─ 0.017 serialize_model_into  onnxscript/ir/serde.py:681
      ├─ 0.016 serialize_graph_into  onnxscript/ir/serde.py:819
      │  ├─ 0.010 serialize_node_into  onnxscript/ir/serde.py:929
      │  │  ├─ 0.003 [self]  onnxscript/ir/serde.py
      │  │  ├─ 0.002 serialize_attribute_into  onnxscript/ir/serde.py:1007
      │  │  │  ├─ 0.001 _fill_in_value_for_attribute  onnxscript/ir/serde.py:1016
      │  │  │  │  └─ 0.001 serialize_tensor_into  onnxscript/ir/serde.py:972
      │  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
      │  │  ├─ 0.002 Value.name  onnxscript/ir/_core.py:908
      │  │  ├─ 0.001 RepeatedCompositeContainer.add  <built-in>
      │  │  ├─ 0.001 Attr.__instancecheck__  typing.py:1490
      │  │  │     [2 frames hidden]  typing, abc
      │  │  └─ 0.001 Node.metadata_props  onnxscript/ir/_core.py:713
      │  ├─ 0.003 serialize_value_into  onnxscript/ir/serde.py:1078
      │  │  ├─ 0.001 serialize_type_into  onnxscript/ir/serde.py:1103
      │  │  │  └─ 0.001 TensorType.dtype  onnxscript/ir/_core.py:748
      │  │  ├─ 0.001 [self]  onnxscript/ir/serde.py
      │  │  └─ 0.001 serialize_shape_into  onnxscript/ir/serde.py:1122
      │  │     └─ 0.001 serialize_dimension_into  onnxscript/ir/serde.py:1128
      │  ├─ 0.001 [self]  onnxscript/ir/serde.py
      │  ├─ 0.001 Value.is_graph_output  onnxscript/ir/_core.py:973
      │  └─ 0.001 _should_create_value_info_for_value  onnxscript/ir/serde.py:715
      └─ 0.001 serialize_function_into  onnxscript/ir/serde.py:864
         └─ 0.001 serialize_node_into  onnxscript/ir/serde.py:929
            └─ 0.001 serialize_attribute_into  onnxscript/ir/serde.py:1007
               └─ 0.001 _fill_in_value_for_attribute  onnxscript/ir/serde.py:1016
                  └─ 0.001 serialize_graph_into  onnxscript/ir/serde.py:819
                     └─ 0.001 serialize_node_into  onnxscript/ir/serde.py:929
                        └─ 0.001 serialize_attribute_into  onnxscript/ir/serde.py:1007
                           └─ 0.001 _fill_in_value_for_attribute  onnxscript/ir/serde.py:1016
                              └─ 0.001 serialize_graph_into  onnxscript/ir/serde.py:819

```

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+            iterate over them as well.
+        - If new elements are inserted before the current node, they will
+            not be iterated over in this iteration.
+        - If the current node is lifted and inserted in a different location,
+            iteration will start from the "next" node at the _original_ location.
+        """
+        box = self._root.next
+        while box is not self._root:
+            if box.owning_list is not self:
+                raise RuntimeError(f"Element {box!r} is not in the list")
```
justinchuby, line 119: Cover this line

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+        Args:
+            box: The box which the new value is to be inserted.
+            new_value: The new value to be inserted.
+            property_modifier: A function that modifies the properties of the new node.
+        """
+        if box.value is new_value:
+            # Do nothing if the new value is the same as the old value
+            return
+        if box.owning_list is not self:
+            raise ValueError(f"Value {box.value!r} is not in the list")
```
justinchuby, line 183: Cover

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+        self._length += 1
+
+    def remove(
+        self, value: TLinkable, property_modifier: Callable[[TLinkable], None] | None = None
+    ) -> None:
+        """Remove a node from the list."""
+        if value._link_box is None:
+            raise ValueError(f"Value {value!r} does not belong to any list")
+        if value._link_box.owning_list is not self:
+            raise ValueError(f"Value {value!r} is not in the list")
```
justinchuby, line 215: Cover

onnxscript/ir/_core.py
```diff
@@ -1027,11 +1072,67 @@ def opset_imports(self) -> dict[str, int]:
...
+            raise ValueError(f"The node {node} does not belong to this graph.")
+        self._nodes.remove(node, property_modifier=self._remove_graph_reference_from_node)
+
+    def insert_after(self, node: Node, new_nodes: Iterable[Node]) -> None:
+        """Insert new nodes after the given node in O(#new_nodes) time."""
+        self._nodes.insert_after(
+            node, new_nodes, property_modifier=self._add_graph_reference_to_node
+        )
+
+    def insert_before(self, node: Node, new_nodes: Iterable[Node]) -> None:
```
BowenBao, line 1124: How to insert only 1 node? `graph.insert_before(n, [new_node])`?

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+        # Remove the link box and detach the value from the box
+        value._link_box.erase()
+
+        # Call the property modifier in case the users want to modify the properties
+        # For example, when a node is removed from a graph, we want to unset its graph property
```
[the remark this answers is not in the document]
BowenBao: nvm, was looking at wrong file. Not same "user"

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+            raise ValueError(f"Value {value!r} is not in the list")
+        insertion_point = value._link_box
+        for new_value in new_values:
+            self._insert_one_after(
+                insertion_point, new_value, property_modifier=property_modifier
+            )
+            assert new_value._link_box is not None
+            insertion_point = new_value._link_box
+
+    def insert_before(
```
BowenBao, line 266: nit: possible to call `insert_after` internally?

justinchuby: Done

onnxscript/ir/_core.py
justinchuby: That's correct. Any thoughts?

onnxscript/ir/_linked_list.py
```diff
@@ -0,0 +1,291 @@
...
+        new_value: TLinkable,
+        property_modifier: Callable[[TLinkable], None] | None = None,
+    ) -> None:
+        """Insert a new value after the given value.
+
+        All insertion methods should call this method to ensure that the list is updated correctly.
+
+        Example::
+            Before: A -> B -> C
+            Call: insert_after(B, D)
```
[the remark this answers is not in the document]
justinchuby: Updated

onnxscript/ir/_protocols.py
```diff
@@ -219,6 +221,33 @@ class NodeProtocol(Protocol):
     version: int | None
     doc_string: str | None
     metadata_props: Mapping[str, str]
+    meta: Mapping[str, Any]
+
+
+@typing.runtime_checkable
+class MutableNodeProtocol(Protocol):
+    """Protocol for a topologically mutable node."""
```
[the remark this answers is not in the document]
justinchuby: Yes. Any suggestions?

justinchuby:
I would like to bring https://github.com/microsoft/onnxscript/pull/1349 into this PR. Created as a separate PR for review.

onnxscript/ir/_core.py
BowenBao: have you considered supporting single node as argument as well and impl with isinstance check?

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
BowenBao: Can we merge and just keep one Node protocol?

onnxscript/ir/_core.py
```diff
@@ -1216,7 +1308,7 @@ def __repr__(self) -> str:
 )"""
 
 
-class Function(_protocols.FunctionProtocol, _display.PrettyPrintable):
+class Function(_protocols.MutableFunctionProtocol, _display.PrettyPrintable):
```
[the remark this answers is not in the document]
justinchuby: I MutableFunctionProtocol has all FunctionProtocol methods. I would like to expose the immutable `FunctionProtocol` for analysis passes when users do not need to modify the topology. Aka the protocol is user facing.

onnxscript/ir/_core.py
```diff
@@ -1027,11 +1072,67 @@ def opset_imports(self) -> dict[str, int]:
 
     @property
     def nodes(self) -> Sequence[Node]:
```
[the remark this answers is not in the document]
justinchuby: I use Sequence because we do not want to make any stronger guarantees than that

onnxscript/ir/_core.py
justinchuby: > have you considered supporting single node as argument as well and impl with isinstance check?

We could but that could be added complexity. Do you have concrete use cases?

justinchuby: > I see torch using
> 
> ```python
> with graph.inserting_after(node):
>     ...
> ```
> 
> Cause they have the need of creating nodes on the fly. I guess we don't have this need? Other than that, I think the imple here is simpler.

Pytorch needs the context manager because its graph is responsible for creating nodes. We don't need the context manager because our nodes are initialized independent of the graph, and can be inserted anywhere in any graph.

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
justinchuby: I would like to expose the immutable `NodeProtocol` for analysis passes when users do not need to modify the topology. Aka the protocol is user facing.

Happy to take better names though. For example we can rename `NodeProtocol` to `NodeViewProtocol` and remake `NodeProtocol` mutable.

onnxscript/ir/_protocols.py
```diff
@@ -254,9 +283,79 @@ class GraphProtocol(Protocol):
...
+
+    def insert_after(self, node: NodeProtocol, new_nodes: Iterator[NodeProtocol]) -> None:
+        """Insert new nodes after the given node."""
+        ...
+
+    def insert_before(self, node: NodeProtocol, new_nodes: Iterator[NodeProtocol]) -> None:
+        """Insert new nodes before the given node."""
+        ...
+
+    def sort(self) -> None:
```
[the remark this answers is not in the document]
justinchuby: The choices are offering as a core api like this, or as a convenience function in a separate module. I think it is reasonable to assume that when users call sort() they are conscious about the runtime implication. It is also an obvious line to be spotted and optimized away.

justinchuby:
I decided to remove the read-only protocols. I realized they are overly complicated as many object fields refer to them and it is very hard to stay consistent.

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
justinchuby: Merged

onnxscript/ir/_core.py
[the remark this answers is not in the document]
justinchuby: Merged instead

onnxscript/ir/_core.py
```diff
@@ -1027,11 +1076,67 @@ def opset_imports(self) -> dict[str, int]:
...
+        # pylint: disable=protected-access
+        if node._graph is not None:
+            node._graph.remove(node)
+        node._graph = self
+        # pylint: enable=protected-access
+
+    # Mutation methods
+    def append(self, node: Node) -> None:
+        """Append a node to the graph in O(1) time."""
+        self._nodes.append(node, property_modifier=self._add_graph_reference_to_node)
```
gramalingam, line 1110: Is the property_modifier outdated? (same below.)

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
BowenBao: I recall we settled on requesting users to maintain topological sorted nodes when writing passes. IMO it is a trade-off between ease of use and performance. I'm still not sure it is worth the extra perf overhead and complexity brought in with sort.

onnxscript/ir/_core.py
justinchuby: Yes. Once I merge #1349 to this pr they will be gone

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
justinchuby: Right - the users are required to maintain the order. They can do it in two ways: (1) maintain the order at all times or (2) insert anywhere then sort. It would be the user's choice on how they want to maintain the order, but we can provide a sort() users can call.

onnxscript/ir/_core.py
justinchuby: Added an isinstance check to support single node

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
BowenBao: Option (2) is not a practical option if we do not provide sort api. Users won't do it if they have to implement sort on our custom linked list data structure themselves.
Both my comments are around concerns of adding sort api that makes option (2) feasible, and whether the trade-off is worth it.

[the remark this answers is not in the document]
gramalingam: Sounds good to me. @BowenBao : would you prefer topological sort be an external utility, not a class method? I am not sure I understand the concern. It is useful to have a topological sort utility somewhere ... ORT's graph has it; onnx users ask for it for onnx models (even though the spec says nodes must be sorted, not all tools produce sorted nodes). It is not being called implicitly, only explicitly by an user.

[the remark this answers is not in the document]
BowenBao: @gramalingam I'm hesitant for this utility to exist. Do we know use cases where sort is practically required, and not just making things more convenient?

BowenBao (approved):

onnxscript/ir/_protocols.py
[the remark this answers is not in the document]
justinchuby: Yeah for now it can stay unimplemented if we prefer. IMO users are going to need the sorting capability. Sometimes it is easier and potentially even cheaper to just sort at the end. I would prefer implementing a correct and tested sort then to let users create multiple versions of sorts on their own which can be not as well tested.

That said, if a user need to implement sort they will not need to understand the linked list. They can simply sort the nodes and call `extend()` and supply the sorted list.

justinchuby:
Thanks all for reviewing. All comments are addressed. I will create follow-ups if there are more comments.

PR #1348, opened 2024-04-06
(DRAFT) RFC: Rewriter design

gramalingam:
We already have a couple of approaches to pattern-matching based rewrite-rules: the original one (restricted to single-output patterns, but supporting other features such as a mechanism to bind variables to specific nodes of the pattern, and using them to check for extra-conditions before applying a rewrite-rule, and allowing for more efficient match-and-replace), and a newer extension to support multiple-output patterns, but not integrated with the other mechanisms.

This is an initial proposal for the API to expose to users writing rewrite-rules.

This meant for discussion, feedback, and comments. It seemed a bit easier to do this with a PR and markdown.

### DIFF
--- a/docs/rewriter/design.md
+++ b/docs/rewriter/design.md
@@ -1,0 +1,164 @@
+# Pattern Matching Based Rewrite Rules
+
+This design document describes a proposed API for specifying rewrite-rules for
+optimizing and transforming ONNX models.
+
+In its simple form, a rewrite-rule contains a source-pattern and a target. The
+source-pattern describes a sub-graph that is to be replaced by the specified
+target. In an extended form, the rewrite-rule can also specify a _condition_
+to be evaluated for every matching sub-graph to decide whether the replacement
+should happen or not. These are referred to as _conditional rewrite rules_.
+
+Here is a simple example of a _pattern_ that looks for a subgraph that is
+gemm-like:
+```py
+def gemm_pattern(alpha, beta, A, B, C):
+    term1 = alpha * op.MatMul (A, B)
+    term2 = beta * C
+    return term1 + term2
+```
+
+A successful match of this pattern against a subgraph in a given model can
+be represented by a dictionary mapping the inputs of this pattern (such
+as `alpha` and `A`) to _values_ in the graph IR. (Note that a value
+in the IR represents either an input of a function/graph or a computed
+value, that is, an output of some node). More generally, the dictionary
+can also store the values bound to intermediate variables such as `term1`.
+We refer to this dictionary as a _match\_binding_.
+
+This allows us to encode extra-conditions that we wish to check after
+a successful match. For example, to replace this subgraph by a single
+call to `Gemm` requires some additional conditions: we may require
+`alpha` to be a scalar constant (so that it can be replaced by an
+attribute in the call to `Gemm`), and we may need to check that there
+is no broadcasting in the `MatMul` since `Gemm` does not support broadcast.
+
+We can encode such conditions using a guard function as below:
+```py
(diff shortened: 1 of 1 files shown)


docs/rewriter/design.md
```diff
@@ -0,0 +1,130 @@
...
+factory (constructor) methods for patterns and rewrite-rules. Specifically,
+these methods will be the developer-facing APIs for expressing rewrite-rules.
+Hence, it is important for us to consolidate these APIs to avoid potential
+confusion among users writing rewrite-rules. (The implementations of the
+pattern-matching can vary, depending on the complexity of the pattern,
+such as whether it is single-output or multiple-output, etc.)
+
+Proposal:
+
+* The rewrite-rule constructor takes two parameters, one to specify the
```
xadupre, line 111: I suggest keeping this extra condition parameter even if it may return True by default. Some verifications are about shapes or types. For example, it is easy to look for nodes Expand and then check that they can be removed if they expand a tensor to the same size.

docs/rewriter/design.md
```diff
@@ -0,0 +1,130 @@
...
+(that is, `**match_bindings`) and returns either `None` (in case of failure)
+or a representation of the modification to the graph (in terms of removed
+and added nodes/values).
+
+* We may benefit from using a trace-mode onnxscript function above. In
+particular, we may be able to improve upon the replacement function by
+integrating the graph-modification cleanly into a generalization of the
+trace-mode onnxscript.
+
+
```
xadupre, line 129: I would add an item about the information the user needs to understand why a pattern did not work as expected. It is better to design it at the beginning than add it later.

gramalingam: That would be great. Any concrete suggestions? Would it be just information printed/logged (controlled by parameters like `verbose` in your implementation, or whenever a match fails in the middle, etc.)?

docs/rewriter/design.md
```diff
@@ -0,0 +1,130 @@
...
+* The rewrite-rule constructor takes two parameters, one to specify the
+pattern, and one to specify the replacement. We can provide a version with
+an extra condition parameter (as a convenience feature).
+
+* The pattern can be either a FunctionProto (or, equivalently, an ONNX Script
+function) or a function that constructs a pattern-graph or a pattern-graph.
+In the first two cases, they are converted into a pattern-graph.
+
+* The replacement is a function that accepts an expanded match-bindings
+(that is, `**match_bindings`) and returns either `None` (in case of failure)
```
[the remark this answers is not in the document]
gramalingam: Is this for the first part (of searching for a pattern-match), which is Xavier's point below? The failure mentioned in the above line is in user's code (the replacement-function), so it is presumably easier for the user to track this (they can return an appropriate error message they choose instead of just "None", I guess).

For the first part, see discussion below.

docs/rewriter/design.md
xadupre: I implemented a mechanism which helped me to understand when a match fails which nodes of pattern are matched, why the match fail and where. The example i added illustrates what the user sees when a match fails.

docs/rewriter/design.md
```diff
@@ -0,0 +1,133 @@
...
+such as whether it is single-output or multiple-output, etc.)
+
+Proposal:
+
+* The rewrite-rule constructor takes two parameters, one to specify the
+pattern, and one to specify the replacement. We can provide a version with
+an extra condition parameter (as a convenience feature).
+
+* The pattern can be either a FunctionProto (or, equivalently, an ONNX Script
+function) or a function that constructs a pattern-graph or a pattern-graph.
```
titaiwangms, line 116: Is `a pattern-graph or a pattern-graph.` a typo?

gramalingam:
Completed.

PR #1349, opened 2024-04-06
Simplify linked list design with a hash table

justinchuby:
Use a dictionary to map the values to the boxes in the doubly linked list implementation to

1. Remove the private pointer from Node to LinkedBox
2. Simplify membership checks
3. Ensure elements in the list are unique

Also changed:
1. The node <-> graph relationships are now managed by Graph directly and not by a property_modifier function. Meaning the user is responsible to remove nodes from other graphs before adding the same node to a new graph. This helps separate the concern of the graph from the linked list, allowing us to treat the linked list as a normal list. The list also doesn't need to handle exclusivity, meaning the same value will not be removed from the original list when being added to a new list and can appear in multiple lists, which is what we expect how lists behave.

### DIFF
--- a/onnxscript/ir/_linked_list.py
+++ b/onnxscript/ir/_linked_list.py
@@ -9,8 +9,7 @@
 
 from __future__ import annotations
 
-from typing import Callable, Generic, Iterable, Iterator, Protocol, Sequence, TypeVar
-
+from typing import Generic, Iterable, Iterator, Sequence, TypeVar
 
 T = TypeVar("T")
 
--- a/onnxscript/ir/_linked_list.py
+++ b/onnxscript/ir/_linked_list.py
@@ -25,19 +25,19 @@ class _LinkBox(Generic[T]):
     list without losing the links. This allows us to remove the object from the list during
     iteration and place the object into a different list without breaking any chains.
 
-    This is an internal class and should only be initialized by the :class:`DoublyLinkedHashList`.
+    This is an internal class and should only be initialized by the :class:`DoublyLinkedSet`.
 
     Attributes:
         prev: The previous box in the list.
         next: The next box in the list.
         erased: A flag to indicate if the box has been removed from the list.
-        owning_list: The :class:`DoublyLinkedHashList` to which the box belongs.
+        owning_list: The :class:`DoublyLinkedSet` to which the box belongs.
         value: The actual object in the list.
     """
 
     __slots__ = ("prev", "next", "value", "owning_list")
 
-    def __init__(self, owner: DoublyLinkedHashList[T], value: T | None) -> None:
+    def __init__(self, owner: DoublyLinkedSet[T], value: T | None) -> None:
         """Create a new link box.
 
         Args:
@@ -49,7 +49,7 @@ class _LinkBox(Generic[T]):
         self.prev: _LinkBox[T] = self
         self.next: _LinkBox[T] = self
         self.value: T | None = value
-        self.owning_list: DoublyLinkedHashList[T] = owner
+        self.owning_list: DoublyLinkedSet[T] = owner
 
     @property
     def erased(self) -> bool:
@@ -69,7 +69,7 @@ class _LinkBox(Generic[T]):
         return f"_LinkBox({self.value!r}, erased={self.erased}, prev={self.prev.value!r}, next={self.next.value!r})"
 
 
-class DoublyLinkedHashList(Generic[T], Sequence[T]):
+class DoublyLinkedSet(Generic[T], Sequence[T]):
     """A doubly linked list of nodes.
 
     Adding and removing elements from the list during iteration is safe. Moving elements
@@ -87,6 +87,8 @@ class DoublyLinkedHashList(Generic[T], Sequence[T]):
         Inserting and removing nodes from the list is O(1). Accessing nodes by index is O(n),
         although accessing nodes at either end of the list is O(1). I.e. `list[0]` and `list[-1]`
         are O(1).
+
+    Values need to be hashable. `None` is not a valid value in the list.
     """
 
     __slots__ = ("_root", "_length", "_values_to_boxes")
@@ -176,6 +178,8 @@ class DoublyLinkedHashList(Generic[T], Sequence[T]):
             box: The box which the new value is to be inserted.
             new_value: The new value to be inserted.
         """
+        if new_value is None:
+            raise TypeError("DoublyLinkedSet does not support None values")
         if box.value is new_value:
             # Do nothing if the new value is the same as the old value
             return box
@@ -269,4 +273,4 @@ class DoublyLinkedHashList(Generic[T], Sequence[T]):
         return self._insert_many_after(insertion_point, new_values)
 
     def __repr__(self) -> str:
-        return f"DoublyLinkedHashList({list(self)})"
+        return f"DoublyLinkedSet({list(self)})"
--- a/onnxscript/ir/_linked_list.py
+++ b/onnxscript/ir/_linked_list.py
@@ -13,16 +13,11 @@ from __future__ import annotations
 from typing import Callable, Generic, Iterable, Iterator, Protocol, Sequence, TypeVar
 
 
-class Linkable(Protocol):
-    # pylint: disable=unused-private-member
-    _link_box: _LinkBox | None
-    # pylint: enable=unused-private-member
+
+T = TypeVar("T")
 
 
-TLinkable = TypeVar("TLinkable", bound=Linkable)
-
-
-class _LinkBox(Generic[TLinkable]):
+class _LinkBox(Generic[T]):
     """A link in a doubly linked list that has a reference to the actual object in the link.
 
     The :class:`_LinkBox` is a container for the actual object in the list. It is used to
@@ -45,7 +40,7 @@ class _LinkBox(Generic[TLinkable]):
 
     __slots__ = ("prev", "next", "value", "owning_list")
 
-    def __init__(self, owner: DoublyLinkedList[TLinkable], value: TLinkable | None) -> None:
+    def __init__(self, owner: DoublyLinkedList[T], value: T | None) -> None:
         """Create a new link box.
 
         Args:
@@ -54,12 +49,10 @@ class _LinkBox(Generic[TLinkable]):
                 the link box is considered erased (default). The root box of the list
                 should be created with a None value.
         """
-        self.prev: _LinkBox[TLinkable] = self
-        self.next: _LinkBox[TLinkable] = self
-        if value is not None:
-            value._link_box = self  # pylint: disable=protected-access
-        self.value: TLinkable | None = value
-        self.owning_list: DoublyLinkedList[TLinkable] = owner
+        self.prev: _LinkBox[T] = self
+        self.next: _LinkBox[T] = self
+        self.value: T | None = value
+        self.owning_list: DoublyLinkedList[T] = owner
 
     @property
     def erased(self) -> bool:
@@ -73,14 +66,13 @@ class _LinkBox(Generic[TLinkable]):
         prev, next_ = self.prev, self.next
         prev.next, next_.prev = next_, prev
         # Detach the value
-        self.value._link_box = None  # pylint: disable=protected-access
         self.value = None
 
     def __repr__(self) -> str:
         return f"_LinkBox({self.value!r}, erased={self.erased}, prev={self.prev.value!r}, next={self.next.value!r})"
 
 
-class DoublyLinkedList(Generic[TLinkable], Sequence[TLinkable]):
+class DoublyLinkedList(Generic[T], Sequence[T]):
     """A doubly linked list of nodes.
 
     Adding and removing elements from the list during iteration is safe. Moving elements
@@ -108,7 +100,7 @@ class DoublyLinkedList(Generic[TLinkable], Sequence[TLinkable]):
         self._root: _LinkBox = root_
         self._length = 0
 
-    def __iter__(self) -> Iterator[TLinkable]:
+    def __iter__(self) -> Iterator[T]:
         """Iterate over the elements in the list.
 
         - If new elements are inserted after the current node, the iterator will
@@ -127,7 +119,7 @@ class DoublyLinkedList(Generic[TLinkable], Sequence[TLinkable]):
                 yield box.value
             box = box.next
 
-    def __reversed__(self) -> Iterator[TLinkable]:
+    def __reversed__(self) -> Iterator[T]:
         """Iterate over the elements in the list in reverse order."""
         box = self._root.prev
         while box is not self._root:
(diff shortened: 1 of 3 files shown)


justinchuby:
Code runs even faster

```
0.045 <module>  testtest.py:1
├─ 0.029 deserialize_model  onnxscript/ir/serde.py:279
│  ├─ 0.026 _deserialize_graph  onnxscript/ir/serde.py:360
│  │  ├─ 0.024 <listcomp>  onnxscript/ir/serde.py:403
│  │  │  ├─ 0.023 _deserialize_node  onnxscript/ir/serde.py:624
│  │  │  │  ├─ 0.007 <listcomp>  onnxscript/ir/serde.py:651
│  │  │  │  │  └─ 0.007 _deserialize_attribute  onnxscript/ir/serde.py:574
│  │  │  │  │     ├─ 0.002 deserialize_tensor  onnxscript/ir/serde.py:535
│  │  │  │  │     │  ├─ 0.001 TensorProtoTensor.__init__  onnxscript/ir/serde.py:83
│  │  │  │  │     │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  │     ├─ 0.002 [self]  onnxscript/ir/serde.py
│  │  │  │  │     ├─ 0.001 AttrTensor.__init__  onnxscript/ir/_core.py:1600
│  │  │  │  │     ├─ 0.001 AttrInt64s.__init__  onnxscript/ir/_core.py:1648
│  │  │  │  │     └─ 0.001 _get_field  onnxscript/ir/serde.py:235
│  │  │  │  │        └─ 0.001 AttributeProto.HasField  <built-in>
│  │  │  │  ├─ 0.007 deserialize_value_info_proto  onnxscript/ir/serde.py:453
│  │  │  │  │  ├─ 0.003 deserialize_type_proto_for_shape  onnxscript/ir/serde.py:464
│  │  │  │  │  │  ├─ 0.002 <listcomp>  onnxscript/ir/serde.py:470
│  │  │  │  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  │  ├─ 0.002 deserialize_type_proto_for_type  onnxscript/ir/serde.py:492
│  │  │  │  │  │  ├─ 0.001 TypeProto.HasField  <built-in>
│  │  │  │  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  │  │  │  └─ 0.002 Value.shape  onnxscript/ir/_core.py:928
│  │  │  │  ├─ 0.005 [self]  onnxscript/ir/serde.py
│  │  │  │  ├─ 0.003 Node.__init__  onnxscript/ir/_core.py:487
│  │  │  │  │  ├─ 0.002 <genexpr>  onnxscript/ir/_core.py:534
│  │  │  │  │  │  ├─ 0.001 [self]  onnxscript/ir/_core.py
│  │  │  │  │  │  └─ 0.001 Value.__init__  onnxscript/ir/_core.py:836
│  │  │  │  │  └─ 0.001 [self]  onnxscript/ir/_core.py
│  │  │  │  └─ 0.001 list.append  <built-in>
│  │  │  └─ 0.001 [self]  onnxscript/ir/serde.py
│  │  └─ 0.002 Graph.__init__  onnxscript/ir/_core.py:1021
│  │     └─ 0.002 DoublyLinkedHashList.__init__  onnxscript/ir/_linked_list.py:91
│  │        └─ 0.002 DoublyLinkedHashList.extend  onnxscript/ir/_linked_list.py:227
│  │           └─ 0.002 DoublyLinkedHashList.append  onnxscript/ir/_linked_list.py:223
│  │              ├─ 0.001 [self]  onnxscript/ir/_linked_list.py
│  │              └─ 0.001 DoublyLinkedHashList._insert_one_after  onnxscript/ir/_linked_list.py:153
│  ├─ 0.002 deserialize_function  onnxscript/ir/serde.py:418
│  │  └─ 0.002 <listcomp>  onnxscript/ir/serde.py:424
│  │     └─ 0.002 _deserialize_node  onnxscript/ir/serde.py:624
│  │        └─ 0.002 Node.__init__  onnxscript/ir/_core.py:487
│  │           └─ 0.002 <genexpr>  onnxscript/ir/_core.py:534
│  └─ 0.001 _deserialized_experimental_value_info_for_function_ir9  onnxscript/ir/serde.py:307
└─ 0.016 serialize_model  onnxscript/ir/serde.py:677
   └─ 0.016 serialize_model_into  onnxscript/ir/serde.py:681
      ├─ 0.015 serialize_graph_into  onnxscript/ir/serde.py:819
      │  ├─ 0.009 serialize_node_into  onnxscript/ir/serde.py:929
      │  │  ├─ 0.005 serialize_attribute_into  onnxscript/ir/serde.py:1007
      │  │  │  └─ 0.005 _fill_in_value_for_attribute  onnxscript/ir/serde.py:1016
      │  │  │     ├─ 0.004 serialize_tensor_into  onnxscript/ir/serde.py:972
      │  │  │     │  └─ 0.004 TensorProto.CopyFrom  <built-in>
      │  │  │     └─ 0.001 [self]  onnxscript/ir/serde.py
      │  │  ├─ 0.002 [self]  onnxscript/ir/serde.py
      │  │  ├─ 0.001 RepeatedScalarContainer.append  <built-in>
      │  │  └─ 0.001 Node.name  onnxscript/ir/_core.py:597
      │  ├─ 0.003 serialize_value_into  onnxscript/ir/serde.py:1078
      │  │  ├─ 0.002 serialize_shape_into  onnxscript/ir/serde.py:1122
      │  │  └─ 0.001 serialize_type_into  onnxscript/ir/serde.py:1103
      │  │     └─ 0.001 TensorType.dtype  onnxscript/ir/_core.py:748
      │  ├─ 0.002 [self]  onnxscript/ir/serde.py
      │  └─ 0.001 Value.is_graph_output  onnxscript/ir/_core.py:973
      │     └─ 0.001 Node.graph  onnxscript/ir/_core.py:719
      └─ 0.001 serialize_function_into  onnxscript/ir/serde.py:864
         └─ 0.001 _should_create_value_info_for_value  onnxscript/ir/serde.py:715
```

onnxscript/ir/_linked_list.py
```diff
@@ -1,24 +1,16 @@
 """Mutable list for nodes in a graph with safe mutation properties."""
```
gramalingam, line 1: Should add usual copyright notice to new files:
```
# -------------------------------------------------------------------------
# Copyright (c) Microsoft Corporation. All rights reserved.
```

onnxscript/ir/_linked_list.py
```diff
@@ -50,12 +42,10 @@ def __init__(self, owner: DoublyLinkedList[TLinkable], value: TLinkable | None)
                 the link box is considered erased (default). The root box of the list
```
gramalingam, line 42: Seems to prevent us from using a container where None is a legal value, preventing its general applicability in other scenarios. Not a major concern right now, but I have a vague feeling that's not a good idea.

gramalingam: I guess it is primarily an optimization right now to avoid a separate `erased` field.

justinchuby: Right I should have called that out. 

>  preventing its general applicability in other scenarios

This is valid. Maybe retaining the erased field is a good idea?

onnxscript/ir/_linked_list.py
```diff
@@ -184,13 +178,13 @@ def _insert_one_after(
             return box
```
gramalingam, line 178: Drop `property_modifier` above?

onnxscript/ir/_linked_list.py
```diff
@@ -269,20 +245,15 @@ def insert_after(
             new_values: The new values to be inserted.
             property_modifier: A function that modifies the properties of the new nodes.
```
justinchuby, line 246:
```suggestion
```

onnxscript/ir/_linked_list.py
```diff
@@ -291,11 +262,10 @@ def insert_before(
...
```
justinchuby, line 263:
```suggestion
```

onnxscript/ir/_linked_list.py
```diff
@@ -184,13 +178,13 @@ def _insert_one_after(
             return box
         if box.owning_list is not self:
             raise ValueError(f"Value {box.value!r} is not in the list")
-        # Remove the new value from the list if it is already in a different list
-        if new_value._link_box is not None:
-            new_value._link_box.owning_list.remove(new_value)
+
+        if new_value in self._values_to_boxes:
```
gramalingam, line 182: The implementation seems to reflect an assumption that this is semantically a "set" as opposed to a "list". While that is true for our intended application, it seems unusual for a doubly-linked-list implementation. Just commenting and still thinking. Is there any difference if we use `id(value)` instead of `value`?

gramalingam: I guess this stems from the methods like remove and insert_after ...

justinchuby: id can be tricky and may cause unexpected behavior:

```
>>> a = (1,)
>>> b = (1,)
>>> id(a)
2649810036928
>>> id(b)
2649815823504
>>> c = "s"
>>> d = "s"
>>> id(c)
2649810832368
>>> id(d)
2649810832368
```

Here to tuples are equal but they have different ids; two strings are equal and they have the same id. I suppose for our use cases (nodes) they would be the same. Using the value as keys makes the "set" behavior uniform across different types.

justinchuby: I renamed the class to "[...]HashList" hoping to reflect this behavior. Didn't use "SetList" because we don't expose methods one would find in a set like `add`. Happy to take a better name.

onnxscript/ir/_linked_list.py
```diff
@@ -133,9 +125,12 @@ def __reversed__(self) -> Iterator[TLinkable]:
             box = box.prev
 
     def __len__(self) -> int:
+        assert self._length == len(
+            self._values_to_boxes
+        ), "Bug in the implementation: length mismatch"
         return self._length
 
-    def __getitem__(self, index: int) -> TLinkable:
+    def __getitem__(self, index: int) -> T:
```
gramalingam, line 133: I wonder if we should forbid this for now, rather than create the impression of a constant-time access ... because we don't expect users to use this kind of access, do we?

onnxscript/ir/_linked_list.py
```diff
@@ -69,14 +59,13 @@ def erase(self) -> None:
...
         # Detach the value
-        self.value._link_box = None  # pylint: disable=protected-access
         self.value = None
 
     def __repr__(self) -> str:
         return f"_LinkBox({self.value!r}, erased={self.erased}, prev={self.prev.value!r}, next={self.next.value!r})"
 
 
-class DoublyLinkedList(Generic[TLinkable], Sequence[TLinkable]):
+class DoublyLinkedHashList(Generic[T], Sequence[T]):
```
gramalingam, line 68: May be DoublyLinkedSet is a better name (focused on the intended semantics that values occur only once in the ordered list ... adding "Hash" may not help, given the ambiguity in its role ("DoublyLinked" is the primary implementation mechanism, Hash gives random-access by value).

onnxscript/ir/_linked_list.py
BowenBao: We can callout that the DoubleLinkList requires and assumes all elements it holds to be unique.

If we really want to make the class behave like a general linked list, we can make `_values_to_boxes` point to a list of boxes.

justinchuby: > We can callout that the DoubleLinkList requires and assumes all elements it holds to be unique.

Right

> If we really want to make the class behave like a general linked list, we can make `_values_to_boxes` point to a list of boxes.

The map is for O(1) retrieval of the boxes from values. I wonder if there are other ways?

I also consider uniqueness to be a feature of this data structure, since it is something we want to maintain in the nodes list.

onnxscript/ir/_linked_list.py
justinchuby: Since the node is a sequence, I find it convenient to find a node by index, for example when I was exploring the model in the console. I think alternatively we can call out the runtime complexity in the docstring?

onnxscript/ir/_linked_list.py
justinchuby: The only concern I have for a "set" is that we would then assume set operations like `add` and `union` etc. The apis we provide resemble that of a list more. Of course that is not a huge concern, since it is a data structure we consume internally.

onnxscript/ir/_linked_list.py
justinchuby: Done

onnxscript/ir/_linked_list.py
justinchuby: Done

onnxscript/ir/_linked_list.py
justinchuby: Renamed to DoublyLinkedSet

onnxscript/ir/_linked_list.py
justinchuby: Documented. I will update the documentation for graph itself in a separate PR.

onnxscript/ir/_linked_list.py
justinchuby: Now raising an error if value is None.

onnxscript/ir/_linked_list.py
gramalingam: > id can be tricky and may cause unexpected behavior:

My interpretation of `id` is that it is similar to a physical-address (pointer) that remains unique for the object throughout its lifetime. It is perfectly consistent with your examples ... using `id` is acceptable when we expect objects to be unique, and value-based-equality/hash etc. is preferable when we expect we may create distinct objects that will compare equal.

What does the default-implementation of hash/equal do for Nodes? Is it the same as equality of id(...) or is it anything different?

gramalingam: I guess the important point is to avoid any complex has/equality on Nodes when using this underlying dictionary.

gramalingam (approved):

onnxscript/ir/_linked_list.py
gramalingam: > The only concern I have for a "set" is that we would then assume set operations like `add` and `union` etc. The apis we provide resemble that of a list more. Of course that is not a huge concern, since it is a data structure we consume internally.

Right. I just feel it helps to clarify the uniqueness of elements. Not a big deal either way once users understand what it is supposed to be. (HashList, on the other hand, might convey a linked-list implementation used for overflow-chains in a hash-table. I guess everything has some minor drawback.)

onnxscript/ir/_linked_list.py
justinchuby: Right. We need to make sure we don't define `__hash__` or equality for Nodes. The default hash of an object is some function of the id(). I am now thinking we should use id() to make this robust. Since we only care about classes we define (not really native python types), id() seems more beneficial.

Issue #997, opened 2023-08-09
Remove dependence on JitScalarType


justinchuby:
It is deprecated. Migrate the methods instead.

https://github.com/microsoft/onnxscript/blob/de52bd67a06503e02cc8a400b69b808818fa8e4e/onnxscript/function_libs/torch_lib/graph_building.py#L156-L160

PR #1354, opened 2024-04-09
Use the new IR for building graphs in torchlib

justinchuby:
Use the new IR for building graphs in torchlib. I created a flag `TORCHLIB_EXPERIMENTAL_USE_IR` to control the feature. When enabled the classes in `graph_building` will be swapped to use the new IR. The exporter should see the same interfaces and not feel anything. After the transition period we can decide where this should live.

Set `TORCHLIB_EXPERIMENTAL_USE_IR=1` to enable this feature.

Fix https://github.com/microsoft/onnxscript/issues/997


thiagocrepaldi (approved):
Overall LGTM. The TorchScript naming should go away, though :)

Maybe some docstrings for both `graph_building/__init__.py` and `ir/__init__.py` exaplining the goal of each module, the basic design idea/mechanics would help a lot the hand off to pytorch and new comers that start on the project

justinchuby:
> The TorchScript naming should go away

Absolutely. We need the exporter to move away from referencing this object and decide how it wants to interface with the IR. Then we can remove this class

PR #1359, opened 2024-04-09
[rewriter] Add Rule Map structure for perf improvement

shubhambhokare1:
Based on issue https://github.com/microsoft/onnx-rewriter/issues/234

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -985,6 +985,13 @@ class RewriteRuleSet:
         if commute:
             rules = list(itertools.chain.from_iterable([rule.commute() for rule in rules]))
         self.rules = rules
+        self.rule_mapping = {}
+        for rule in self.rules:
+            target_node_op_type = rule._target_node_pattern.op._value
+            if target_node_op_type not in self.rule_mapping:
+                self.rule_mapping[target_node_op_type] = [rule]
+            else:
+                self.rule_mapping[target_node_op_type].append(rule)
 
     def _apply_to_graph_or_function(
         self,
@@ -999,6 +1006,11 @@ class RewriteRuleSet:
         for rule in self.rules:
             deltas = []
             for i, node in enumerate(graph_or_function.nodes):
+                if (
+                    node.op_type not in self.rule_mapping
+                    or rule not in self.rule_mapping[node.op_type]
+                ):
+                    continue
                 if hasattr(rule, "pattern"):
                     from onnxscript.rewriter.generic_pattern import (
                         GenericRewriteRule,
--- a/onnxscript/rewriter/rule_mapping_test.py
+++ b/onnxscript/rewriter/rule_mapping_test.py
@@ -1,0 +1,35 @@
+import unittest
+
+import onnx.parser
+
+from onnxscript._legacy_ir import irbuilder
+from onnxscript.rewriter.pattern import RewriteRuleSet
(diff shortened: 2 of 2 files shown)


onnxscript/rewriter/rule_mapping_test.py
justinchuby, line 1: I would move the content in this file to `pattern_test.py` so it's easy to find

onnxscript/rewriter/rule_mapping_test.py
```diff
@@ -0,0 +1,35 @@
...
+            }
+        """
+        )
+
+        ir = irbuilder.build_ir(model)
+        new_rules = broadcast_to_matmul.rules.rules + [gemm_to_matmul_add.rule]
+        rule_set = RewriteRuleSet(new_rules)
+        count = rule_set.apply_to_model(ir)
+        self.assertEqual(count, 1)
+        self.assertEqual(len(ir.graph.nodes), 4)
```
justinchuby, line 31: It seems it's not immediately clear what we need to assert. Could you explain? Thanks!

onnxscript/rewriter/pattern.py
```diff
@@ -985,6 +985,13 @@ def __init__(self, rules: Sequence[RewriteRule], *, commute: bool = False) -> No
         if commute:
             rules = list(itertools.chain.from_iterable([rule.commute() for rule in rules]))
         self.rules = rules
+        self.rule_mapping = {}
+        for rule in self.rules:
+            target_node_op_type = rule._target_node_pattern.op._value
```
gramalingam, line 990: We should not assume that the target node pattern will have a definitive op value. (A recent PR allows us to create a node pattern that will match against all calls to ops/functions with a name that matches a prefix for example.)

However, in the common case it will. Hence, to optimize this, we will need a combined scheme: for the common case we should partition the rules using a dictionary (as in this PR). But we should have a fallback list of rules to be applied to all nodes (when the rule doesn't correspond to a single optype).

onnxscript/rewriter/pattern.py
```diff
@@ -999,6 +1006,11 @@ def _apply_to_graph_or_function(
         for rule in self.rules:
             deltas = []
             for i, node in enumerate(graph_or_function.nodes):
+                if (
+                    node.op_type not in self.rule_mapping
+                    or rule not in self.rule_mapping[node.op_type]
```
gramalingam, line 1011: This doesn't really give us much of a performance advantage. This is the unfortunate result of a concurrent PR change that swapped the order of the loops above (the earlier iteration was over nodes, and for every node iterate through matching rules).

We will need to swap the order of loops back to get the perf improvement. We will be able to do that by migrating to the new IR. The new IR will allow us to make a change to the graph, and continue iterating over the rest of the graph.

justinchuby:
For any perf work, I would make sure to use pyinstrument or other tools to show and validate the improvement.

onnxscript/rewriter/pattern.py
xadupre: I think you can switch the order of the loops. The second part of the loop is looking for conflicts. If two rules try to rewrite the same node, one will be applied. The other one will be tried again in the next iteration.

BowenBao: cc @titaiwangms

gramalingam:
I suggest closing this. I think this is worth doing, but the underlying code has changed significantly since this was written. So, it may be better off redoing this when we need it

PR #1364, opened 2024-04-10
[IR] Create GraphView for read-only view on a subgraph

justinchuby:
A read-only view on a graph.

The GraphView is useful for analysis of a subgraph. It can be initialized
with a subset of nodes from a :class:`Graph`. Creating GraphView does not
change the ownership of the nodes, and so it is possible to create multiple
GraphViews that contain the same nodes.

The graph view can be serialized to ONNX::

        graph_proto = ir.serde.serialize_graph(graph_view)

It can also be used to create a model::

        model = ir.Model(graph_view, ir_version=8)
        model_proto = ir.serde.serialize_model(model)

Example usage:

```python
import onnx

from onnxscript import ir

model = onnx.load("torchscript_model.onnx")
m = ir.serde.deserialize_model(model)
graph = m.graph

nodes = [
    node for node in graph if node.name is not None and node.name.startswith("/model/layers.0")
]

inputs = set()
outputs = set()
for node in nodes:
    inputs.update(node.inputs)
    outputs.update(node.outputs)

graph_inputs = sorted(inputs - outputs, key=lambda val: val.name)
graph_outputs = sorted(outputs - inputs, key=lambda val: val.name)

graph_view = ir.GraphView(graph_inputs, graph_outputs, nodes=nodes)

print(graph_view)

model = ir.Model(graph_view, ir_version=8)
onnx_model = ir.serde.serialize_model(model)
onnx.save(onnx_model, "torchscript_model_layer_0.onnx")
```

image


tests/ir/graph_view_test.py
```diff
@@ -0,0 +1,41 @@
...
+        outputs = set()
+        for node in nodes:
+            inputs.update(node.inputs)
+            outputs.update(node.outputs)
+
+        graph_inputs = sorted(inputs - outputs, key=lambda val: val.name)
+        graph_outputs = sorted(outputs - inputs, key=lambda val: val.name)
+
+        graph_view = ir.GraphView(graph_inputs, graph_outputs, nodes=nodes)
+        model = ir.Model(graph_view, ir_version=8)
```
BowenBao, line 35: Is cloning happening in this step?

onnxscript/ir/_core.py
```diff
@@ -508,7 +511,19 @@ def _quoted(string: str) -> str:
 
 
 class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
-    """IR Node."""
+    """IR Node.
+
+    When the node is initialized, it does not belong to any graph. It is the
+    responsibility of the caller to add the node to a graph, by calling :func:`Graph.absorb_nodes`.
```
BowenBao, line 517: `absorb_nodes` this seems new to me, maybe I missed it in previous PRs.

tests/ir/graph_view_test.py
justinchuby: There will not be cloning. It will just be a reference

BowenBao: What happens then if nodes inside the new model model are mutated? e.g., inputs rewired

justinchuby: The graphview will change as well if the underlying nodes change, Making that clear

onnxscript/ir/_core.py
justinchuby: Updated.

tests/ir/graph_view_test.py
justinchuby: Updated documentation.

> If the underlying nodes / connections are mutated, the mutation will be reflected in all views as well.

BowenBao, line 35: can it support local function nodes?

justinchuby: One won't be able to insert nodes into m2. They will only be able to insert nodes to `m1.graph`.

BowenBao: great, let's put that in docstring as well.

justinchuby: It can. We can create the same graphview on function nodes.

BowenBao: How will the new `ir.Model` instance populate its 'functions' field?

BowenBao (approved):
LGTM

tests/ir/graph_view_test.py
justinchuby: We can initialize a function using a graphview, or pass in functions from the original models, or operate directly on the graphview, depending on what we want to do.

justinchuby: I think it is clear in the following sense:
1. Models do not have mutation methods.
2. A model built with a graph view do not have mutation methods in its graph.

so users won't be able to modify the graph.

onnxscript/ir/_core.py
```diff
@@ -1254,61 +1259,172 @@ def metadata_props(self) -> dict[str, str]:
...
+    are mutated, the mutation will be reflected in all views as well.
+
+    The graph view can be serialized to ONNX::
+
+            graph_proto = ir.serde.serialize_graph(graph_view)
+
+    It can also be used to create a model::
+
+            model = ir.Model(graph_view, ir_version=8)
+            model_proto = ir.serde.serialize_model(model)
```
justinchuby, line 1343:
```suggestion
            model_proto = ir.serde.serialize_model(model)
            
        The model created with a GraphView will have a fixed topology, and its graph will stay read-only as a GraphView.
```

onnxscript/ir/_core.py
```diff
@@ -1254,61 +1259,174 @@ def metadata_props(self) -> dict[str, str]:
...
+            
+    The model created with a GraphView will have a fixed topology, and its graph will stay read-only as a GraphView. No copying will be done during the initialization process
```
justinchuby, line 1345:
```suggestion
    The model created with a GraphView will have a fixed topology, and its graph
    will remain read-only as a GraphView. No copying will be done during the
    initialization process.
```

tests/ir/graph_view_test.py
justinchuby: Added:

>    The model created with a GraphView will have a fixed topology, and its graph
>    will remain read-only as a GraphView. No copying will be done during the
>    initialization process.

tests/optimizer/test_models.py
```diff
@@ -30,7 +30,10 @@ def test_model_runs_and_matches_accuracy_after_optimization(self, model_name):
             self.skipTest(skip_reason)
 
         model_dir = f"{model_folder_path}/{model_name}/dynamo"
-        model = onnx.load(f"{model_dir}/{model_name}_dynamo.onnx")
+        model_path = f"{model_dir}/{model_name}_dynamo.onnx"
+        if not pathlib.Path(model_path).exists():
+            self.skipTest(f"Model {model_name} does not exist")
```
xadupre, line 35: {model_name!r}

onnxscript/ir/_core.py
```diff
@@ -1553,7 +1673,7 @@ def __str__(self) -> str:
         inputs_text = ",\n".join(str(x) for x in self.inputs)
         outputs_text = ",\n".join(str(x) for x in self.outputs)
         attributes_text = ",\n".join(
-            attr.name + f": {attr.type}" + f"= {attr.value}" * (attr.value is None)
+            attr.name + f": {attr.type}" + f" = {attr.value}" * (attr.value is None)
```
xadupre, line 1676: Why not using a single string instead of a concatenation?

onnxscript/ir/_core.py
```diff
@@ -1254,61 +1259,176 @@ def metadata_props(self) -> dict[str, str]:
...
+    Attributes:
+        name: The name of the graph.
+        inputs: The input values of the graph.
+        outputs: The output values of the graph.
+        nodes: All nodes visible in this view. They do not have to be sorted.
+        initializers: The initializers in the graph.
+        doc_string: Documentation string.
+        opset_imports: Opsets imported by the graph.
+        metadata_props: Metadata.
+    """
```
xadupre, line 1358: That would be great to add an example of the outputs in the documentation to understand what it looks like.

onnxscript/ir/_core.py
```diff
@@ -1254,61 +1259,176 @@ def metadata_props(self) -> dict[str, str]:
...
+            if initializer.name is None:
+                raise ValueError(f"Initializer must have a name: {initializer}")
+        self.initializers = {tensor.name: tensor for tensor in initializers}
+        self.doc_string = doc_string
+        self.opset_imports = opset_imports or {}
+        self._metadata: _metadata.MetadataStore | None = None
+        self._metadata_props: dict[str, str] | None = None
+        self.nodes: tuple[Node, ...] = tuple(nodes)
+
+    def __getitem__(self, index: int) -> Node:
```
xadupre, line 1396: If it is a view just meant to be displayed, maybe we can remove all these short methods.

onnxscript/ir/_core.py
justinchuby: Do you mean what the object prints out, or the graph outputs?

onnxscript/ir/_core.py
justinchuby: The view is designed for analysis/exploration etc. (like string_view for strings), and so one can use the same read-only methods on a Graph to read a GraphView. Was I missing anything?

onnxscript/ir/_core.py
justinchuby: Good point. Let me try that

tests/optimizer/test_models.py
justinchuby: Done

onnxscript/ir/_core.py
justinchuby: I think I understand now: The GraphView is not just meant to be displayed. I touch the repr methods because I wanted to share them between Graph and GraphView. I will add an example in the repr function later.

onnxscript/ir/_core.py
justinchuby: I merged the first parts. Seems a little hard to merge `+ f" = {attr.value}" * (attr.value is None)`

justinchuby:
Merging for now. I will create follow-ups if new comments come in

PR #1367, opened 2024-04-11
Support pattern rewriting with parameters

xadupre:


onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
         return proto
 
 
+class PatternMatchResult:
```
justinchuby, line 297: I like this structured way of returning results.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+    """Stores information about a match if a match was successful."""
+
+    def __init__(
+        self,
+        pattern: "GenericPattern",
```
justinchuby, line 302:
```suggestion
        pattern: GenericPattern,
```

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+        model_nodes: typing.Sequence[oir.Node],
```
justinchuby, line 303:
```suggestion
        model_nodes: Sequence[oir.Node],
```

nit: I would import Sequence from typing instead, for simplicity

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+        pattern_nodes: typing.Sequence[oir.Node],
+        pattern_input_names: typing.Sequence[str],
```
justinchuby, line 305: @titaiwangms is migrating the pattern module to use the new IR. We may need some coordination here

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+        pattern_output_names: typing.Sequence[str],
+    ):
+        assert len(model_nodes) == len(pattern_nodes), (
```
justinchuby, line 308: Is this an error that can be caused by user code? If so I would raise a ValueError instead of asserting.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+        self.pattern_nodes = pattern_nodes
+        self.pattern_input_names = pattern_input_names
+        self.pattern_output_names = pattern_output_names
+        self.kwargs = {}
+
+        matched_pattern_to_model_name: dict[str, str] = {}
+        for gn, pn in zip(model_nodes, pattern_nodes):
+            assert (
+                gn.op_type == pn.op_type
+            ), f"Unexpected type mismatch {gn.op_type!r} != {pn.op_type!r}"
```
justinchuby, line 324: nit: Avoid abbreviations

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+            assert len(gn.input_names) == len(
+                pn.input_names
+            ), f"Unexpected number of inputs for type {gn.op_type}"
+            for a, b in zip(gn.input_names, pn.input_names):
```
justinchuby, line 328: nit: Avoid abbreviations

onnxscript/rewriter/generic_pattern.py
justinchuby, line 297: Is this a user facing class? Can it be private? Or can the whole module be private?

onnxscript/rewriter/generic_pattern.py
```diff
@@ -477,10 +549,19 @@ def match_pattern(
             f"Class {cls.__name__!r} must overwrite method match_pattern."
         )
 
-    @classmethod
     def _build_pattern(
-        cls, g: ModelWithGraphStructure, fct: typing.Callable
+        self,
+        g: ModelWithGraphStructure,
+        fct: typing.Callable | None = None,
```
justinchuby, line 555: Good time to rename to avoid abbreviations

onnxscript/rewriter/generic_pattern.py
```diff
@@ -1062,46 +1088,68 @@ class OnnxGenericPattern(GenericPattern):
     :param match_proto: the onnx function defining the matching pattern
     :param apply_proto: the onnx function defining the new pattern
     :param validate_mapping: the function used to validate a pattern
+    :param use_onnxscript: tells if the apply_proto is an onnxscript function or
```
justinchuby, line 1091: Prefer google style docstring. We can do a sweep later

onnxscript/rewriter/generic_pattern.py
```diff
@@ -1062,46 +1088,68 @@ class OnnxGenericPattern(GenericPattern):
...
-            raise AssertionError(
-                f"Function {fct} is not {self.match_pattern} or {self.apply_pattern}."
-            )
+            key = id(g), match, str(kwargs)
+            assert not callable(self.apply_proto) or isinstance(kwargs, dict)
+        if key in self._cache:
+            return self._cache[key]
+
+        if match:
+            onx = self.match_proto
```
justinchuby, line 1138: Avoid abbreviations

onnxscript/rewriter/generic_pattern.py
xadupre: Yes it is part of the public API, when the user writes the validation function.

onnxscript/rewriter/generic_pattern.py
xadupre: Class GenericPattern is defined after this line.

onnxscript/rewriter/generic_pattern.py
xadupre: I use typing everything and you always say not to do "from ... import ...". So I'm confused.

onnxscript/rewriter/generic_pattern.py
xadupre: No it cannot be created by the user.

onnxscript/rewriter/generic_pattern.py
justinchuby: If so I would be very careful in designing the fields and making sure we are able to evolve them

onnxscript/rewriter/generic_pattern.py
justinchuby: That's ok. `from __future__ import annotations` will quote this for us so we don't have to.

onnxscript/rewriter/generic_pattern.py
justinchuby: For the full text: https://google.github.io/styleguide/pyguide.html#22-imports

onnxscript/rewriter/generic_pattern.py
justinchuby: Or maybe call out in the assert string that this would be a bug?

onnxscript/rewriter/generic_pattern.py
titaiwangms: I am integrating as soon as I can. I can see some conflicts to fix..

onnxscript/_legacy_ir/protobuilder.py
```diff
@@ -108,6 +108,9 @@ def process_ir_node(self, ir_node: ir.Node, node_proto: onnx.NodeProto) -> onnx.
         for attr in ir_node.attributes.items():
             attr_proto = self.process_attribute(attr)
             node_proto.attribute.append(attr_proto)
+        if len(ir_node.attributes) == 0 and len(ir_node.original_node_proto.attribute) > 0:
+            # ir_node.attributes is never populated.
```
BowenBao, line 112: Should this be fixed in `irbuilder.py`? But I think node attrs should be covered there, otherwise unittests should have caught it. What is the particular case that the attrs weren't populated?

titaiwangms (approved):
Please make sure all the comments are addressed, and I will be integrating "multiple outputs" with new IR.

cc @gramalingam if there is anything to add.

onnxscript/rewriter/generic_pattern.py
gramalingam: This is good. Ideally, we should unify this and the other pattern-match-result [MatchBinding](https://github.com/microsoft/onnxscript/blob/e3bbfaa9e1ba91fa2e21e812fc378ce780babdac/onnxscript/rewriter/pattern.py#L278) to avoid confusion for users ... I am not saying we need to do this in this PR, but eventually. Since we have lot of concurrent work, it is ok to do things piece-meal for efficiency.

justinchuby: Do we have other structured representation too? Sorry I wasn't aware. I think we should not create new structures that duplicate current structures.

onnxscript/_legacy_ir/protobuilder.py
gramalingam: May be ok. Anyway, all of this will disappear after Titai's refactoring. The legacy IR is just a wrapper around a original NodeProto ... see Xavier's TODO comment in file above.

onnxscript/rewriter/generic_pattern.py
justinchuby: It also looks like MatchResult serves as an input to user code. PatternMatchResult is used to carry the rewritten results? In both cases I am still not sure why a user would need to initialize this object.

onnxscript/rewriter/generic_pattern.py
gramalingam: nit: if abbreviations are to be used, would be helpful to change `gn` to `mn` ... to match the first letter of `model_nodes` ... or change `model_nodes` to `graph_nodes` (which is where the `g` is coming from, I guess).

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+        for gn, pn in zip(model_nodes, pattern_nodes):
+            assert (
+                gn.op_type == pn.op_type
+            ), f"Unexpected type mismatch {gn.op_type!r} != {pn.op_type!r}"
+            assert len(gn.input_names) == len(
+                pn.input_names
+            ), f"Unexpected number of inputs for type {gn.op_type}"
+            for a, b in zip(gn.input_names, pn.input_names):
+                if b == "":
+                    # optional input or not an interesting input
```
gramalingam, line 330: This could end up being a problematic convention: users may want to check that an optional input has the "" value sometimes and sometimes they may want to ignore an uninteresting input. It is better to fix the meaning of "" in a pattern-node-input: I think it is better to treat it as a requirement that the matching graph-node-input must also be "". (The users can always use an extra input symbol for the dont-care situation.)

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+            f"{len(pattern_nodes)}."
+        )
+        self.pattern = pattern
+        self.model_nodes = model_nodes
+        self.pattern_nodes = pattern_nodes
+        self.pattern_input_names = pattern_input_names
+        self.pattern_output_names = pattern_output_names
+        self.kwargs = {}
+
+        matched_pattern_to_model_name: dict[str, str] = {}
```
gramalingam, line 320: The other matcher uses a map from (pattern) names to `ir.Value`. I think that may be preferable (it will avoid an extra unnecessary lookup to convert the name into a `ir.Value` when users want to use it).

onnxscript/rewriter/generic_pattern.py
```diff
@@ -294,6 +294,78 @@ def make_node(
...
+            assert len(gn.output_names) == len(
+                pn.output_names
+            ), f"Unexpected number of outputs for type {gn.op_type}"
+            for a, b in zip(gn.output_names, pn.output_names):
+                if b == "":
+                    # Only final outputs are interesting.
+                    continue
+                assert a != "", f"{a!r} cannot be optional"
+                if b in matched_pattern_to_model_name:
+                    assert matched_pattern_to_model_name[b] == a, (
```
gramalingam, line 349: So, these asserts mean that the pattern-matcher should have stopped earlier (and skipped the match) in the absence of bugs ... is that right?

gramalingam (approved):
My comments are non-blocking. I think we will need more refactoring to integrate the two pattern-matchers, but happy to do it incrementally.

onnxscript/_legacy_ir/protobuilder.py
BowenBao: Agree no need to update code here since refactor will replace this. I guess I'm more interested in the particular case in case it is an edge case that were not considered, since in general attributes should be populated by irbuilder ...

onnxscript/rewriter/generic_pattern.py
xadupre: What do you mean by 'able to evolve them'?

justinchuby: So if we add or remove fields in the initializer, older usages won't break.

onnxscript/rewriter/generic_pattern.py
xadupre: done

onnxscript/rewriter/generic_pattern.py
xadupre: I'll remove it. I prefer long error messages but that's your call.

onnxscript/rewriter/generic_pattern.py
xadupre: I'm still not sure about you mean. You obviously don't break anything by adding more fields. It is very likely to break if an attribute is removed.

onnxscript/rewriter/generic_pattern.py
xadupre: I'd rather create another PR to do that since this PR was started: https://github.com/microsoft/onnxscript/pull/1369.

onnxscript/rewriter/generic_pattern.py
xadupre: I'd rather create another PR to do that since this PR was started: https://github.com/microsoft/onnxscript/pull/1369.

onnxscript/rewriter/generic_pattern.py
xadupre: The replacement needs this dictionary. It is more efficient in my case. The pattern names never change, the model names do.

onnxscript/rewriter/generic_pattern.py
gramalingam: Created issue #1385

onnxscript/rewriter/generic_pattern.py
gramalingam: A repeated variable X (representing a value) does have a valid interpretation: for example, if I want to rewrite "X*A + X*B" into "X*(A+B)". The meaning is that all occurrences of X represent the same value. My question really is that this check will happen earlier (in the pattern-matching algorithm) ... so if we see a subgraph in the model that looks like "W*A + V*B", then the pattern-matcher will fail when the second-occurrence of X is matched against V (which is not same as W). I am saying this is not an error that should cause a runtime-exception, but it is a failed-match that will be skipped.

gramalingam: Created issue #1386

onnxscript/rewriter/generic_pattern.py
gramalingam: I don't understand. My point is not about pattern name and model names. The key used for the dictionary in both forms of MatchResult is the same (pattern name). But I am suggesting that the value returned by the dictionary can be an IR-value instead of an IR-value's name

gramalingam: Created issue #1387

PR #1369, opened 2024-04-11
[Rewriter] Integrate ONNX IR with rewriter pattern match

titaiwangms:


onnxscript/ir/_ir_utils_temp.py
```diff
@@ -0,0 +1,66 @@
...
+            attr_value = node.attributes.get(attr_name)
+            if attr_value is not None:
+                ir_value.const_value = attr_value.value  # type: ignore[union-attr]
+                break
+    return ir_value
+
+
+def get_numpy_from_ir_value(value: ir.Value) -> np.ndarray | None:
+    constant_value = value.const_value
+    if constant_value is not None:
```
justinchuby, line 46: All const value will have numpy() defined when it is not none.

onnxscript/ir/_ir_utils_temp.py
```diff
@@ -0,0 +1,66 @@
...
+        if isinstance(constant_value, serde.TensorProtoTensor):
```
justinchuby, line 47: So we don’t need this check

onnxscript/rewriter/pattern.py
```diff
@@ -1029,8 +1029,8 @@ def _apply_deltas(
             for old_node in deleted_nodes:
                 graph_or_function.remove(old_node)
 
-    for position, insert in sorted(to_insert.items()):
-        for v in insert:
+    for position, insert in sorted(to_insert.items(), reverse=True):
+        for v in reversed(insert):
             graph_or_function.insert_after(graph_or_function.nodes[position], v)
```
justinchuby, line 1034: Position may change. I would try getting the node at the position first, as anchors. Before doing any mutation. 

Insert after supports inserting multiple nodes at once so it may be possible to save the loop

PR #1370, opened 2024-04-11
[IR] Make `add_user` in Value private

justinchuby:
[IR] Make `add_user` in Value private and note that they are only used by the `Node` class.

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -605,7 +605,7 @@ class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
         # Add the node as a user of the inputs
         for i, input_value in enumerate(self._inputs):
             if input_value is not None:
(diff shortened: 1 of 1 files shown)


titaiwangms (approved):

onnxscript/ir/_core.py
```diff
@@ -951,13 +951,25 @@ def def_index(self) -> int | None:
...
+        """Add a user node.
+
+        This is an internal method. It should only be called by the Node class.
+        """
         self._users.add((user, index))
 
-    def remove_user(self, user: Node, index: int) -> None:
-        """Reduce a user node."""
+    def _remove_user(self, user: Node, index: int) -> None:
+        """Reduce a user node.
```
justinchuby, line 969:
```suggestion
        """Remove a node from the users of this value.
```

PR #1371, opened 2024-04-11
[IR] Assign default names to nodes and values

justinchuby:
Assign default names to nodes and values if they do not have one. 

The class does not keep track of the names it has given, so it is possible to
generate names that conflicts with existing names. It is the responsibility of the
user to ensure that the names are unique (typically by running a name-fixing pass
on the graph).

### DIFF
--- a/onnxscript/ir/_name_authority.py
+++ b/onnxscript/ir/_name_authority.py
@@ -1,0 +1,30 @@
+"""Auxiliary class for managing names in the IR."""
+
+from __future__ import annotations
+
+from onnxscript.ir import _core
+
+
+class NameAuthority:
+    """Class for giving names to values and nodes in the IR.
+
+    The names are generated in the format ``val_{value_counter}`` for values and
+    ``node_{op_type}_{node_counter}`` for nodes. The counter is incremented each time
+    a new value or node is named.
+
+    The class does not keep track of the names it has given, so it is possible to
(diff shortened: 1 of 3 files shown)


onnxscript/ir/_name_authority.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def __init__(self):
+        self._value_counter = 0
+        self._node_counter = 0
+
+    def name_value(self, value: _core.Value) -> None:
+        value.name = f"val_{self._value_counter}"
+        self._value_counter += 1
+
+    def name_node(self, node: _core.Node):
```
justinchuby, line 29:
```suggestion
    def name_node(self, node: _core.Node) -> None:
```

BowenBao (approved):
Would be nice to have a linter checking if there are naming violations or name overrides inside subgraphs. Not sure if checker covers all of that.

justinchuby:
> Would be nice to have a linter checking if there are naming violations or name overrides inside subgraphs. Not sure if checker covers all of that.

Good idea. I can create one

titaiwangms:
Should this PR fixes the unnamed Node output names? I am still getting None if I don't set it manually.

justinchuby:
> Should this PR fixes the unnamed Node output names? I am still getting None if I don't set it manually.

Yes it should

PR #1377, opened 2024-04-13
[CI] Fix test path to run tests/ in CI

justinchuby:
Run tests under the `tests/` directory. Previously we migrated `tests/` from `onnxscript/tests` but I did not realize nox was not properly configured to discover them. cc @BowenBao 

1. Fix a serialization / deserialization issue with shape/type in the IR
2. Fix an import error in test

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -59,8 +59,7 @@ def test(session):
     )
     session.install(".", "--no-deps")
     session.run("pip", "list")
-    session.run("pytest", "onnxscript", *session.posargs)
-    session.run("pytest", "docs/test", *session.posargs)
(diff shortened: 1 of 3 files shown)


tests/optimizer/test_models.py
```diff
@@ -40,27 +40,27 @@ def test_model_runs_and_matches_accuracy_after_optimization(self, model_name):
         )
 
         with tempfile.TemporaryDirectory() as tmp_folder:
```
justinchuby, line 42:
```suggestion
        with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmp_folder:
```

titaiwangms (approved):

PR #1395, opened 2024-04-17
Integrate new IR with rewriter

justinchuby:
Continuation of https://github.com/microsoft/onnxscript/pull/1369.

- Change pattern and generic_pattern to use the new IR
- Use ir.Tensor to specify constant tensor patterns
- Remove legacy IR builders and protobuilder

NOTE: In `pattern.py`, there are some ignored MYPY rules that we should gradually enable when we are migrating `generic_pattern.py` and `pattern.py`.

Author @titaiwangms; co-authored by @justinchuby


titaiwangms:
@justinchuby Py38 is failing with unsupported annotation. It looks like

```python
```

is not saving us in this case.

justinchuby:
Fixed: Changed import from collections to typing to fit py38

onnxscript/ir/_enums.py
```diff
@@ -82,6 +82,18 @@ def numpy(self) -> np.dtype:
 
         return onnx.helper.tensor_dtype_to_np_dtype(self)
 
+    @classmethod
+    def from_numpy(cls, dtype: np.dtype) -> DataType:
```
justinchuby, line 86: Prob bad idea?

titaiwangms: Because of adding dependency on onnx.helper?

examples/pattern_rewriting.py
```diff
@@ -60,7 +60,7 @@ def get_rotary_model(bad_model=False):
 
 
 model = get_rotary_model()
-ir_model = oir.irbuilder.build_ir(model)
+ir_model = ir.serde.deserialize_model(model)
```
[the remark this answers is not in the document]
gramalingam: More a naming question really ... whether an alternative name might be better than serialize/deserialize.

onnxscript/rewriter/__init__.py
```diff
@@ -27,18 +25,17 @@ def rewrite(
     function_rewrite_rules: Sequence[type[FunctionRewriteRule]] = (),
     pattern_rewrite_rules: Sequence[PatternRewriteRule] = (),
 ) -> onnx.ModelProto:
+    print(f"len(value_info): {len(model.graph.value_info)}")
+    model_ir = ir.serde.deserialize_model(model)
```
gramalingam, line 29: (Non-blocking comment: ok to create an issue to track this.) Maybe useful to support another scenario: in the context of the optimizer, we don't want to keep converting back and forth between IR and Proto in intermediate states ... we should do the conversion only once, in the beginning and end.

Supporting inputs which are both ModelProto and IR is easy and worth doing (just as a convenience for user really ... allowing IR input is important, allowing ModelProto input is convenience.)

For the output, however, we may need to make a choice. Either return IR always (reasonable first choice) or something more complicated (just as a convenience).

onnxscript/ir/_enums.py
justinchuby: Right. But it’s ok I guess?

examples/pattern_rewriting.py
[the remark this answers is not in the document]
justinchuby: Sure - do you have suggestions? Convert? Just thinking a name that is not overloaded would be preferable

onnxscript/rewriter/__init__.py
justinchuby: https://github.com/microsoft/onnxscript/issues/1401

onnxscript/rewriter/broadcast_to_matmul.py
```diff
@@ -31,7 +31,9 @@ def check_if_need_reshape(match_bindings: dict[str, ir.Value | Any]) -> bool:
     """
     input_a_shape = match_bindings["input_a"].shape
     input_b_shape = match_bindings["input_b"].shape
-    shape_c = match_bindings["shape_c"].value_as_np_array
+    # TODO: Get a helper func to get const_value
+    shape_c_value = _ir_utils.propagate_const_value(match_bindings["shape_c"])
```
gramalingam, line 35: Trying to understand this ... I think I would push this into the rewriter engine, instead of burdening the rule-writer. Ideally, we would have a separate pass to compute constant-values (eg., as done by the optimizer). Then, whenever the rewriter modifies the graph, it can invoke computation of constant values on newly inserted nodes? For now, we can probably do the second step? (first is a separate issue, anyway, and requires migration of optimizer to the new IR).

onnxscript/rewriter/pattern.py
```diff
@@ -23,21 +23,96 @@
 # appropriate).
 
 
-class ConstantPattern:
-    def __init__(self, value: int | str | list) -> None:
+class PythonPattern:
```
gramalingam, line 26: nit: Why `PythonPattern`? Not that it matters, since this is outside the scope of this PR. We will need to clean the pattern IR next, separately.

onnxscript/rewriter/pattern.py
```diff
@@ -927,57 +1023,38 @@ def _apply_deltas(
...
+            for old_node in deleted_nodes:
+                graph_or_function.remove(old_node)
+
+    for replaced_node, inserted_nodes in to_insert:
+        graph_or_function.insert_after(replaced_node, inserted_nodes)
+        # TODO: improve this
+        # This is updating the graph/function outputs to use the new outputs
+        for inserted_node in inserted_nodes:
+            for new_output in inserted_node.outputs:
+                if (index := new_output.meta.get(_ir_utils.GRAPH_OUTPUT_META_KEY)) is not None:
```
gramalingam, line 1053: What is this for? Feeling a bit lost here.

onnxscript/rewriter/pattern.py
```diff
@@ -927,57 +1023,38 @@ def _apply_deltas(
...
-            position_to_delete.append(i)
-
-    for p in reversed(position_to_delete):
-        del nodes[p]
+            # Reconnect the users of the deleted node to use the new outputs
+            for last_deleted_output, last_inserted_output in zip(
+                last_deleted.outputs, last_inserted.outputs
+            ):
+                for node, index in last_deleted_output.consumers():
+                    node.replace_input_with(index, last_inserted_output)
```
gramalingam, line 1032: Great if this works. Just trying to understand what the IR allows us to do: at this point `last_inserted_output` is computed by a node that has not been inserted into the graph yet (I guess that happens below). Is that safe/permitted?

gramalingam: Or, is it better to insert the nodes first, before updating the uses?

justinchuby: Yes! The value-node connections are independent of the graph. This means we will need to maintain connections for them to be valid, but validity is easy to check

onnxscript/rewriter/pattern.py
```diff
@@ -879,7 +977,8 @@ def replace_pattern(new_pattern):
 
 def _apply_deltas(
     graph_or_function: ir.Graph | ir.Function,
-    deltas: list[tuple[int, tuple[list[ir.Node], list[ir.Node]]]],
+    # TODO(jutinchuby): Use a more descriptive data structure to store deltas
+    deltas: Sequence[tuple[int, tuple[Sequence[ir.Node], Sequence[ir.Node]]]],
```
gramalingam, line 981: I think we should also be simplifying this function (several aspects). But, as this is a complicated enough PR, I am in favor of not adding more restructuring yet in this PR ...

onnxscript/rewriter/pattern.py
```diff
@@ -23,21 +23,96 @@
...
-    def matches(self, value: int | str | list) -> bool:
+    @property
+    def name(self) -> str | None:
+        return self._name
+
+    def matches(self, value: int | str | Sequence) -> bool:
         return value == self.value
 
-    def to_ir(self, model, bindings=None) -> int | str | list:
+    def to_ir(self, model, bindings=None) -> int | str | Sequence:
```
gramalingam, line 42: Hmmm ... I hadn't thought of the `to_ir` methods (being impacted here). I have been thinking we should probably get rid of these (which exists only as a sort of a poor-substitute for a trace-mode onnxscript integrated into the pattern-IR). Anyway, that is for later, not an issue here.

onnxscript/rewriter/pattern.py
```diff
@@ -193,13 +293,17 @@ def matches(self, opset):
 
     def to_ir(self, model, bindings=None) -> str:
         domain = self.domain_pattern.to_ir(model, bindings)
+        assert isinstance(domain, str), f"Expected str, got {type(domain)}"
         # TODO: Should we ban other custom domains?
-        if domain not in model.version_map:
-            model.version_map[self.domain_pattern.value] = self.version_pattern.value
+        if domain not in model.opset_imports:
+            assert not isinstance(
```
gramalingam, line 299: Wonder if it is better to assert what instance it is expected to be?

onnxscript/rewriter/pattern.py
```diff
@@ -193,13 +293,17 @@ def matches(self, opset):
...
+                self.version_pattern, AnyPattern
+            ), f"Cannot match against any version of custom domain {domain}"
```
gramalingam, line 301: The error is in converting it to IR (not in matching).

onnxscript/rewriter/pattern.py
justinchuby: > Or, is it better to insert the nodes first, before updating the consumers?

I think they are equivalent? Is there any pros to requiring this?

onnxscript/rewriter/pattern.py
justinchuby: @titaiwangms

onnxscript/rewriter/pattern.py
justinchuby: Here we had the issue of we don't know which matched pattern node should be an output to the graph with how the nodes are passed in to this function right now. So we use a meta field to store the info and update the graph output as needed.

Alternatively, we can pass a mapping of {old_value: new_value} to simply call `old_value.is_graph_output()` when updating the connections.

onnxscript/rewriter/pattern.py
justinchuby: @titaiwangms

onnxscript/rewriter/pattern.py
justinchuby: @titaiwangms

onnxscript/rewriter/pattern.py
titaiwangms: Referring to python type: int/str/float, etc. And these are not going to be serialized.

examples/pattern_rewriting.py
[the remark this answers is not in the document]
justinchuby: We can create a class that defines methods convert_model_from_proto/convert_model_to_proto etc.

onnxscript/rewriter/broadcast_to_matmul.py
titaiwangms: The utils under _ir_utils are all temporary hack, and wait for the implementation in IR. I think this will be removed once we have propagation in IR.

onnxscript/rewriter/pattern.py
titaiwangms: I wonder if in the next PR (combine generic_pattern.py and pattern.py) we would also like to merge function rewriter. function rewriter is currently using protobuf.

onnxscript/rewriter/pattern.py
titaiwangms: Curious about how we could remove this as to_ir converts Pattern to IR?

gramalingam: I created [issue 1403](https://github.com/microsoft/onnxscript/issues/1403) to explain things a bit more. Currently, we use the same `op` object to build the source pattern, as well as to specify the replacement pattern. That has one advantage, namely that the user doesn't have to bother about the distinction. But, in reality, we are doing different things when we are specifying a source-pattern and when we are building the replacement sub-graph. We can use different "builder" objects for the two, and that would avoid the need to couple both of these.

examples/pattern_rewriting.py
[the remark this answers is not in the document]
gramalingam: Yes, `convert_X_to_Y` is reasonable for specific X. If we want a flexible version (allowing multiple source types), something like `convert_to_Y` or just `to_Y` is also fine. 

Re. the naming convention (I saw the issue you created for this): using a vanilla `model` has some disadvantages (especially since a lot of pre-existing code will contain references to vanilla `model` meaning a model in proto format or some other representation). Creating a more specific name (like `model_ir`) is also useful, as it can be used unambiguously where desired.

For the convert methods, using a name like `convert_proto_to_ir` or `convert_model_proto_to_ir` might be preferable to `convert_proto_to_model` (at least, that's my personal take).

onnxscript/rewriter/pattern.py
gramalingam: I see. Ok, I guess we can deal with this in the next phase/PR for simplification (of the API for this function).

onnxscript/rewriter/generic_pattern.py
```diff
@@ -553,33 +297,36 @@ def none(
...
+    ) -> ir.Graph:
         del match
-        assert fct, f"Not implemented if fct is None in class {self.__class__.__name__}"
-        assert not kwargs, (
-            f"Not implemented when kwargs is not empty but {kwargs} "
-            f"in class {self.__class__.__name__}"
-        )
+        if func is None:
+            raise NotImplementedError(
+                f"Not implemented if fct is None in class {self.__class__.__name__}"
```
gramalingam, line 318: nit: fct vs. func

onnxscript/rewriter/pattern.py
titaiwangms: Done

onnxscript/rewriter/pattern.py
titaiwangms: Done

onnxscript/rewriter/broadcast_to_matmul.py
gramalingam: https://github.com/microsoft/onnxscript/issues/1404

onnxscript/rewriter/generic_pattern.py
```diff
@@ -1093,136 +844,73 @@ def make_rule(self) -> orp.RewriteRule:
...
-    :param use_onnxscript: tells if the apply_proto is an onnxscript function or
-        a regular function returning a FunctionProto
-    :param opsets: opset to consider when converting the function into ONNX,
-        if not specified, it is opset 18 for the main opset, and opset 1
-        for domain com.microsoft.
-    :param verbose: in [0, 10], increase the verbosity to understand why a pattern
-        does not match
-    """
-
+class FunctionPattern(GenericPattern):
```
gramalingam, line 847: What's this for? What's the difference between FunctionPattern and GenericPattern?

gramalingam (approved):

onnxscript/rewriter/pattern.py
gramalingam: I would recommend doing those two separately, they are distinct.

onnxscript/rewriter/pattern.py
gramalingam: Well, in one sense, it is a well-formed graph when we insert new nodes that compute new values. And then if we replace old-value with new-value in some node that consumers old-value. But if we replace old-value with new-value, where new-value is not yet a valid value in the graph, we have a temporarily invalid graph. That's why I was asking (since it could make a difference in some implementations, but I guess it does not in the current one.)

onnxscript/rewriter/generic_pattern.py
justinchuby: @titaiwangms we can use func everywhere

onnxscript/rewriter/generic_pattern.py
titaiwangms: It was `OnnxGenericPattern`. It's a feature supporting Callable and FunctionProto to become Pattern. I will add the description.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -728,126 +459,128 @@ def _match_backward(
...
+            matched: nodes of the pattern matched as already matched
+            stack: next node to look into
+            graph_node: node coming from the graph
+            pattern_node: node coming from the pattern
+
+        Returns:
+            number of matched nodes to continue, None or False to indicate a failed match
         """
-        res = 0
+        match_countch_count = 0
```
gramalingam, line 493: change to `match_count` (globally)

titaiwangms: Done

onnxscript/rewriter/generic_pattern.py
titaiwangms: Done

onnxscript/rewriter/generic_pattern.py
titaiwangms: Done

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -450,138 +422,121 @@ def validate_rotary_mapping(g, match_result) -> bool:
...
         # back to ir
-        ir_model = oir.irbuilder.build_ir(model)
+        ir_model = ir.serde.deserialize_model(model)
 
         # starts matching
-        rule = org.make_pattern_rule(
+        rule = generic_pattern.make_pattern_rule(
             transpose_transpose_pattern,
-            transpose_transpose_apply_pattern,
+            transpose_transpose_apply_pattern(perm=[2, 0, 1]),
```
xadupre, line 529: The test does not test the same thing anymore. By defining the permutation here, it cannot be guessed from the matched nodes.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -1093,136 +844,86 @@ def make_rule(self) -> orp.RewriteRule:
...
-
-    def validate_mapping(self, g: oir.Model, match_result: PatternMatchResult) -> bool:
-        """Evaluates the consistency of the replacements."""
-        return self._validate_mapping(g, match_result)
+        self.match_pattern = match_pattern
+        self.apply_pattern = apply_pattern
+        self.validate_mapping = validate_mapping
+        self.verbose = verbose
 
-    def _build_pattern(
```
xadupre, line 1134: This function should not have been removed. It creates the replacements nodes based on the parameters obtained after the matching.

onnxscript/rewriter/generic_pattern_test.py
justinchuby: Good call, thanks. Building patterns based on match results is support by pattern.py, could you see if that fulfills your need (as described by Rama in a separate PR)? Since the feature has been integrated to pattern.py, we removed it from generic_pattern

PR #1398, opened 2024-04-17
[IR] Improve APIs 3/n

justinchuby:
- Use dictionary keys to store consumers of a Value to preserve insertion order so that the visiting order is deterministic
- Update the `is_graph_output` method to make it more robust
- Use UserDict to simplify MetadataStore
- Reenforce that `Function` and `Graph` are `Sequence[Node]` by removing the `node` attribute, since accessing `node` is `O(# nodes)` which can be expensive. Uses should treat a `Graph` as a sequence of nodes.
- Create tests for Tensor and fix its dlpack implementation


onnxscript/ir/_core.py
```diff
@@ -419,6 +420,51 @@ def __ge__(self, other: object) -> bool:
             raise TypeError(f"Expected other to be Dimension or int, got {type(other)}")
         return int(self) >= int(other)
 
+    def __neg__(self) -> int:
```
justinchuby, line 423: Needs tests

onnxscript/ir/_core.py
```diff
@@ -436,6 +437,51 @@ def __ge__(self, other: object) -> bool:
...
+        return -int(self)
+
+    def __add__(self, other):
```
gramalingam, line 443: My intuition is that this is not worth adding. It is likely to be misused. Logically, what is useful (eg., in shape-inference context) is something that handles the symbolic case also ... even if it is to just produce an "unknown" Dimension value (represented as None). Eg.: something like [this](https://github.com/onnx/onnx/blob/c459890aa266d74ca612f334d0f3e869dcdbb597/onnx/defs/shape_inference.h#L175) for multiplication.

justinchuby: In building the graph for torchlib, I have seen traced only functions doing `shape[0] // 2`, which prompted this update

justinchuby: We can also return dim.value when a user is indexing on shape; and provide another api for symbolic dimensions

justinchuby: But if we use `.dims` the getter `shape.dims[...]` will not be symmetrical with the setter `shape[...]`

gramalingam: (i) We can't assume self is an int here. If it is not, there is a way to return a valid Dimension.

(ii) Wrt to particular use-case you mention: if a traceonly function is doing `shape[0] // 2`, then what do we want to do? The general solution is to emit ONNX nodes/ops that represent this computation (with a statically known value being a special case). 

Basically, when `shape` itself is computed, it should NOT be an instance of `ir.Shape`. It should be an instance of `ir.Value` ... and `shape[0]` should also be an instance of `ir.Value` (not an `ir.Dimension`). Or, am I mistaken?

justinchuby: We can flatten the Shape class a little. Use int for known dimensions, and use Dim(str|None) for symbolic dimensions (this way two strings are not accidentally compared).

We can then store the denotations in the Shape class.

onnxscript/ir/_protocols.py
```diff
@@ -14,8 +14,8 @@
 
 import typing
 from typing import (
-    AbstractSet,
```
gramalingam, line 17: I thought that you were planning to eliminate the protocols?

onnxscript/ir/_core.py
justinchuby: So the use case above is for static evaluation; we didn't want to emit dynamic code in that particular instance of the trace function. The result of `shape[0] // 2` was used as a constant

justinchuby: When a user wishes to have a dynamic shape, I would expect they do `op.Shape(<ir.Value>)`; whereas `ir.Value.shape` returns static shape information that is part of its type.

onnxscript/ir/_protocols.py
justinchuby: No I wasn't planning to eliminate the protocols. In the last comment I made I decided to not separate the read-only protocols from the mutable protocols because if we do so they will have to be used together and conflict.

Here all protocols define mutation methods so we are good.

onnxscript/ir/_core.py
justinchuby: I have reverted this part of the change and will work on it in a separate PR.

justinchuby: https://github.com/microsoft/onnxscript/pull/1399

onnxscript/ir/_core.py
```diff
@@ -496,8 +497,8 @@ def __len__(self) -> int:
     def __iter__(self) -> Iterator[Dimension]:
         return iter(self._dims)
 
-    def __getitem__(self, index: int) -> Dimension:
-        return self._dims[index]
+    def __getitem__(self, index: int) -> int:
+        return self.numpy()[index]
```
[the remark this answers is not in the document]
justinchuby: Sorry you are right. This should be reverted. The other PR is ready for review. I will revert and rebase this PR on https://github.com/microsoft/onnxscript/pull/1399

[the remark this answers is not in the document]
justinchuby: Updated

gramalingam (approved):

onnxscript/ir/_protocols.py
```diff
@@ -236,29 +241,32 @@ class GraphProtocol(Protocol):
     allows different subgraphs to import different opsets. It is the responsibility
     of the deserializer to reconcile the different opsets.
 
-    The :attr:`nodes` are not guaranteed to be topologically sorted. But the
+    The `nodes` are not guaranteed to be topologically sorted. But the
```
justinchuby, line 244:
```suggestion
    The nodes are not guaranteed to be topologically sorted. But the
```

PR #1399, opened 2024-04-17
[IR] Improve shape handling

justinchuby:
- In Shape, store integers as plain python integers; rename Dimension to SymbolicDim to hold {str, None} or any other symbols. This way `shape[...]`, when it is an int, can be used freely for computation and we don't need to worry about arithmetic operations on Dimension.
- Move denotation from Dimension to Shape to accommodate the change.
- Shape can be set to frozen when it is initialized from a tensor
- Implement slicing on shapes


onnxscript/ir/_core.py
```diff
@@ -175,14 +175,19 @@ def __init__(
...
-                "Please specify the shape explicitly."
-            )
+        if not hasattr(value, "shape"):
+            if shape is None:
+                raise ValueError(
+                    f"Expected an object with a shape attribute, but {type(value)} does not have shape. "
+                    "Please specify the shape explicitly."
+                )
+            else:
+                self._shape = shape
```
justinchuby, line 185:
```suggestion
            self._shape = shape
```

gramalingam:
I am not sure I understand the motivation for this change. The existing version has the advantage of being aligned with the ONNX proto definition ... so, it is easy to use for anybody familiar with ONNX proto. There should be a good reason/advantage to altering it (and losing this advantage). I don't see it yet.

onnxscript/ir/_core.py
```diff
@@ -389,89 +394,78 @@ def metadata_props(self) -> dict[str, str]:
...
-        for dim in dims:
-            if dim is not None and not isinstance(dim, (int, str, Dimension)):
-                raise TypeError(f"Expected int, str, None or Dimension, got '{type(dim)}'")
-        self._dims: list[Dimension] = [
-            dim if isinstance(dim, Dimension) else Dimension(dim) for dim in dims
+    __slots__ = ("_dims", "_frozen")
+
+    def __init__(
+        self,
+        dims: Iterable[int | SymbolicDim | Any],
```
gramalingam, line 434: Why is `Any` here?

justinchuby: Good catch - will remove

gramalingam: Is it supposed to be `str | None` ?

justinchuby: Actually it is because: When a ``dim`` is not an integer or a SymbolicDim, it is converted to a SymbolicDim.

justinchuby: This allows users to create Shape(["a", "b"])

onnxscript/ir/_core.py
```diff
@@ -482,36 +476,57 @@ def rank(self) -> int:
...
-    def __setitem__(
-        self, index: int, value: _protocols.DimensionProtocol | int | str | None
-    ) -> None:
-        if isinstance(value, Dimension):
-            self._dims[index] = value
-            return
-        if isinstance(value, (int, str, type(None))):
-            self._dims[index] = Dimension(value)
-            return
+    def __setitem__(self, index: int, value: int | SymbolicDim) -> None:
```
gramalingam, line 493: Should it support `| str | None` to be consistent with constructor (and old version)?

gramalingam (approved):
Minor comments above

onnxscript/ir/_core.py
justinchuby: Yes `str | None` is better. I was one step ahead and thinking about sympy objects. But we don't care about them now.

justinchuby:
> I am not sure I understand the motivation for this change. The existing version has the advantage of being aligned with the ONNX proto definition ... so, it is easy to use for anybody familiar with ONNX proto. There should be a good reason/advantage to altering it (and losing this advantage). I don't see it yet.

Recapping our discussion for future reference, this allows ONNX Script and torchlib etc. to subclass value and present the `shape` interface with familiar behavior (compared with numpy, pytorch). Users can check the type of a particular dimension {int, SymbolicDim} to know if it is static, during graph analysis.

onnxscript/ir/_core.py
justinchuby: Done

onnxscript/ir/_core.py
justinchuby: Done

justinchuby:
CausalLM
image

onnxscript/ir/_core.py
```diff
@@ -543,36 +534,65 @@ def rank(self) -> int:
...
-            self._dims[index] = Dimension(value)
-            return
+    @typing.overload
+    def __getitem__(self, index: slice) -> tuple[int | SymbolicDim, ...]: ...
 
-        raise TypeError(
-            f"Value must be int, str, None or DimensionProtocol. Got '{type(value)}'"
-        )
+    def __getitem__(self, index):
+        return tuple(self._dims)[index]
```
gramalingam, line 555: Non-blocking question: why is a conversion to tuple needed? I assume tuple and list behave differently for some input of interest here ... I guess the returned value will be a tuple when slicing, and a list otherwise ... but should that matter? Or, is there some other difference?

PR #1405, opened 2024-04-19
[IR] Avoid creating empty dictionaries during deserialization

justinchuby:
Avoid creating empty dictionaries for metadata_props during deserialization.

When deserializing a 350k-node model, not creating the dictionary by default should save us 84mb of memory just on the nodes, according to https://lerner.co.il/2019/05/12/python-dicts-and-memory-usage/.

350k * 240 bytes = 84 mb

### DIFF
--- a/onnxscript/ir/serde.py
+++ b/onnxscript/ir/serde.py
@@ -483,7 +483,9 @@ def deserialize_value_info_proto(
         value.name = proto.name
     value.shape = deserialize_type_proto_for_shape(proto.type)
     value.type = deserialize_type_proto_for_type(proto.type)
-    value.metadata_props.update(deserialize_metadata_props(proto.metadata_props))
+    metadata_props = deserialize_metadata_props(proto.metadata_props)
+    if metadata_props is not None:
+        value.metadata_props.update(metadata_props)
     return value
 
 
--- a/onnxscript/ir/serde.py
+++ b/onnxscript/ir/serde.py
@@ -87,7 +87,9 @@ class TensorProtoTensor(_core.TensorBase):
 
     def __init__(self, proto: onnx.TensorProto) -> None:
         self._proto = proto
-        self._metadata_props: dict[str, str] = deserialize_metadata_props(proto.metadata_props)
+        self._metadata_props: dict[str, str] | None = deserialize_metadata_props(
+            proto.metadata_props
+        )
         self._metadata: _metadata.MetadataStore | None = None
 
     @property
@@ -142,6 +144,8 @@ class TensorProtoTensor(_core.TensorBase):
 
     @property
     def metadata_props(self) -> dict[str, str]:
+        if self._metadata_props is None:
+            self._metadata_props = {}
         return self._metadata_props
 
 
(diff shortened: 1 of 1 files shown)


onnxscript/ir/serde.py
```diff
@@ -617,7 +619,9 @@ def deserialize_tensor(
     )
 
 
-def deserialize_metadata_props(proto: Sequence[onnx.StringStringEntryProto]) -> dict[str, str]:
+def deserialize_metadata_props(proto: Sequence[onnx.StringStringEntryProto]) -> dict[str, str] | None:
+    if len(proto) == 0:
+        return None
```
justinchuby, line 624:
```suggestion
        # Avoid creating an empty dictionary to save memory
        return None
```

onnxscript/ir/serde.py
```diff
@@ -483,7 +483,9 @@ def deserialize_value_info_proto(
         value.name = proto.name
     value.shape = deserialize_type_proto_for_shape(proto.type)
     value.type = deserialize_type_proto_for_type(proto.type)
-    value.metadata_props.update(deserialize_metadata_props(proto.metadata_props))
+    metadata_props = deserialize_metadata_props(proto.metadata_props)
+    if metadata_props is not None:
+        value.metadata_props.update(metadata_props)
```
gramalingam, line 488: nit: Doesn't this involve a copy? Why wouldn't we deserialize the props in-place?

justinchuby: It does, but since I expect this to be rare I think it is ok. It is done for two reasons

1. value.metadata_props is a read-only property, meaning we cannot assign a new dictionary to it (so we need .update)
2. I could write a for loop to update the dictionary and avoid creating a copy first, but reusing `deserialize_metadata_props` and not having to duplicate the logic seems to be a good idea

gramalingam: I meant make deserialize_metadata_props take the target dictionary as a parameter ... just the way (I believe) deserialization of a parent passes the child as a parameter to the child's deserialization method. Granted, you might want to move the length=0 optimization up to the caller to do this. 

350K nodes is rare (but is the motivation here). There could be a similarly rare model with 350K entries inside the dictionary. (Actually, if a model decides to store node-specific info in this field, like layer-info for each node, this model would have 350K entries in the dictionary.)

justinchuby: That makes sense, thanks. I will send a patch

justinchuby: I made an attempt, but realized having the caller to do the length check doesn't seem clean, since there a several callers. I think having to the dictionary is acceptable, because the act of copying is not slow, and we just add 84mb to the peak memory (which is gc'ed right after). That seems acceptable to trade with the added complexity on the caller site.

justinchuby:
```
import pyinstrument
d = {i:i for i in range(350000)}
p=pyinstrument.Profiler()
p.start()
{}.update(d)
p.stop()

p.print()
```

```
0.009 <module>  test2.py:1
├─ 0.007 dict.update  <built-in>
└─ 0.001 [self]  test2.py
```

PR #1409, opened 2024-04-19
[WIP] Use IR in onnxscript

justinchuby:


justinchuby:
Should stmt subclass node or own a node?

gramalingam:
I think this is outdated and can be closed, right @justinchuby ?

justinchuby:
I will come back to this after exporter. I think some of the code is still relevant

PR #1411, opened 2024-04-19
Refactor pattern matcher

gramalingam:
Part 1 of refactoring the pattern matcher (towards a unified API and implementation of existing two pattern matchers).
* Eliminate separate "Var" class for patterns. (Every value-pattern will have an optional name, used to bind to corresponding ir-value in a match.)
* Eliminate the delay-run (vs. eager-run) distinction. Everything is used in delayed-run mode now.
* Switch all replacement-builders to use context-parameter to directly build IR nodes for replacement.
* Cleanup some rewrite rules
* Get rid of the to_ir methods


onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -12,58 +13,34 @@
...
-def cast_constant_of_shape_without_value(
-    shape: Sequence[int],
-    dtype: int,
-    match_bindings: dict[str, ir.Value | Any] | None = None,
-) -> pattern.OpPattern:
-    del match_bindings  # Unused
+def fused_cast_constant_of_shape(op, shape: ir.Value, scalar: ir.Attr, dtype: ir.Attr, **_):
+    # Cast scalar (a TensorProto attribute) to the specified dtype
+    scalar_value = scalar.value.numpy().item()
+    cast_value = onnx.helper.make_tensor("value", dtype.value, (), [scalar_value])
```
justinchuby, line 24: Use ir.Tensor?

justinchuby: In fact `scalar.value` should be a Tensor already?

onnxscript/rewriter/no_op.py
```diff
@@ -24,7 +24,7 @@ def div_by_1(x):
 
 
 # Replacement
-def identity(x):
+def identity(op, x):
```
justinchuby, line 27: Should we add type?

onnxscript/tests/export/fun_with_attr_param.py
justinchuby, line 1: Tests have been moved to /tests: https://github.com/microsoft/onnxscript/tree/rama/refactor-pattern/tests

onnxscript/tests/export/fun_with_attr_param.py
```diff
@@ -0,0 +1,15 @@
+import numpy
+from onnx import TensorProto
+from onnx.helper import make_tensor
+from onnxscript import script, external_tensor
```
justinchuby, line 4:
```suggestion
import onnxscript
```

onnxscript/tests/export/fun_with_attr_param.py
```diff
@@ -0,0 +1,15 @@
+import numpy
+from onnx import TensorProto
+from onnx.helper import make_tensor
```
justinchuby, line 3:
```suggestion
import onnx.helper
```

nit: modules only

onnxscript/tests/export/fun_with_attr_param.py
justinchuby: Was this file auto generated?

onnxscript/rewriter/cast_constant_of_shape.py
gramalingam: I haven't tried your latest changes (today), but ir.Tensor was failing because convert_attributes was expecting a TensorProto (if I remember correctly)

justinchuby: I see. It should be fixed now.

gramalingam: > In fact `scalar.value` should be a Tensor already?

Yes, but we want to change/cast its type.

onnxscript/rewriter/no_op.py
gramalingam: Once things settle down (and there is agreement on the type, which is a newly added one ... eg., should it be a `RewriterContext` as proposed or an `OpBuilder` ... I think for long run, context is better if we want to add more info in the future. Should the object still be called `op` (more aligned with the onnxscript usage, and existing use)?

onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -12,58 +13,34 @@
...
-    return op.ConstantOfShape(shape, value=casted_val)
-
-
-def cast_constant_of_shape_without_value(
-    shape: Sequence[int],
-    dtype: int,
-    match_bindings: dict[str, ir.Value | Any] | None = None,
-) -> pattern.OpPattern:
-    del match_bindings  # Unused
+def fused_cast_constant_of_shape(op, shape: ir.Value, scalar: ir.Attr, dtype: ir.Attr, **_):
```
gramalingam, line 21: TODO: Existing rules (here and elsewhere) don't handle the possibility of AttrRefs ... when we apply a rule to a function graph.

gramalingam:
Why did tests/export get in here, even though it is in .gitignore? Also, how can I get rid of this folder, `git rm` does not seem to recognize it (maybe because of .gitignore, but I even deleted that line from .gitignore with no effect).

gramalingam:
> Not very sure. I would try delete the files, force add the deletion to git, and they should not get picked up again?

Hmmm ... after changing .gitignore (locally), I was able to add/update these files, commit and push them into this PR ... I then deleted and pushed the delete into the PR ... however, this weirdly seems to revert the files back to their original form ... instead of removing them.

justinchuby:
I did it with the github UI

onnxscript/rewriter/no_op.py
justinchuby: That's a good question. I am happy to explore different options later

onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -12,58 +11,34 @@
...
+    # Cast scalar (a TensorProto attribute) to the specified dtype
+    scalar_value = scalar.value.numpy().item()
+    cast_value = onnx.helper.make_tensor("value", dtype.value, (), [scalar_value])
```
justinchuby, line 22:
```suggestion
    cast_value = ir.Tensor(scalar.value.numpy().astype(ir.DataType(dtype.value).numpy()))
```

I think we can do this

onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -12,58 +11,34 @@
...
-def fused_cast_constant_of_shape_without_value(
-    shape: Sequence[int], dtype: int, match_bindings: dict[str, ir.Value | Any]
-) -> pattern.OpPattern:
-    del dtype  # Unused
-    v_dtype = match_bindings["dtype"]
-    v_dtype = ir.DataType(v_dtype.value).numpy()  # type: ignore[union-attr]
-    val = ir.Tensor(np.zeros(1, dtype=v_dtype))
-    return op.ConstantOfShape(shape, value=val)
+def fused_cast_constant_of_shape_without_value(op, shape, dtype, **_):
+    zero = onnx.helper.make_tensor("value", dtype.value, (), [0])
```
justinchuby, line 32:
```suggestion
    zero = ir.Tensor(np.zeros(1, dtype=ir.DataType(dtype.value)).numpy())
```

onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, Sequence
 
-import numpy as np
+import onnx.helper
```
justinchuby, line 5:
```suggestion
```

maybe not needed?

onnxscript/rewriter/cast_constant_of_shape.py
justinchuby: And we can probably make `ir.DataType(dtype.value).numpy())` a helper function

gramalingam: A couple of questions/comments:
* The API should allow a TensorProto created in any form (as it does, as seen here, I think).
* From a readability standpoint, the make_tensor seems easier to read (though, admittedly, it may be because of greater familiarity with that long-used API)
* What happens to types that are not supported in numpy()? Just wondering

justinchuby: Right - the api does support tensor protos being created. I just felt it would be a good idea to avoid moving protos around, from a consistency point of view. But it is an issue when the types are not supported by numpy.

I think eventually we will need to replicate some logic from make_tensor to the Tensor class.

justinchuby: In terms of readability, we can create the same interface w/o the need for serialization to bytes.

justinchuby: I can do this in a separate PR

onnxscript/rewriter/broadcast_to_matmul.py
```diff
@@ -140,10 +140,7 @@ def two_reshapes_matmul_reshape_pattern(input_a, input_b, shape_a, shape_b, shap
     return op.Reshape(matmul, shape_c)
 
 
-def matmul_with_two_shape_inputs(input_a, input_b, shape_a, shape_b, shape_c):
-    del shape_a  # Unused
-    del shape_b  # Unused
-    del shape_c  # Unused
+def matmul(op, input_a, input_b, **_):
```
xadupre, line 143: Do we need **_? If not, let's remove it.

onnxscript/rewriter/cast_constant_of_shape.py
```diff
@@ -12,58 +11,34 @@
...
-) -> pattern.OpPattern:
-    del match_bindings  # Unused
+def fused_cast_constant_of_shape(op, shape: ir.Value, scalar: ir.Attr, dtype: ir.Attr, **_):
+    # Cast scalar (a TensorProto attribute) to the specified dtype
+    scalar_value = scalar.value.numpy().item()
+    cast_value = onnx.helper.make_tensor("value", dtype.value, (), [scalar_value])
+    return op.ConstantOfShape(shape, value=cast_value)
+
+
+def cast_constant_of_shape_without_value(shape, dtype):
```
xadupre, line 26: Can we create only one node here instead of two?

onnxscript/rewriter/pattern.py
```diff
@@ -800,63 +678,68 @@ def __init__(self, function: Callable) -> None:
...
+
+    # TODO(justinchuby): Merge with the rest of pattern building methods
+    def __init__(self):
+        self.tape = _tape.Tape()
+
+    def __getattr__(self, op_type: str) -> Any:
+        return lambda *args, **kwargs: self._make_node(op_type, args, kwargs)
+
+    def _make_node(self, op_type: str, inputs: Sequence[ir.Value], kwargs: dict[str, Any]):
+        domain = kwargs.pop("domain", "")
```
titaiwangms, line 697: Would versioning be a problem when it comes to custom domain? Now we don't have version info to add into model.opset_imports?

gramalingam: yes, we also need to support version as a parameter. That's pending on my list, will handle it.

onnxscript/rewriter/pattern.py
```diff
@@ -729,15 +759,13 @@ def __init__(self, function) -> None:
...
         match_bindings: dict[str, ir.Value | Any] | None = None,
-    ) -> tuple[NodePattern | None, int | None]:
-        if match_bindings is None:
-            return None, None
+    ) -> ReplacementSubgraph:
         context = RewriterContext()
         new_values = self._function(context, **match_bindings)
-        return context.nodes
+        if not isinstance(new_values, Sequence):
+            new_values = [new_values]
```
justinchuby, line 767: The new values are in fact new outputs of the subgraph?

gramalingam: Right. Maybe new_outputs would be a better name

onnxscript/rewriter/cast_constant_of_shape.py
gramalingam: https://github.com/microsoft/onnxscript/issues/1442

Issue #1324, opened 2024-04-01
IR Roadmap


justinchuby:
- [x] Data model https://github.com/microsoft/onnx-rewriter/pull/232 https://github.com/microsoft/onnx-rewriter/pull/257 
	- [x] Support tensor protos cleanly https://github.com/microsoft/onnx-rewriter/pull/254 https://github.com/microsoft/onnx-rewriter/pull/264
	- [ ] Support dtypes that are not in numpy
    - [ ] User defined data types
    - [x] Typing system: Unify Sequence types and Optional types https://github.com/microsoft/onnx-rewriter/pull/254 https://github.com/microsoft/onnx-rewriter/pull/264
    - [x] Tensor: implement nbytes, size and itemsize https://github.com/microsoft/onnx-rewriter/pull/254
    - [x] Use slots https://github.com/microsoft/onnx-rewriter/pull/254
    - [x] ~Shared AttrRef in functions~
    - [ ] ~Make `Function` callable.` __call__(self, *values, *attrs) -> Sequence[Value]` (but still need to get the nodes); we can also just call a function to inline it?~ Or we can "expand" it
- [ ] Use [weakref](https://docs.python.org/3/library/weakref.html#module-weakref) for some bidirectional pointers / dictionaries
    - [ ] For example, pointer in DiagnosticMessage?
- [x] Serialization to proto https://github.com/microsoft/onnx-rewriter/pull/267
    - [x] Be careful to avoid copying
- [x] Deserialization from proto https://github.com/microsoft/onnx-rewriter/pull/254 
	- [x] Test serde with the ONNX test suite https://github.com/microsoft/onnx-rewriter/pull/267
    - [x] ~User defined external data format in deserializer?~ They should handle them in a pass
    - [ ] Handle unsorted nodes
- [ ] Mutation methods for graph; sorting @justinchuby "I am here"
    - https://github.com/microsoft/onnx-rewriter/pull/273
    - [ ] Mutable protocols
    - [ ] FX graph
    - [ ] Topo sorting
    - [ ] Insertion
    - [ ] Iterator
    - [ ] Validation scenarios
        - [ ] Replace subgraph with new node
        - [ ] Build new graph from a list of nodes
        - [ ] Drop nodes
        - [ ] Insert nodes
- [ ] Initialization / mutation hooks
- [ ] Sparse tensor
- [ ] Naming authority: `must_be_named()`
- [x] Function ref in nodes (Just create new constants)
[...]

PR #1414, opened 2024-04-20
[IR] Implement replace_all_uses_with

justinchuby:
Create a convenience function `replace_all_uses_with` to replace uses of values. Test with doctest.

Rename convenience to _convenience. We should expose some of the methods once they are proven to be useful.


onnxscript/ir/_convenience.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any, Mapping, Sequence
+
+from onnxscript.ir import _core, _protocols
+
+
+def convert_attributes(attrs: Mapping[str, Any]) -> list[_core.Attr]:
```
gramalingam, line 18: Is this the same as [the one in _tape.py](https://github.com/microsoft/onnxscript/blob/5a0cd8ef59179e8e53221c7a53fbf437c42c942a/onnxscript/rewriter/_tape.py#L15) ? We can presumably use this there?

onnxscript/ir/_convenience.py
```diff
@@ -0,0 +1,93 @@
...
+        values: The value or values to be replaced.
+        replacements: The new value or values to use as inputs.
+    """
+    if not isinstance(values, Sequence):
+        values = (values,)
+    if not isinstance(replacements, Sequence):
+        replacements = (replacements,)
+    if len(values) != len(replacements):
+        raise ValueError("The number of values and replacements must match.")
+    for value, replacement in zip(values, replacements):
```
gramalingam, line 91: A question about the IR: how is a (missing) optional output represented? Does it have a corresponding Value object or is it represented as None in the node's outputs? Wondering if we should handle the possibility of None being here.

gramalingam (approved):

onnxscript/ir/_convenience.py
justinchuby: I wasn't aware that an empty output was possible, so I didn't handle them anywhere. I will create a pr to implement empty output support.

justinchuby: https://github.com/microsoft/onnxscript/issues/1324

onnxscript/ir/_convenience.py
justinchuby: Sure! I will create a follow up

onnxscript/ir/_convenience.py
justinchuby: Correction: an empty output is represented as a Value whose name is `""`. It will not have any users so we don't need to do anything about it.

PR #1415, opened 2024-04-20
[IR] Add a safe option to graph.remove

justinchuby:
When safe=True, `Graph.remove()` performs the following actions before removal:

1. It checks to make sure there are no users of the node that are not
    to be removed before removing it.
2. It checks the node does not contribute to any graph outputs.
3. It removes references to all inputs so it is no longer a user of other nodes.

The additional check is typically O(1), assuming the number of consumers of nodes is small

This option will break all interconnections (input/output) of the node to remove.

Also changed:

1. Allow remove to accept a collection of nodes.


onnxscript/rewriter/pattern.py
```diff
@@ -1050,8 +1047,7 @@ def _apply_deltas(
                 if (index := new_output.meta.get(_ir_utils.GRAPH_OUTPUT_META_KEY)) is not None:  # type: ignore[assignment]
                     graph_or_function.outputs[index] = new_output
 
-    for n in to_delete:
-        graph_or_function.remove(n)
+    graph_or_function.remove(to_delete, safe=True)
```
justinchuby, line 1050: Otherwise we need to set all inputs in to_delete to None to detach the usages.

onnxscript/ir/_core.py
```diff
@@ -1236,6 +1237,42 @@ def __init__(
...
+        if output in graph_outputs:
+            raise ValueError(
+                f"Node {node} is still an output of the graph and cannot be removed when safe=True."
+            )
+        for node, _ in output.consumers():
+            if node in to_remove:
+                continue
+            raise ValueError(
+                f"Node {node} is still being used by other nodes that are not to be "
+                f"removed: {output.consumers()}"
```
xadupre, line 1272: ``consumers`` returns ``self._consumers.keys()``, which is an iterator. I wonder about the error message. It should write something like ``dict_keys``.

onnxscript/ir/_core.py
```diff
@@ -1390,19 +1427,46 @@ def extend(self, nodes: Iterable[Node], /) -> None:
...
+            nodes = {nodes}
+        else:
+            nodes = frozenset(nodes)
+        graph_outputs = frozenset(self.outputs)
+        for node in nodes:
+            if node.graph is not self:
+                raise ValueError(f"The node {node} does not belong to this graph.")
+            if safe:
+                # Check 1, 2
+                _check_node_safe_to_remove(node, nodes, graph_outputs)
```
xadupre, line 1461: ``nodes`` can be an iterator here and ``_check_node_safe_to_remove`` checks that the consumers of `node` belongs to ``nodes``. Maybe Iterator is not the right annotation?

justinchuby: I think it is made a set in https://github.com/microsoft/onnxscript/pull/1415/files#diff-33a5ea7182ac2a3ba4127e9fea3cd2db2dbe54810994ca210727d28fccf2b5cfR1451-R1455. Do you prefer the new variable to be renamed for clarity?

onnxscript/ir/_core.py
justinchuby: Updated to a tuple so that it is displayed

titaiwangms (approved):

onnxscript/ir/_core.py
justinchuby: Renamed to nodes_set

PR #1417, opened 2024-04-22
[IR] Implement `convert_attributes`

justinchuby:
Implement `convert_attribute` and `convert_attributes` as convenience functions for the IR.

Also updates repr for Attr to make it more succinct.

### DIFF
--- a/onnxscript/ir/_convenience.py
+++ b/onnxscript/ir/_convenience.py
@@ -164,7 +164,7 @@ def convert_attributes(
         [AttrInt64('int', INT, 1), AttrFloat32('float', FLOAT, 1.0), AttrString('str', STRING, 'hello'), AttrInt64s('ints', INTS, [1, 2, 3]), AttrFloat32s('floats', FLOATS, [1.0, 2.0, 3.0]), AttrStrings('strings', STRINGS, ['hello', 'world']), AttrTensor('tensor', TENSOR, Tensor<INT64,[3]>(array([1, 2, 3]))), AttrTensor('tensor_proto', TENSOR, TensorProtoTensor<FLOAT,[3]>('proto'))]
 
     Args:
-        attrs: A dictionary of {<attribute name>: <python objects> to convert.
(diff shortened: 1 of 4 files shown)


onnxscript/ir/_convenience.py
```diff
@@ -10,30 +10,168 @@
...
+            raise ValueError(
+                f"Attribute type '{attr.type}' does not match provided type '{attr_type}'"
+            )
+        return attr
+
+    if attr_type is None:
+        attr_type = _infer_attribute_type(attr)
+
+    else:
+        if isinstance(attr, (_core.Attr, _core.RefAttr)):
```
gramalingam, line 105: Checks seem duplicated above and here

titaiwangms (commented):
I think it's been used by `generic_pattern.py`, so the file should also be updated?

onnxscript/ir/_convenience.py
```diff
@@ -10,30 +10,168 @@
...
+        return _core.AttrStrings(name, attr)  # type: ignore
+    if attr_type == _enums.AttributeType.TENSOR:
+        if isinstance(attr, (_core.TensorBase, _protocols.TensorProtocol)):
+            return _core.AttrTensor(name, attr)
+        if isinstance(attr, onnx.TensorProto):
+            return _core.AttrTensor(name, serde.TensorProtoTensor(attr))
+    raise TypeError(f"Unsupported attribute type: '{type(attr)}'")
+
+
+def convert_attributes(
```
gramalingam, line 135: nit: I wonder if we should (later) simplify this double-conversion (from a dictionary/map to list here, and then back from list to dictionary when we construct a node)?

gramalingam (approved):

onnxscript/ir/_convenience.py
justinchuby: Good catch thanks

justinchuby:
> I think it's been used by `generic_pattern.py`, so the file should also be updated?

I think it's updated? In _tape right?

PR #1418, opened 2024-04-22
Use new IR in function rewriter

titaiwangms:
Update function rewriter to use new graph IR. This avoids serialize/deserialize between rewriting. Before this PR, it's using protobuf to insert/replace functions.

Summary:
1. Instead of updating old function to use new function content, this PR uses new function to fully replace the old one in the model.
2. Avoid passing around protobuf, and modify within protobuf
3. Apply IR APIs


onnxscript/rewriter/function_rule.py
```diff
@@ -8,7 +8,7 @@
 
 import onnxscript
 from onnxscript import ir
-from onnxscript._legacy_ir import visitor
+from onnxscript.ir import serde
```
justinchuby, line 11: Prefer using ir.serde, as a convention

justinchuby:
Can we make the passes return an ir function? Doesnt have to be this PR.

titaiwangms:
> Can we make the passes return an ir function? Doesnt have to be this PR.

In which we should return ir.Function?

onnxscript/rewriter/function_rule.py
```diff
@@ -128,24 +127,14 @@ def _find_node_contains_key_in_name(
         return None
 
     def _find_node_by_type(
```
titaiwangms, line 129: @justinchuby Do we have a better approach for node searching?

onnxscript/rewriter/function_rule.py
titaiwangms: Done

onnxscript/rewriter/function_rule.py
justinchuby: Nothing better in the IR. Would it be helpful to build an auxiliary data structure here? Eg a dictionary?

titaiwangms: I think we can extend it when we have other needs.

onnxscript/rewriter/onnxruntime/transformers/biassplitgelu.py
```diff
@@ -29,4 +29,4 @@ def ggelu(input, weight, bias):
             return msft_opset.BiasSplitGelu(matmul_input, bias)
 
         function_proto = onnxscript.script(default_opset=op)(ggelu).to_function_proto()  # type: ignore[arg-type]
```
justinchuby, line 31: Here for example we are returning protos. Would it make sense to return an ir function?

titaiwangms: I see. That makes sense. ir.Function goes in, and we should return the same type.

titaiwangms: Done

onnxscript/rewriter/function_rule.py
justinchuby, line 1: Are we able to delete _legacy_ir/ now?

onnxscript/rewriter/function_rule.py
```diff
@@ -130,27 +127,17 @@ def _find_node_contains_key_in_name(
...
-            for output in node.output:
-                if output == value_name:
-                    return node
-        return None
-
     def compose_new_function(
-        self, old_function: onnx.FunctionProto, pkg_version: version.Version | None
-    ) -> tuple[onnx.FunctionProto, tuple[onnx.OperatorSetIdProto]]:
+        self, old_function: ir.Function, pkg_version: version.Version | None
+    ) -> tuple[ir.Function]:
```
justinchuby, line 140: Is the return type accurate? I would check if mypy is enabled for this file in lintrunner.toml

onnxscript/rewriter/function_rule.py
titaiwangms: optimizer/evaluator and visitor is still using legacy_ir to do graph analysis. I think that part remains unresolved. Either we decide to extend the current IR to do graph analysis or we keep the old one.

https://github.com/microsoft/onnxscript/blob/15d295a833bde1de688328be53675406eccbb82c/onnxscript/optimizer/evaluator.py#L52

@gramalingam can provide more details on this I think.

onnxscript/rewriter/function_rule.py
titaiwangms: It looks like most/all of files under optimizer and rewriter is not covered in MYPY and PYLINT. Pattern-writing breaks MYPY.

titaiwangms: I enabled MYPY to cover this file.

onnxscript/rewriter/function_rule.py
```diff
@@ -146,7 +147,8 @@ def compose_new_function(
         Raises:
             FunctionRewriteError: If the rewrite fails.
         """
-        func = self._version_controller.dispatch(pkg_version)
+        # self._version_controller is created in the subclass
```
justinchuby, line 150: hmm. That's a little strange? But we don't need to worry too much about it here

justinchuby (approved):

onnxscript/rewriter/function_rule.py
```diff
@@ -185,57 +175,53 @@ def try_rewrite_function(
...
-        function.output.extend(new_function.output)
+        new_function.name = function.name
+        new_function.domain = function.domain
 
-        del function.node[:]
-        function.node.extend(nodes)
-        for new_opset in opset_imports:
-            function.opset_import.append(new_opset)
-            if new_opset.domain not in self._opset_imports:
-                model.opset_import.append(new_opset)
```
titaiwangms, line 214: This PR has a regression here. I didn't add the new opset version into opset_imports..

titaiwangms: @justinchuby Even with replace the whole new function, I think we still have to manually update opset_imports to model.opset_imports, right?

justinchuby: That's right. We could automate this process though.

justinchuby: https://github.com/microsoft/onnxscript/issues/1444

PR #1422, opened 2024-04-23
[IR] Rename Value.consumers() to Value.uses()

justinchuby:
A second renaming PR, hopefully more stable afterward... 

1. The consumer() naming is not as accurate as it can be, because while a `Node` is a consumer/user, a `(Node, index)` is a "use". At the same time ONNX C++ IR (https://github.com/onnx/onnx/blob/c459890aa266d74ca612f334d0f3e869dcdbb597/onnx/common/ir.h#L288) as well as torch fx (https://github.com/pytorch/pytorch/blob/ea61c9cb299b6dfebc57dc9d8821c34321d568ab/torch/fx/node.py#L226) both uses the `user` / `use` concept. FX calls it `user` because it is a Node-Node relationship. We have a Node-(Node (user), index) relationship.

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -638,7 +638,7 @@ class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
     After the node is initialized, it will add itself as a user of the input values.
 
     The output values of the node are created during node initialization and are immutable.
-    To change the output values, create a new node and replace the each of the inputs of ``output.uses`` with
+    To change the output values, create a new node and replace the each of the inputs of ``output.uses()`` with
     the new output values by calling :meth:`replace_input_with` on the using nodes
     of this node's outputs.
     """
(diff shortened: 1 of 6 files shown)


onnxscript/ir/_core.py
```diff
@@ -635,11 +635,11 @@ class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
     user is responsible to call ``graph.append(node)`` (or other mutation methods
     in :class:`Graph`) to add the node to the graph.
 
-    After the node is initialized, it will add itself as a consumer of the input values.
+    After the node is initialized, it will add itself as a user of the input values.
 
     The output values of the node are created during node initialization and are immutable.
-    To change the output values, create a new node and replace the each of the inputs of ``output.consumers`` with
-    the new output values by calling :meth:`replace_input_with` on the consumer nodes
+    To change the output values, create a new node and replace the each of the inputs of ``output.uses`` with
```
justinchuby, line 641:
```suggestion
    To change the output values, create a new node and replace the each of the inputs of ``output.uses()`` with
```

gramalingam (approved):

gramalingam:
Speaking of other terminology: the concept of def-use is common in compiler analysis: eg: https://en.wikipedia.org/wiki/Use-define_chain

PR #1437, opened 2024-04-23
[IR] Create readme

justinchuby:
Initial readme for the IR. Describes the file structure.

### DIFF
--- a/onnxscript/ir/README.md
+++ b/onnxscript/ir/README.md
@@ -2,7 +2,7 @@
 
 An in-memory IR that supports the full ONNX spec, designed for graph construction, analysis and transformation.
 
-## Features
+## Features ✨
 
 - Full ONNX spec support: all valid models representable by ONNX protobuf, and a subset of invalid models (so you can load and fix them).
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):

xadupre:
It would be great to have all the md files compiled into a documentation. That way, this page can refer to the documentation of each class. You should also add short examples of code.

onnxscript/ir/README.md
```diff
@@ -0,0 +1,21 @@
+# ONNX IR
+
+An in-memory IR that supports the full ONNX spec, designed for graph construction, analysis and transformation.
+
+## Features
+
+- Full ONNX spec support: all valid models representable by ONNX protobuf, and a subset of invalid models (so you can load and fix them).
```
xadupre, line 7: I would remove `and a subset of invalid models (so you can load and fix them)`. Either it can load every saved onnx models even if they are invalid (they were saved so the IR should be able to load them), either it cannot then the tool cannot be used to fix invalid models.

justinchuby: From teams discussion: It is still useful that we can process some invalid graph, so users can fix them without leaving the IR. However, it is not realistic to represent the full space of invalid graphs, because the IR does not mirror the protobuf data structure. We can choose to support invalid graphs that are still meaningful

PR #1451, opened 2024-04-25
Update function opset imports in rewriter

titaiwangms:
Previous to this PR, if rewrite pattern includes custom domain, and it's rewriting nodes inside a function. The invalid model will be generated because of lack of registered domain.

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -768,7 +768,7 @@ class ReplacementPatternFunction:
         return ReplacementSubgraph(new_values, context.nodes, context.used_opsets)
 
 
-def _update_opset_impots(
+def _update_opset_imports(
     graph_or_function: ir.Graph | ir.Function, delta: ReplacementSubgraph
 ):
     imports = graph_or_function.opset_imports
@@ -864,8 +864,8 @@ class RewriteRule:
                 # appears incorrect for single-pattern matcher. Best to alter iteration to apply
                 # each rewrite immediately, instead of accumulating them.
                 # (iv) return delta here
-                _update_opset_impots(graph_or_function, delta)
-                _update_opset_impots(model.graph, delta)
+                _update_opset_imports(graph_or_function, delta)
+                _update_opset_imports(model.graph, delta)
                 return match.values, delta.new_nodes
         return None
 
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -768,6 +768,21 @@ class ReplacementPatternFunction:
         return ReplacementSubgraph(new_values, context.nodes, context.used_opsets)
 
 
+def _update_opset_impots(
+    graph_or_function: ir.Graph | ir.Function, delta: ReplacementSubgraph
+):
+    imports = graph_or_function.opset_imports
+    for domain, version in delta.used_opsets:
+        if domain not in imports:
+            # use 1 as default version if not explicitly specified
+            imports[domain] = version if version is not None else 1
+        elif version is not None and version != imports[domain]:
+            raise ValueError(
+                f"Multiple versions of opset {domain} used. "
+                f"Expected version {imports[domain]}, but got {version}."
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -768,6 +768,21 @@ def get_replacement(
         return ReplacementSubgraph(new_values, context.nodes, context.used_opsets)
 
 
+def _update_opset_impots(
```
gramalingam, line 771: nit: typo, "impots" => "imports"

onnxscript/rewriter/pattern.py
```diff
@@ -975,7 +982,7 @@ def _apply_to_graph_or_function(
         for rule in self.rules:
             deltas = []
             for i, node in enumerate(graph_or_function):
-                delta = rule.try_rewrite(model, node)
+                delta = rule.try_rewrite(model, graph_or_function, node)
```
gramalingam, line 985: I think generic_pattern.py also has an implementation of `try_rewrite` ... at least the signature will have to be changed there?

titaiwangms: You are right. I don't find generic_pattern.py updates opset_imports. Do you recall where it might be, or you add it somewhere?

gramalingam: Right, I don't think the generic pattern matcher is handling the introduction of new opsets. Agree it can be done separately (ideally by using same implementation).

onnxscript/rewriter/pattern_test.py
```diff
@@ -323,6 +323,39 @@ def double(op, x):
...
+        """
+        )
+        model = ir.serde.deserialize_model(model_proto)
+        count = pattern.RewriteRuleSet([rule], commute=True).apply_to_model(model)
+        self.assertEqual(count, 1)
+        self.assertEqual(len(model.functions), 1)
+        self.assertEqual(model.graph.opset_imports["custom.domain"], 10)
+        self.assertEqual(
+            model.functions[("pkg.custom", "afunction", "")].opset_imports["custom.domain"], 10
+        )
```
gramalingam, line 357: I think it is a good idea to deserialize model and call checker ... Just realizing this after Xavier's bug reports, it will help us catch other bugs.

gramalingam (approved):

onnxscript/rewriter/pattern.py
titaiwangms: I tried but we would get this instead:

```
onnxscript/rewriter/pattern_test.py:360: in test_opset_import_in_function
    onnx.checker.check_model(model_ir)
../anaconda3/envs/main-dev/lib/python3.11/site-packages/onnx/checker.py:179: in check_model
    C.check_model(
E   onnx.onnx_cpp2py_export.checker.ValidationError: No Op registered for afunction with domain_version of 17
```

It lookes like checker does not like the idea of ("", 17)?

onnxscript/rewriter/pattern_test.py
titaiwangms: I tried but we would get this instead:

```
onnxscript/rewriter/pattern_test.py:360: in test_opset_import_in_function
    onnx.checker.check_model(model_ir)
../anaconda3/envs/main-dev/lib/python3.11/site-packages/onnx/checker.py:179: in check_model
    C.check_model(
E   onnx.onnx_cpp2py_export.checker.ValidationError: No Op registered for afunction with domain_version of 17
```

It lookes like checker does not like the idea of ("", 17)? Or it's complaining about we delete the used opset?

titaiwangms: I will raise an issue about adding checker into optimizer/rewriter.

gramalingam: Oh, line 340 should be `z = pkg.custom.afunction(x)`

titaiwangms: I will do a follow up

PR #1452, opened 2024-04-25
[IR] Test with onnx backend test files

justinchuby:


onnxscript/ir/_core.py
```diff
@@ -453,6 +455,116 @@ def meta(self) -> _metadata.MetadataStore:
...
+        *,
+        shape: Shape | None = None,
+        name: str = "",
+        doc_string: str | None = None,
+        metadata_props: dict[str, str] | None = None,
+    ) -> None:
+        """Initialize a tensor.
+
+        Args:
+            value: The backing data of the tensor. It can be a numpy array or a Sequence of strings.
```
justinchuby, line 482:
```suggestion
            value: The backing data of the tensor. It can be a numpy array or a Sequence of bytes.
```

onnxscript/ir/_core.py
```diff
@@ -453,6 +455,116 @@ def meta(self) -> _metadata.MetadataStore:
...
+        """The data type of the tensor. Immutable."""
+        return _enums.DataType.STRING
+
+    @property
+    def shape(self) -> Shape:
+        """The shape of the tensor. Immutable."""
+        return self._shape
+
+    @property
+    def raw(self) -> str:
```
justinchuby, line 533:
```suggestion
    def raw(self) -> Sequence[bytes] | npt.NDArray[np.bytes_]:
```

onnxscript/ir/_core.py
```diff
@@ -453,6 +455,116 @@ def meta(self) -> _metadata.MetadataStore:
...
+        """Backing data of the tensor. Immutable."""
+        return self._raw  # type: ignore[return-value]
+
+    def numpy(self) -> np.ndarray:
```
justinchuby, line 537:
```suggestion
    def numpy(self) -> npt.NDArray[np.bytes_]:
```

onnxscript/ir/serde.py
```diff
@@ -1195,12 +1259,18 @@ def serialize_type_into(type_proto: onnx.TypeProto, from_: _protocols.TypeProtoc
...
+    value_field = type_proto.WhichOneof("value")
+    tensor_type = getattr(type_proto, value_field)
+    while not isinstance(tensor_type.elem_type, int):
+        # Find the leaf type that has the shape field
+        type_proto = tensor_type.elem_type
+        value_field = type_proto.WhichOneof("value")
+        tensor_type = getattr(type_proto, value_field)
     # When from is empty, we still need to set the shape field to an empty list by touching it
-    tensor_type_proto.shape.ClearField("dim")
+    tensor_type.shape.ClearField("dim")
```
justinchuby, line 1270: Check naming

onnxscript/ir/serde.py
```diff
@@ -1092,37 +1123,69 @@ def _fill_in_value_for_attribute(
...
+        if value.shape is not None:
+            serialize_shape_into(attribute_proto.tp, value.shape)
+        attribute_proto.type = onnx.AttributeProto.TYPE_PROTO
+    elif type_ == _enums.AttributeType.TYPE_PROTOS:
+        for ir_type in value:
+            # ir_type: _core.TypeAndShape
+            type_proto = attribute_proto.type_protos.add()
+            if ir_type.type is not None:
+                serialize_type_into(type_proto, ir_type.type)
+            # Need to create the type _before_ writing the shape
```
justinchuby, line 1185:
```suggestion
            # Need to create the type _before_ writing the shape so that the shape can be written to the leaf type proto
```

onnxscript/ir/serde.py
```diff
@@ -1169,10 +1232,11 @@ def serialize_value_into(
         value_info_proto.name = from_.name
     if from_.metadata_props:
         _serialize_metadata_props_into(value_info_proto.metadata_props, from_.metadata_props)
-    if from_.shape is not None:
-        serialize_shape_into(value_info_proto.type, from_.shape)
     if from_.type is not None:
         serialize_type_into(value_info_proto.type, from_.type)
+    # Need to create the type _before_ writing the shape
```
justinchuby, line 1237:
```suggestion
    # Need to create the type _before_ writing the shape so that the shape can be written to the leaf type proto
```

onnxscript/testing/__init__.py
```diff
@@ -448,6 +449,14 @@ def assert_onnx_proto_equal(
                         error_message = f"Field {field} index {i} in sequence not equal. type(a_value_i): {type(a_value_i)}, type(b_value_i): {type(b_value_i)}, a_value_i: {a_value_i}, b_value_i: {b_value_i}"
                         raise AssertionError(error_message) from e
                 elif a_value_i != b_value_i:
+                    if (
+                        isinstance(a_value_i, float)
+                        and isinstance(b_value_i, float)
+                        and math.isnan(a_value_i)
+                        and math.isnan(b_value_i)
+                    ):
+                        # Consider NaNs equal
```
titaiwangms, line 458: Does this mean ONNX IR could get different results/values from the original one (protobuf?)

onnxscript/ir/serde.py
```diff
@@ -1213,4 +1283,6 @@ def serialize_dimension_into(
     if isinstance(dim, int):
         dim_proto.dim_value = dim
     elif isinstance(dim, (_core.SymbolicDim, _protocols.SymbolicDimProtocol)):
-        dim_proto.dim_param = str(dim.value)
+        if dim.value is not None:
+            # TODO(justinchuby): None is probably not a valid value for dim_param
```
titaiwangms, line 1287: Should we do something when it's None?

onnxscript/ir/serde.py
```diff
@@ -1169,10 +1232,11 @@ def serialize_value_into(
         value_info_proto.name = from_.name
     if from_.metadata_props:
         _serialize_metadata_props_into(value_info_proto.metadata_props, from_.metadata_props)
-    if from_.shape is not None:
-        serialize_shape_into(value_info_proto.type, from_.shape)
     if from_.type is not None:
         serialize_type_into(value_info_proto.type, from_.type)
+    # Need to create the type _before_ writing the shape so that the shape can be written to the leaf type proto
```
titaiwangms, line 1237: Should we show this dependency in code? Like putting from_.shape under from_.type is not None branch?

justinchuby: Do you mean line 1238-1239?

onnxscript/testing/__init__.py
justinchuby: There was a test cases in ONNX that contained NaN. Since NaN != NaN when directly compared, I created this to handle it. They should be considered equal for our purposes.

onnxscript/ir/serde.py
titaiwangms: Yes

justinchuby: I see what you are saying. Currently we just assume the type to be a type that is an unknown tensor dtype with a known shape, if type is None. I think that's ok?

onnxscript/ir/serde.py
justinchuby: Looks like we are ok doing nothing. The None was from a test case. We should add a pass to check that no dims are None in rewriter produced values. etc.

titaiwangms (approved):
CI fails not only on macOS, and py38 Ubuntu/Win

gramalingam (approved):

justinchuby:
> CI fails not only on macOS, and py38 Ubuntu/Win

Those are more spurious errors. We need a better way to handle them perhaps.

Issue #1455, opened 2024-04-25
Transformations that impact graph outputs


gramalingam:
Currently, there is a problem when a rewrite-rule transforms how a graph-output is computed. Specifically, it ends up with a different name for the graph-output, and further there was no associated type for that value.

* Ideally, we should not change graph output names
* We need to ensure update types for graph outputs (via onnx inference, if needed)

PR #1459, opened 2024-04-25
Propagate relevant info from old values to new values

gramalingam:
When we create new values to replace old ones, we lose some of the information associated with the old values. Propagate this information. (See https://github.com/microsoft/onnxscript/issues/1455)

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -903,13 +903,21 @@ def _apply_deltas(
         else:
             deleted_nodes, inserted_nodes = delta
             # Replace deleted nodes with inserted nodes.
-            # However, we merge the last deleted node and last inserted node
-            # to avoid replacing the values produced by the last deleted node
-            # in all places where they are used. So, we reuse the output
-            # values from the last deleted node and replace the node itself
             # TODO: simplify this
             last_deleted = deleted_nodes[-1]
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -903,13 +903,21 @@ def _apply_deltas(
...
+            for old_value, new_value in zip(last_deleted.outputs, last_inserted.outputs):
+                # Propagate relevant info from old value to new value
+                # TODO(Rama): Perhaps we should merge old and new types. As of now, new
+                # values don't have type information. Note that this could be a problem
+                # for semantics-altering rewrite-rules: we should allow users to override
+                # this for such rules.
+                new_value.type = old_value.type
+                new_value.shape = old_value.shape
+                new_value.const_value = old_value.const_value
+                new_value.name = old_value.name
```
gramalingam, line 919: @justinchuby : is this safe (per the naming authority)?

justinchuby: Yes!

justinchuby (approved):

onnxscript/rewriter/pattern.py
justinchuby: https://github.com/microsoft/onnxscript/issues/1460

PR #1464, opened 2024-04-26
Cleanup apply_deltas

gramalingam:
Next step in cleanup of pattern-matcher:

* Each rewrite-transformation is applied immediately when it is identified (since the IR now allows us to do this).
* Eliminate the extra checks to avoid overlapping matches (because of above)
* Use a structured representation of the delta
* Enable function constructing the replacement to return None (failure of rewrite)
* Other naming cleanup

TODO:
* Consider extending delta representation to capture old/new output-values

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -320,7 +320,7 @@ class MatchResult:
     """Represents the result of a match operation.
 
     A match can either succeed or fail.
-    If it succeeds, it returns a list of IR values that matched the pattern
+    If it succeeds, it returns a list of nodes that matched the pattern
     and a set of bindings for the variables in the pattern.
 
     Example:
@@ -330,21 +330,20 @@ class MatchResult:
             t2 = op.Reshape(t1, shape2)
             return t2
     The above pattern matches a sequence of two Reshape ops.
-    The matched_values will contain the values representing the (output of)
-    the two Reshape ops, and the bindings will contain the values that
-    are bound to the variables `x`, `shape1`, and `shape2`.
+    The matched_nodes will contain the two Reshape ops, and the bindings will
+    contain the values that are bound to the variables `x`, `shape1`, and `shape2`.
     """
 
     def __init__(
-        self, matched_values=None, bindings: dict[str, ir.Value | Any] | None = None
+        self, matched_nodes=None, bindings: dict[str, ir.Value | Any] | None = None
     ) -> None:
-        assert matched_values is None or isinstance(matched_values, list)
-        self.success: bool = matched_values is not None
-        # For a successful match, matched_values is a list of values that matched the pattern.
+        assert matched_nodes is None or isinstance(matched_nodes, list)
+        self.success: bool = matched_nodes is not None
+        # For a successful match, matched_nodes is a list of values that matched the pattern.
         # These include the internal nodes of the pattern that were matched, but not
         # the leaves (sub-trees) that match against the variables in the pattern.
         # These represent the values that will be replaced by the replacement pattern.
-        self.matched_values: Sequence[Any] | None = matched_values
+        self.matched_nodes: Sequence[ir.Node] | None = matched_nodes
         # For a successful match, bindings is a dictionary of mapping pattern-variable-names
         # to values.
         self.bindings: dict[str, Any] = bindings if bindings is not None else {}
@@ -357,12 +356,12 @@ class MatchResult:
         return cls(None)
 
     @property
-    def values(self) -> Sequence[Any] | None:
-        return self.matched_values
+    def nodes(self) -> Sequence[ir.Node] | None:
+        return self.matched_nodes
 
     def fail(self):
         self.success = False
-        self.matched_values = None
+        self.matched_nodes = None
         self.bindings = {}
 
     def extend(self, other: MatchResult | bool):
@@ -381,8 +380,8 @@ class MatchResult:
                     return
             else:
                 self.bindings[var] = val
-        assert self.matched_values is not None, "matched_values should not be None."
-        self.matched_values.extend(other.matched_values)  # type: ignore[attr-defined]
+        assert self.matched_nodes is not None, "matched_nodes should not be None."
+        self.matched_nodes.extend(other.matched_nodes)  # type: ignore[attr-defined]
 
 
 class ValuePattern:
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -563,10 +558,15 @@ def __init__(
...
-        return MatchResult.FAIL()
+    def match_scalar(self, scalar_value):
+        status = math.isclose(
+            scalar_value, self.value, rel_tol=self.rel_tol, abs_tol=self.abs_tol
+        )
+        # Note: If the value is produced by a Constant node, we could include
+        # the Constant node in the return_value list. However, we don't do that.
+        # Instead, we will rely on DCE to remove the constant node if it is not
+        # used elsewhere.
+        return MatchResult(status)
```
titaiwangms, line 569:
```suggestion
        return MatchResult(success=status)
```

I somehow feel like the arg in MatchResult would be not so obvious if we don't specify it's success.

onnxscript/rewriter/pattern.py
```diff
@@ -467,7 +462,7 @@ def matches_node(self, node: ir.Node, model: ir.Model) -> MatchResult:
             return MatchResult.FAIL()
         if not self.op.matches(node.op_type):
             return MatchResult.FAIL()
-        match = MatchResult([])
+        match = MatchResult(success=True)
```
titaiwangms, line 465: If the usage is to assume it's a success in the begining, should we set a default to MatchResult?

titaiwangms (approved):
LG with non-blocking comments

onnxscript/rewriter/pattern.py
```diff
@@ -330,59 +330,53 @@ def pattern(x, shape1, shape2):
...
     @classmethod
     def FAIL(cls):
-        return cls(None)
+        return cls(False)
 
     @property
-    def values(self) -> Sequence[Any] | None:
-        return self.matched_values
+    def nodes(self) -> Sequence[ir.Node]:
+        return self.matched_nodes
```
justinchuby, line 357: Does it mean match_nodes can be private?

onnxscript/rewriter/pattern.py
```diff
@@ -330,59 +330,53 @@ def pattern(x, shape1, shape2):
...
-        self.matched_values = None
-        self.bindings = {}
+    def bind(self, var: str, value: Any):
+        self.bindings[var] = value
 
     def extend(self, other: MatchResult | bool):
         if not self.success:
             return
         if not other:
-            self.fail()
```
justinchuby, line 372: Curious how `fail()` was designed and what are the thoughts for setting a boolean state instead? Just for my learning. Thanks!

onnxscript/rewriter/pattern.py
```diff
@@ -491,11 +486,11 @@ def matches_node(self, node: ir.Node, model: ir.Model) -> MatchResult:
...
             match.extend(sub_match)
         for name in node.attributes:
-            # TODO: Support matching default values for attributes.
+            # TODO: Support matching default nodes for attributes.
             if name not in self.attributes:
                 return MatchResult.FAIL()
-        assert match.values is not None, "Matched values should not be None."
-        match.values.append(node)  #  type: ignore[attr-defined]
+        assert match.nodes is not None, "Matched nodes should not be None."
+        match.nodes.append(node)  #  type: ignore[attr-defined]
```
justinchuby, line 493:
```suggestion
        match.nodes.append(node)

File: onnxscript/rewriter/pattern.py [hunk 6]
```diff
@@ -330,59 +330,53 @@ def pattern(x, shape1, shape2):
...
 
     @classmethod
     def FAIL(cls):
-        return cls(None)
+        return cls(False)
 
     @property
-    def values(self) -> Sequence[Any] | None:
-        return self.matched_values
+    def nodes(self) -> Sequence[ir.Node]:
```
justinchuby, line 356:
```suggestion
    def nodes(self) -> MutableSequence[ir.Node]:
```

since we can append to it

onnxscript/rewriter/pattern.py
```diff
@@ -827,25 +821,22 @@ def try_rewrite(
...
         if match:
-            assert match.values is not None, "Matched values should not be None."
-            if _valid_to_replace(match.values):
-                # bindings will be consumed by the replacement function
-                delta = self._replacement_pattern.get_replacement(match.bindings)
-                if len(delta.new_values) != self._target_num_outputs:
+            assert match.nodes is not None, "Matched values should not be None."
+            if _valid_to_replace(match.nodes):
+                delta = self._replacement_pattern.get_replacement(match)
+                if delta is None:
```
justinchuby, line 827:
```suggestion
                if replacement_subgraph is None:
```

for clarity?

onnxscript/rewriter/pattern.py
```diff
@@ -891,64 +883,53 @@ def _apply_deltas(
...
+        graph_or_function.insert_after(node, inserted_nodes)
         # TODO: improve this
         # This is updating the graph/function outputs to use the new outputs
         for inserted_node in inserted_nodes:
             for new_output in inserted_node.outputs:
                 if (index := new_output.meta.get(_ir_utils.GRAPH_OUTPUT_META_KEY)) is not None:  # type: ignore[assignment]
                     graph_or_function.outputs[index] = new_output
 
-    graph_or_function.remove(to_delete, safe=True)
+        for d in deleted_nodes:
```
justinchuby, line 903: Bookmark for self, to resume

onnxscript/rewriter/pattern.py
```diff
@@ -891,64 +883,53 @@ def _apply_deltas(
...
+        for old_value, new_value in zip(node.outputs, last_inserted.outputs):
+            # Propagate relevant info from old value to new value
+            # TODO(Rama): Perhaps we should merge old and new types. As of now, new
+            # values don't have type information. Note that this could be a problem
+            # for semantics-altering rewrite-rules: we should allow users to override
+            # this for such rules.
+            new_value.type = old_value.type
+            new_value.shape = old_value.shape
+            new_value.const_value = old_value.const_value
+            new_value.name = old_value.name
```
justinchuby, line 920: Should we extract is as a helper for now?

onnxscript/rewriter/pattern.py
justinchuby, line 903: I would rename deleted_nodes to nodes_to_remove or something like it because they are not removed yet. Non blocking

justinchuby (approved):

onnxscript/rewriter/pattern.py
gramalingam: No deep thought here. Not sure about your question, I am not altering the behavior/approach here, merely inlining the definition of fail(). On second thoughts, retaining the method might be useful ... let me think, anyway, there's more to come, since I'd like to merge this with Xavier's MatchResult definition eventually.

onnxscript/rewriter/pattern.py
gramalingam: Done (not pushed yet). Minor nit: this will have to be replaced in all the places where it appears ... making it a "committable suggestion" can backfire if we push the "commit suggestion" button (since it looks like it updates only one occurrence).

onnxscript/rewriter/pattern.py
gramalingam: Renamed to `matched_nodes` to match usage elsewhere

Issue #1443, opened 2024-04-25
optimizer fails on this particular model


xadupre:
The optimizer fails for the attach model (so dort fails as well). It was obtained with the latest onnx, onnxscript and torch nightly.

[dump3bug.zip](https://github.com/microsoft/onnxscript/files/15106272/dump3bug.zip)

To replicate:

```python
import onnx
from onnxscript import optimizer
onx = onnx.load(model)
optimized = optimizer.optimize(onx)
```

It is coming from the following graph module.

```
graph():
    %primals_7 : [num_users=1] = placeholder[target=primals_7]
    %primals_1 : [num_users=1] = placeholder[target=primals_1]
    %primals_2 : [num_users=1] = placeholder[target=primals_2]
    %primals_3 : [num_users=1] = placeholder[target=primals_3]
    %primals_4 : [num_users=1] = placeholder[target=primals_4]
    %primals_5 : [num_users=1] = placeholder[target=primals_5]
    %add : [num_users=2] = call_function[target=torch.ops.aten.add.Tensor](args = (%primals_7, %primals_1), kwargs = {})
    %_native_batch_norm_legit_no_training : [num_users=1] = call_function[target=torch.ops.aten._native_batch_norm_legit_no_training.default](args = (%add, %primals_2, %primals_3, %primals_4, %primals_5, 0.1, 1e-05), kwargs = {})
    %getitem : [num_users=1] = call_function[target=operator.getitem](args = (%_native_batch_norm_legit_no_training, 0), kwargs = {})
    return (add, getitem)
```

Error:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "check_model.py", line 43, in <module>
    optimized = optimizer.optimize(onx)
  File "onnxscript/onnxscript/optimizer/__init__.py", line 61, in optimize
    model = onnx.shape_inference.infer_shapes(
  File "onnx/onnx/shape_inference.py", line 46, in infer_shapes
    inferred_model_str = C.infer_shapes(
[...]

PR #1467, opened 2024-04-26
[ShapeInferenceError] Inferred shape and existing shape differ in dimension 0

xadupre:
Creates a unit test running DORT and showing the following error:

```
[ShapeInferenceError] Inferred shape and existing shape differ in dimension 0
```

Can be used to replicate issue #1443


gramalingam (approved):
nit: I wonder if it would help to make the tests conditional on the presence of torch (for managing the CI matrix).

xadupre:
Right. So far, I did not fix onnx-script. I just created the unit test which fails.

justinchuby:
We always have torch so I think we are ok

tests/dort_test.py
```diff
@@ -0,0 +1,202 @@
...
+import sys
+import unittest
+import warnings
+from typing import Optional
+
+import onnxruntime
+import torch
+import torch.nn as nn
+import torch.onnx
+from onnx.inliner import inline_local_functions
```
justinchuby, line 13:
```suggestion
import onnx.inliner
```

tests/dort_test.py
```diff
@@ -0,0 +1,202 @@
...
+        del first_model_proto.graph.node[:]
+        del first_model_proto.functions[:]
+        del first_model_proto.graph.initializer[:]
+        del first_model_proto.opset_import[:]
+        first_model_proto.graph.node.extend(next_model.graph.node)
+        first_model_proto.functions.extend(next_model.functions)
+        first_model_proto.graph.initializer.extend(next_model.graph.initializer)
+        first_model_proto.opset_import.extend(next_model.opset_import)
+
+        return first_model_proto
```
justinchuby, line 56: Avoid nested declarations of functions if it doesn't close over local variables?

tests/dort_test.py
```diff
@@ -0,0 +1,202 @@
...
+class TestSimpleDort(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        torch._dynamo.reset()  # pylint: disable=protected-access
+        # hides the warnings
+        # torch.onnx.dynamo_export only implements opset version 18 for now.
+        # torch.library.impl_abstract was renamed to torch.library.register_fake.
+        warnings.simplefilter("ignore", (UserWarning, DeprecationWarning))
+
+    def assertONNX(
```
justinchuby, line 105: I would consider a more descriptive name. Does it assert on that the output of the onnx model will match the torch output, for example? Also snake cases for all function names

tests/dort_test.py
```diff
@@ -0,0 +1,202 @@
...
+import torch
+import torch.nn as nn
+import torch.onnx
+from onnx.inliner import inline_local_functions
+
+
+def make_aot_ort(
+    dynamic: bool = False,
+    ort_optimization_level: Optional[str] = None,
+) -> tuple:
```
justinchuby, line 19:
```suggestion
) -> tuple:
```

Is the return type accurate?

tests/dort_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+        compiled_model = torch.compile(
+            copy.deepcopy(model),
+            backend=local_aot_ort,
+            dynamic=dynamic_axes is not None,
+            fullgraph=fullgraph,
+        )
+
+        baseline_result = model(*args)
+        result = compiled_model(*args)
```
titaiwangms, line 140: @xadupre I tried to put everything on CUDA and run the code, but even though compiled_model is on CUDA, this result will not be on CUDA. Do you know why?

PR #1441, opened 2024-04-24
[IR] Allow tensor created with numpy unsupported dtypes

justinchuby:
Support all ONNX types in `ir.Tensor` and fix re-serialization of tensor protos that does not use the `raw_data` field to store data. Previously the *_data to raw_data conversion was incorrect when the data type size doesn't match the *_data type size.

For numpy unsupported dtypes, the array is represented as the unsigned int type of the same size as the dtype. For example, uint8 for float8, uint16 for bfloat16; with an exception for 4bit types.

This PR creates a more performant and robust method to convert tensor proto to numpy array than `onnx.numpy_helper.to_array`. We will upstream the implementation in the future.

Additionally
- Consolidate all tensorproto classes to the TensorProtoTensor class

Tested: unit tests

PR #1481, opened 2024-04-30
[IR] Create documentation for tensors

justinchuby:
Document the TensorProtocol and the various tensor classes. When we create a helper function to create tensors in the IR, we should document that as well.


thiagocrepaldi (approved):

docs/intermediate_representation/tensors.md
```diff
@@ -0,0 +1,154 @@
+# Tensor representation in the IR
+
+The ONNX IR offers the :class:`TensorProtocol` interface for usings different data structures as backing data for tensors. Besides the traditional ``onnx.TensorProto``, you can also use ``np.ndarray``, ``torch.Tensor``, and virtually anything else to represent tensors in the graph. This allows for  They can be accessed and serialized via the same `TensorProtocol` interface.
```
gramalingam, line 3: change "They can be" to "them to be"?

docs/intermediate_representation/tensors.md
```diff
@@ -0,0 +1,154 @@
...
+    tensor = ir.TensorProtoTensor(tensor_proto)
+    print(tensor: ", tensor)  # TensorProtoTensor<INT16,[3]>(name='tensor')
+    print("shape: ", tensor.shape)  # ir.Shape([3])
+    print("dtype: ", tensor.dtype)  # ir.DataType.INT16
+    print(tensor.raw == tensor_proto)  # The raw field is the exact tensor_proto provided at initialization
+    print("tobytes: ", tensor.tobytes())  # b'\x01\x00\x02\x00\x03\x00'
+    print("numpy: ", tensor.numpy())  # array([1, 2, 3], dtype=int16)
+```
+
+When tensors store their data in a different data field like `int32_data`, use the corresponding class as a wrapper:
```
gramalingam, line 36: Unclear. This looks like the same wrapper as the one above? But the description makes it sound like it is a different wrapper?

docs/intermediate_representation/tensors.md
```diff
@@ -0,0 +1,154 @@
...
+
+```{eval-rst}
+.. exec_code::
+    import onnx
+    from onnxscript import ir
+
+    tensor_proto = onnx.helper.make_tensor(
+        "tensor", onnx.TensorProto.INT16, (3,), b'\x01\x00\x02\x00\x03\x00', raw=True
+    )
+    tensor = ir.TensorProtoTensor(tensor_proto)
```
gramalingam, line 27: Does the API allow users to create this tensor without using `onnx.helper.make_tensor`? How would they do that?

When users create the IR representation from a pre-existing tensor-proto, ideally they should call a single API method (without having to handle the different cases of a TensorProto themselves).

justinchuby: Good point - I am going to combine all tensor proto classes and simplify as part of https://github.com/microsoft/onnxscript/pull/1441

justinchuby: Done: https://github.com/microsoft/onnxscript/pull/1441/commits/040827821422f192cf2febccfbdd890ea7c60df7

justinchuby: > Does the API allow users to create this tensor without using onnx.helper.make_tensor? 

Yes - all users need is a tensor proto as input

I expect to create a similar (make_tensor) api that will produce ir tensor for us

docs/intermediate_representation/tensors.md
justinchuby: Removed

docs/intermediate_representation/tensors.md
justinchuby: Thanks for the catch. That was incomplete 😅

gramalingam (approved):

justinchuby:
1. single backtick in markdown
2. {py:class}`ir.Value <onnxscript.ir._core.Value>`

PR #1483, opened 2024-04-30
Add device to TorchScriptTensor

titaiwangms:
In torchlib, we have [device specific onnx function](https://github.com/microsoft/onnxscript/blob/8dba367fb000e3696c79b618638861b5cdf759dc/onnxscript/function_libs/torch_lib/ops/core.py#L5797-L5804), therefore, we need `TorchScriptTensor` to carry device property, so converter dispatcher can dispatch the best fitted function to the ATen op.


justinchuby:
Lgtm. Could you create a test?

justinchuby (approved):

titaiwangms:
> Lgtm. Could you create a test?

Yes. could you point out the TorchScriptTensor related tests at?

justinchuby:
I would just create one here, if possible: https://github.com/microsoft/onnxscript/blob/main/onnxscript/function_libs/torch_lib/graph_building/graph_building_test.py

Issue #1470, opened 2024-04-26
[torchlib] native_layer_norm_float32 is failing on MacOS CI with significant absolute difference


justinchuby:
Observations: Doesn't fail on other platforms. Started to fail only after the Mac runners on Github were down and brought back up a day ago. All failing tests have 1D single element inputs.

Example https://github.com/microsoft/onnxscript/actions/runs/8840793727/job/24276753517

```
_ TestOutputConsistencyEagerCPU.test_output_match_opinfo__native_layer_norm_cpu_float32 (inputs="['Tensor<torch.Size([1]), dtype=torch.float32>', (1,), 'Tensor<torch.Size([1]), dtype=torch.float32>', 'Tensor<torch.Size([1]), dtype=torch.float32>', 1e-05]", kwargs='{}', sample_num=8) _
[gw1] darwin -- Python 3.10.11 /Users/runner/work/onnxscript/onnxscript/.nox/test/bin/python
tests/function_libs/torch_lib/ops_test.py:279: in run_test_output_match
    torch.testing.assert_close(
E   AssertionError: Tensor-likes are not close!
E   
E   Mismatched elements: 1 / 1 (100.0%)
E   Greatest absolute difference: 1.154693603515625 at index (0,) (up to 0.00018 allowed)
E   Greatest relative difference: 0.0036514615640044212 at index (0,) (up to 3.7e-05 allowed)
```

cc @thiagocrepaldi @titaiwangms @xiaowuhu @fatcat-z

justinchuby:
Pasting the script here for access

```
import os
import onnx
import torch
import numpy as np
import onnxruntime as ort

current_folder = os.path.dirname(os.path.abspath(__file__))
onnx_file_path = current_folder + "/test_mac_issue.onnx"

def prepare_onnx_inputs(input_node_names, torch_tensor):
    onnx_inputs = {}
    onnx_tensor = torch_tensor.detach().cpu().numpy()

    onnx_inputs[input_node_names[0]] = onnx_tensor
    onnx_inputs[input_node_names[1]] = [1, 2, 3]
    onnx_inputs[input_node_names[2]] = onnx_tensor
    onnx_inputs[input_node_names[3]] = onnx_tensor

    return onnx_inputs

onnx_model = onnx.load(onnx_file_path)
onnx.checker.check_model(onnx_model)

providers = ['CPUExecutionProvider']
opt_without_optimization = ort.SessionOptions()
[...]

PR #1488, opened 2024-05-02
Fix native_layer_norm on mac

justinchuby:
Fix https://github.com/microsoft/onnxscript/issues/1470


justinchuby:
Assigned myself on https://github.com/microsoft/onnxscript/issues/1470 instead

PR #1489, opened 2024-05-02
[IR] Test IR roundtrip with model zoo

justinchuby:
The test is created as a local script so it doesn't become a long running CI test.

image


gramalingam (approved):
So, do all of them pass? (Just wondering about the image in the description of the PR.)

justinchuby:
Some of them don’t pass. I will need https://github.com/microsoft/onnxscript/pull/1452 and test again with the fixes

Issue #1381, opened 2024-04-14
Naming convention for common objects in rewritter/ir


justinchuby:
1. ONNX protobuf objects: `[type]_proto`. E.g. `model_proto`, `graph_proto`, `value_info_proto`
2. ONNX IR objects: `type`, E.g. `model`, `graph`

PR #1490, opened 2024-05-02
First series of P0 patterns to optimize llama

xadupre:


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,42 @@
+import onnxscript.rewriter.no_op as no_op
+import onnxscript.rewriter.pattern as orp
+from onnxscript.rewriter import pattern
+
+op = pattern.onnxop
+
+
+def transpose_identity(x):
+    return op.Transpose(x, perm=[0, 1, 2])
```
gramalingam, line 9: Hi Xavier, you can write this pattern to get the perm-attribute value from input model as below: using this to illustrate how we get values from the model itself:
```py
import onnxscript.rewriter as rewriter
import onnxscript.rewriter.pattern as orp
import onnxscript.ir as ir
from onnxscript import script, opset17, FLOAT
import onnx.printer

op = orp.onnxop

def transpose(x, perm):
    return op.Transpose(x, perm=perm)

def check_identity_perm_impl(x: ir.Value, perm: ir.Attr | ir.RefAttr):
    # Inside a function, an attribute may be a RefAttr and its value may not be available.
    if isinstance(perm, ir.RefAttr):
        return False
    if perm.type == ir.AttributeType.INTS:
        if perm.value == list(range(len(perm.value))):
            return True
    return False

def check_identity_perm(match_bindings):
    # Current API uses this interface. We would like to switch to the one in "check_identity_perm_impl".
    return check_identity_perm_impl(**match_bindings)

def identity_perm(op, x: ir.Value, perm: ir.Attr | ir.RefAttr):
    return op.Identity(x)

rule = orp.RewriteRule(transpose, identity_perm, check_identity_perm)

rules = orp.RewriteRuleSet([rule])

@script()
def example(X: FLOAT[2, 4, 1024]) -> FLOAT[2, 4, 1024]:
    return opset17.Transpose(X, perm=[0, 1, 2])

model = example.to_model_proto()

optimized = rewriter.rewrite(model, pattern_rewrite_rules=rules)

print(onnx.printer.to_text(optimized))

@script()
def example2(X: FLOAT[2, 1024]) -> FLOAT[2, 1024]:
    return opset17.Transpose(X, perm=[1, 0])

model2 = example2.to_model_proto()

optimized2 = rewriter.rewrite(model2, pattern_rewrite_rules=rules)

print(onnx.printer.to_text(optimized2))

```

gramalingam: Adding type annotations to the pattern, it will look like the below (just FYI):
```py
def transpose(x: orp.Var, perm: orp.Var) -> orp.ValuePattern:
    return op.Transpose(x, perm=perm)
```

xadupre:
``def example2(X: FLOAT[2, 1024]) -> FLOAT[2, 1024]:``, the pattern works for any type, any 2D shape, so I would prefer to avoid to be that precise.

xadupre:
I tried to follow the recommandations. I get the following error:

```
FAILED onnxscript/rewriter/llama_rule_sets_tests.py::LlamaRuleSetsTest::test_llama_p0_rule_set_identity - TypeError: Cannot convert <class 'onnxscript.ir._core.Input'> to ValuePattern
FAILED onnxscript/rewriter/llama_rule_sets_tests.py::LlamaRuleSetsTest::test_llama_p0_rule_set_transpose_transpose - TypeError: Cannot convert <class 'onnxscript.ir._core.Input'> to ValuePattern
```

CI is passing...

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,93 @@
...
+def transpose_identity_check(x: ir.Value, perm: ir.Attr | ir.RefAttr) -> bool:
+    if isinstance(perm, ir.RefAttr):
+        return False
+    if perm.type == ir.AttributeType.INTS:
+        if perm.value == list(range(len(perm.value))):
+            return True
+    return False
+
+
+def transpose_identity_rewrite(x: ir.Value, perm: ir.Attr):
```
gramalingam, line 24: I think you need to add `op` as the first parameter here ... and not use the global `op`.

Sorry about the confusing notation ... I am also trying to unify the conventions so that `op` is the first parameter for both functions uniformly (WIP).

onnxscript/rewriter/llama_rule_sets_tests.py
justinchuby, line 1: I would rename the test file since it's not picked up by pytest right now: llama_rule_sets_tests.pyllama_rule_sets_test.py

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def llama_p0_rule_set(verbose: int = 0) -> orp.RewriteRuleSet:
+    """Returns a set of rules which should be applied
+    before any other one as they usually remove unnecessary computation
+    such as the multiplication by 1 or two consecutive transpose.
+
+    Args:
+        verbose: verbosity
+    Returns:
```
justinchuby, line 79:
```suggestion
        verbose: verbosity

    Returns:
```

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
+from __future__ import annotations
+
+import onnxscript.ir as ir
+import onnxscript.rewriter.no_op as no_op
+import onnxscript.rewriter.pattern as orp
+from onnxscript.rewriter import pattern
```
justinchuby, line 6: Looks like the same module is imported under different name. Consider removing one of them?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _apply_transpose(perm: tuple[int, ...], on: list[int]) -> list[int]:
+    assert len(perm) == len(on), "length mismatch"
+    res = [-1 for i in on]
+    for i, p in enumerate(perm):
+        res[i] = on[p]
+    return res
+
+
+def _apply_transposes(perms: list[tuple[int, ...]], on: list[int] | None = None) -> list[int]:
```
justinchuby, line 48:
```suggestion
def _apply_transposes(perms: Sequence[tuple[int, ...]], on: Sequence[int] | None = None) -> list[int]:
```

nit: Prefer accepting Sequence

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
...
+
+transpose_identity_rule = pattern.RewriteRule(
+    transpose_identity, transpose_identity_rewrite, transpose_identity_check
+)
+transpose_transpose_rule = pattern.RewriteRule(
+    transpose_transpose, transpose_transpose_rewrite, transpose_transpose_check
+)
+
+
+def llama_p0_rule_set(verbose: int = 0) -> orp.RewriteRuleSet:
```
justinchuby, line 72:
```suggestion
def llama_p0_rule_set() -> orp.RewriteRuleSet:
```

Looks like verbose is unused?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
...
+from onnxscript.rewriter import pattern
+
+_op = pattern.onnxop
+
+
+def transpose_identity(x, perm):
+    return _op.Transpose(x, perm=perm)
+
+
+def transpose_identity_check(x: ir.Value, perm: ir.Attr | ir.RefAttr) -> bool:
```
justinchuby, line 15:
```suggestion
def _transpose_identity_check(x: ir.Value, perm: ir.Attr | ir.RefAttr) -> bool:
```

Can this be private?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
...
+
+def transpose_transpose_check(
+    x: ir.Value, perm1: ir.Attr | ir.RefAttr, perm2: ir.Attr | ir.RefAttr
+) -> bool:
+    if isinstance(perm1, ir.RefAttr) or isinstance(perm2, ir.RefAttr):
+        return False
+    return True
+
+
+def _apply_transpose(perm: tuple[int, ...], on: list[int]) -> list[int]:
```
justinchuby, line 40: Just wondering what `on` is and where it comes from?

onnxscript/rewriter/llama_rule_sets.py
justinchuby, line 48: Helpful to add a docstring?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _apply_transposes(perms: list[tuple[int, ...]], on: list[int] | None = None) -> list[int]:
+    if on is None:
+        on = list(range(len(perms[0])))
+    for p in perms:
+        on = _apply_transpose(p, on)
+    return on
+
+
+def transpose_transpose_rewrite(op, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr):
```
justinchuby, line 56: Helpful to add docstrings to functions?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -0,0 +1,92 @@
+from __future__ import annotations
```
justinchuby, line 1: I would create a docstring for the module to describe all rewrites involved.

onnxscript/rewriter/llama_rule_sets_tests.py
```diff
@@ -0,0 +1,128 @@
...
+            ),
+        ]
+        return models
+
+    def test_llama_p0_rule_set_identity(self):
+        for model in self._identity_models():
+            ir_model = ir.serde.deserialize_model(model)
+            rule_set = llama_rule_sets.llama_p0_rule_set()
+            rule_set.apply_to_model(ir_model)
+            rewritten_model = ir.serde.serialize_model(ir_model)
```
justinchuby, line 94: Prefer `model_proto` for proto messages, `model` for ir.Model 
- https://github.com/microsoft/onnxscript/issues/1381

onnxscript/rewriter/llama_rule_sets_tests.py
```diff
@@ -0,0 +1,128 @@
...
+                    [onnx.helper.make_tensor_value_info("X", FLOAT, [None, None, None])],
+                    [onnx.helper.make_tensor_value_info("Y", FLOAT, [None, None, None])],
+                ),
+                opset_imports=[onnx.helper.make_opsetid("", 18)],
+            ),
+        ]
+        return models
+
+    def test_llama_p0_rule_set_transpose_transpose(self):
+        for model in self._transpose_transpose_models():
```
justinchuby, line 117: Can this be parameterized so we can test all models in parallel?

onnxscript/rewriter/llama_rule_sets_tests.py
```diff
@@ -0,0 +1,128 @@
...
+        for model in self._identity_models():
+            ir_model = ir.serde.deserialize_model(model)
+            rule_set = llama_rule_sets.llama_p0_rule_set()
+            rule_set.apply_to_model(ir_model)
+            rewritten_model = ir.serde.serialize_model(ir_model)
+
+            self.assertEqual(["Identity"], [n.op_type for n in rewritten_model.graph.node])
+            self._check_model(model, rewritten_model)
+
+    def _transpose_transpose_models(self):
```
justinchuby, line 99: We can mode this method to the module level since it doesn't depend on self. This also makes it easier to parameterize the test methods.

justinchuby (approved):
lgtm with minor comments. Thanks!

onnxscript/rewriter/llama_rule_sets.py
xadupre: apply transpose ... on ...

onnxscript/rewriter/llama_rule_sets.py
xadupre: I hesitated. I used list only.

xadupre: done

onnxscript/rewriter/llama_rule_sets.py
xadupre: I prefer to keep the same prefix for the three functions. There will be 20 functions in this file. It seems better to keep the same prefix for the three functions related to the same rule. I wanted to group them in a class but I was sure onnxscript would allow that.

onnxscript/rewriter/llama_rule_sets.py
xadupre: Is there a place where I can document a rule? The documentation should be shared accross the three functions implementing a rule.

onnxscript/rewriter/llama_rule_sets.py
xadupre: I need a mechanism to let the rule prints out more information to help the investigation. I was hoping to connect this parameter to the existing mechanism if it exists.

onnxscript/rewriter/llama_rule_sets_tests.py
xadupre: Good catch. That explains why CI was not failing.

onnxscript/rewriter/llama_rule_sets_tests.py
xadupre: done

onnxscript/rewriter/llama_rule_sets_tests.py
xadupre: I made them as class method. I prefer to keep them here. It is close to the test calling the function so that I don't have to scroll too much.

onnxscript/rewriter/llama_rule_sets_tests.py
xadupre: There are other tests. I guess the parallelism will apply at test level. I usually keep this syntax because when it fails, it does not run the other models. But the main reason is I prefer to keep the model close to the test because I don't have to look for it.

onnxscript/rewriter/llama_rule_sets.py
xadupre: Where should I put this information, top of the module, inside the functions? It the documentation is published, that would help. For the time being, my goal to get better performance. The code may change.

onnxscript/rewriter/llama_rule_sets.py
xadupre: removed

onnxscript/rewriter/llama_rule_sets.py
xadupre: I'll do it once all patterns are done.

PR #1492, opened 2024-05-02
Add bfloat16 to float16 converter in onnxruntime specific rewriter

titaiwangms:
Previous to this PR, as numpy currently does not support bfloat16, onnxruntime python interface would not be able to execute bfloat16 input/output models. This PR provides temporary solution that convert model input/output to float16, and add op.Cast after/before them to enable it. The pass lives under onnxruntime rewriter defaulting off.

NOTE: We should delete this offline pass once we have bfloat16 supportingin numpy.
BLOCKED: onnxruntime doesn't seem to have implementation of bfloat16 on op.Add and op.Mul on both CUDA and CPU(see test case.)


onnxscript/rewriter/onnxruntime/__init__.py
```diff
@@ -27,6 +28,7 @@ def rewrite(
     /,
     function_rules: list[type[function_rule.FunctionRewriteRule]] | None = None,
     pattern_rules: list[pattern.RewriteRule] | None = None,
+    convert_bloat16_to_float16: bool = False,
```
justinchuby, line 31: I am conscious that the interface is changed. I personally don't think this api should be modified

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    if node.graph is None:
+        logger.warning("Node %s has no graph", node)
+        return
+    # Update graph/function outputs
+    for idx, graph_or_function_output in enumerate(node.graph.outputs):
+        if graph_or_function_output == node.outputs[index]:
+            node.graph.outputs[idx] = cast.outputs[0]
+
+
+def dtype_adapter_for_bfloat16_model(model: ir.Model) -> ir.Model:
```
justinchuby, line 62:
```suggestion
def dtype_adapter_for_bfloat16_model(model: ir.Model) -> None:
```

Model is updated in-place so we don't need to return, for clarity.

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _convert_inputs_from_bfloat16_to_float16(value: ir.Input):
+    if value.dtype == ir.DataType.BFLOAT16:
+        value.dtype = ir.DataType.FLOAT16
+        value_users = list(value.uses())
+        for node, index in value_users:
+            _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node, index)
```
justinchuby, line 13:
```suggestion
        for node, index in tuple(value.uses()):
            _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node, index)
```

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+        if node is None or index is None:
+            logger.warning("Output value %s has no producer or index", value)
+            return
+        _insert_cast_nodes_for_bfloat16_to_float16_to_outputs(node, index)
+
+
+def _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node: ir.Node, index: int):
+    attr = ir.AttrInt64(name="to", value=ir.DataType.BFLOAT16)
+    cast = ir.Node(
+        domain=node.domain,
```
justinchuby, line 29:
```suggestion
        domain="",
```

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    cast.outputs[0].dtype = ir.DataType.BFLOAT16  # type: ignore[union-attr]
+    cast.outputs[0].shape = node.inputs[index].shape  # type: ignore[union-attr]
+    node.prepend(cast)
+    node.replace_input_with(index, cast.outputs[0])
+
+
+def _insert_cast_nodes_for_bfloat16_to_float16_to_outputs(node: ir.Node, index: int):
+    attr = ir.AttrInt64(name="to", value=ir.DataType.FLOAT16)
+    cast = ir.Node(
+        domain=node.domain,
```
justinchuby, line 44:
```suggestion
        domain="",
```

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+        op_type="Cast",
+        inputs=[node.outputs[index]],
+        num_outputs=1,
+        attributes=[attr],
+    )
+    cast.outputs[0].dtype = ir.DataType.FLOAT16
```
justinchuby, line 50: Do we need to provide an option for users to choose float16 vs float32?

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    cast.outputs[0].shape = node.outputs[index].shape
+    node.append(cast)
+    if node.graph is None:
+        logger.warning("Node %s has no graph", node)
+        return
```
justinchuby, line 55: Node.append should check for this already.

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    """Adapt the model datatype if it's bfloat16.
+
+    Because onnxruntime does not support bfloat16 as input/output datatype, we need to
+    convert the bfloat16 datatype to float16. This function will convert the bfloat16
+    datatype to float16 and insert Cast nodes to convert float16 to bfloat16.
+
+    Model:
+        inputs(float16) -> Cast(bfloat16) -> nodes(bfloat16) -> Cast(float16) -> outputs(float16)
+
+    TODO: Delete this function after onnxruntime supports bfloat16.
```
justinchuby, line 72: Probably useful to offer just as a small tool

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+def _convert_inputs_from_bfloat16_to_float16(value: ir.Input):
+    if value.dtype == ir.DataType.BFLOAT16:
+        value.dtype = ir.DataType.FLOAT16
+        value_users = list(value.uses())
+        for node, index in value_users:
+            _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node, index)
+
+
+def _convert_outputs_from_bfloat16_to_float16(value: ir.Value):
+    if value.dtype == ir.DataType.BFLOAT16:
```
justinchuby, line 17:
```suggestion
    if value.dtype != ir.DataType.BFLOAT16:
        return
```

negate and return early to reduce nest level

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    attr = ir.AttrInt64(name="to", value=ir.DataType.FLOAT16)
+    cast = ir.Node(
+        domain=node.domain,
+        op_type="Cast",
+        inputs=[node.outputs[index]],
+        num_outputs=1,
+        attributes=[attr],
+    )
+    cast.outputs[0].dtype = ir.DataType.FLOAT16
+    cast.outputs[0].shape = node.outputs[index].shape
```
justinchuby, line 51: We may need to retain the output names. Potentially rename the original value so that it doesn't collide?

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    cast = ir.Node(
+        domain=node.domain,
+        op_type="Cast",
+        inputs=[node.inputs[index]],
+        num_outputs=1,
+        attributes=[attr],
+    )
+    cast.outputs[0].dtype = ir.DataType.BFLOAT16  # type: ignore[union-attr]
+    cast.outputs[0].shape = node.inputs[index].shape  # type: ignore[union-attr]
+    node.prepend(cast)
```
justinchuby, line 37: nice!

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
titaiwangms: This is for MYPY. Do you have another way around it?

justinchuby: Just do an assert?

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,84 @@
...
+    if value.dtype == ir.DataType.BFLOAT16:
+        node = value.producer()
+        index = value.index()
+        if node is None or index is None:
+            logger.warning("Output value %s has no producer or index", value)
+            return
+        _insert_cast_nodes_for_bfloat16_to_float16_to_outputs(node, index)
+
+
+def _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node: ir.Node, index: int):
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
titaiwangms: Changed it to tuple

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
titaiwangms: Done

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
titaiwangms: Done

onnxscript/rewriter/onnxruntime/__init__.py
titaiwangms: I am not sure how this tool is going to be used actually. @thiagocrepaldi might have an idea.

titaiwangms: But this is indeed not a long term argument. Maybe we should hide it.

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
titaiwangms: Swap the output value names you meant?

justinchuby: > Swap the output value names you meant?

That works

titaiwangms: What difference does it make? The newly created node is not the same naming authority as the graph? If that's the case, Isn't swapping making no difference in terms of the duplication of names in the same graph?

justinchuby: So I was thinking the output names need to be preserved. For example, consider

```
v_n -> output 
```

where output is graph.output[0]. Then after adding cast we have

```
v_n -> output -> v0
```

where v0 is the name of the new graph output

we should now rename "v0" to "output", so that the graph output name stays unchanged. Like you suggested swapping the names may be sufficient:

```
v_n -> v0 -> output
```

otherwise we can also rename output->output_bfloat16 and v0->output

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -57,6 +57,12 @@ def _insert_cast_nodes_for_bfloat16_to_float16_to_outputs(node: ir.Node, index:
     for idx, graph_or_function_output in enumerate(node.graph.outputs):
         if graph_or_function_output == node.outputs[index]:
             node.graph.outputs[idx] = cast.outputs[0]
+    # Swap the output name of the node with the output name of the cast node to
+    # preserve the output name in the graph
+    node.outputs[index].name, cast.outputs[0].name = (
```
justinchuby, line 62: One caveat is if the incoming graph is produced by IR and contains names like val_0; we would have a name collision since we don't currently ensure name uniqueness in the IR, and the new output's name will again be val_0; so _maybe_ adding the _bfloat16 suffix can be a better solution for now; the alternative is to create a pass to reconcile name collisions; or improve the naming authority to produce unique names

titaiwangms: Maybe I should use graph.insert_before/after? Those should reasonably cover the same naming authority?

justinchuby: prepend and append calls the graphs methods internally so they should be the exact same.

titaiwangms: I see. I thought naming authority lived inside the graph, and managed them not to collide.

justinchuby: > I see. I thought naming authority lived inside the graph

this is true

> and managed them not to collide.

This is not implemented yet

onnxscript/rewriter/onnxruntime/__init__.py
gramalingam: Yes, we need a better strategy here ... especially for utils like this. How is this intended to be used? Do we expect the torch exporter to invoke this at export time, or will it be done by the users separately?

Here, one option is to introduce the notion of a pass (which is essentially an in-place transformer of an ir.Model), and expose this conversion-utility as a pass, and allow the input to simply be a list of passes. We should be able to create a pass out of a list of rewrite-rules (of different kinds). Eventually, this and the optimizer API should be unified.

gramalingam: For now, exposing as a separate utility called by user might be okay, rather than changing this API frequently?

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -0,0 +1,92 @@
...
+CREATED_CAST_BFLOAT16_NAME_SUFFIX = "_cast_bfloat16"
+
+
+def _convert_inputs_from_bfloat16_to_float16(value: ir.Input) -> None:
+    if value.dtype != ir.DataType.BFLOAT16:
+        return
+    value.dtype = ir.DataType.FLOAT16
+    value_users = tuple(value.uses())
+    for node, index in value_users:
+        _insert_cast_nodes_for_float16_to_bfloat16_to_inputs(node, index)
```
gramalingam, line 15: If the var is used in 100 nodes, we don't want to create 100 cast nodes? We should create 1 cast node, and replace all uses of original_var by cast_var.

titaiwangms: Make sense

justinchuby (approved):
lgtm with api change to rewriter undone. Thanks!

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -10,38 +10,47 @@ def _convert_inputs_from_bfloat16_to_float16(value: ir.Input) -> None:
...
     )
-    cast.outputs[0].dtype = ir.DataType.BFLOAT16  # type: ignore[union-attr]
-    cast.outputs[0].shape = node.inputs[index].shape  # type: ignore[union-attr]
-    node.prepend(cast)
-    node.replace_input_with(index, cast.outputs[0])
+    cast.outputs[0].dtype = ir.DataType.BFLOAT16
+    cast.outputs[0].shape = value.shape
+
+    for node, index in tuple(value.uses()):
+        if node == cast:
```
justinchuby, line 37:
```suggestion
        if node is cast:
```

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_bfloat16_converted_model_runtime(self):
+        model_proto = ir.serde.serialize_model(self.model)
+        model_proto_filled_shape_type = onnx.shape_inference.infer_shapes(
+            model_proto, check_type=True, strict_mode=True, data_prop=True
+        )
+        onnx.checker.check_model(model_proto_filled_shape_type, full_check=True)
+        try:
+            ort_session = onnxruntime.InferenceSession(
+                model_proto_filled_shape_type.SerializeToString()
```
fatcat-z, line 78: We need to specify providers for InferenceSession() method.

titaiwangms: https://github.com/microsoft/onnxscript/pull/1513

PR #1495, opened 2024-05-03
Unify the two pattern-matchers

gramalingam:
Update generic pattern matcher to support the unified API.

* Update generic_pattern_test cases to use unified API.
* Update the generic_pattern implementation.

Still todo:
* generic_pattern doesn't support attributes yet
* pattern hasn't been updated yet to support extra first-parameter for pattern-builder and checker-function
* unify PatternMatchResult and MatchResult
* More redundancies to be eliminated between the two matchers. Eg., generic-pattern doesn't yet update opset-imports (unlike pattern)


onnxscript/rewriter/pattern.py
```diff
@@ -621,10 +623,17 @@ def _make_node(self, op_type: str, inputs: Sequence[ir.Value], kwargs: dict[str,
             assert isinstance(outputs, int)
             num_outputs = outputs
         if num_outputs == 1:
-            return self._tape.op(op_type, inputs=inputs, attributes=kwargs, domain=domain)
-        return self._tape.op_multi_output(
+            value = self._tape.op(op_type, inputs=inputs, attributes=kwargs, domain=domain)
+            if isinstance(outputs, Sequence):
+                value.name = outputs[0]
+            return value
+        values = self._tape.op_multi_output(
```
titaiwangms, line 630: Q: Do we have multi-outputs with pattern.py now? If so, should we remove generic_pattern.py to avoid the potential confusion that users might have?

onnxscript/rewriter/generic_pattern.py
```diff
@@ -46,7 +42,7 @@ def __init__(
         self.pattern_outputs = pattern_outputs
         self.kwargs: dict[str, Any] = {}
 
-        matched_pattern_to_model_value: dict[ir.Value, ir.Value] = {}
+        matched_pattern_to_model_value: dict[str, ir.Value] = {}
```
justinchuby, line 45: Is it possible when the name is None?

onnxscript/rewriter/generic_pattern.py
```diff
@@ -832,68 +679,75 @@ class FunctionPattern(GenericPattern):
...
         return self.match_pattern
 
 
+def _build_pattern(match_pattern_function: Callable) -> ir.Graph:
+    kwargs = {}
+    args = []
+
+    # There should be a better way.
+    sig = inspect.signature(match_pattern_function)
+    for i, p in enumerate(sig.parameters.values()):
```
justinchuby, line 701:
```suggestion
    for i, parameter in enumerate(sig.parameters.values()):
```

justinchuby (approved):

onnxscript/rewriter/generic_pattern.py
gramalingam: Good question. For now, there is an assert. May be better to make it conditional update to map. Eventually, this should be replaced by the pattern IR (instead of onnx IR).

titaiwangms (approved):

onnxscript/rewriter/generic_pattern.py
gramalingam: In the end, we care about binding only users want, for which they will give a name

onnxscript/rewriter/pattern.py
gramalingam: No. This is different (it's about creating nodes with multiple outputs).

PR #1508, opened 2024-05-07
[IR] Implement to_proto and from_proto convenience functions

justinchuby:
* __->__ #1508


titaiwangms (commented):
I feel they are overlapped ai with deserialize/serialize

justinchuby:
This is a convenience for users to not have to decide which ser/de function to use. Any suggestions?

titaiwangms:
> This is a convenience for users to not have to decide which ser/de function to use. Any suggestions?

Deprecate deserialize/serialize to avoid confusion? Having duplicated api seems weird.

titaiwangms (approved):
I think users would reference on how we use these api to start with. I suggest we keep one?

justinchuby:
> > This is a convenience for users to not have to decide which ser/de function to use. Any suggestions?
> 
> Deprecate deserialize/serialize to avoid confusion? Having duplicated api seems weird.

But we still need the implementation somewhere. to/from_proto will pick a ser/de function. I agree we don't prefer duplication. Maybe we don't have to create to/from_proto?

titaiwangms:
Maybe from original onnx user perspective, they would prefer apis with proto in names, so this is not a bad idea after all.
cc @gramalingam

justinchuby:
@gramalingam I will merge if looks good to you?

shubhambhokare1 (approved):
lgtm! 
nit: Are these functions supposed to be for internal use? If these are to be exposed to user maybe adding a couple unit tests could be a good idea?

justinchuby:
> lgtm! nit: Are these functions supposed to be for internal use? If these are to be exposed to user maybe adding a couple unit tests could be a good idea?

Yes, will do, thanks!

Issue #1426, opened 2024-04-23
[IR] Handle unsorted nodes


justinchuby:
Handle unsorted nodes during deserialization from proto

PR #1509, opened 2024-05-07
[IR] Support deserializing unsorted graphs

justinchuby:
* #1510
* #1508
* __->__ #1509

When a graph is unsorted, we initialize the node input value first and add it to the value pull. Producer nodes of this value will find it from the pool when the node is initialized later, with a warning message.

```
Input 'val_1' of node 'node_1(::Node1:)' not found in any scope. The graph may be unsorted. Creating a new input (current depth: 1) .
```

We needed this feature because we also need to be able to convert invalid models into IR and be also to fix them.

Fix https://github.com/microsoft/onnxscript/issues/1426


titaiwangms (approved):
Temp workaround because we don't have sorting yet?

justinchuby:
> Temp workaround because we don't have sorting yet?

We needed this feature because we also need to be able to convert invalid models into IR and be also to fix them. Added the explanation to PR description.

PR #1513, opened 2024-05-07
Add provider argument to bfloat16 converter test

titaiwangms:
In test, adding missed provider information into ORT inference session.


justinchuby (approved):
I thought that's the default but lgtm. Do you know what the default is?

titaiwangms:
> I thought that's the default but lgtm. Do you know what the default is?

I guess it's just being specific.

https://github.com/microsoft/onnxruntime/blob/aff04ba08a1c223aa7ed7d62335073551ea726f3/onnxruntime/python/onnxruntime_inference_collection.py#L369-L372

PR #1515, opened 2024-05-08
Unify single-output and multi-output pattern matchers

gramalingam:
* Migrate all rewrite-rule functions to uniformly use an extra first-parameter "op/context".
* Align and factor out common logic of the two rewrite-algorithms. Now, they differ only in the core-matching algorithm, and share all the rest.


onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +169,43 @@ class OpPatternBuilder:
...
+        if "version" in kwargs:
+            raise ValueError(
+                "The pattern builder does not support 'version' keyword argument."
+            )
+        if "domain" in kwargs:
+            domain = kwargs.pop("domain")
+            if isinstance(domain, str):
+                opset_pattern = OpsetPatternBuilder(domain)
+            else:
+                raise ValueError("domain must be a string.")
```
justinchuby, line 190:
```suggestion
                raise TypeError("domain must be a string.")
```

onnxscript/rewriter/pattern.py
```diff
@@ -504,13 +572,44 @@ def commute(self) -> list[ValuePattern]:
...
+    node_patterns.reverse()
+    return node_patterns
+
+
+def _compute_uses(nodepatterns: Sequence[NodePattern]) -> None:
+    """Compute the uses of values in the pattern."""
+    for nodepattern in nodepatterns:
+        for index, value in enumerate(nodepattern.inputs):
+            if value is not None:
+                value.uses().append((nodepattern, index))
```
justinchuby, line 597: mutating uses() seems a little strange...

onnxscript/rewriter/pattern.py
```diff
@@ -504,13 +572,44 @@ def commute(self) -> list[ValuePattern]:
...
 
-    def __init__(self, outputs: Sequence[ValuePattern]) -> None:
+    def __init__(
+        self, inputs: Sequence[ValuePattern], outputs: Sequence[ValuePattern]
+    ) -> None:
+        self.inputs = inputs
         self.outputs = outputs
         if len(outputs) == 0:
             raise ValueError("GraphPattern must have at least one output")
+        self.nodes = _nodes_in_pattern(outputs)
```
justinchuby, line 610: Thoughts on making as many attributes private / immutable? I assume the patterns are not to be mutated once created?

onnxscript/rewriter/pattern.py
```diff
@@ -659,6 +778,10 @@ class ReplacementSubgraph:
     used_opsets: UsedOpsets
 
 
+def always_true(*args, **kwargs) -> bool:
```
justinchuby, line 781: Helpful to add a docstring on when it should be used?

onnxscript/rewriter/pattern.py
```diff
@@ -694,12 +817,57 @@ def _update_opset_imports(
             )
 
 
+class PatternMatcher(ABC):
```
justinchuby, line 820:
```suggestion
class PatternMatcher(abc.ABC):
```

onnxscript/rewriter/pattern.py
```diff
@@ -4,9 +4,11 @@
 import inspect
 import itertools
 import math
+from abc import ABC, abstractmethod
```
justinchuby, line 7:
```suggestion
import abc
```

nit

onnxscript/rewriter/generic_pattern.py
```diff
@@ -102,61 +82,13 @@ def _to_match_result(pmr: PatternMatchResult) -> orp.MatchResult:
     result = orp.MatchResult(success=True)
     result.nodes.extend(pmr.model_nodes)
     for var, val in pmr.matched_pattern_to_model_value.items():
-        result.bind(var, val)
-    result.outputs.extend(
-        [pmr.matched_pattern_to_model_value[v.name] for v in pmr.pattern_outputs]
-    )
+        if var.name is not None:
+            result.bind(var.name, val)
+    result.outputs.extend([pmr.matched_pattern_to_model_value[v] for v in pmr.pattern.outputs])
```
titaiwangms, line 87: What happens when var.name is None?

onnxscript/rewriter/generic_pattern.py
```diff
@@ -247,8 +182,9 @@ def none(
                     f"{os.path.split(self.__class__.__module__)[-1]}, "
                     f"op_type={node.op_type}{msg}{msg2}"
                 )
+        return None
 
-    def print_match(self, n1: ir.Node, n2: ir.Node) -> str:
+    def print_match(self, n1: ir.Node, n2: orp.NodePattern) -> str:
```
titaiwangms, line 187: Maybe change n2 to something else, as now it's not ir.Node

onnxscript/rewriter/generic_pattern.py
```diff
@@ -527,47 +472,50 @@ def _match_forward(
...
         if self.verbose > 5 and match_count > 0:
-            print(f"[GenericPattern._match_forward] add {match_count} nodes")
+            print(f"[GenericPatternMatcher._match_forward] add {match_count} nodes")
         return match_count
 
     def match(
         self,
-        g: ir.Graph | ir.GraphView,
+        model: ir.Model,
+        graph: ir.Graph | ir.Function,
```
titaiwangms, line 481: nit: Use `graph_or_function` to maintain the convention naming in rewriter?

onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +170,43 @@ class OpPatternBuilder:
...
         self.opset_pattern = opset_pattern
-        self.op_name_pattern = op_name_pattern
+        self.op_name = op_name
 
     def __call__(self, *args, **kwargs):
-        # TODO(rama): Unify with convention used elsewhere.
         if "_num_outputs" in kwargs:
-            num_outputs = kwargs["_num_outputs"]
-            del kwargs["_num_outputs"]
+            raise ValueError("Use 'outputs' instead of '_num_outputs'.")
```
titaiwangms, line 181: Is this for future usage? I don't see outputs is used in any case?

onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +170,43 @@ class OpPatternBuilder:
...
+        if "version" in kwargs:
+            raise ValueError(
+                "The pattern builder does not support 'version' keyword argument."
+            )
+        if "domain" in kwargs:
```
titaiwangms, line 186: Should we define the expected kwargs on top of this function or class as a note/comment? It seems like there are multiple kwargs could be passed in.

onnxscript/rewriter/pattern.py
```diff
@@ -308,6 +324,14 @@ class ValuePattern:
 
     def __init__(self, name: str | None) -> None:
         self.name = name
+        # Note: uses will be computed only when the full graph-pattern is constructed.
+        self._uses: list[tuple[NodePattern, int]] = []
+
+    def producer(self) -> None | NodePattern:
+        return None
```
titaiwangms, line 331: I wonder if this raises formatting error.

onnxscript/rewriter/pattern.py
```diff
@@ -365,18 +389,56 @@ class NodePattern:
...
 
-    def matches_node(self, node: ir.Node) -> MatchResult:
-        """Examine if the IR node matches the self pattern."""
+    def op_identifier(self) -> Tuple[str, str, str]:
+        return self._op_identifier
+
+    @property
+    def op_type(self) -> str:
+        if self._op_identifier is not None:
+            return self._op_identifier[1]
```
titaiwangms, line 417: Should we just return self.op?

onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +170,43 @@ class OpPatternBuilder:
...
-        opset_pattern: Pattern[str],
-        op_name_pattern: Pattern[str],
+        opset_pattern: OpsetPatternBuilder,
+        op_name: str | Pattern[str],
     ) -> None:
         self.opset_pattern = opset_pattern
-        self.op_name_pattern = op_name_pattern
+        self.op_name = op_name
 
     def __call__(self, *args, **kwargs):
```
justinchuby, line 179:
```suggestion
    def __call__(self, *args, domain: str | None = None, **kwargs):
```

onnxscript/rewriter/pattern.py
justinchuby: I think we can capture it explicitly. Commented above

onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +170,43 @@ class OpPatternBuilder:
...
+        if "domain" in kwargs:
+            domain = kwargs.pop("domain")
+            if isinstance(domain, str):
+                opset_pattern = OpsetPatternBuilder(domain)
+            else:
+                raise TypeError("domain must be a string.")
         else:
-            num_outputs = 1
+            opset_pattern = self.opset_pattern
+        outputs = kwargs.pop("outputs", 1)
```
justinchuby, line 194: Define known kwargs explicitly?

onnxscript/rewriter/pattern.py
```diff
@@ -169,28 +170,43 @@ class OpPatternBuilder:
...
+        opset_pattern: OpsetPatternBuilder,
+        op_name: str | Pattern[str],
     ) -> None:
         self.opset_pattern = opset_pattern
-        self.op_name_pattern = op_name_pattern
+        self.op_name = op_name
 
     def __call__(self, *args, **kwargs):
-        # TODO(rama): Unify with convention used elsewhere.
         if "_num_outputs" in kwargs:
```
justinchuby, line 180: Helpful to a todo to remove this check, maybe after the deprecation period is passed?

onnxscript/rewriter/generic_pattern.py
gramalingam: Consider an intermediate statement in a pattern like below
```py
     XY = op.MatMul (X, Y) 
```
The user will not have access to this intermediate value `XY` in their condition-function for validating the match. But if they want access to it, they can indicate that by specifying an output-name for this as below:
```py
     XY = op.MatMul (X, Y, output=["XY"]) 
```
Then, they be able to access it in their condition function as below:
```py
   def check_condition(context, XY, ...):
```
In the first case, var.name is None in the pattern, and we will not bind any value in the match-binding we create. In the second case, we will.

gramalingam: @shubhambhokare1 : this is worth documenting eventually, once the PR is merged.

onnxscript/rewriter/pattern.py
gramalingam: It is used in the examples/test-cases in generic_pattern_test.

onnxscript/rewriter/pattern.py
gramalingam: Agree with the suggestion to capture it explicitly. I think it is also important to describe it in the documentation of the rewriter. The users will be those who write rewrite-rules, who may not be looking at this class, which is a builder class.

A related question (this is the same convention/trick used in _tape too): these names can clash with attribute-names used in ONNX spec. I wonder if we should use names like `_domain` which are less likely to clash with ONNX attribute-names?

onnxscript/rewriter/pattern.py
gramalingam: What formatting error? If you mean the type-signature being more general, it is to reminder the reader that in derived class a NodePattern may be returned.

onnxscript/rewriter/pattern.py
gramalingam: No ... remember that we can even specify that a pattern-node can match any node whose op has "Mat" as a prefix. So, the pattern's op is a predicate (an object with a `match` method). The op-identifier has the opname, when it is available.

onnxscript/rewriter/pattern.py
gramalingam: I think I should remove it ... it was just for my own reminder during the refactoring.

onnxscript/rewriter/pattern.py
gramalingam: Yes, I should clean this up!

onnxscript/rewriter/pattern.py
gramalingam: I guess I could add an append_use method ... we can construct the uses only finally, after whole pattern-graph has been constructed.

onnxscript/rewriter/pattern.py
titaiwangms: I personally think that _domain is better. _ also warns users that they are modifying something important. But I don't have strong preference.

onnxscript/rewriter/pattern.py
titaiwangms: I see. I didn't see NodeOutputPattern is also using this.

onnxscript/rewriter/pattern.py
titaiwangms: I still don't get it. I thought domain is the only arg that we support for prefix? Could you share me the example pattern/test?

onnxscript/rewriter/pattern.py
justinchuby: Just ideating: can we make op a callable that creates a partially applied function? E.g.

```
op(domain="com.microsoft").Gelu(...)
```

but I see we also have `outputs`, which doesn't fit the semantics of parameterizing op

onnxscript/rewriter/pattern.py
gramalingam: Good question ... you are right, we can keep updating it, as and when new nodes are added, but we won't have it when the value is created first (so, I guess that part is common). Right now, uses() is computed finally ... we could change it to do it incrementally when nodes are added. Let me think. This is not required by the single-output-matcher, only by the multi-output matcher.

gramalingam: Done

onnxscript/rewriter/pattern.py
gramalingam: The example is the one using [submodule](https://github.com/microsoft/onnxscript/blob/b4dd7774f166e46984aebc6d3626cd5edae3dcd5/onnxscript/rewriter/onnxruntime/group_normalization_merge_silu.py#L31) ... I think you added it?

justinchuby (approved):

onnxscript/rewriter/pattern.py
```diff
@@ -504,13 +585,35 @@ def commute(self) -> list[ValuePattern]:
         return [self]
 
 
+def _nodes_in_pattern(outputs: Sequence[ValuePattern | None]) -> list[NodePattern]:
+    """Computes inputs and nodes from outputs of a pattern."""
```
justinchuby, line 589: What does "computes" refer to? Does it mean "find"

onnxscript/rewriter/pattern.py
```diff
@@ -504,13 +585,35 @@ def commute(self) -> list[ValuePattern]:
...
+    node_patterns: list[NodePattern] = []
+
+    def visit(value_patterns: Sequence[ValuePattern]) -> None:
+        for value_pattern in value_patterns:
+            if isinstance(value_pattern, NodeOutputPattern):
+                node_pattern = value_pattern.producer()
+                if node_pattern not in node_patterns:
+                    node_patterns.append(node_pattern)
+                    visit(node_pattern.inputs)
```
justinchuby, line 598: Would there be stack overflow?

onnxscript/rewriter/pattern.py
```diff
@@ -523,19 +626,35 @@ def __init__(self, outputs: Sequence[ValuePattern]) -> None:
...
+    def __len__(self) -> int:
+        return len(self.nodes)
+
+    def __iter__(self) -> Iterator[NodePattern]:
+        return iter(self.nodes)
+
+    def __reversed__(self) -> Iterator[NodePattern]:
+        return reversed(self.nodes)
+
+    @property
```
justinchuby, line 646: Should this be a method instead?

onnxscript/rewriter/pattern.py
titaiwangms: I see. I forgot op_type is PrefixPattern. I thought this was the same op_type we have in ir.Node.

onnxscript/rewriter/pattern.py
gramalingam: No cycles are possible. Pattern lengths are not expected to be extremely long. Are you recommending a worklist based algorithm?

PR #1516, opened 2024-05-08
[IR] Implement `node` and `num_nodes` on Graph

justinchuby:
- `node()` to get node by index or name
- `num_nodes()` to obtain the node counts


titaiwangms (commented):
Should we also have a method to overview what nodes we have in the graph, so that we will know which node we are aiming to? I usually need to iterate it in pdb, and then figure out which node I am looking for.

gramalingam (approved):

justinchuby:
> Should we also have a method to overview what nodes we have in the graph, so that we will know which node we are aiming to? I usually need to iterate it in pdb, and then figure out which node I am looking for.

I think that can be a util somewhere outside

titaiwangms (approved):

PR #1521, opened 2024-05-10
Update example

gramalingam:
The example in `examples\pattern_rewriting.py` was not updated when the pattern-matchers API were unified. Fix this.

Question: do we want this standalone example? I think the documentation folder examples are a better place to add anything we want, and this can be removed. (But leaving it here for now.)

TODO: Improve the logging info produced in verbose mode, especially for the pattern graph.

### DIFF
--- a/examples/pattern_rewriting.py
+++ b/examples/pattern_rewriting.py
@@ -13,7 +13,6 @@ import onnx
 import onnx.helper as oh
 import onnx.numpy_helper as onh
 
-import onnxscript
 from onnxscript import ir
 from onnxscript.rewriter import generic_pattern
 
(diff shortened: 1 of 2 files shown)


justinchuby:
>  I think the documentation folder examples are a better place to add anything

I think so too!

examples/pattern_rewriting.py
```diff
@@ -67,45 +67,27 @@ def get_rotary_model(bad_model=False):
...
-def rotary_match_pattern(x, pos_ids, axis):
+def rotary_match_pattern(op, x, pos_ids, axis):
     """The pattern to match."""
     unsqueeze = op.Unsqueeze(x, axis)
     cast = op.Cast(unsqueeze, to=onnx.TensorProto.FLOAT)
 
     matmul = op.MatMul(pos_ids, cast)
     transpose = op.Transpose(matmul)
-    output, length = msft_op.ConcatTraining(transpose, transpose)
+    output, length = op.ConcatTraining(transpose, transpose, domain="com.microsoft", outputs=2)
```
justinchuby, line 78: The IR uses `outputs` to accept a list of Values and num_outputs for an int. Thoughts on consistency?

gramalingam: Hmmm ... I used to have `_num_outputs` (in building a pattern) ... I switched to `outputs` to match what was there (after the update to generic-matcher to use the new IR ... not sure what the origin of this was)!

I think that the overloaded use of `outputs` to denote either the number of outputs or names of outputs is reasonable ... we could use distinct names too, but then would have to check that they are not used inconsistently. Typically, we expect only one or the other to be used.

But whatever we decide, agree we should have a consistent one.

justinchuby (approved):

PR #1523, opened 2024-05-10
Minor cleanup for multi-output-matcher verbose trace output

gramalingam:
Add `__str__` methods to pattern objects and use them in the multi-output-matcher trace output.

### DIFF
--- a/onnxscript/rewriter/generic_pattern.py
+++ b/onnxscript/rewriter/generic_pattern.py
@@ -88,6 +88,26 @@ def _to_match_result(pmr: PatternMatchResult) -> orp.MatchResult:
     return result
 
 
+def _value_to_str(value: ir.Value | orp.ValuePattern) -> str:
+    return value.name if value.name is not None else "anonymous:" + str(id(value))
+
+
+def _opt_value_to_str(value: ir.Value | orp.ValuePattern | None) -> str:
+    return _value_to_str(value) if value is not None else "None"
+
+
+def _node_to_str(node: ir.Node) -> str:
+    inputs = ", ".join(_opt_value_to_str(input) for input in node.inputs)
+    outputs = ", ".join(_opt_value_to_str(output) for output in node.outputs)
+    return f"{outputs} = {node.op_type}({inputs})"
+
+
+def _pattern_node_to_str(node: orp.NodePattern) -> str:
+    inputs = ", ".join(_opt_value_to_str(input) for input in node.inputs)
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/generic_pattern.py
```diff
@@ -88,6 +88,26 @@ def _to_match_result(pmr: PatternMatchResult) -> orp.MatchResult:
...
+
+
+def _opt_value_to_str(value: ir.Value | orp.ValuePattern | None) -> str:
+    return _value_to_str(value) if value is not None else "None"
+
+
+def _node_to_str(node: ir.Node) -> str:
+    inputs = ", ".join(_opt_value_to_str(input) for input in node.inputs)
+    outputs = ", ".join(_opt_value_to_str(output) for output in node.outputs)
+    return f"{outputs} = {node.op_type}({inputs})"
```
justinchuby, line 102: Do we care about domains etc?

onnxscript/rewriter/generic_pattern.py
```diff
@@ -88,6 +88,26 @@ def _to_match_result(pmr: PatternMatchResult) -> orp.MatchResult:
...
 
+def _value_to_str(value: ir.Value | orp.ValuePattern) -> str:
+    return value.name if value.name is not None else "anonymous:" + str(id(value))
+
+
+def _opt_value_to_str(value: ir.Value | orp.ValuePattern | None) -> str:
+    return _value_to_str(value) if value is not None else "None"
+
+
+def _node_to_str(node: ir.Node) -> str:
```
justinchuby, line 99:
```suggestion
def _node_to_short_str(node: ir.Node) -> str:
```

maybe something like this to distinguish from the native `__str__`?

justinchuby (approved):

onnxscript/rewriter/generic_pattern.py
gramalingam: Good point. I will update it ... though it may not matter much in a debugging context.

onnxscript/rewriter/generic_pattern.py
gramalingam: In a given context like this, it is (implicitly) the specialized representation desired by that context ... I think what you are suggesting would make sense if we try to share this, eg., move this to be a method into `ir.Node` as a `to_short_str()` method?

justinchuby: Sounds reasonable, thanks!

PR #1537, opened 2024-05-14
[IR] Improve name authority to generate unique names

### LINKED ISSUE #1535: [IR] Enhanced name authority
Make sure the new names generated do not collide with the existing in the graph. This can be done inexpensively by tracking all values that starts with `val_`, because this is the only thing the name authority can generate. The name authority still does not care about user specified names.

justinchuby:
- Store all names from the graph for generating unique names for new values.
- Also allow values to be initialized with no arguments.

Fix #1535

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1764,6 +1764,8 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def append(self, node: Node, /) -> None:
         """Append a node to the graph in O(1) time.
 
+        A unique name will be assigned to the node and its values is not yet specified.
+
         Args:
             node: The node to append.
 
@@ -1775,6 +1777,8 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
 
     def extend(self, nodes: Iterable[Node], /) -> None:
         """Extend the graph with the given nodes in O(#new_nodes) time.
+
+        A unique name will be assigned to the node and its values is not yet specified.
 
         Args:
             nodes: The nodes to extend the graph with.
@@ -1828,6 +1832,8 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def insert_after(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None:
         """Insert new nodes after the given node in O(#new_nodes) time.
 
+        A unique name will be assigned to the node and its values is not yet specified.
+
         Args:
             node: The node to insert after.
             new_nodes: The new nodes to insert.
@@ -1842,6 +1848,8 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
 
     def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None:
         """Insert new nodes before the given node in O(#new_nodes) time.
+
+        A unique name will be assigned to the node and its values is not yet specified.
 
         Args:
             node: The node to insert before.
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1565,9 +1565,7 @@ class Input(Value):
         type: _protocols.TypeProtocol | None = None,
         doc_string: str | None = None,
     ) -> None:
-        super().__init__(
-            None, index=None, name=name, shape=shape, type=type, doc_string=doc_string
-        )
+        super().__init__(name=name, shape=shape, type=type, doc_string=doc_string)
 
 
 def _check_node_safe_to_remove(
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1712,11 +1712,9 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
                 f"The node '{node!r}' belongs to another graph. Please remove it first with Graph.remove()."
             )
         # Give the node and its output values names if they don't not have one
-        if node.name is None:
-            self._name_authority.name_node(node)
+        self._name_authority.register_or_name_node(node)
         for value in node._outputs:  # pylint: disable=protected-access
-            if value.name is None:
-                self._name_authority.name_value(value)
+            self._name_authority.register_or_name_value(value)
         node.graph = self
         return node
 
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1764,7 +1764,7 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def append(self, node: Node, /) -> None:
         """Append a node to the graph in O(1) time.
 
-        A unique name will be assigned to the node and its values is not yet specified.
+        Unique names will be assigned to the node and its values if any name is ``None``.
 
         Args:
             node: The node to append.
@@ -1778,7 +1778,7 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def extend(self, nodes: Iterable[Node], /) -> None:
         """Extend the graph with the given nodes in O(#new_nodes) time.
 
-        A unique name will be assigned to the node and its values is not yet specified.
+        Unique names will be assigned to the node and its values if any name is ``None``.
 
         Args:
             nodes: The nodes to extend the graph with.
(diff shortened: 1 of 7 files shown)


onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,43 @@ class NameAuthority:
...
-    generate names that conflicts with existing names. It is the responsibility of the
-    user to ensure that the names are unique (typically by running a name-fixing pass
-    on the graph).
+    This class does keep track of the names it has given and all the existing names
+    in the graph to prevent producing duplicated names.
+
+    However, if a value/node is already named when added to the graph,
+    the name authority will not change its name.
+    It is the responsibility of the user to ensure that the names are unique
+    (typically by running a name-fixing pass on the graph).
```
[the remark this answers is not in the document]
justinchuby: Sure - we don't have this pass yet. We can add when it is created.

onnxscript/ir/_core.py
```diff
@@ -1845,6 +1849,8 @@ def insert_after(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None:
     def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None:
         """Insert new nodes before the given node in O(#new_nodes) time.
 
+        A unique name will be assigned to the node and its values is not yet specified.
```
shubhambhokare1, line 1852: Not quite sure what the (is not yet specified) part means.

onnxscript/ir/_name_authority.py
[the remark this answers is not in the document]
shubhambhokare1: Cool! Feel free to add a TODO there.

onnxscript/ir/_core.py
justinchuby: Thanks, updated to "Unique names will be assigned to the node and its values if any name is ``None``." Does that look better?

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,52 @@ class NameAuthority:
...
```
justinchuby, line 21:
```suggestion
    (typically by running a name-fixing pass on the graph).
    
    TODO(justichuby): Describe the pass when we have a reference implementation.
```

onnxscript/ir/_name_authority.py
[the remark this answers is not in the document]
justinchuby: Done

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,54 @@ class NameAuthority:
...
+
```
justinchuby, line 22:
```suggestion

```

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
```diff
@@ -61,9 +61,6 @@ def _insert_cast_nodes_for_bfloat16_to_float16_to_outputs(value: ir.Value) -> No
     )
     cast.outputs[0].dtype = ir.DataType.FLOAT16
     cast.outputs[0].shape = node.outputs[index].shape
-    # To prevent naming conflicts, we need to append suffix to the output name of the cast node
-    # TODO: Remove this after naming authority covers this case
-    cast.outputs[0].name = node.outputs[index].name + CREATED_CAST_BFLOAT16_NAME_SUFFIX  # type: ignore[operator]
```
titaiwangms, line 66: Could you also remove the varialble: CREATED_CAST_BFLOAT16_NAME_SUFFIX

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,53 @@ class NameAuthority:
...
+            if name not in self._node_names:
+                return name
 
-    def name_value(self, value: _core.Value) -> None:
-        value.name = f"val_{self._value_counter}"
-        self._value_counter += 1
+    def register_or_name_value(self, value: _core.Value) -> None:
+        # TODO(justinchuby): Record names of the initializers and graph inputs
+        if value.name is None:
+            value.name = self._unique_value_name()
```
titaiwangms, line 50: We need a demo/test for this. I suppose that one of the usage if that it happens after node is created without named inputs/outputs?

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,53 @@ class NameAuthority:
...
-        node.name = f"node_{node.op_type}_{self._node_counter}"
-        self._node_counter += 1
+    def register_or_name_node(self, node: _core.Node) -> None:
+        if node.name is None:
+            node.name = self._unique_node_name(node.op_type)
+        # If the name is already specified, we do not change it because keeping
+        # track of the used names can be costly when nodes can be removed from the graph:
+        # How do we know if a name is no longer used? We cannot reserve unused names
+        # because users may want to use them.
+        self._node_names.add(node.name)
```
titaiwangms, line 64: Add a test

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,53 @@ class NameAuthority:
...
     """
 
     def __init__(self):
         self._value_counter = 0
         self._node_counter = 0
+        self._value_names: set[str] = set()
+        self._node_names: set[str] = set()
+
+    def _unique_value_name(self) -> str:
+        """Generate a unique name for a value."""
```
gramalingam, line 32: Would it make sense to allow callers to specify a desired name, which can be used as a prefix (instead of `val`)? Or, does that functionality reside elsewhere? That is, if we have rewrite-rule that introduces names like `shape` and `dim`, how do we turn them into `shape_102` instead of `val_102`?

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,53 @@ class NameAuthority:
...
+        while True:
+            name = f"val_{self._value_counter}"
+            self._value_counter += 1
+            if name not in self._value_names:
+                return name
+
+    def _unique_node_name(self, op_type: str) -> str:
```
gramalingam, line 39: Can this just be any prefix that the user chooses? If so, we can rename this to be `preferred_name`. Trying to separate the mechanism here from the naming-policy (which a user might want to control). `node_` could be a default-value for this  prefix.

onnxscript/ir/_name_authority.py
justinchuby: Currently it’s not handled. Happy to explore options

onnxscript/ir/_core.py
```diff
@@ -1766,6 +1764,8 @@ def num_nodes(self) -> int:
     def append(self, node: Node, /) -> None:
         """Append a node to the graph in O(1) time.
 
+        Unique names will be assigned to the node and its values if any name is ``None``.
```
gramalingam, line 1767: Are these the only places where the naming happens? When a node is added into a graph? Just trying to get the big picture. It makes sense.

onnxscript/ir/_name_authority.py
```diff
@@ -12,20 +12,53 @@ class NameAuthority:
     ``node_{op_type}_{node_counter}`` for nodes. The counter is incremented each time
     a new value or node is named.
 
-    The class does not keep track of the names it has given, so it is possible to
-    generate names that conflicts with existing names. It is the responsibility of the
-    user to ensure that the names are unique (typically by running a name-fixing pass
-    on the graph).
+    This class does keep track of the names it has given and all the existing names
```
gramalingam, line 15: nit rewording suggestion: `This class keeps tracks of the names it has generated and existing names`

onnxscript/ir/_core.py
gramalingam: Does deserialization of a proto also call these node-insertion methods when creating the graph?

justinchuby: > Does deserialization of a proto also call these node-insertion methods when creating the graph?

Yes they do call

gramalingam (approved):

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter.py
justinchuby: Done

onnxscript/ir/_core.py
justinchuby: > Are these the only places where the naming happens? When a node is added into a graph? Just trying to get the big picture. It makes sense.

Yes, it only happens when a node is added into a graph currently

onnxscript/ir/_name_authority.py
justinchuby: Done

onnxscript/ir/_name_authority.py
justinchuby: From discussion, users can use a pass to reconcile conflicts if they decide to specify their own names, or they can build an auxiliary data structure to keep track of names with stronger assumptions, for now. We can evolve the mechanism later when needed.

onnxscript/ir/_name_authority.py
justinchuby: Done

onnxscript/ir/_name_authority.py
justinchuby: Done

PR #1538, opened 2024-05-14
[fix ci] Disable macos tests for native_layer_norm_float32

shubhambhokare1:


tests/function_libs/torch_lib/ops_test_common.py
```diff
@@ -56,7 +57,8 @@
 )
 
 TEST_OPSET_VERSION = 18
-IS_WINDOWS = os.name == "nt"
+IS_MACOS = sys.platform.startswith('darwin')
+IS_WINDOWS = sys.platform.startswith('win32')
```
justinchuby, line 61: I think the original check for windows is reliable; or are there exceptions not handled?

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -1932,6 +1932,10 @@ def _where_input_wrangler(
         core_ops.aten_native_layer_norm,
         trace_only=True,
         tolerance={torch.float32: (3.7e-5, 1.8e-4), torch.float16: (1e-1, 7e-4)},
+    ).xfail(
+        dtypes=(torch.float32,),
```
justinchuby, line 1936: I would check the input size. Only skip the size-1 inputs.

justinchuby:
Thanks for creating the fix! Minor comments

tests/function_libs/torch_lib/ops_test_common.py
shubhambhokare1: Had changed the format for the sake of uniformity. But the previous check was sufficient. Reverting back to that.

justinchuby (approved):
Thank you!

PR #1542, opened 2024-05-15
[IR] Fix broadcast_to_matmul

### LINKED ISSUE #1541: crash in rewriter/broadcast_to_matmul.py (proposed fix)
My model fails during torch.onnx.dynamo_export here with "AttributeError: 'NoneType' object has no attribute 'numpy'": 
https://github.com/microsoft/onnxscript/blob/fe9f29a61745962d18773e1d73d3751988fff8cd/onnxscript/rewriter/broadcast_to_matmul.py#L31
Obviously, "if shape_c is None:" check should have been before that line, not after. 
Here is a one-line diff that fixes the problem. Resulting ONNX then works and compares with PyTorch results:
 
```
--- broadcast_to_matmul.bak     2024-05-14 19:22:51.489899788 +0000
+++ broadcast_to_matmul.py      2024-05-14 19:49:17.761779925 +0000
@@ -28,9 +28,10 @@
     input_b_shape = input_b.shape
     # TODO: Get a helper func to get const_value
     shape_c_value = _ir_utils.propagate_const_value(shape_c)
-    shape_c = shape_c_value.const_value.numpy()  # type: ignore[union-attr]
+    shape_c = shape_c_value.const_value  # type: ignore[union-attr]
     if shape_c is None:
         return False
+    shape_c = shape_c.numpy()  # type: ignore[union-attr]
     if not isinstance(shape_c, np.ndarray):
         logger.info("Unexpected shape_c value. Expected np.ndarray, got %s", type(shape_c))
         return False
```

justinchuby:
- Check whether the shape tensor is constant before using it in the logic. Exiting early if needed.
- Handle cases when the input is 1d or 0d

Thanks @borisfom for the proposed fix!

Fix #1541


onnxscript/rewriter/broadcast_to_matmul.py
```diff
@@ -21,65 +22,70 @@ def check_if_not_need_reshape(context, input_a, input_b, shape_c, **_) -> bool:
...
     # NOTE: When there is a subset match with a pattern. The MatchResult won't have the shape
     # information. So, we need to check if the shape is None and return False.
-    if input_a_shape is None or input_b_shape is None or shape_c is None:
+    if input_a_shape is None or input_b_shape is None:
         logger.info("Shape information is not available for the inputs and outputs.")
         return False
-    input_a_shape = list(input_a_shape)
-    input_b_shape = list(input_b_shape)
+    input_a_shape = input_a_shape.numpy()
+    input_b_shape = input_b_shape.numpy()
```
justinchuby, line 51: Handle symbolic values

onnxscript/ir/_convenience.py
```diff
@@ -190,7 +190,7 @@ def convert_attributes(
         ...     "type_protos": [ir.TensorType(ir.DataType.FLOAT), ir.TensorType(ir.DataType.FLOAT)],
         ... }
         >>> convert_attributes(attrs)
-        [AttrInt64('int', 1), AttrFloat32('float', 1.0), AttrString('str', 'hello'), AttrInt64s('ints', [1, 2, 3]), AttrFloat32s('floats', [1.0, 2.0, 3.0]), AttrStrings('strings', ['hello', 'world']), AttrTensor('tensor', Tensor<DOUBLE,[3]>(array([1., 2., 3.]), name='')), AttrTensor('tensor_proto', TensorProtoTensor<FLOAT,[3]>(name='proto')), AttrInt64s('graph', Graph(
+        [AttrInt64('int', 1), AttrFloat32('float', 1.0), AttrString('str', 'hello'), AttrInt64s('ints', [1, 2, 3]), AttrFloat32s('floats', [1.0, 2.0, 3.0]), AttrStrings('strings', ['hello', 'world']), AttrTensor('tensor', Tensor<DOUBLE,[3]>(array([1., 2., 3.]), name=None)), AttrTensor('tensor_proto', TensorProtoTensor<FLOAT,[3]>(name='proto')), AttrInt64s('graph', Graph(
```
titaiwangms: What's the difference between None and '' in here?

titaiwangms: None is better for conditional check (if/else)?

onnxscript/ir/_core.py
```diff
@@ -1520,8 +1518,11 @@ def const_value(
     @const_value.setter
     def const_value(
         self,
-        value: _protocols.TensorProtocol | Sequence[_protocols.TensorProtocol] | None,
```
titaiwangms: Is Sequence[...] not necessary?

onnxscript/rewriter/_ir_utils.py
```diff
@@ -10,24 +12,41 @@
...
+                    np.array(attr_value.value, dtype=np.int64), name=ir_value.name
+                )
+            elif attr_name in {"value_string", "value_strings"}:
+                const_value = ir.StringTensor(
+                    np.array(attr_value.value, dtype=np.bytes_), name=ir_value.name
+                )
+            elif attr_name == "value":
+                const_value = typing.cast(ir.TensorProtocol, attr_value.value)
+
+            ir_value.const_value = const_value
```
titaiwangms, line 46: Should we reassign shape and dtype when value is assigned? Then we can avoid the situation that users forget to reassign shape/dtype when they update the const_value?

titaiwangms (approved):

onnxscript/rewriter/_ir_utils.py
justinchuby: Good idea

onnxscript/ir/_core.py
justinchuby: Using the sequence annotation made operations like value.const_value.name=… difficult because we would need to assert that const_value is not a sequence everytime. I am removing this for now and will consider better ways when we need const value for sequences

onnxscript/ir/_convenience.py
justinchuby: Since the the values can take None as name and we are assigning the same name to the tenor when we’re assigning names to the value I decided to support None as tensor names as well

onnxscript/rewriter/_ir_utils.py
justinchuby: But what happens when the original value has a symbolic shape? Do we want to narrow the symbols as concrete values when assigning constants?

onnxscript/ir/_core.py
```diff
@@ -1528,6 +1530,9 @@ def const_value(
                 raise TypeError(
                     f"Expected value to be a TensorProtocol or None, got '{type(value)}'"
                 )
+        if value is not None:
+            self.shape = value.shape
+            self.dtype = value.dtype
```
justinchuby, line 1535: Probably not possible because when we roundtrip from ONNX files, some models will have initializers but not ValueInfoProto for certain values. We need to respect the model file and not add valueinfo for these values. Probably good to leave it to some (micro) shape inference passes @titaiwangms

onnxscript/rewriter/broadcast_to_matmul.py
justinchuby: Done

onnxscript/rewriter/_ir_utils.py
justinchuby: Reverted the change because initializers may not have the value information when deserializing, but this change will add value information to the initializer ir.Value objects, breaking the serde roundtrip. We can think about a better way of doing this.

PR #1547, opened 2024-05-15
Add unique op

a-gardner1:
Add support for exporting `torch.unique` following the conclusion of https://github.com/pytorch/pytorch/issues/113118.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8359,6 +8360,63 @@ def aten_unique_consecutive(
...
+        if dim is None:
+            unique_values, inverse_indices, counts = aten_unique2(self, sorted, return_inverse, return_counts)
+        else:
+            unique_values, inverse_indices, counts = aten_unique_dim(self, dim, sorted, return_inverse, return_counts)
+    except Exception as e:
+        # try to provide a more informative error message
+        if _NOT_IMPLEMENTED_UNIQUE.search(str(e)) is not None:
+            raise NotImplementedError(
+                f"'onnxruntime' does not yet support Unique(11) operator with dtype={self.dtype}'"
+                ) from e
```
justinchuby, line 8390: I would remove this try-catch as the function here is symbolic

justinchuby (commented):
Thanks for your contribution! Could you follow the CLA bot's instruction to get that cleared?

a-gardner1:
> Thanks for your contribution! Could you follow the CLA bot's instruction to get that cleared?

Yea, I may have jumped the gun a bit. Working on officially getting permission from my employer.

a-gardner1:
> @a-gardner1 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="Radiance Technologies"]

@microsoft-github-policy-service agree company="Microsoft"

a-gardner1:
@microsoft-github-policy-service agree company="Radiance Technologies"

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -438,6 +438,34 @@ def _where_input_wrangler(
     return args, kwargs
 
 
+def _unique_unsorted_xfail_matcher(
```
a-gardner1, line 441: @justinchuby I'm not sure what the preferred behavior is here. Should we match `torch.unique` and ignore the `sorted` argument (i.e., always sort in `aten_unique`) or respect the argument and deviate in accordance with this matcher?

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: Addressed in b528a6ae902f3d7de87920ee1482a5a888cd14bc

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8381,7 +8427,16 @@ def aten_unique_dim(
 ) -> tuple[TensorType, TensorType, TensorType]:
     """unique_dim(Tensor self, int dim, bool sorted=True, bool return_inverse=False, bool return_counts=False) -> (Tensor, Tensor, Tensor)"""
 
-    raise NotImplementedError()
+    unique_values, indices, inverse_indices, counts = op.Unique(self, axis=dim, sorted=sorted)
+    # HACK: force indices to be in the graph so that it gets a name during optimization
+    # Otherwise an error will be raised in `onnxscript.Scope.lookup_or_create`
+    indices_size = op.Shape(indices)
+    counts = op.Reshape(counts, indices_size)
```
a-gardner1, line 8434: I want to note that the way that this function was written in 1d74d591ad65eb0b01cbd241f72c2b80277e1c4c is functionally equivalent but yields an error in `onnxscript.Scope.lookup_or_create` because it causes [`modified` to be True in `onnxscript.optimizer.optimize`](https://github.com/microsoft/onnxscript/blob/d71b74fb0b194e718a1fa78eddef7d89b57cf4a1/onnxscript/optimizer/__init__.py#L71), thus causing a second loop of optimization that crashes in the first call to `inline_simple_functions`. 

This seems indicative of a potential bug to me, but I am not knowledgeable enough about the codebase to suggest a cause or fix.

justinchuby:
Thanks for completing the CLA. I will take a look next week

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+    if return_inverse:
+        if return_counts:
+            result = unique_values, inverse_indices, counts
+        else:
+            result = unique_values, inverse_indices
+    elif return_counts:
+        result = unique_values, counts
+    else:
+        result = unique_values
+    return result
```
justinchuby, line 8398: I think we need to always return the same number of values. Consider returning None when they are not available?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+def aten_unique(
+    self: TensorType,
+    sorted: bool = True,
+    return_inverse: bool = False,
+    return_counts: bool = False,
+    dim: Optional[int] = None,
+) -> tuple[TensorType, TensorType, TensorType]:
+    """unique(Tensor self, bool sorted=True, bool return_inverse=False, bool return_counts=False, int? dim=None) -> (Tensor, Tensor?, Tensor?)"""
+
+    if dim is None:
```
justinchuby, line 8385: Codecov reported that this function is not covered. Is there a way to unit test this function?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+    else:
+        result = unique_values
+    return result
+
+
+@torch_op("aten::_unique2")
+def aten_unique2(
+    self: TensorType,
+    sorted: bool = True,
+    return_inverse: bool = False,
```
justinchuby, line 8405: Does this argument affect the operator's behavior? If it is intended to be unused, I would add ` # pylint: disable=unused-argument` to allow linter to pass

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+            result = unique_values, inverse_indices
+    elif return_counts:
+        result = unique_values, counts
+    else:
+        result = unique_values
+    return result
+
+
+@torch_op("aten::_unique2")
+def aten_unique2(
```
justinchuby, line 8402:
```suggestion
def aten__unique2(
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+    self: TensorType,
+    sorted: bool = True,
+    return_inverse: bool = False,
+    return_counts: bool = False
+) -> tuple[TensorType, TensorType, TensorType]:
+    """unique(Tensor self, bool sorted=True, bool return_inverse=False, bool return_counts=False) -> (Tensor, Tensor, Tensor)"""
```
justinchuby, line 8408: https://github.com/pytorch/pytorch/blob/cf35a591b95220aa1bfcc04ff8a943efd1d6d6eb/aten/src/ATen/native/native_functions.yaml#L6329

```suggestion
    """_unique2(Tensor self, bool sorted=True, bool return_inverse=False, bool return_counts=False) -> (Tensor, Tensor, Tensor)"""
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8372,6 +8372,52 @@ def aten_unique_consecutive(
...
+
+    unique_values, indices, inverse_indices, counts = op.Unique(self, axis=None, sorted=sorted)
+    # HACK: force indices to be in the graph so that it gets a name during optimization
```
justinchuby, line 8411: Is this a bug we should fix elsewhere?

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: cc @gramalingam

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -438,6 +438,34 @@ def _where_input_wrangler(
...
+    # equivalent to the output of the ONNX op
+    expect_fail = False
+    if sample.kwargs.get("sorted", None) is False and sample.input.numel() > 1:
+        # sorted == None is equivalent to True
+        # the result will be mismatched if the input is not sorted
+        # To expect equality, we must verify that the first appearance
+        # of each unique value is in ascending order in the input
+        test_kwargs = dict(sample.kwargs)
+        test_kwargs['return_inverse'] = True
+        test_kwargs['return_counts'] = False
```
justinchuby, line 454: A comment on why forcing these arguments?

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -438,6 +438,34 @@ def _where_input_wrangler(
...
+        _, inverse = torch.unique(sample.input, **test_kwargs)
+        observed = set()
+        max_observed: Optional[int] = None
+        for inv_idx in inverse.flatten().tolist():
+            if inv_idx not in observed:
+                if max_observed is not None and inv_idx < max_observed:
+                    expect_fail = True
```
justinchuby, line 461: Possible to just return here?

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -438,6 +438,34 @@ def _where_input_wrangler(
...
+                    break
+                observed.add(inv_idx)
+                if max_observed is None or inv_idx > max_observed:
+                    max_observed = inv_idx
+    return expect_fail
```
justinchuby, line 466:
```suggestion
    return False
```

Possible to simplify?

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: I think this could possibly be considered a different bug. The other one is a side-effect of  `onnxscript.optimizer.constant_folding.fold_constants`, whereas this one is a side-effect of the function linked below, which converts the names of unused outputs to empty strings but only removes them if they are trailing. Since `inverse_indices` and `counts` are used, it leads to an error being raised in [`onnxscript.Scope.lookup_or_create`](https://github.com/microsoft/onnxscript/blob/69ae7f421e2f1931e353949fa5e3a0fb23dbe622/onnxscript/_legacy_ir/visitor.py#L194) due to the empty string name given to `indices`.
https://github.com/microsoft/onnxscript/blob/69ae7f421e2f1931e353949fa5e3a0fb23dbe622/onnxscript/optimizer/remove_unused.py#L14

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: Doing so deviates from the behavior of `torch.unique` and causes this assertion in the unit tests to fail:
https://github.com/microsoft/onnxscript/blob/69ae7f421e2f1931e353949fa5e3a0fb23dbe622/tests/function_libs/torch_lib/ops_test.py#L251

Please advise on how to address this.

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: I think this is a false positive. The function is definitely tested as shown by the assertion error indicated in https://github.com/microsoft/onnxscript/pull/1547#discussion_r1607006325 that arises if one changes it.

tests/function_libs/torch_lib/ops_test_data.py
a-gardner1: Addressed in 6f89cef1fd0fca0eb463c700a5316d03d3795d85

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Does `torch.ops.aten.unique` exhibit the same behavior? If it always returns three variables, consider creating a new OpInfo for `torch.ops.aten.unique` similar to https://github.com/microsoft/onnxscript/blob/a5ed07981f1e7c100c5abeb2340a841b9e81e878/tests/function_libs/torch_lib/extra_opinfo.py#L2105-L2114. You may remove the xfail with the custom OpInfo too because you may simply remove the xfail cases. 

You may adapt the sample function from https://github.com/pytorch/pytorch/blob/b948b1ad7a9cf61c9692506c60c295fd40e00f43/torch/testing/_internal/common_methods_invocations.py#L3346-L3372

tests/function_libs/torch_lib/ops_test_data.py
justinchuby: I wonder if the argument leads to different behavior in cuda/cpu etc? I assume sorted=False means it _can_ be sorted, but it doesn't need to be; and there are some potential performance gain by turning it off. If that's the interpretation I would keep the argument. Otherwise ignoring the argument and matching behavior would also be nice.

a-gardner1: I am investigating differences in behavior between cuda/cpu and have found at least one already (`unique_dim` on CPU ignores the `return_inverse` and `return_counts` arguments whereas the CUDA impl does not). How should these differences be handled? Can the op registration be conditioned by the device somehow?

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: Thanks for the pointer to `extra_opinfo`.  It turns out `torch.ops.aten.unique` does not exist, but `torch.ops.aten._unique` does. Added `OpInfo` for it, `_unique2`, and `unique_dim` in 14d03b539095a80df070a2d570e023db2a4bee58

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: This function no longer exists.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8380,8 +8380,21 @@ def aten__unique(
 ) -> tuple[TensorType, TensorType]:
     """_unique(Tensor self, bool sorted=True, bool return_inverse=False) -> (Tensor, Tensor)"""
 
-    unique_values, _, inverse_indices, _ = op.Unique(self, axis=None, sorted=True)
+    unique_values, indices, inverse_indices, _ = op.Unique(self, axis=None, sorted=True)
+    # HACK: force indices to be in the graph so that it gets a name during optimization
```
justinchuby, line 8384: I suggest removing all hacks. I will go fix what's necessary where the bug is. We are also moving to prefer `trace_only=True` for new functions so if you can include the flag in `@torch_op` that would be awesome.

tests/function_libs/torch_lib/ops_test_data.py
justinchuby: Matching CUDA for now is preferable. Thanks!

onnxscript/function_libs/torch_lib/ops/core.py
a-gardner1: That would be awesome. The hacks are definitely getting out of hand. I'll wait for that fix so that I can continue to test with this locally.

justinchuby: Do you have a short script handy that will reproduce the error?

a-gardner1:
```python
if __name__ == '__main__':
    import logging
    import torch
    import numpy as np
    import onnx
    import onnxruntime as ort
    for i in range(16):
        sorted = bool(i & 1)
        return_inverse = bool((i & 2) > 1)
        return_counts = bool((i & 4) > 1)
        dim = 0 if bool((i & 8) > 1) else None

        def test_function(
                x: torch.Tensor,
                s: bool = sorted,
                ri: bool = return_inverse,
                rc: bool = return_counts,
                d: int | None = dim) -> Any:
            result = torch.unique(
                x,
                sorted=s,
                return_inverse=ri,
                return_counts=rc,
                dim=d)
            if ri or rc:
                result = result[1]
            return result

        onnx_program = torch.onnx.dynamo_export(
            test_function,
            torch.arange(10),
            export_options=torch.onnx.ExportOptions(
                dynamic_shapes=True,
                diagnostic_options=torch.onnx.DiagnosticOptions(
                    verbosity_level=logging.DEBUG)))
        onnx_program.save("torch_unique.onnx")
        onnx_inputs = onnx_program.adapt_torch_inputs_to_onnx(torch.arange(10))
        onnx_outputs = onnx_program(*onnx_inputs)
        loaded_onnx_program = onnx.load("torch_unique.onnx")
        onnx.checker.check_model(loaded_onnx_program)
        ort_session = ort.InferenceSession("torch_unique.onnx")
        inputs = np.random.randint(0, 10, 10)
        print(f"Inputs: {inputs}")
        outputs = ort_session.run(None,
                                  {"l_x_": inputs})
        print(f"Outputs: {outputs}")
    print("Success")
```

a-gardner1: Oh, you should also test using the nightly release of PyTorch with the changes in https://github.com/pytorch/pytorch/pull/126561.

a-gardner1: Is `trace_only=True` expected to require significant changes to the way one implements an op? It appears that enabling the flag breaks passing a value to `op.ConstantOfShape` and also breaks indexing a shape.

For example,  `op.ConstantOfShape([0], value=[0])` must become `op.Cast(op.ConstantOfShape([0]), to=INT64.dtype)`, and ` output_size[dim]` must become `op.Slice(output_size, [dim], [dim+1])`.

justinchuby: Your observation is correct. This may be the case because the gaps in implementation we have. Bridging the gaps is in our roadmap but is not the highest priority for the team.

a-gardner1:
Circling back around to this @justinchuby. At the time, I had been waiting for you to resolve the bug that required a [hacky workaround](https://github.com/microsoft/onnxscript/pull/1547/files/29c4f78bfbe58feac5fbf6c19b86399130c9dcb0..1310262d66914d9deb9d46a0d30b81203a60c0e3#r1609009212), but I realize that might not be clear. 

There were also a couple of other potential unresolved bugs outside the scope of this PR, e.g., [this comment](https://github.com/microsoft/onnxscript/pull/1547#discussion_r1605539265). 

How would you like to proceed?

justinchuby:
Sorry for missing the clarity. I would suggest that you remove all the hacks so that the code is at its desirable state. If tests fail because of that, that’s ok. I will then go ahead to fix what’s needed. (After I’m back from vacation)

kabyanil:
I am implementing a CTC decoder class in pytorch -

```
class GreedyCTCDecoder(torch.nn.Module):
    def __init__(self, labels, blank=0):
        super().__init__()
        self.labels = labels
        self.blank = blank

    def forward(self, emission: torch.Tensor) -> List[str]:
        """Given a sequence emission over labels, get the best path
        Args:
          emission (Tensor): Logit tensors. Shape `[num_seq, num_label]`.

        Returns:
          List[str]: The resulting transcript
        """
        indices = torch.argmax(emission, dim=-1)  # [num_seq,]
        indices = torch.unique_consecutive(indices, dim=-1)
        indices = [i for i in indices if i != self.blank]
        joined = "".join([self.labels[i] for i in indices])
        return joined.replace("|", " ").strip().split()


greedy_decoder = GreedyCTCDecoder(tokens)
```
I'm not able to export this class to onnx. Here is my error -

```
/usr/local/lib/python3.10/dist-packages/torch/onnx/_internal/_exporter_legacy.py:116: UserWarning: torch.onnx.dynamo_export only implements opset version 18 for now. If you need to use a different opset version, please register them with register_custom_op.
  warnings.warn(
---------------------------------------------------------------------------
DynamicOutputShapeException               Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/torch/_dynamo/utils.py](https://localhost:8080/#) in run_node(tracer, node, args, kwargs, nnmodule)
   2131             if op == "call_function":
-> 2132                 return node.target(*args, **kwargs)
   2133             elif op == "call_method":

53 frames
DynamicOutputShapeException: aten.unique_consecutive.default

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
RuntimeError: Failed running call_function <function boolean_dispatch.<locals>.fn at 0x7ba8c8f9dd80>(*(FakeTensor(..., size=(s0,), dtype=torch.int64),), **{'dim': -1}):
aten.unique_consecutive.default

During handling of the above exception, another exception occurred:

Unsupported                               Traceback (most recent call last)
Unsupported: dynamic shape operator: aten.unique_consecutive.default; Operator does not have a meta kernel that supports dynamic output shapes, please report an issue to PyTorch

from user code:
   File "<ipython-input-43-c1d03e7f78a6>", line 16, in forward
    indices = torch.unique_consecutive(indices, dim=-1)

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information


The above exception was the direct cause of the following exception:

OnnxExporterError                         Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/torch/onnx/_internal/_exporter_legacy.py](https://localhost:8080/#) in dynamo_export(model, export_options, *model_args, **model_kwargs)
   1231             f"Please report a bug on PyTorch Github: {_PYTORCH_GITHUB_ISSUES_URL}"
   1232         )
-> 1233         raise errors.OnnxExporterError(message) from e
   1234 
   1235 

OnnxExporterError: Failed to export the model to ONNX. Generating SARIF report at 'report_dynamo_export.sarif'. SARIF is a standard format for the output of static analysis tools. SARIF logs can be loaded in VS Code SARIF viewer extension, or SARIF web viewer (https://microsoft.github.io/sarif-web-component/). Please report a bug on PyTorch Github: https://github.com/pytorch/pytorch/issues
```

How can I resolve this?

tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -1950,6 +1950,30 @@ def shape(size, rank, with_batch_channel=True):
...
+            yield sample
+
+
+def sample_inputs_unique_dim(op_info, device, dtype, requires_grad, **kwargs):
+    for sample in common_methods_invocations.sample_inputs_unique(
+            op_info, device, dtype, requires_grad, **kwargs):
+        # take only those samples that ask for a dim
+        if sample.kwargs.get('dim') is not None:
+            yield sample
 def sample_inputs_upsample_trilinear3d_vec(op_info, device, dtype, requires_grad, **kwargs):
```
justinchuby, line 1977:
```suggestion

def sample_inputs_upsample_trilinear3d_vec(op_info, device, dtype, requires_grad, **kwargs):
```

justinchuby:
@a-gardner1 sorry for the delay. I think this PR is now good to merge (started the merge)

Issue #1501, opened 2024-05-06
[IR] Values that are initialized but not graph inputs do not have their ValueInfo preserved


justinchuby:
Currently the values in the graph that are not inputs and do not have producing nodes are not registered with the graph. Their value info is not preserved at serialization. We need a way to easily access them and be able to serialize them.

gramalingam:
Just to clarify, I guess these are initializers (that are not inputs). I realize the title says this, but I wasn't looking at that first!

How do users create them? (In previous frameworks/approaches, typically we used to create "Constant()" op nodes, depending on subsequent processing, if desired, to convert them into initializers.)

justinchuby:
> Just to clarify, I guess these are initializers (that are not inputs). I realize the title says this, but I wasn't looking at that first!

Thanks for clarifying - the title was referring to "initialized objects", but yes they are typically initializers.

> How do users create them? (In previous frameworks/approaches, typically we used to create "Constant()" op nodes, depending on subsequent processing, if desired, to convert them into initializers.)

These values are simply ir.Value as inputs to nodes. When deserializing they are created on the fly, but they are not referenced by the graph, and the serializers only create ValueInfo for node outputs and not inputs.

I think we can simply create ValueInfo for all initializers. I will double check if we are doing this already.

PR #1552, opened 2024-05-16
[IR][fix] Save value info for initializers

justinchuby:
Previously initializers are not included in the graph value_info because they are not easily accessible from the Graph object. Now what we store all the Values for initializers, we can serialize the value information into the graph.

Fix https://github.com/microsoft/onnxscript/issues/1501

### DIFF
--- a/onnxscript/ir/serde.py
+++ b/onnxscript/ir/serde.py
@@ -1093,17 +1093,20 @@ def serialize_graph_into(
         graph_proto.doc_string = from_.doc_string
     for input_ in from_.inputs:
         serialize_value_into(graph_proto.input.add(), input_)
+    input_names = {input_.name for input_ in from_.inputs}
     # TODO(justinchuby): Support sparse_initializer
-    for initializer in from_.initializers.values():
-        if initializer.const_value is None:
+    for value in from_.initializers.values():
+        if _should_create_value_info_for_value(value) and value.name not in input_names:
+            # Serialize information about all initializers into value_info,
+            # except for those that are also graph inputs
+            serialize_value_into(graph_proto.value_info.add(), value)
+        if value.const_value is None:
             # Skip initializers without constant values
-            logger.warning(
-                "Initializer '%s' does not have a constant value set.", initializer.name
-            )
+            logger.warning("Initializer '%s' does not have a constant value set.", value.name)
             continue
         # Make sure the tensor's name is the same as the value's name
-        initializer.const_value.name = initializer.name
-        serialize_tensor_into(graph_proto.initializer.add(), from_=initializer.const_value)
+        value.const_value.name = value.name
+        serialize_tensor_into(graph_proto.initializer.add(), from_=value.const_value)
     for node in from_:
         serialize_node_into(graph_proto.node.add(), from_=node)
         for node_output in node.outputs:


onnxscript/ir/serde.py
```diff
@@ -1093,17 +1110,20 @@ def serialize_graph_into(
...
             # Skip initializers without constant values
-            logger.warning(
-                "Initializer '%s' does not have a constant value set.", initializer.name
-            )
+            logger.warning("Initializer '%s' does not have a constant value set.", value.name)
             continue
         # Make sure the tensor's name is the same as the value's name
-        initializer.const_value.name = initializer.name
-        serialize_tensor_into(graph_proto.initializer.add(), from_=initializer.const_value)
+        value.const_value.name = value.name
```
justinchuby, line 1125: Different values can share the same tensor. We need to find a way to create tensor alias for this purpose, so we do not need to worry about tensor names

justinchuby: https://github.com/microsoft/onnxscript/issues/1554

onnxscript/ir/serde.py
```diff
@@ -1093,17 +1110,20 @@ def serialize_graph_into(
...
+            # except for those that are also graph inputs
+            serialize_value_into(graph_proto.value_info.add(), value)
+        if value.const_value is None:
             # Skip initializers without constant values
-            logger.warning(
-                "Initializer '%s' does not have a constant value set.", initializer.name
-            )
+            logger.warning("Initializer '%s' does not have a constant value set.", value.name)
             continue
         # Make sure the tensor's name is the same as the value's name
```
justinchuby, line 1124:
```suggestion
        # Make sure the tensor's name is the same as the value's name
        # TODO(#1554): Handle tensor alias better
```

onnxscript/ir/serde.py
```diff
@@ -527,38 +527,55 @@ def _deserialize_graph(
...
+
     initializer_values = []
-    for tensor in initializer_tensors:
-        if tensor.name in values:
+    for i, tensor in enumerate(initializer_tensors):
+        initializer_name = tensor.name
+        if not initializer_name:
+            logger.warning(
+                "Initializer tensor must have a name but the %s-th initializer does not. Skipping this initializer.",
+                i,
```
titaiwangms, line 545: Should we indicate it's index? Like f"initiliizer[{i}]"

titaiwangms (commented):
Ci fails

gramalingam:
Can this be closed, or is this still relevant?

justinchuby:
This is still relavent. We need to update our test data first.

onnxscript/ir/serde.py
justinchuby: `%s-th` shows it's an index

onnxscript/ir/serde.py
```diff
@@ -625,36 +625,50 @@ def _deserialize_graph(
...
                 None,
                 index=None,
-                name=tensor.name,
-                # TODO(justinchuby): Fix type hinting for shape and dtype
-                shape=tensor.shape,  # type: ignore
-                type=_core.TensorType(tensor.dtype),
+                name=initializer_name,
+                # Do not include shape or type as we need to respect the ONNX file
+                # if the shape or type is not provided as ValueInfoProto
+                # The shape/type information will be filled in in the subsequent ValueInfoProto
```
justinchuby, line 658: But that's inconvenient. Maybe not do this

onnxscript/ir/serde.py
```diff
@@ -625,36 +625,50 @@ def _deserialize_graph(
...
+                # deserialization step
                 const_value=tensor,
             )
+            if initializer_name in value_info:
+                # This is where we fill in the shape and type information for the initializer
+                deserialize_value_info_proto(value_info[initializer_name], initializer_value)
+            values[initializer_name] = initializer_value  # type: ignore[index]
```
justinchuby, line 665: move to below

onnxscript/ir/serde.py
```diff
@@ -625,36 +625,50 @@ def _deserialize_graph(
...
             if initializer_value.name in quantization_annotations:
                 _deserialize_quantization_annotation(
                     quantization_annotations[initializer_value.name], initializer_value
                 )
-            values[tensor.name] = initializer_value  # type: ignore[index]
         initializer_values.append(initializer_value)
```
justinchuby, line 671: keep?

onnxscript/ir/serde.py
```diff
@@ -625,36 +625,50 @@ def _deserialize_graph(
...
 
-    # Add ValueInfos for this graph scope
-    value_info = {info.name: info for info in proto.value_info}
```
justinchuby, line 656: Keep?

onnxscript/ir/serde.py
```diff
@@ -1284,18 +1301,22 @@ def serialize_graph_into(
...
             # Skip initializers without constant values
-            logger.warning(
-                "Initializer '%s' does not have a constant value set.", initializer.name
-            )
+            logger.warning("Initializer '%s' does not have a constant value set.", value.name)
             continue
         # Make sure the tensor's name is the same as the value's name
-        initializer.const_value.name = initializer.name
-        serialize_tensor_into(graph_proto.initializer.add(), from_=initializer.const_value)
+        # TODO(#1554): Handle tensor alias better
```
justinchuby, line 1317:
```suggestion
```

titaiwangms (approved):
Even the model is updated, should we assert valueinfo in the tests?

onnxscript/ir/serde.py
```diff
@@ -1284,18 +1298,21 @@ def serialize_graph_into(
...
             # the initializer list is tracked
             _maybe_add_quantization_annotation(graph_proto, input_)
+    input_names = {input_.name for input_ in from_.inputs}
     # TODO(justinchuby): Support sparse_initializer
-    for initializer in from_.initializers.values():
-        _maybe_add_quantization_annotation(graph_proto, initializer)
-        if initializer.const_value is None:
+    for value in from_.initializers.values():
+        _maybe_add_quantization_annotation(graph_proto, value)
+        if _should_create_value_info_for_value(value) and value.name not in input_names:
```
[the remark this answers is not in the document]
justinchuby: It's like a parameter having a default value. Any input can be initialized if an initializer of the same name is in the graph. Users can choose to overwrite the initializer by providing their own input.

justinchuby:
> Even the model is updated, should we assert valueinfo in the tests?

Good point. I will create a follow up to update some of the test behaviors, thanks.

PR #1553, opened 2024-05-16
remove get_numpy_from_ir_value

### LINKED ISSUE #1550: Remove `get_numpy_from_ir_value`
`get_numpy_from_ir_value` in `_ir_utils` is unnecessary and can be removed.

titaiwangms:
Fix #1550

### DIFF
--- a/onnxscript/rewriter/_ir_utils.py
+++ b/onnxscript/rewriter/_ir_utils.py
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-import numpy as np
-
 from onnxscript import ir
 
 GRAPH_OUTPUT_META_KEY = "pkg.onnxscript.rewriter.generic_pattern.graph_output"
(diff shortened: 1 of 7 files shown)


onnxscript/rewriter/onnxruntime/instance_to_group_normalization.py
```diff
@@ -41,10 +41,10 @@ def check_if_simulated_instance_norm_is_used(
         bool: True if the simulated instance normalization is used, False otherwise.
     """
     weight_for_norm = _ir_utils.propagate_const_value(weight_for_norm)
-    weight_for_norm = _ir_utils.get_numpy_from_ir_value(weight_for_norm)
+    weight_for_norm = weight_for_norm.const_value
 
     bias_for_norm = _ir_utils.propagate_const_value(bias_for_norm)
-    bias_for_norm = _ir_utils.get_numpy_from_ir_value(bias_for_norm)
+    bias_for_norm = bias_for_norm.const_value
```
justinchuby, line 47: I would call .numpy() if we need a numpy value; Besure to check for None before calling numpy; assign it to a new variable to prevent mypy complaining about the type changing for bias_for_norm;

onnxscript/rewriter/_ir_utils.py
```diff
@@ -29,12 +27,3 @@ def propagate_const_value(ir_value: ir.Value) -> ir.Value:
...
                     break
     return ir_value
-
-
-def get_numpy_from_ir_value(value: ir.Value) -> np.ndarray | None:
-    constant_value = value.const_value
-    if constant_value is not None:
-        if isinstance(constant_value, ir.serde.TensorProtoTensor):
-            return constant_value.numpy()
-        return np.array(constant_value)
```
titaiwangms, line 39: @justinchuby based on your comment, sounds like it's this line we don't want. We should delete the np.array() call and keep this function?

PR #1555, opened 2024-05-17
Add verbose support in single-output pattern-matcher

gramalingam:
Next step in unifying the two pattern-matchers:

* Refactor the pattern-matching algorithm out of the pattern-IR classes
* Add support for verbose-flag: will print info about status during algorithm
* Unify the constructors for rewrite-rule


onnxscript/rewriter/pattern_test.py
```diff
@@ -57,6 +59,16 @@ def test_failed_match(self):
         self.assertEqual(count, 0)
         self.assertEqual(len(model.graph), 4)
 
+        # Test verbose output produces something:
+        # TODO(rama): Need a better way to test this.
+        # Well-defined error-codes and messages would be helpful.
+
+        buffer = io.StringIO()
+        with contextlib.redirect_stdout(buffer):
+            self.rule().apply_to_model(model, verbose=5)
```
justinchuby, line 68: Possible to specify verbosity levels and standardize it?

gramalingam: Yes, that would be useful. For now, I am just doing something similar to Xavier's implementation, so that this doesn't become a reason to choose one implementation over another. I am happy to leave it to Xavier, as the primary user who uses this, to flesh this out depending on his usage experience.

onnxscript/rewriter/pattern.py
```diff
@@ -604,27 +570,30 @@ def __init__(
...
         # TODO (rama): allow users to specify shape requirement, if desired.
         if constant_value.size != 1:
-            return MatchResult.FAIL()
+            return match.fail(f"Value is not a scalar, expecting {self.value}.")
 
-        return self.match_scalar(constant_value.item())
+        if not math.isclose(
+            constant_value.item(), self._value, rel_tol=self._rel_tol, abs_tol=self._abs_tol
+        ):
+            match.fail(f"Value mismatch: expected {self._value}, got {constant_value.item()}.")
```
justinchuby, line 590: Is it every possible for fail to be called multiple times on the same match result? That way the result will be overwritten.

onnxscript/rewriter/pattern.py
```diff
@@ -912,23 +876,140 @@ def __init__(self, pattern: GraphPattern) -> None:
...
+
+    def _match_constant(self, pattern_constant: Constant, value: ir.Value) -> bool:
+        """Match a Constant pattern against a value.
+
+        If the constant value is produced by a Constant node, we do not include
+        the constant node as part of the matched graph. Thus, it will not be deleted,
+        if subgraph replacement happens. But subsequent DCE will remove the constant
+        node if it is not used elsewhere.
+        """
+        value = _ir_utils.propagate_const_value(value)
```
justinchuby, line 895: I would add a comment to remove these two lines when we sort this out

onnxscript/rewriter/pattern_test.py
justinchuby: SG. Thanks!

onnxscript/rewriter/generic_pattern.py
```diff
@@ -633,6 +634,10 @@ def make_pattern_rule(
         the rewriting rule
     """
 
+    warnings.warn(
+        "make_pattern_rule(...) is deprecated, use pattern.RewriteRule(...) instead",
+        DeprecationWarning,
```
justinchuby, line 639: FutureWarning for deprecations: https://docs.python.org/3/library/exceptions.html#FutureWarning

onnxscript/rewriter/pattern.py
gramalingam: Not by design. The first failure terminates that particular search.

onnxscript/rewriter/pattern.py
gramalingam: Not sure I understand. It serves as documentation, in case a reader starts wondering about this. If we change this design, then we would need to change both the code and documentation.

onnxscript/rewriter/generic_pattern.py
gramalingam: Thanks, changed

onnxscript/rewriter/pattern.py
justinchuby: I mean ` value = _ir_utils.propagate_const_value(value)` is temporary and we should remove _ir_utils usage when ready

PR #1561, opened 2024-05-21
Simplify get_numpy_from_ir_value

### LINKED ISSUE #1550: Remove `get_numpy_from_ir_value`

titaiwangms:
Fix #1550 

NOTE: from https://github.com/microsoft/onnxscript/pull/1553#discussion_r1604205079, I think we can hide the None check in this function.

### DIFF
--- a/onnxscript/rewriter/_ir_utils.py
+++ b/onnxscript/rewriter/_ir_utils.py
@@ -47,7 +47,5 @@ def propagate_const_value(ir_value: ir.Value) -> ir.Value:
 def get_numpy_from_ir_value(value: ir.Value) -> np.ndarray | None:
     constant_value = value.const_value
     if constant_value is not None:
-        if isinstance(constant_value, ir.serde.TensorProtoTensor):
-            return constant_value.numpy()
-        return np.array(constant_value)
+        return constant_value.numpy()
     return constant_value
--- a/onnxscript/rewriter/_ir_utils.py
+++ b/onnxscript/rewriter/_ir_utils.py
@@ -42,10 +42,3 @@ def propagate_const_value(ir_value: ir.Value) -> ir.Value:
     ir_value.shape = const_value.shape  # type: ignore
     ir_value.dtype = const_value.dtype
     return ir_value
-
-
-def get_numpy_from_ir_value(value: ir.Value) -> np.ndarray | None:
-    constant_value = value.const_value
-    if constant_value is not None:
-        return constant_value.numpy()
-    return constant_value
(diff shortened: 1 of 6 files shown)


onnxscript/rewriter/_ir_utils.py
```diff
@@ -47,7 +47,5 @@ def propagate_const_value(ir_value: ir.Value) -> ir.Value:
 def get_numpy_from_ir_value(value: ir.Value) -> np.ndarray | None:
     constant_value = value.const_value
     if constant_value is not None:
-        if isinstance(constant_value, ir.serde.TensorProtoTensor):
-            return constant_value.numpy()
-        return np.array(constant_value)
+        return constant_value.numpy()
     return constant_value
```
justinchuby, line 51: Instead of this, I suggest removing the function and move the None check before the call site, since we have to check for None anyways

titaiwangms: Before we call const_value?

justinchuby: Yes. So

if value.const_value is None:
    # do something
else:
    something_else = value.const_value.numpy()

justinchuby: However if we feat the `numpy()` call may fail, we can create a 

```
def try_get_numpy(value: ir.Value) -> np.ndarray:
    try:
        ...
    except ...:
        return None
```

justinchuby: But it doesn't feel very elegant :/

onnxscript/rewriter/onnxruntime/transformers/multihead_attention.py
```diff
@@ -109,7 +109,12 @@ def infer_attn_size_config(self, function: ir.Function) -> AttnSizeConfig:
                     constant_node.op_type == "Constant"
                 ), "Expected the second input to Reshape to be a Constant node."
                 value = _ir_utils.propagate_const_value(reshape_node.inputs[1])
-                constant_numpy_value = _ir_utils.get_numpy_from_ir_value(value)
+                constant_value = value.const_value
+                if constant_value is None:
+                    raise function_rule.FunctionRewriteError(
```
justinchuby, line 114: Will this fail the whole thing or is it captured?

onnxscript/rewriter/pattern_test.py
```diff
@@ -258,9 +257,11 @@ def identity(op, x, newshape):
         def check_for_redundant_reshape(context, x, newshape):
             oldshape = x.shape
             newshape = _ir_utils.propagate_const_value(newshape)
-            newshape = _ir_utils.get_numpy_from_ir_value(newshape)
-            if not isinstance(newshape, np.ndarray):
+            newshape_cont_value = newshape.const_value
```
justinchuby, line 260: nit: typo?

justinchuby (approved):

onnxscript/rewriter/onnxruntime/transformers/multihead_attention.py
titaiwangms: Captured. The rule would be treated as no matched.

titaiwangms:
Merged as I think CI failue is unrelated.

Issue #1559, opened 2024-05-21
how to set opset in generated model by to_model_proto?


vincentme:
# behavior

The auto-generated opset in the model by to_model_proto is not consistent with the opset set in individual subfunction by @script. The former is opset_import: ["this" : 1, "" : 21] and the latter is say 20, so the onnx.checker complains. 

I tried to set opset in @script or  to_model_proto(), but seems does not work. 

# code to reproduce

```python
from onnxscript import opset20 as op
from onnxscript import FLOAT
from onnxscript import script
from onnx import TensorProto

@script()
def to_float32(x):
    return op.Cast(x, to = TensorProto.FLOAT)

@script()
def self_norm(x):
    x_max = op.ReduceMax(x)
    x_min = op.ReduceMin(x)
    y = (x-x_min)/(x_max-x_min) # (x-x_min)/(x_max-x_min)
    
    return y, x_max, x_min

@script()
def algorithm(img_in: FLOAT['row', 'col']) -> tuple[FLOAT['row', 'col'], FLOAT[1, 1], FLOAT[1, 1]]:
    img_out, img_in_max, img_in_min = self_norm(to_float32(img_in))
    return img_out, img_in_max, img_in_min


if __name__ == '__main__':
    import numpy as np
    import onnx
    
    img_in_np = (np.random.random_sample((3, 4))*100).astype(np.uint16)
    
    img_out, img_in_max, img_in_min = algorithm(img_in_np)
    
    model = algorithm.to_model_proto()
    model = onnx.shape_inference.infer_shapes(model)
    print(onnx.printer.to_text(model))
    onnx.checker.check_model(model)
``` 

# error message

```
ValidationError: Opset import for domain  in function op Castis not compatible with the version imported by model. FunctionOp imports version 20 whereas model imports version 21
```

justinchuby:
That seems to be a bug in `to_model_proto`

justinchuby:
@vincentme this week is hectic so please expect delayed response. Thank you for your patience!

cc @gramalingam in case you have an idea, thanks!

vincentme:
No problem!

gramalingam:
A bit strange. Seems like a simple example. Likely a bug since there is no explicit op used in main model

PR #1564, opened 2024-05-22
Use default opset only if functions don't use opset

gramalingam:
Fix bug reported in Issue #1559 :  look at opsets imported by functions _before_ using a default.

### DIFF
--- a/onnxscript/irbuilder.py
+++ b/onnxscript/irbuilder.py
@@ -374,6 +374,10 @@ class IRFunction:
         for proto in functions:
             if proto.domain not in opsets:
                 opsets[proto.domain] = 1
+            # TODO(rama): Handle conflicts with appropriate error/warning message.
+            for opset in proto.opset_import:
+                if opset.domain not in opsets:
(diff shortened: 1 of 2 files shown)


onnxscript/irbuilder.py
```diff
@@ -370,13 +370,15 @@ def to_proto(f):
         for n in self.stmts:
             if n.callee.opset.domain not in opsets:
                 opsets[n.callee.opset.domain] = n.callee.opset.version
+
```
justinchuby, line 373: I would add a comment about the significance of the ordering here.

justinchuby (approved):

onnxscript/irbuilder.py
gramalingam: Hmmm ... actually, I didn't check this carefully ... changed it to WIP ... need to check a couple of things

gramalingam: Ok, fixed it, hopefully it is okay now.

Issue #1556, opened 2024-05-19
Create a node iterator


justinchuby:
Create `traversal.py` and a node iterator to replace NodeTransformer.

```python
class _RecursiveNodeIterator(Iterator):
    def __call__(self): ...
    def __reversed__(self): ...

def recursive_node_iter(graph, enter_graph_handler, exit_graph_handler, recursive: Callable[[Node], bool]):
    return _RecursiveNodeIterator(...)

# Usage
scope = []
for node in recursive_node_iter(
    graph,
    lambda graph: scope.append(graph)
    lambda graph: scope.pop()
    lambda node: True
):
    ...
```

PR #1565, opened 2024-05-23
Create a recursive graph iterator and use it to refactor UnusedFunctionRemover

justinchuby:
- Create `traversal.py` for graph traversal utilities and implemented `RecursiveGraphIterator`. Expose `traversal` to the `ir` module. Fixes https://github.com/microsoft/onnxscript/issues/1556
- Remove `NodeTransformer` because `RecursiveGraphIterator` is more flexible.
- Refactor remove_unused_function.py to use `RecursiveGraphIterator`

### DIFF
--- a/onnxscript/ir/traversal.py
+++ b/onnxscript/ir/traversal.py
@@ -25,8 +25,6 @@ class RecursiveGraphIterator(Iterator[_core.Node], Reversible[_core.Node]):
 
         Args:
             graph: The graph to traverse.
-            enter_graph_handler: A callback that is called when a subgraph is entered.
-            exit_graph_handler: A callback that is called when a subgraph is exited.
             recursive: A callback that determines whether to recursively visit the subgraphs
                 contained in a node. If not provided, all nodes in subgraphs are visited.
             reverse: Whether to iterate in reverse order.
--- a/onnxscript/ir/traversal.py
+++ b/onnxscript/ir/traversal.py
@@ -1,3 +1,7 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
 """Utilities for traversing the IR graph."""
 
 from __future__ import annotations
--- a/onnxscript/ir/traversal.py
+++ b/onnxscript/ir/traversal.py
@@ -54,7 +54,7 @@ class RecursiveGraphIterator(Iterator[_core.Node]):
         if self._enter_graph_handler is not None:
             self._enter_graph_handler(graph)
         iterable = reversed(graph) if self._reversed else graph
-        for node in iterable:
+        for node in iterable:  # type: ignore[union-attr]
             yield node
             if self._recursive is not None and not self._recursive(node):
                 continue
--- a/onnxscript/ir/traversal.py
+++ b/onnxscript/ir/traversal.py
@@ -2,46 +2,66 @@
 
 from __future__ import annotations
 
+from typing_extensions import Self
+
 from onnxscript.ir import _core, _enums
 
 __all__ = [
-    "recursive_node_iter",
+    "RecursiveNodeIterator",
 ]
 
-from typing import Callable, Iterable, Iterator
+from typing import Callable, Iterator
 
 
-class _RecursiveNodeIterator(Iterable[_core.Node]):
-    def __init__(self,
+class RecursiveNodeIterator(Iterator[_core.Node]):
+    def __init__(
+        self,
         graph: _core.Graph,
         enter_graph_handler: Callable[[_core.Graph], None],
         exit_graph_handler: Callable[[_core.Graph], None],
         recursive: Callable[[_core.Node], bool],
         reversed: bool = False,
     ):
+        """Iterate over the nodes in the graph, recursively visiting subgraphs.
+
+        Args:
+            graph: The graph to traverse.
+            enter_graph_handler: A callback that is called when a subgraph is entered.
+            exit_graph_handler: A callback that is called when a subgraph is exited.
+            recursive: A callback that determines whether to recursively visit a node.
+            reversed: Whether to iterate in reverse order.
+        """
         self._graph = graph
         self._enter_graph_handler = enter_graph_handler
         self._exit_graph_handler = exit_graph_handler
         self._recursive = recursive
         self._subgraph_stack = []
         self._reversed = reversed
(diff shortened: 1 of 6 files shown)


onnxscript/ir/traversal.py
```diff
@@ -0,0 +1,91 @@
...
+            exit_graph_handler: A callback that is called when a subgraph is exited.
+            recursive: A callback that determines whether to recursively visit a node. If
+                not provided, all nodes are visited.
+            reversed: Whether to iterate in reverse order.
+        """
+        self._graph = graph
+        self._enter_graph_handler = enter_graph_handler
+        self._exit_graph_handler = exit_graph_handler
+        self._recursive = recursive
+        self._subgraph_stack = []
```
justinchuby, line 39: Remove

onnxscript/ir/traversal.py
```diff
@@ -0,0 +1,99 @@
...
+
+from onnxscript.ir import _core, _enums
+
+
+class RecursiveGraphIterator(Iterator[_core.Node], Reversible[_core.Node]):
+    def __init__(
+        self,
+        graph: _core.Graph | _core.Function | _core.GraphView,
+        *,
+        enter_graph_handler: Callable[[_core.Graph | _core.Function | _core.GraphView], None]
```
gramalingam, line 21: I am unclear about the pros/cons of some of the design aspects here. 
* I can see the value in a utility that allows users to write ` for node in ir.traversal.RecursiveGraphIterator(model.graph)`.
* But, the moment users start providing enter_graph / exit_graph handlers, I personally feel that the advantage shifts to a design where the user defines a subclass (say, of RecursiveGraphIterator, which seems to be a Visitor pattern of some kind) by over-riding enter_graph / exit_graph methods. So, I would omit those parameters (in favor of user-defined sub-classes ... which may still be generic enough to be used by many, encoding a specific style of visiting).

gramalingam: I think that the idea of decoupling `Pass` from a `GraphIterator\Visitor` makes sense. They are two separate infrastructure components.

justinchuby: That works too. Do you envision the following: to record the current scope in the iteration, one would do:

```python
class ScopeRecursiveGraphIterator(RecursiveGraphIterator):
    def __init__(self, scope):
        self.scope = []

	def enter_graph_handler(graph):
	    scopes.append(graph.name)

	def exit_graph_handler(graph):
	    scopes.pop()

scope = []
for node in ScopeRecursiveGraphIterator(graph, scope):
    print(scopes)
```

instead of

```python
scopes = []

def enter_graph_handler(graph):
    scopes.append(graph.name)

def exit_graph_handler(graph):
    scopes.pop()

for node in traversal.RecursiveGraphIterator(
    graph, enter_graph_handler=enter_graph_handler, exit_graph_handler=exit_graph_handler
):
    print(scopes)
```

?

justinchuby: A challenge is that in the recursive calls, we don't know how to initialize the user defined classes.

gramalingam: Perhaps providing just the basic GraphIterator (without the ability to specify enter/exit graph handlers) would be a reasonable starting point? As we complete the other applications, such as migrating the Optimizer etc., we may find what extensions we need.

justinchuby: Done

onnxscript/ir/traversal.py
```diff
@@ -0,0 +1,85 @@
...
+        iterable = reversed(graph) if self._reverse else graph
+        for node in iterable:  # type: ignore[union-attr]
+            yield node
+            if self._recursive is not None and not self._recursive(node):
+                continue
+            yield from self._iterate_subgraphs(node)
+
+    def _iterate_subgraphs(self, node: _core.Node):
+        iterator = (
+            reversed(node.attributes.values()) if self._reverse else node.attributes.values()
```
gramalingam, line 60: This may be unnecessary. This will change things only for a IF-node. When users want to iterate forward or backward, I don't think they care what order the THEN-branch and ELSE-branch are visited in (as long as internally those branches are visited in the right order). Semantically, it is a reasonable spec that the two branches may be visited in either order.

If there is a cost to reversing the attributes, we may as well avoid it.

gramalingam: What users may care more about is pre-order or post-order: are the THEN/ELSE branches visited before or after the IF node itself. But that would mean another option ... so, perhaps better to wait for a use-case before bothering about it.

justinchuby: Done

onnxscript/optimizer/remove_unused_function.py
```diff
@@ -13,41 +13,49 @@
...
+
+
+class UnusedFunctionRemover(ir.passes.PassBase):
     def __init__(self):
         super().__init__()
         self.used: set[ir.OperatorIdentifier] = set()
 
-    def _call_function(self, function: ir.Function) -> None:
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        for node in ir.traversal.RecursiveGraphIterator(model.graph):
```
gramalingam, line 32: May be useful to move the initialization of `self.used` here, and clear it at the end of the function?

onnxscript/optimizer/remove_unused_function.py
```diff
@@ -13,41 +13,49 @@
...
 
-    def call_node(self, node: ir.Node) -> None:
+    def _call_node(self, model: ir.Model, node: ir.Node) -> None:
         op_identifier = node.op_identifier()
-        if op_identifier in self.model.functions:
-            self._call_function(self.model.functions[op_identifier])
+        if op_identifier in model.functions:
+            self._call_function(model, model.functions[op_identifier])
         else:
             self.used.add(op_identifier)
```
gramalingam, line 57: I wonder why this is necessary? It is sufficient if `used` tracks only functions in the model?

gramalingam (approved):
Just a couple of minor suggestions added in the comments

onnxscript/optimizer/remove_unused_function.py
justinchuby: Great point, thanks

justinchuby: Done

onnxscript/optimizer/remove_unused_function.py
justinchuby: Removed

PR #1588, opened 2024-06-04
Address missing license

xadupre:
Many other places are missing. This should be address automatically or maybe add a unit test to detect missing places.


justinchuby:
I can add a check. Thanks!

justinchuby:
Can be replaced with https://github.com/microsoft/onnxscript/pull/1589

PR #1594, opened 2024-06-08
FixOp (argmax, argmin) | feat(torchlib)

titaiwangms:
aten_argmax and aten_argmin have "dim=None" as default. 

https://github.com/pytorch/pytorch/blob/2369c719d485af0787d95668947125a5605bed88/aten/src/ATen/native/native_functions.yaml#L810

Previous to "trace all traceable functions" PR, scripted function manages to handle unamtched attributes if they are None, but in traced function, this becomes errors of unrecognized arguments to the function.


justinchuby (approved):
Thanks!

PR #1595, opened 2024-06-10
Update installed dependencing for tests

xadupre:


onnxscript/testing/type_double.py
justinchuby, line 1: `testing/` is for utils exposed to users and not test data or internal helpers. We can discuss where this fits

tests/converter_test.py
justinchuby, line 1: Unit tests should stay in place next to source

onnxscript/testing/type_double.py
justinchuby: So two criteria for code in testing/: 1. user facing (public apis); 2. utilities

xadupre:
Closing this one. Too many lint issues to fix. scikit-learn is using the same convention, tests are close to the file to test (they have many subfolder tests) but tests are included in the package so that users can check the package works by running them after it is installed. It is confusing to have tests inside and outside the package so I thought tests inside the package would be part of the released package. That makes sense I think if we want users to quickly check the package is working with the version of pytorch and transformers they have. I would consider that approach.

PR #1596, opened 2024-06-10
Proposal to group function defining a pattern into a class.

xadupre:


onnxscript/rewriter/pattern.py
```diff
@@ -1119,6 +1119,58 @@ def replace_pattern(new_pattern):
...
+    @classmethod
+    def rewrite(cls, op, *_) -> Any:
+        raise NotImplementedError("Method 'rewrite' must be overwritten.")
+
+    @classmethod
+    def check(cls, context, *_) -> bool:
+        return True
+
+
+def make_rewrite_rule_from_class(rule_class: type | RewriteRuleAsClass) -> RewriteRule:
```
gramalingam, line 1141: Is the plan to support both forms? Or, switch everything to the Class form?

gramalingam (approved):

onnxscript/rewriter/pattern.py
xadupre: Support both forms. The one I propose is just to make it easier to read if multiple pattern are implemented in the same file.

PR #1601, opened 2024-06-11
Add simple patterns for llama

xadupre:


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,73 @@
...
+import onnx.numpy_helper
+
 import onnxscript.ir as ir
 import onnxscript.rewriter.no_op as no_op
 import onnxscript.rewriter.pattern as orp
 
 op = orp.onnxop
 
 
+class CastIdentity(orp.RewriteRuleAsClass):
```
gramalingam, line 15: I am surprised if ORT doesn't implement this already. Still, I can see it being useful here.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,73 @@
...
+    """Replaces ``Cast(., to=to)`` by ``Identity`` if possible."""
+
+    @classmethod
+    def pattern(cls, op, x, to: int):
```
gramalingam, line 19: minor nit: the `int` annotation is not strictly correct here, or down below ... it will be a pattern-variable here, which will be bound to an attribute (of int type) in the methods below.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,73 @@
...
+    @classmethod
+    def rewrite(cls, op, x: ir.Value, to: int):
+        return op.Identity(x)
+
+    @classmethod
+    def check(cls, context, x, to: int) -> bool:
+        return x.dtype == to.value
+
+
+class CastCast(orp.RewriteRuleAsClass):
```
gramalingam, line 31: This is not strictly the same, right? If the first cast loses precision? Eg., what happens if you cast int32 to int16 and then back to int32? Would it be correct?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,73 @@
...
+    def pattern(cls, op, x, shape):
+        return op.Expand(x, shape)
+
+    @classmethod
+    def rewrite(cls, op, x: ir.Value, shape: ir.Value):
+        return op.Identity(x)
+
+    @classmethod
+    def check(cls, context, x, shape) -> bool:
+        shape_x = x.shape
```
gramalingam, line 56: I think you need an extra check to check that `shape` is a constant?

gramalingam: We might have some to-do task relating to this, let me search

onnxscript/rewriter/llama_rule_sets.py
xadupre: It probably does. But my patterns work like layers. You need this one so that the next one cast triggers.

onnxscript/rewriter/llama_rule_sets.py
xadupre: True, I can add some verifications to check it is not just about rounding. Cast + Cast is something happening a lot due to pytorch ExportedProgram and onnxscript as well.

onnxscript/rewriter/llama_rule_sets.py
xadupre: It would be better. I just hope it fails right now.

gramalingam: I see some example [here](https://github.com/microsoft/onnxscript/blob/47fb031b98da291d005a71a782f29f72514db806/onnxscript/rewriter/pattern_test.py#L261) ... so, two points: 
(i) Checking required is simpy checking `shape.const_value is not None`
(ii) The TODO item is to integrate constant-propagation into the IR/rewriter. For now, the workaround used is as shown in above example to use `_ir_utils.propagate_const_value`.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,73 @@
...
+        return shape_x.dims == tuple(shape.const_value.numpy().tolist())
+
+
+class ReshapeReshape(orp.RewriteRuleAsClass):
+    """Replaces ``Reshape(Reshape(X, ...), shape)`` by ``Reshape(X, shape)``."""
```
gramalingam, line 61: I think this is not always valid ... specifically, when -1 and 0 are is used in the shape. For example, consider `Reshape(Reshape(X, [128, 128, 128]), [256, 0, -1]))` Would checking that there are no 0 and -1 in shape work?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -76,8 +136,39 @@ def rewrite(cls, op, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr):
...
+    """Replaces ``Unsqueeze(Unsqueeze(., axes1), axes2)``
+    with one Unsqueeze.
+    """
+
+    @classmethod
+    def pattern(cls, op, x, axes1, axes2):
+        return op.Unsqueeze(op.Unsqueeze(x, axes1), axes2)
+
+    @classmethod
+    def _combine1(cls, axes1: np.ndarray, axes2: np.ndarray) -> np.ndarray:
```
gramalingam, line 149: Does this handle negative values as axes?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -9,6 +9,34 @@
...
+    def pattern(cls, op, x, to: int):
+        return op.Cast(x, to=to)
+
+    @classmethod
+    def rewrite(cls, op, x: ir.Value, to: int):
+        return op.Identity(x)
+
+    @classmethod
+    def check(cls, context, x, to: int) -> bool:
+        return x.dtype == to.value
```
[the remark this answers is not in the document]
xadupre: done

onnxscript/rewriter/llama_rule_sets.py
xadupre: True, I noticed but I forgot to change the annotation.

onnxscript/rewriter/llama_rule_sets.py
xadupre: Let's check it is a constant. We may do more if needed later.

onnxscript/rewriter/llama_rule_sets.py
xadupre: Right, I forgot that. I'll add those verification.

onnxscript/rewriter/llama_rule_sets.py
xadupre: No. I'll remove that as well.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,86 @@
...
+    @classmethod
+    def check(cls, context, x, to) -> bool:
+        return x.dtype == to.value
+
+
+class CastCast(orp.RewriteRuleAsClass):
+    """Replaces ``Cast(Cast(X, ...), to=to)`` by ``Cast(X, to=to)``."""
+
+    @classmethod
+    def pattern(cls, op, x, to: int, to0):
```
justinchuby, line 35: Maybe a more meaningful name for `to0`?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,86 @@
...
+
 import onnxscript.ir as ir
 import onnxscript.rewriter.no_op as no_op
 import onnxscript.rewriter.pattern as orp
 
 op = orp.onnxop
 
 
+class CastIdentity(orp.RewriteRuleAsClass):
+    """Replaces ``Cast(., to=to)`` by ``Identity`` if possible."""
```
justinchuby, line 16: Is it possible to elide the node in the optimizer?

onnxscript/rewriter/llama_rule_sets_test.py
```diff
@@ -127,6 +130,197 @@ def test_llama_p0_rule_set_transpose_transpose(self):
...
+                            np.array([5, 4, 3], dtype=np.int64), name="shape"
+                        ),
+                    ],
+                ),
+                opset_imports=[onnx.helper.make_opsetid("", 18)],
+            ),
+        ]
+        return models
+
+    def test_llama_p0_rule_set_resshape_reshape(self):
```
justinchuby, line 312: I would test this rule more to make robust

onnxscript/rewriter/llama_rule_sets.py
xadupre: Not sure. Using idenity avoids renaming everything and we can still remove all identity nodes after every iteration. It is very fast.

onnxscript/rewriter/llama_rule_sets.py
xadupre: ok

onnxscript/rewriter/llama_rule_sets_test.py
xadupre: Not really, if the second shape has only positive values, there is no ambiguity.

justinchuby:
For all test models, I would use onnxscript to create them when possible to establish it as the high level api for builiding onnx models.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -2,13 +2,84 @@
...
+    def check(cls, context, x, shape) -> bool:
+        if shape.const_value is None:
+            # Shape is not a constant and cannot be guessed.
+            return False
+        shape_x = x.shape
+        return shape_x.dims == tuple(shape.const_value.numpy().tolist())
+
+
+class ReshapeReshape(orp.RewriteRuleAsClass):
+    """Replaces ``Reshape(Reshape(X, ...), shape)`` by ``Reshape(X, shape)``."""
```
justinchuby, line 64: I would explain handling of `-1`, `0` shapes as part of the docstring.

onnxscript/rewriter/llama_rule_sets_test.py
```diff
@@ -19,7 +19,10 @@ class LlamaRuleSetsTest(unittest.TestCase):
     def _get_random_inputs(self, model: onnx.ModelProto) -> dict[str, Any]:
         feeds: dict[str, Any] = {}
         for i in model.graph.input:
-            shape = tuple(d + 2 for d in range(len(i.type.tensor_type.shape.dim)))
+            ish = tuple(i.type.tensor_type.shape.dim)
+            shape = tuple(
+                (d.dim_value if d.dim_value > 0 else i + 2) for i, d in enumerate(ish)
+            )
```
justinchuby, line 25: I would improve readability of this logic: `ish`, and why the change is needed

onnxscript/rewriter/llama_rule_sets_test.py
```diff
@@ -127,6 +130,217 @@ def test_llama_p0_rule_set_transpose_transpose(self):
...
+    def test_llama_p0_rule_set_reshape_reshape(self):
+        for model_proto in self._reshape_reshape_models():
+            ir_model = ir.serde.deserialize_model(model_proto)
+            rule_set = llama_rule_sets.llama_p0_rule_set()
+            rule_set.apply_to_model(ir_model)
+            rewritten_model = ir.serde.serialize_model(ir_model)
+
+            self.assertEqual(
+                ["Reshape"], [n.op_type for n in rewritten_model.graph.node]
+            )
```
justinchuby, line 341: I would leverage the IR for comparisons and avoid checks on the proto unless necessary.

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -76,8 +145,49 @@ def rewrite(cls, op, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr):
...
+    """Replaces ``Unsqueeze(Unsqueeze(., axes1), axes2)``
+    with one Unsqueeze.
+    """
+
+    @classmethod
+    def pattern(cls, op, x, axes1, axes2):
+        return op.Unsqueeze(op.Unsqueeze(x, axes1), axes2)
+
+    @classmethod
+    def _combine1(cls, axes1: np.ndarray, axes2: np.ndarray) -> np.ndarray:
```
justinchuby, line 158: `_combine1` -> meaningful name and docstring

onnxscript/rewriter/llama_rule_sets_test.py
xadupre: The condition applies on the proto because this is what we use at the end. Not sure it would be better with the ir.

onnxscript/rewriter/llama_rule_sets.py
xadupre: Ok. I added a comment. But the pattern is really really simple.

onnxscript/rewriter/llama_rule_sets_test.py
xadupre: I added a comment. About readibility, what do you have in mind? It looks ok to me.

onnxscript/rewriter/llama_rule_sets.py
xadupre: done

xadupre:
> For all test models, I would use onnxscript to create them when possible to establish it as the high level api for builiding onnx models.

I know well the onnx api but not yet onnxscript. The documentation is missing very easy example I can copy paste and change.

onnxscript/rewriter/llama_rule_sets_test.py
justinchuby: We should be confident that checking the IR _is_ checking the proto; this way the proto is only used as a serialization format and nothing more.

onnxscript/rewriter/llama_rule_sets_test.py
```diff
@@ -127,6 +133,264 @@ def test_llama_p0_rule_set_transpose_transpose(self):
...
+                        onnx.helper.make_node("Cast", ["X"], ["Y"], to=onnx.TensorProto.FLOAT),
+                    ],
+                    "name",
+                    [onnx.helper.make_tensor_value_info("X", FLOAT, [None, None, None])],
+                    [onnx.helper.make_tensor_value_info("Y", FLOAT, [None, None, None])],
+                ),
+                opset_imports=[onnx.helper.make_opsetid("", 18)],
+            ),
+        ]
+        return models
```
justinchuby, line 187:
```suggestion
        @onnxscript.script
        def model(x: FLOAT["a", "b", "c"]) -> FLOAT["a", "b", "c"]:
        y = op.Cast(x, to=onnx.TensorProto.FLOAT)
        return y
        
        proto = model.to_model_proto()
```

onnxscript/rewriter/llama_rule_sets_test.py
justinchuby: `["Reshape"], [n.op_type for n in ir_model.graph]`

onnxscript/rewriter/llama_rule_sets_test.py
```diff
@@ -127,6 +136,256 @@ def test_llama_p0_rule_set_transpose_transpose(self):
...
+            rule_set.apply_to_model(ir_model)
+            rewritten_model = ir.serde.serialize_model(ir_model)
+
+            self.assertEqual(["Cast"], [n.op_type for n in rewritten_model.graph.node])
+            self._check_model(model_proto, rewritten_model)
+
+    @classmethod
+    def _cast_identity_models(cls):
+        @onnxscript.script()
+        def model(x: ot.FLOAT["a", "b", "c"]) -> ot.FLOAT["a", "b", "c"]:
```
justinchuby, line 178:
```suggestion
        def model(x: ot.FLOAT["a", "b", "c"]) -> ot.FLOAT["a", "b", "c"]:  # noqa: F821, UP037
```

onnxscript/rewriter/llama_rule_sets.py
gramalingam: Is this a TODO? I don't see any checks. It would help to make sure we don't forget this.

gramalingam (approved):

onnxscript/rewriter/llama_rule_sets.py
xadupre: done, sorry, I forgot.

onnxscript/rewriter/pattern.py
```diff
@@ -1137,11 +1149,13 @@ def rewrite(cls, op, *_) -> Any:
...
 
     @classmethod
-    def check(cls, context, *_) -> bool:
+    def check(cls, context, *_, **__) -> bool:
         return True
 
 
-def make_rewrite_rule_from_class(rule_class: type | RewriteRuleAsClass) -> RewriteRule:
+def make_rewrite_rule_from_class(
+    rule_class: type | RewriteRuleAsClass, generic: bool = False
```
justinchuby, line 1157: Wanted to understand: why do we need this option to distinguish between them? I would think we have already merged the two algorithms and the algo choice was internal.

onnxscript/rewriter/pattern.py
```diff
@@ -1171,7 +1185,22 @@ def rewrite(cls, op, x: ir.Value, perm: ir.Attr | None = None):
...
-    return RewriteRule(rule_class.pattern, rule_class.rewrite, rule_class.check)
+    if generic:
+        import onnxscript.rewriter.generic_pattern as orpp
+
+        return RewriteRule(
+            rule_class.pattern,
+            rule_class.rewrite,
+            rule_class.check,
+            orpp.GenericPatternMatcher,
+            name=rule_class.__name__,  # type: ignore[union-attr]
```
justinchuby, line 1196: I am concerned about breaking this into different paths again.

onnxscript/rewriter/pattern.py
justinchuby: Ideally this shouldn't be a user choice, as seen here.

onnxscript/rewriter/pattern.py
```diff
@@ -1048,6 +1049,7 @@ def __init__(
             matcher: The pattern matcher that will be used to match the pattern.
                 If not provided, a default matcher will be used.
             verbose: The verbosity level of the rule.
+            name: for debugging purpose
```
justinchuby, line 1052: Would there be a way for formalize this param? Maybe include a recommended way of using it? Something along the lines of "An optional name for the pattern that will show up in verbose logging" etc.

onnxscript/rewriter/pattern.py
xadupre: why?

onnxscript/rewriter/pattern.py
xadupre: done

onnxscript/rewriter/pattern.py
xadupre: Two algorithms not doing exactly the same thing. The can switch if one is failing.

PR #1603, opened 2024-06-12
[INV] Add a test with a model using FFT

xadupre:
The model is decomposed using fft_r2c, fft_c2r. The first function outputs complex. Then for some operators, pytorch and onnx disagree on the result type (complex for pytorch and float for onnx).


justinchuby:
The disagreement is expected because ONNX does not produce any complex types. We need to reinterpret the floating-point outputs from ONNX when we expect complex outputs.

When the complex values are intermediate values, the exporter is able to match the dtypes and interpret the ONNX values correctly. When they are outputs of a model, the user needs to reinterpret the values.

PR #1593, opened 2024-06-07
Add llama model to benchmark

xadupre:

PR #1606, opened 2024-06-13
[torchlib] Fix names for registered functions

justinchuby:
Fix the name for `getitem` and `adaptive*` ops.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3669,8 +3669,7 @@ def aten_ger(self: TensorType, vec2: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-# NOTE: The name is made up for `getitem` to be included in the registry
-@torch_op("aten::getitem")
+@torch_op("_operator::getitem")
 def aten_getitem(self: Sequence[TTensor], i: INT64) -> TTensor:
     return op.SequenceAt(self, i)
 
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3560,7 +3560,7 @@ def aten_full(size: INT64, fill_value: FLOAT, dtype: int = FLOAT.dtype):
 
 
 @torch_op("aten::full_like")
-def aten_full_like(self, fill_value: TTensor) -> TTensor:
+def aten_full_like(self: TTensor, fill_value: TTensor) -> TTensor:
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -40,7 +40,7 @@
 TFloatUnlessFloat32 = TypeVar("TFloatUnlessFloat32", bound=Union[BFLOAT16, FLOAT16, DOUBLE])
 
 
-@torch_op("aten::aten_adaptive_avg_pool1d", traceable=True)
+@torch_op("aten::adaptive_avg_pool1d", traceable=True)
```
justinchuby, line 43: Is there decomp for these?

titaiwangms: Yes, it's forced to be decomposed, just like upsampling.

justinchuby: If that's the case, we may be better off forcing the decomp since we don't have full support for adaptive pooling

titaiwangms: It usually starts with `@register_decomposition()`, but for this one, I haven't found it yet.

justinchuby:
Merging for now

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3669,8 +3669,7 @@ def aten_ger(self: TensorType, vec2: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-# NOTE: The name is made up for `getitem` to be included in the registry
-@torch_op("aten::getitem")
+@torch_op("_operator::getitem")
```
titaiwangms, line 3672: Although it's fixed by #1593, aten::getitem was used in https://github.com/pytorch/pytorch/blob/ee140a198fbe8e7edb5f51806f63e9fa80f523fe/torch/onnx/_internal/fx/onnxfunction_dispatcher.py#L300-L303. This made a regression, and would probably break this week torchbench.

titaiwangms: Is this part of the refactoring you are doing on dispatcher?

justinchuby: hmm ok. We can add the original back

justinchuby: Looks like fixed in https://github.com/microsoft/onnxscript/pull/1593

PR #1608, opened 2024-06-14
Fix SliceSplit generic pattern

xadupre:


onnxscript/rewriter/generic_pattern.py
```diff
@@ -333,8 +333,8 @@ def _match_forward(
         starting_node: ir.Node,
         matched: dict[orp.NodePattern, ir.Node],
         stack: list[orp.NodePattern],
-        graph_node: ir.Node,
-        pattern_node: orp.NodePattern,
+        graph_node: ir.Node | ir.TensorProtocol,
```
justinchuby, line 336: Curious why we see tensors here?

xadupre:
Closing at #1620 is handling this.

PR #1609, opened 2024-06-14
Add CI to check with onnxruntime-traning and transformers

xadupre:


justinchuby (changes requested):
Please leverage nox to ensure we manage tests in a single place and allow the same test environment to be created locally.

tests/ir/serde_roundtrip_test.py
justinchuby, line 1: pyinstrument is part of the dev dependency. So we expect to have it.

xadupre: Can we remove the test? I only found one test and it is not checking a functionality of onnxscript but just checking pyinstrument is working with onnxscript. So if it fails, it is mostly because of pyinstrument and not because of onnxscript. According to the documentation, it can be very slow in a docker: https://github.com/joerick/pyinstrument?tab=readme-ov-file#known-issues.

justinchuby: Sure! That works too

justinchuby (approved):
LGTM thanks! There is a step in the ci job

```
      - name: Upload IR profiling results
        if: matrix.name == 'py311' || matrix.name == 'py311-onnx-weekly'
        uses: actions/upload-artifact@v3
        with:
          name: IR profiling results
          path: tests/ir/serde_test_profiles
```

That can be removed too.

gramalingam (approved):

justinchuby:
https://github.com/microsoft/onnxscript/issues/1635

PR #1610, opened 2024-06-14
Miscellaneous imporovements

justinchuby:
- Correctly register some torchlib functions
- Make `aten::cat` trace only
- Capture error stacks in ir.serde for a better view of errors
- Add function support in ir.to_proto / ir.from_proto
- Fix producer display in Value repr
- Add repr to ONNXFunction

### DIFF
--- a/onnxscript/ir/serde.py
+++ b/onnxscript/ir/serde.py
@@ -12,6 +12,7 @@
 #     plain Python objects.
 
 from __future__ import annotations
+
 import functools
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


justinchuby:
Replaced by https://github.com/microsoft/onnxscript/pull/1612

onnxscript/ir/serde.py
```diff
@@ -509,6 +544,7 @@ def deserialize_graph(proto: onnx.GraphProto) -> _core.Graph:
     return _deserialize_graph(proto, [])
 
 
+@_capture_errors(lambda proto: proto.name)
```
gramalingam, line 547: Shouldn't the lambda have the same arguments as the function below, which has an extra scoped_values?

justinchuby: Good catch thanks!

PR #1611, opened 2024-06-14
Print onnxfunction repr

justinchuby:


gramalingam (approved):

justinchuby:
I think the main thing was merged so this is good.

PR #1612, opened 2024-06-14
Miscellaneous imporovements

justinchuby:
- Correctly register some torchlib functions
- Make `aten::cat` trace only
- Capture error stacks in ir.serde for a better view of errors
- Add function support in ir.to_proto / ir.from_proto
- Fix producer display in Value repr
- Add repr to ONNXFunction

### DIFF
--- a/onnxscript/ir/serde.py
+++ b/onnxscript/ir/serde.py
@@ -544,7 +544,7 @@ def deserialize_graph(proto: onnx.GraphProto) -> _core.Graph:
     return _deserialize_graph(proto, [])
 
 
-@_capture_errors(lambda proto: proto.name)
+@_capture_errors(lambda proto, scoped_values: proto.name)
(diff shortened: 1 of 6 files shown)


onnxscript/ir/serde.py
```diff
@@ -717,6 +758,7 @@ def deserialize_dimension(
     return _core.SymbolicDim(None), denotation
 
 
+@_capture_errors(lambda proto: proto.name)
```
gramalingam, line 761: Here too: `lambda proto, *_:` ?

gramalingam (approved):

onnxscript/ir/serde.py
justinchuby: I did anther pass to fix everything (had more errors). PTAL. Thanks!

PR #1613, opened 2024-06-14
[torchlib] Fix more signatures

justinchuby:
Fix more signatures in torchlib that were previously overlooked


gramalingam (approved):

titaiwangms (approved):
Please run this change with main torch to make sure there is no broken op tests.

PR #1617, opened 2024-06-15
Add utility to remove unused values/nodes in IR

gramalingam:
Mostly a re-implementation of the existing proto-based optimization to remove unused-values/nodes to use the IR.


onnxscript/optimizer/remove_unused_ir.py
```diff
@@ -0,0 +1,93 @@
...
+        def is_used_output(i: int) -> bool:
+            if i < len(node.outputs):
+                val = node.outputs[i]
+                return val in graph_outputs or val.uses()
+            return False
+
+        if is_used_output(1) or is_used_output(2):
+            return
+        node.outputs[1].name = ""
+        node.outputs[2].name = ""
```
gramalingam, line 36: The IR does not seem to allow us to remove entries from node.outputs. (The alternative seems to be to create a new node to replace the existing node.) For now, hope this works.

onnxscript/optimizer/remove_unused_ir.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def process_function_or_graph(function_or_graph: ir.Function | ir.Graph) -> int:
+    graph_outputs = frozenset(function_or_graph.outputs)
+    onnx_opset_version = function_or_graph.opset_imports.get("", None)
+    count = 0
+    for node in reversed(function_or_graph):
+        removable = True
+        for output in node.outputs:
+            if output in graph_outputs or output.uses():
```
gramalingam, line 62: Maybe it would be useful to add an utility method to ir.Graph/Function to abbreviate this check: a method `is_used(self, value)` ?

justinchuby (approved):

onnxscript/optimizer/remove_unused_proto.py
justinchuby, line 1: Should we make these two new files private?

onnxscript/optimizer/remove_unused_ir.py
```diff
@@ -0,0 +1,93 @@
...
+            for attr in node.attributes.values():
+                if isinstance(attr, ir.AttrGraph):
+                    count += process_function_or_graph(attr.value)
+                elif isinstance(attr, ir.AttrGraphs):
+                    for graph in attr.value:
+                        count += process_function_or_graph(graph)
+    return count
+
+
+def remove_unused_nodes(model: ir.Model) -> None:
```
titaiwangms, line 80: We should have tests to guard this api, although it's fixed in https://github.com/microsoft/onnxscript/pull/1628

titaiwangms: This is called in torch converter.

PR #1620, opened 2024-06-16
Fix multi-output pattern-matcher bug

gramalingam:
See unit test below for example of pattern not handled by matcher. Basically, match-forward needs to be done for "values" as well, in addition to, "nodes".

### DIFF
--- a/onnxscript/rewriter/generic_pattern.py
+++ b/onnxscript/rewriter/generic_pattern.py
@@ -532,7 +532,7 @@ class GenericPatternMatcher(orp.PatternMatcher):
             if result is None:
                 return result
             match_count += result
-            
+
         if self.verbose > 5 and match_count > 0:
             print(f"[GenericPatternMatcher._match_forward] add {match_count} nodes")
         return match_count
--- a/onnxscript/rewriter/generic_pattern.py
+++ b/onnxscript/rewriter/generic_pattern.py
@@ -300,7 +300,11 @@ class GenericPatternMatcher(orp.PatternMatcher):
             # TODO(rama): Handle constant-pattern
             pattern_pred = pattern_value.producer()
             if pattern_pred is None:
-                # pattern_pred is None means the pattern ends here.
+                # pattern_pred is None means the pattern backward search ends here.
+                result = self._match_values_forward(starting_node, matched, stack, graph_value, pattern_value)
+                if result is None:
+                    return result
+                match_count += result
                 continue
             graph_pred = graph_value.producer()
             if graph_pred is None:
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/generic_pattern.py
```diff
@@ -300,7 +300,11 @@ def _match_backward(
             # TODO(rama): Handle constant-pattern
             pattern_pred = pattern_value.producer()
             if pattern_pred is None:
-                # pattern_pred is None means the pattern ends here.
+                # pattern_pred is None means the pattern backward search ends here.
+                result = self._match_values_forward(starting_node, matched, stack, graph_value, pattern_value)
```
xadupre, line 304: I fixed the same bug here https://github.com/microsoft/onnxscript/pull/1608 but I inserted a similar call at a different place. I think we can bypass this call if ``len(graph_value.uses()) == 1`` to avoid adding too much time.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -328,6 +332,164 @@ def _match_backward(
             print(f"[GenericPatternMatcher._match_backward] add {match_count} nodes")
         return match_count
 
+    def _match_values_forward(
```
xadupre, line 335: Did you change anything here? It is difficult to check since the code was moved.

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -246,6 +247,40 @@ def get_rotary_model(self):
...
+            """
+            <ir_version: 7, opset_import: [ "" : 17]>
+            agraph (float[N] y) => (float[N] z)
+            {
+                temp1 = Sin(y)
+                temp2 = Cos(y)
+                z = Add(temp1, temp2)
+            }
+        """
+        )
```
justinchuby, line 274: Currently we have three ways to build test models: onnx helper, text repr, and ONNX Script. I think it would be good if we consolidate to use one, preferably ONNX Script.

xadupre: I think for unit tests people are using whatever is faster for them. For this particular case, using something else than onnxscript may be better. You need to make sure that the optimized graph is exactly what you have in mind. When you write something like ``op.Split(x, num_outputs=2, axis=-1)``, you need to make sure there are two outputs. onnxscript may have a bug plus you can choose the node and results names more easily.

gramalingam: I personally find both the text repr and onnxscript equally compact and readable. But, as Xavier mentions, with the text repr we know exactly what the test-model is (onnxscript could do some expansion, etc.).

Anyway, hope this is not a blocking comment. The PR is ready to be merged with an approval

justinchuby: Sg. Thanks!

justinchuby (approved):

PR #1622, opened 2024-06-17
Add first rewriting patterns for llama adding onnxruntime contrib ops

xadupre:


titaiwangms (commented):
Currently, we separate ort rewriter to this folder https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/onnxruntime/__init__.py

onnxscript/rewriter/onnxruntime/fused_matmul_rule_sets.py
```diff
@@ -0,0 +1,179 @@
...
+        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
+        for name in ["transA", "transB", "transBatchA", "transBatchB"]:
+            att = node.attributes.get(name)
+            if att:
+                kwargs[name] = att.value
+        return op.FusedMatMul(x, y, **kwargs, domain="com.microsoft")
+
+
+class _TransposeMatMulBase(orp.RewriteRuleAsClass):
+    _pos: ClassVar = 1
```
gramalingam, line 67: Maybe add a comment documenting this? Something like "Indicates the input of MatMul that is transposed."

gramalingam: So you have patterns where only one of the two inputs is transposed? Wondering if eventually you would need to support both inputs being transposed.

gramalingam (approved):

onnxscript/rewriter/onnxruntime/fused_matmul_rule_sets.py
xadupre: I have but I'll add them in another PR.

PR #1629, opened 2024-06-17
Support optional attribute checking in matcher

gramalingam:
Extend matcher to allow users to specify whether all attributes must be exactly as in pattern. Change default-value to allow extra-attributes in actual node, not specified in pattern.

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -427,6 +427,15 @@ class NodePattern:
     This differs from a NodeOutputPattern in that it matches against a node (which
     may produce 1 or more outputs), whereas a NodeOutputPattern matches against
     a specific output of a node.
+
+    Args:
+        domain: pattern to match against the domain of the node.
+        op: pattern or string constant to match against the op_type of the node.
+        inputs: sequence of ValuePatterns (or constants) to match against the inputs of the node.
+        attributes: dictionary of attribute patterns to match against the attributes of the node.
+        outputs: specifies pattern-variable-name for outputs (or None)
+        allow_other_attributes: specifies whether other attributes (not mentioned in `attributes`)
+          are allowed in the node.
     """
 
     def __init__(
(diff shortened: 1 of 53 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -205,6 +205,7 @@ def __call__(
         domain: str | None = None,
         version: int | None = None,
         outputs: int | list[str | None] = 1,
+        _allow_other_attributes: bool | None = None,
```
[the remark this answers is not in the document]
gramalingam: Not really.  It is more to avoid conflicts with ONNX attribute names. We discussed doing something similar for other names as well, like domain/version/outputs. Any thoughts about this? May be a good time to do it.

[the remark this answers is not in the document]
justinchuby: I see. I don't have a better idea yet but I am wishing for one

justinchuby (approved):
LGTM based on discussion. I would document the decision and guideline on how we name the parameters

onnxscript/rewriter/pattern.py
```diff
@@ -228,7 +229,9 @@ def __call__(
             raise ValueError("outputs must be an int or a list[str|None].")
         inputs = [_to_value_pattern(x) for x in args]
         attributes = {name: _to_attr_pattern(value) for (name, value) in kwargs.items()}
-        node_pattern = NodePattern(opset_pattern, self.op_name, inputs, attributes, outputs)
+        node_pattern = NodePattern(
+            opset_pattern, self.op_name, inputs, attributes, outputs, _allow_other_attributes
```
justinchuby, line 233: Nit: call with keywords for kwargs

onnxscript/rewriter/pattern.py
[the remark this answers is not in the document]
gramalingam: (Post internal discussion comment): My personal preference and recommendation is to use kwargs with a leading underscore for domain/version/outputs/ etc. This
* Meets the goal of distinguishing ONNX op attributes from builder's custom options,
* Involves the least typing effort for users and least cognitive overhead in reading-or-writing (except for using names like domain/version, which fails the first criterion).

justinchuby:
Should the `NodePattern` class use the underscore version too? I think we can remove underscore there because attributes are not unpacked.

 https://github.com/microsoft/onnxscript/blob/95927da57f79cc7a575aaaa500e1dd7709935139/onnxscript/rewriter/pattern.py#L439C9-L439C32

justinchuby:
I would also add a docstring for NodePattern

onnxscript/rewriter/pattern.py
```diff
@@ -205,6 +205,7 @@ def __call__(
         domain: str | None = None,
         version: int | None = None,
         outputs: int | list[str | None] = 1,
+        _allow_other_attributes: bool | None = None,
         **kwargs,
     ):
```
justinchuby, line 210: I would add a docstring for this to explain the params

onnxscript/rewriter/pattern.py
```diff
@@ -433,11 +436,16 @@ def __init__(
         inputs: Sequence[int | float | ValuePattern | None],
         attributes: dict[str, AttrPattern],
         outputs: Sequence[str | None],
+        _allow_other_attributes: bool | None,
```
justinchuby, line 439: docstring

onnxscript/rewriter/pattern_test.py
```diff
@@ -368,6 +368,58 @@ def double(op, x):
...
+                z = ConcatFromSequence (t)
+            }
+        """
+        )
+        model = ir.serde.deserialize_model(model_proto)
+        count = rule.apply_to_model(model)
+        self.assertEqual(count, 1)
+        self.assertEqual(len(model.graph), 1)
+        self.assertEqual(model.graph[0].op_type, "Concat")
+        self.assertFalse("axis" in model.graph[0].attributes)
```
justinchuby, line 421: assertNotIn

PR #1636, opened 2024-06-19
Add optimization patterns fusing ops into custom ops

xadupre:


onnxscript/rewriter/generic_pattern.py
```diff
@@ -297,20 +299,33 @@ def _match_backward(
...
-                    graph_node,
-                )
-                return self.none(starting_node, inspect.currentframe().f_lineno)
+        has_predecessor_in_pattern = any(
+            pattern_value.producer() is not None for pattern_value in pattern_node.inputs
+        )
+
+        if has_predecessor_in_pattern:
+            for graph_input, pattern_input in zip(graph_node.inputs, pattern_node.inputs):
+                if len(list(graph_input.uses())) != len(list(pattern_input.uses())):
```
justinchuby, line 308:
```suggestion
                if len(graph_input.uses()) != len(pattern_input.uses()):
```

onnxscript/rewriter/generic_pattern.py
```diff
@@ -297,20 +299,33 @@ def _match_backward(
...
+                        graph_node,
+                    )
+                    return self.none(starting_node, inspect.currentframe().f_lineno)
 
         for graph_value, pattern_value in zip(graph_node.inputs, pattern_node.inputs):
-            # TODO(rama): Handle constant-pattern
             pattern_pred = pattern_value.producer()
+            if pattern_pred is not None and len(list(graph_value.uses())) != len(
+                list(pattern_value.uses())
+            ):
```
justinchuby, line 322:
```suggestion
            if pattern_pred is not None and len((graph_value.uses()) != len(
                (pattern_value.uses()
            ):
```

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
```diff
@@ -0,0 +1,495 @@
...
+
+
+class CombinedAddAdd1(_CombineBinary):
+    @classmethod
+    def pattern(cls, op, x, y, z):
+        return op.Add(x, op.Add(y, z))
+
+    @classmethod
+    def rewrite(cls, op, x, y, z):
+        return op.AddAdd(x, y, z, domain="ai.onnx.contrib")
```
justinchuby, line 58: What is the `ai.onnx.contrib` domain?

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
```diff
@@ -0,0 +1,495 @@
...
+
+def llm_rule_set_cuda() -> orp.RewriteRuleSet:
+    """Returns a set of rules fusing nodes into custom kernels.
+
+    Returns:
+        RewriteRuleSet
+    """
+    return orp.RewriteRuleSet(
+        [
+            orp.make_rewrite_rule_from_class(AddSharedInput1, True),
```
justinchuby, line 461: Prefer calling by keyword for boolean args to make it more readable

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
xadupre: kernel implemented in onnxruntime-extensions

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
xadupre: done

justinchuby:
Helpful to create a Pr description? Thanks! https://google.github.io/eng-practices/review/developer/cl-descriptions.html

onnxscript/rewriter/generic_pattern.py
```diff
@@ -297,20 +299,33 @@ def _match_backward(
...
-                    pattern_node,
-                    "-- model",
-                    graph_node,
-                )
-                return self.none(starting_node, inspect.currentframe().f_lineno)
+        has_predecessor_in_pattern = any(
+            pattern_value.producer() is not None for pattern_value in pattern_node.inputs
+        )
+
+        if has_predecessor_in_pattern:
```
gramalingam, line 306: Is this condition meant to be an optimization (with no effect in behavior)? Trying to understand

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
```diff
@@ -0,0 +1,495 @@
...
+
+    @classmethod
+    def rewrite(cls, op, x, y):
+        return op.ReplaceZero(x, y, equal=0, domain="ai.onnx.contrib")
+
+
+class Rotary1(orp.RewriteRuleAsClass):
+    @classmethod
+    def pattern(cls, op, x):
+        x1, x2 = op.Split(x, axis=-1, outputs=2, num_outputs=2)
```
gramalingam, line 323: This reminds me: I think the generic pattern matcher doesn't check for attribute-values like "axis=-1". It is easy to add (the underlying pattern object has a match method to do it), but I think this is still a to-do. Trying to understand whether you want the check here? If so, it can't be working right now?

onnxscript/rewriter/onnxruntime/fused_matmul_rule_sets.py
```diff
@@ -167,13 +167,13 @@ def fused_matmul_rule_sets() -> orp.RewriteRuleSet:
...
         [
-            orp.make_rewrite_rule_from_class(FusedMatMulDiv1, True),
-            orp.make_rewrite_rule_from_class(FusedMatMulDiv2, True),
-            orp.make_rewrite_rule_from_class(FusedMatMulTranspose, True),
-            orp.make_rewrite_rule_from_class(MatMulTranspose, True),
-            orp.make_rewrite_rule_from_class(TransposeMatMul1, True),
-            orp.make_rewrite_rule_from_class(TransposeFusedMatMul1, True),
-            orp.make_rewrite_rule_from_class(TransposeMatMul2, True),
-            orp.make_rewrite_rule_from_class(TransposeFusedMatMul2, True),
+            orp.make_rewrite_rule_from_class(FusedMatMulDiv1, generic=True),
```
gramalingam, line 170: Trying to understand why you have this `generic=True`? Is it for debugging?

xadupre: No it is using the generic pattern matcher (multi output pattern)

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
gramalingam: Technically, the ONNX org owns the `ai.onnx` prefix. It is probably not a great idea for microsoft to use that without some approval from some ONNX authority ... not sure why they didn't use `com.microsoft.contrib` ?

gramalingam: Idea being to avoid name conflicts ... e.g., if AMD, Nvidia, etc. start doing the same thing, it would be a problem.

onnxscript/rewriter/generic_pattern.py
gramalingam: I think this fix might be related to an issue Jingyan is running into. Does this fix allow pattern-inputs to have other uses outside of the pattern?

gramalingam: I think so. I think we should change this slightly, by applying the condition for each pattern_value separately, as mentioned in comment below.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -297,20 +299,33 @@ def _match_backward(
...
+            for graph_input, pattern_input in zip(graph_node.inputs, pattern_node.inputs):
+                if len(graph_input.uses()) != len(pattern_input.uses()):
```
gramalingam, line 308: Instead of the has_predecessor_in_pattern check above, I think we should do the check here: add `if pattern_input.producer() is not None` here ... that will handle the case where some inputs have producer and others don't.

onnxscript/rewriter/generic_pattern.py
```diff
@@ -297,20 +299,33 @@ def _match_backward(
...
+                        pattern_node,
+                        "-- model",
+                        graph_node,
+                    )
+                    return self.none(starting_node, inspect.currentframe().f_lineno)
 
         for graph_value, pattern_value in zip(graph_node.inputs, pattern_node.inputs):
-            # TODO(rama): Handle constant-pattern
             pattern_pred = pattern_value.producer()
+            if pattern_pred is not None and len(graph_value.uses()) != len(
```
gramalingam, line 320: I am confused ... isn't this already handled above? Why this check on the number of uses again?

onnxscript/rewriter/onnxruntime/fused_matmul_rule_sets.py
gramalingam: If the pattern has multiple outputs, the implementation should automatically choose the generic matcher ... user doesn't have to specify it.

onnxscript/rewriter/custom_ops/llm_rule_sets_cuda.py
xadupre: You are right, it does not. I forgot to add that verification.

xadupre: done

onnxscript/rewriter/generic_pattern.py
xadupre: Fill free to rewrite the code. If the unit test are passing, it means it is good.

onnxscript/rewriter/generic_pattern.py
xadupre: changed

onnxscript/rewriter/generic_pattern.py
xadupre: The rest of the loop can be executed for the first input if it satisfies the conditions and there is a case where it should not be done.

xadupre: Fill free to find a better solution. The unit test should catch any weird behaviour.

gramalingam (approved):

PR #1704, opened 2024-06-24
[IR] Create a convenience function to create name->value mappings

justinchuby:
cc @gramalingam . I realized we also cannot maintain the mapping in the IR, because we don’t require values to always have a name while in IR.

### DIFF
--- a/onnxscript/ir/_convenience.py
+++ b/onnxscript/ir/_convenience.py
@@ -373,6 +373,8 @@ def tensor(
 def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
     """Return a dictionary mapping names to values in the graph.
 
+    The mapping does not include values from subgraphs.
+
     Args:
         graph: The graph to extract the mapping from.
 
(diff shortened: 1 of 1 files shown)


onnxscript/ir/_convenience.py
```diff
@@ -368,3 +368,29 @@ def tensor(
             doc_string=name,
         )
     return tensor_
+
+
+def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
+    """Return a dictionary mapping names to values in the graph.
+
+    The mapping does not include values from subgraphs.
```
gramalingam, line 376: Maybe it's useful to consider how to handle subgraphs? Or, to be more precise, outer graphs? Eg.: if we are given a subgraph, should we construct a dictionary that includes outer scope names visible?

But it's not a priority, since we don't have a specific need yet.

gramalingam (approved):

onnxscript/ir/_convenience.py
justinchuby: Good point thanks! Currently we don't have a pointer from subgraphs to outer graphs yet. It's something I am thinking to add.

PR #1708, opened 2024-06-25
Test torch.onnx.export(..., dynamo=True)

xadupre:


gramalingam (approved):
Some description of the PR (what is it trying to fix?) would be helpful. I am a bit unclear about this.

gramalingam:
We should also find out about python 3.8 support in pytorch ... if it is necessary there, we will need to continue supporting it

PR #1709, opened 2024-06-27
Add unit test to check backward function for conv, checks there is no graph breaks

xadupre:


onnxscript/function_libs/torch_lib/backward_test.py
justinchuby, line 1: In the OpInfo data structure, I have seen a field that says supports_grad or something which may make it easier for us to generate backward tests. @xiaowuhu do you have some ideas?

xiaowuhu: This seems be a different scenario than the OpInfo way. Here, we need to go through the aot-compile-training-backward process which is an e2e scenario, although it is not a straight forward way.  But this requirement will only benefit not more than 20 backward functions, so I think it is OK.

justinchuby: SG. Thanks!

PR #1716, opened 2024-07-03
Add aten::scatter.value function

xiaowuhu:
This new aten function will fix a lot missing backward function issue related to "softmaxentropy".

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7369,6 +7369,19 @@ def aten_scalar_tensor_sym_number(
     return common_ops.cast_to(s, dtype=dtype)
 
 
+@torch_op("aten::scatter.value")
+def aten_scatter(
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7369,6 +7369,19 @@ def aten_scalar_tensor_sym_number(
     return common_ops.cast_to(s, dtype=dtype)
 
 
+@torch_op("aten::scatter.value")
```
justinchuby, line 7372: traceable

justinchuby, line 7372: traceable

justinchuby (approved):
Thank you!

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Added.

PR #1724, opened 2024-07-08
Add aten_prod function

xiaowuhu:
The backward routine need aten_prod.dim_int function.
Todo: add test case for this function.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6583,10 +6583,11 @@ def aten_prelu_backward(
     raise NotImplementedError()
 
 
-@torch_op(("aten::prod", "aten::prod.dim_int"), trace_only=True)
+@torch_op(("aten::prod.dim_int"), trace_only=True)
 def aten_prod(self: TReal, dim: int, keepdim: bool = False) -> TReal:
     """prod(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
 
+    # Todo: add test for this function later
     return op.ReduceProd(self, axes=[dim], keepdims=keepdim)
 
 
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6583,10 +6583,11 @@ def aten_prelu_backward(
     raise NotImplementedError()
 
 
-def aten_prod(self: TensorType, dtype: Optional[int] = None) -> TensorType:
+@torch_op(("aten::prod", "aten::prod.dim_int"), trace_only=True)
+def aten_prod(self: TReal, dim: int, keepdim: bool = False) -> TReal:
     """prod(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.ReduceProd(self, axes=[dim], keepdims=keepdim)
```
justinchuby, line 6590: Are tests available?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6583,10 +6583,11 @@ def aten_prelu_backward(
     raise NotImplementedError()
 
 
-def aten_prod(self: TensorType, dtype: Optional[int] = None) -> TensorType:
+@torch_op(("aten::prod", "aten::prod.dim_int"), trace_only=True)
```
justinchuby, line 6586:
```suggestion
@torch_op("aten::prod.dim_int", trace_only=True)
```

this should be good enough, since prod does not have a `.default` overload

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
xiaowuhu: Added 'Todo: add test for this function later'. Will do that after finish all the backward functions.

Issue #1701, opened 2024-06-22
Use IR in the optimizer


justinchuby:
And improve robustness so it is reliable enough to be turned on by default.

justinchuby:
The ExternalTensor support is working only when code is launched at the same directory as the model file. aka. we read the external data file using relative path. This is because when a model is loaded as a proto, its path information is lost to the IR. We need to design a way to handle when the model file is somewhere else.

This is relevant because I assume the optimizer will sometimes fold external tensors. Some examples may include `Transpose(x)` and element wise computation, casting, etc.

gramalingam:
Related: https://github.com/microsoft/onnxscript/issues/1436

Issue #1438, opened 2024-04-23
[IR] Support symbolic values?


justinchuby:
Create a field in Value to support representing equivalence relations among values.

1. Value' is an alias of Value, e.g. output of identity or `+0`
2. Value is a Sequence containing known values as elements

justinchuby:
Before this is created, I recommend using a key in the `meta` dictionary to store this information. It is possible to create a helper function to interact with the dictionary. Replacing it with the new field when the field is implemented should be easy.

Another question is how we can make the field robust. Is there a lightweight pass we can run every time the graph is changed? Do we need to create any invariance on this field?

justinchuby:
cc @gramalingam 

I came up with a new idea for this. It may be helpful to create an auxiliary data structure for this, rather than making it a field in values since it is more usage specific. Consider a data structure with these apis:

1. `mark_equivalence(value_1, value_2)` and `mark_equivalence(value_1, [<sequence of values>])` Establishes that two values are equivalent
1. `unmark_equivalence(value_1, value_2)` breaks the equivalence relationship
1. `assign(value, ir.Tensor)` binds a constant tensor to a value, and internally also binds it to all values that share the equivalence relation
1. `are_equivalent(value_1, value_2)` returns whether two values are the same
1. etc.

This way it is very easy to establish and manage equivalence relationships. Internally it can be represented as a tree like structure.

gramalingam:
Keeping it a separate data-structure makes sense: making it clear it is the user's responsibility in all regards. Agree with that. However, a dictionary mapping an IR value to its symbolic value would work better for this particular usage than an equivalence-relation data-structure. I think I will try that.

justinchuby:
Closed. Users will maintain their own data structure for this information. See the constant folding pass.

PR #1725, opened 2024-07-09
Migrating optimizer to new IR (part 1)

gramalingam:
Migrated the core logic. 

TO DO: decide how to handle functions. Optimizer currently incorporates function-specialization. Need to choose between function-specialization and function-inlining.


onnxscript/optimizer/constant_folding_ir.py
justinchuby, line 1: Can this be private?

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+import math
+from typing import Any, Callable, Sequence, Union
+
+import numpy as np
+import onnx
+import onnx.reference.ops
+
+import onnxscript.ir as ir
+import onnxscript.ir._convenience as _convenience
+import onnxscript.ir._enums as _enums
```
justinchuby, line 19: enums are exposed in `ir`

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+import onnxscript.ir.serde as serde
```
justinchuby, line 20: Conventionally I suggest using `ir.serde` for consistency
```suggestion
```

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+logger = logging.getLogger(__name__)
+
+# "Standard" evaluators are used to perform constant-folding.
+# The API below works only for non-control-flow ops (ops without any graph-attributes).
+# This currently used ONNX's reference implementation. But we could also
+# use ORT's implementation if we want to.
+
+
+class ReferenceEvaluator:
+    def get_evaluator(self, domain: str, op: str, version: int) -> callable | None:
```
justinchuby, line 55:
```suggestion
    def get_evaluator(self, domain: str, op: str, version: int) -> Callable | None:
```

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+
+    def evaluate(self, domain: str, op: str, version: int, *args, **kwargs) -> Any:
+        logger.debug("Evaluating %s::%s", domain, op)
+        evaluator = self.get_evaluator(domain, op, version)
+        if evaluator is None:
+            return None
+        return evaluator(*args, **kwargs)
+
+
+reference_evaluator = ReferenceEvaluator()
```
justinchuby, line 70: Can this be private?

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+    if val is None:
+        return None
+    if hasattr(val, "symbolic_value"):
+        return val.symbolic_value
+    return None
+
+
+def _get_numpy_value(val: ir.Value) -> np.ndarray | None:
+    const_value = val.const_value
+    if hasattr(const_value, "numpy"):
```
justinchuby, line 162: This should always be true?

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+def _update_type(value: ir.Value, type: ir.TypeProtocol) -> None:
+    # TODO: merge types
+    value.type = type
+
+
+def _get_input_element_type(node: ir.Node, index: int) -> int:
+    input = _get_input(node, index)
+    if input is not None and input.type is not None:
+        return input.type.dtype.value
+    return _enums.DataType.UNDEFINED.value
```
justinchuby, line 203: _enums.DataType.UNDEFINED can be used as an int so `.value` may not be necessary

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+        attr = node.attributes[name]
+        if isinstance(attr, ir.AttrInt64):
+            return attr.value
+        return None
+    return default
+
+
+# TODO(rama): The following should not be necessary. Generic incremental shape-inference
+# should handle this. This essentially implements type/shape-inference for Cast op.
+@register("Cast")
```
justinchuby, line 217: Note to self: I will resume from here

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+        for old_value, new_value in zip(old_values, new_values):
+            # Propagate relevant info from old value to new value
+            # TODO(Rama): Perhaps we should merge old and new types. As of now, new
+            # values don't have type information. Note that this could be a problem
+            # for semantics-altering rewrite-rules: we should allow users to override
+            # this for such rules.
+            new_value.type = old_value.type
+            new_value.shape = old_value.shape
+            new_value.const_value = old_value.const_value
+            new_value.name = old_value.name
```
justinchuby, line 639: We can create a copy() method for nodes

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+    return None
+
+
+class ConstantFolder:
+    opset_imports: dict[str, int]
+
+    def __init__(
+        self,
+        external_data_folder: str,
+        do_shape_inference: bool,
```
justinchuby, line 464:
```suggestion
        shape_inference: bool,
```

nit: simplicity

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+            return op.Identity(result)
+    return None
+
+
+class ConstantFolder:
+    opset_imports: dict[str, int]
+
+    def __init__(
+        self,
+        external_data_folder: str,
```
justinchuby, line 463:
```suggestion
        external_data_folder: Literal["some_options", "some_other_options"],
```

justinchuby, line 463: I would support os.PathLike as well (in general for paths)

justinchuby: folder -> dir ?

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+registry: PartialEvaluatorRegistry = PartialEvaluatorRegistry()
+
+register = registry.register
+
+
+def _get_sym_value(val: ir.Value | None) -> ir.Value | None:
+    if val is None:
+        return None
+    if hasattr(val, "symbolic_value"):
+        return val.symbolic_value
```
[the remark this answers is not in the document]
gramalingam: The last time we discussed this, you were ok with extending IR value to support symbolic_value, or so I thought. How does updating the IR class documentation/typing annotation to add symbolic_value compare with using meta?

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+    input = _get_input(node, index)
+    if input is not None and input.type is not None:
+        return input.type.dtype.value
+    return _enums.DataType.UNDEFINED.value
+
+
+def _get_int_attribute(node: ir.Node, name: str, default: int | None = None) -> int | None:
+    if name in node.attributes:
+        attr = node.attributes[name]
+        if isinstance(attr, ir.AttrInt64):
```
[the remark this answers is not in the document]
gramalingam: Hmmm ... so there is no guarantee that the specialized derived classes of ir.Attr will be created? That seems potentially confusing.

onnxscript/optimizer/constant_folding_ir.py
```diff
@@ -0,0 +1,718 @@
...
+
+
+def _get_output(node: ir.Node, index: int) -> ir.Value | None:
+    if index < len(node.outputs):
+        return node.outputs[index]
+    return None
+
+
+def _update_type(value: ir.Value, type: ir.TypeProtocol) -> None:
+    # TODO: merge types
```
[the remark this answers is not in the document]
gramalingam: Ideally, we want to combine the pre-existing type/shape information with newly inferred type/shape information to retain the most information (may be the exporter has more more precise shape info or may be the inference produces more precise info) ... this requires adding check for consistency between the two, raising errors, etc. also.

onnxscript/optimizer/constant_folding_ir.py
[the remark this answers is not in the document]
justinchuby: > potentially confusing

that's true. Do you have suggestions? We can turn the specialized classes into functions that will produce Attr objects, for example.

onnxscript/optimizer/constant_folding_ir.py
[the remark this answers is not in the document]
justinchuby: So the latest idea was: https://github.com/microsoft/onnxscript/issues/1438#issuecomment-2094560289 . 

Wondering if you have thoughts?

onnxscript/optimizer/constant_folding_ir.py
gramalingam: `shape_inference` by itself sounds ambiguous to me ... it is unclear what the option means. Maybe it is just me

onnxscript/optimizer/constant_folding_ir.py
gramalingam: yes, at least, we need some common utility we can call here.

onnxscript/optimizer/constant_folding_ir.py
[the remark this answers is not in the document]
gramalingam: Making it a transient data-structure, used only by this analysis, makes sense. Will do that.

onnxscript/optimizer/constant_folding_ir.py
gramalingam: So, this seems unused currently! Previously, it was used when converting an external tensor into a numpy array (using ONNX's numpy_helper, which only supports "str" for the external-dir incidentally). Now, this uses IR to convert to numpy array. What happens in the IR when external-tensors are converted to numpy array? Does the IR remember the external-data-folder? Or, do we need pass it as a parameter?

justinchuby: It is currently incomplete, we need to update the deserialize function: https://github.com/microsoft/onnxscript/issues/1701#issuecomment-2189866219

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
...
+
+registry: PartialEvaluatorRegistry = PartialEvaluatorRegistry()
+
+register = registry.register
+
+
+def _get_numpy_value(val: ir.Value) -> np.ndarray | None:
+    const_value = val.const_value
+    if hasattr(const_value, "numpy"):
+        return const_value.numpy()
```
justinchuby, line 170: `val.const_value` should always be an `ir.TensorProto` type or None so it should always have `numpy()` when it is not None. Currently the type is not enforced for runtime efficiency, but we can add more type checking under the `DEBUG` mode

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
...
+    return None
+
+
+def _get_bool_value(val: ir.Value | None) -> bool | None:
+    if val is None:
+        return None
+    val = _get_numpy_value(val)
+    if val is None:
+        return None
+    if isinstance(val, bool):
```
justinchuby, line 180: Curious when this is needed? Do we need to update the typing annotation or tighten the invariance?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
...
+        return node.outputs[index]
+    return None
+
+
+def _update_type(value: ir.Value, type: ir.TypeProtocol) -> None:
+    # TODO: merge types
+    value.type = type
+
+
+def _get_input_element_type(node: ir.Node, index: int) -> int:
```
justinchuby, line 206:
```suggestion
def _get_input_element_type(node: ir.Node, index: int) -> ir.DataType:
```

seems appropriate based on usage

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
...
+    def __init__(self):
+        self.op_evaluators: dict[tuple[str, str], list[PartialEvaluator]] = {}
+
+    def lookup_evaluators(self, domain: str, opname: str, version: int):
+        evaluator_list = self.op_evaluators.get((domain, opname), [])
+        return [
+            evaluator.function for evaluator in evaluator_list if evaluator.valid_for(version)
+        ]
+
+    def register(self, opname: str, domain: str = "", version=None):
```
justinchuby, line 140:
```suggestion
    def register(self, opname: str, domain: str = "", version=None) -> Callable[[T], T]:
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
...
+
+
+class OptimizerState:
+    def __init__(self):
+        self._sym_value_map: dict[ir.Value, Any] = {}
+
+    def get_sym_value(self, value: ir.Value | None) -> Any:
+        if value is None:
+            return None
+        return self._sym_value_map.get(value, None)
```
justinchuby, line 88:
```suggestion
        return self._sym_value_map.get(value)
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -0,0 +1,712 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# -------------------------------------------------------------------------
```
justinchuby, line 4:
```suggestion
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```

justinchuby (approved):
lgtm w/ comments. In particular, I think `const_value` should always be `ir.TensorProtocol` and we can rely on this invariance.

onnxscript/optimizer/_constant_folding.py
gramalingam: done

onnxscript/optimizer/_constant_folding.py
gramalingam: skipping this for now. (will require dropping the `.value` in the implementation in two places as well.)

onnxscript/optimizer/constant_folding_test.py
```diff
@@ -3,12 +3,29 @@
...
+    def _fold(self, model: onnx.ModelProto, onnx_shape_inference=False):
+        if self.using_ir:
+            ir_model = serde.deserialize_model(model)
+            _constant_folding.fold_constants(
+                ir_model, onnx_shape_inference=onnx_shape_inference
+            )
+            optimizer.remove_unused_nodes(ir_model)
+            return serde.serialize_model(ir_model)
+        else:
+            constant_folding.fold_constants(model, onnx_shape_inference=onnx_shape_inference)
```
titaiwangms, line 25: Will we replace constant_folding with _constant_folding eventually? If so, _constant_folding should not import stuffs from constant_folding?

gramalingam: yes ... they will migrate into this file eventually

PR #1731, opened 2024-07-16
Add test case for unsupported input-variable reuse in multi-output pattern matcher

gramalingam:
The multi-output pattern does not support a useful case, where the pattern's input variables are reused elsewhere.

This PR adds a test-case illustrating the scenario. I was hoping that some changes in the in-progress PR #1636 by Xavier might fix this. But I don't think it does. Adding the test-case for now. Need to figure out how to support this.

### DIFF
--- a/onnxscript/rewriter/generic_pattern_test.py
+++ b/onnxscript/rewriter/generic_pattern_test.py
@@ -282,7 +282,7 @@ class GenericPatternTest(unittest.TestCase):
         self.assertEqual(graph.node[0].op_type, "SinCos")
 
     @unittest.skip("Input variable reuse not supported yet")
-    def test_shared_root_value_extra_use_test(self):
+    def test_shared_root_value_extra_use(self):
         def match_pattern(op, x):
             t1 = op.Sin(x)
             t2 = op.Cos(x)
--- a/onnxscript/rewriter/generic_pattern_test.py
+++ b/onnxscript/rewriter/generic_pattern_test.py
@@ -281,6 +281,7 @@ class GenericPatternTest(unittest.TestCase):
         self.assertEqual(len(graph.node), 2)
         self.assertEqual(graph.node[0].op_type, "SinCos")
 
+    @unittest.skip("Input variable reuse not supported yet")
     def test_shared_root_value_extra_use_test(self):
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -281,6 +281,42 @@ def apply_pattern(op, x, **_):
         self.assertEqual(len(graph.node), 2)
         self.assertEqual(graph.node[0].op_type, "SinCos")
 
+    @unittest.skip("Input variable reuse not supported yet")
+    def test_shared_root_value_extra_use_test(self):
```
justinchuby, line 285:
```suggestion
    def test_shared_root_value_extra_use(self):
```

justinchuby (approved):

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -281,6 +281,42 @@ def apply_pattern(op, x, **_):
         self.assertEqual(len(graph.node), 2)
         self.assertEqual(graph.node[0].op_type, "SinCos")
 
+    @unittest.skip("Input variable reuse not supported yet")
```
justinchuby, line 284: We can xfail this so we know when it is fixed

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -281,6 +281,42 @@ def apply_pattern(op, x, **_):
...
+        """
+        )
+        onnx.checker.check_model(model_proto)
+        model = onnx.shape_inference.infer_shapes(model_proto)
+        ir_model = ir.serde.deserialize_model(model)
+        rule.apply_to_model(ir_model)
+        rewritten_model = ir.serde.serialize_model(ir_model)
+        graph = rewritten_model.graph
+        self.assertEqual(len(graph.node), 3)
+        self.assertEqual(graph.node[0].op_type, "SinCos")
```
justinchuby, line 318:
```suggestion
        rule.apply_to_model(ir_model)
        graph = ir_model.graph
        self.assertEqual(len(graph), 3)
        self.assertEqual(graph.node[0].op_type, "SinCos")
```

We can test the ir model directly

PR #1733, opened 2024-07-17
[torchlib] Add missing operators (set 2)

shubhambhokare1:
- [x] aten.special.expm1
- [x] aten.sort

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7710,7 +7710,7 @@ def aten_sort(
         self = op.Unsqueeze(self, op.Constant(value_ints=[0]))
     shape = op.Shape(self)
     dim_size = op.Gather(shape, dim, axis=0)
-    dim_size = op.Reshape(op.Cast(dim_size, to=INT64.dtype), op.Constant(value_ints=[1]))
+    dim_size = op.Reshape(dim_size, op.Constant(value_ints=[1]))
     values, indices = op.TopK(self, dim_size, axis=dim, largest=descending, sorted=True)
     if self_is_scalar:
         values = op.Squeeze(values, op.Constant(value_ints=[0]))
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7707,7 +7707,7 @@ def aten_sort(
 
     self_is_scalar = IsScalar(self)
     if self_is_scalar:
-        return op.Identity(self), op.Squeeze(op.Constant(value_ints=[0]))
+        return op.Identity(self), op.Constant(value_int=0)
     shape = op.Shape(self)
     dim_size = op.Gather(shape, dim, axis=0)
     dim_size = op.Reshape(dim_size, op.Constant(value_ints=[1]))
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7699,7 +7699,7 @@ def aten_softmax_no_dtype(self: TFloatOrBFloat16, dim: int) -> TFloatOrBFloat16:
     return result
 
 
-@torch_op("aten::sort", traceable=True)
+@torch_op("aten::sort", trace_only=True)
 def aten_sort(
     self: TReal, dim: int = -1, descending: bool = False, stable: bool = False
 ) -> tuple[TReal, INT64]:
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7699,12 +7699,23 @@ def aten_softmax_no_dtype(self: TFloatOrBFloat16, dim: int) -> TFloatOrBFloat16:
...
+) -> tuple[TReal, INT64]:
+    """sort(Tensor self, int dim=-1, bool descending=False, bool stable=False) -> (Tensor values, Tensor indices)"""
 
-    raise NotImplementedError()
+    self_is_scalar = IsScalar(self)
+    if self_is_scalar:
+        self = op.Unsqueeze(self, op.Constant(value_ints=[0]))
+    shape = op.Shape(self)
+    dim_size = op.Gather(shape, dim, axis=0)
+    dim_size = op.Reshape(op.Cast(dim_size, to=INT64.dtype), op.Constant(value_ints=[1]))
```
justinchuby, line 7713:
```suggestion
    dim_size = op.Reshape(dim_size, op.Constant(value_ints=[1]))
```

op.Shape should always be int64

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7699,12 +7699,23 @@ def aten_softmax_no_dtype(self: TFloatOrBFloat16, dim: int) -> TFloatOrBFloat16:
...
 def aten_sort(
-    self: TensorType, dim: int = -1, descending: bool = False
-) -> tuple[TensorType, TensorType]:
-    """sort(Tensor self, int dim=-1, bool descending=False) -> (Tensor values, Tensor indices)"""
+    self: TReal, dim: int = -1, descending: bool = False, stable: bool = False
+) -> tuple[TReal, INT64]:
+    """sort(Tensor self, int dim=-1, bool descending=False, bool stable=False) -> (Tensor values, Tensor indices)"""
 
-    raise NotImplementedError()
+    self_is_scalar = IsScalar(self)
```
justinchuby, line 7708: if self is scalar, is it possible to just return self? e.g. op.Identity(self) ?

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
shubhambhokare1: Had tried that initially, but looks like we cannot return from inside a control-flow block

gramalingam: Should it be made traceonly? The only control-flow logic is rank-dependent, and we are anyway assuming that is known as export-time. Then, you can do what Justin suggests.

justinchuby: Alternatively, you can assign the output of identity to result, then return result outside of the control flow. But trace only makes sense

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7699,12 +7699,20 @@ def aten_softmax_no_dtype(self: TFloatOrBFloat16, dim: int) -> TFloatOrBFloat16:
...
+    if self_is_scalar:
+        return op.Identity(self), op.Squeeze(op.Constant(value_ints=[0]))
```
justinchuby, line 7710:
```suggestion
        return op.Identity(self), op.Constant(value_int=0)
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7699,12 +7699,20 @@ def aten_softmax_no_dtype(self: TFloatOrBFloat16, dim: int) -> TFloatOrBFloat16:
...
+) -> tuple[TReal, INT64]:
+    """sort(Tensor self, int dim=-1, bool descending=False, bool stable=False) -> (Tensor values, Tensor indices)"""
 
-    raise NotImplementedError()
+    self_is_scalar = IsScalar(self)
+    if self_is_scalar:
+        return op.Identity(self), op.Constant(value_int=0)
+    shape = op.Shape(self)
+    dim_size = op.Gather(shape, dim, axis=0)
+    dim_size = op.Reshape(dim_size, op.Constant(value_ints=[1]))
```
justinchuby, line 7713: Use a slice to replace the two ops?

shubhambhokare1: Oh, think I missed this comment before merge. Making a note of this as a #TODO for the PR covering the next set of operators.

PR #1734, opened 2024-07-17
Extend basic matcher to handle multiple-output-nodes

gramalingam:
This PR extends the basic matcher to handle multiple output nodes. This provides an alternative to the generic-matcher algorithm, which is incomplete and fails in some circumstances. This can also be useful in debugging match-failures (when it is unclear if the failure is valid or due to limitations of the matching algorithm). The drawback is that this algorithm can, in some cases, be expensive, especially when the number of output-nodes is large and the graph size is large.  (So far, however, we haven't encountered patterns with more than 2 output-nodes.)

### DIFF
--- a/onnxscript/rewriter/generic_pattern.py
+++ b/onnxscript/rewriter/generic_pattern.py
@@ -298,15 +298,7 @@ class GenericPatternMatcher(orp.PatternMatcher):
             return self.none(starting_node, inspect.currentframe().f_lineno)
 
         for graph_input, pattern_input in zip(graph_node.inputs, pattern_node.inputs):
-            # Intermediate values in the pattern must have the same number of uses
-            # in the graph for a valid match. By design, this esnures patterns where
-            # intermediate values have extra uses in the graph do NOT match.
-            # However, pattern-input-values may have extra uses. This is because
-            # pattern-inputs will not be removed by pattern-replacement, but
-            # intermediate values will be removed.
-            if pattern_input.producer() is not None and len(list(graph_input.uses())) != len(
-                list(pattern_input.uses())
-            ):
+            if len(graph_input.uses()) != len(pattern_input.uses()):
                 self._hint(
                     "BACKWARD: one input is used outside the pattern",
                     "-- pattern",
--- a/onnxscript/rewriter/generic_pattern.py
+++ b/onnxscript/rewriter/generic_pattern.py
@@ -298,7 +298,15 @@ class GenericPatternMatcher(orp.PatternMatcher):
             return self.none(starting_node, inspect.currentframe().f_lineno)
 
         for graph_input, pattern_input in zip(graph_node.inputs, pattern_node.inputs):
-            if len(list(graph_input.uses())) != len(list(pattern_input.uses())):
+            # Intermediate values in the pattern must have the same number of uses
+            # in the graph for a valid match. By design, this esnures patterns where
+            # intermediate values have extra uses in the graph do NOT match.
+            # However, pattern-input-values may have extra uses. This is because
+            # pattern-inputs will not be removed by pattern-replacement, but
+            # intermediate values will be removed.
+            if pattern_input.producer() is not None and len(list(graph_input.uses())) != len(
+                list(pattern_input.uses())
+            ):
                 self._hint(
                     "BACKWARD: one input is used outside the pattern",
                     "-- pattern",
@@ -423,12 +431,12 @@ class GenericPatternMatcher(orp.PatternMatcher):
             return match_count
         if len(free) < len(pattern_node_users_not_matched):
             # Not enough successors to match the remaining patterns.
-            return self.none(node, inspect.currentframe().f_lineno)
+            return self.none(starting_node, inspect.currentframe().f_lineno)
         if len(pattern_node_users_not_matched) == len(free) == 1:
             # Only one option again.
             graph_node = free[0]
             if pattern_node_users_not_matched[0].op_identifier() != graph_node.op_identifier():
-                return self.none(node, inspect.currentframe().f_lineno)
+                return self.none(starting_node, inspect.currentframe().f_lineno)
 
             key = pattern_node_users_not_matched[0]
             if self.verbose >= 10:
@@ -461,11 +469,11 @@ class GenericPatternMatcher(orp.PatternMatcher):
                 "-- model-matched",
                 pattern_node_users_matched,
             )
-            return self.none(node, inspect.currentframe().f_lineno)
+            return self.none(starting_node, inspect.currentframe().f_lineno)
         for k, v in ec.items():
             if gc[k] < v:
                 # Not enough types to match.
-                return self.none(node, inspect.currentframe().f_lineno)
+                return self.none(starting_node, inspect.currentframe().f_lineno)
 
         # At this stage, we know matching the types is possible.
         # We first mark whatever is possible.
--- a/onnxscript/rewriter/generic_pattern_test.py
+++ b/onnxscript/rewriter/generic_pattern_test.py
@@ -285,8 +285,10 @@ class GenericPatternTest(unittest.TestCase):
         self.assertEqual(len(graph.node), 2)
         self.assertEqual(graph.node[0].op_type, "SinCos")
 
-    @unittest.skip("Input variable reuse not supported yet")
     def test_shared_root_value_extra_use(self):
+        if self.matcher_algo is generic_pattern.GenericPatternMatcher:
+            raise unittest.SkipTest("GenericPatternMatcher does not support extra uses yet.")
+
         def match_pattern(op, x):
             t1 = op.Sin(x)
             t2 = op.Cos(x)
@@ -298,7 +300,7 @@ class GenericPatternTest(unittest.TestCase):
         rule = pattern.RewriteRule(
             match_pattern,
             apply_pattern,
-            matcher=generic_pattern.GenericPatternMatcher,
+            matcher=self.matcher_algo,
         )
         model_proto = onnx.parser.parse_model(
             """
@@ -318,7 +320,7 @@ class GenericPatternTest(unittest.TestCase):
         rule.apply_to_model(ir_model)
         graph = ir_model.graph
         self.assertEqual(len(graph), 3)
-        self.assertEqual(graph.node[0].op_type, "SinCos")
+        self.assertEqual(graph.node(0).op_type, "SinCos")
 
     def test_rotary_embedding(self):
         # The test work on a model if it has the expected name.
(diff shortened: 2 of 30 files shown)


onnxscript/rewriter/generic_pattern.py
```diff
@@ -298,7 +298,15 @@ def _match_backward(
...
-            if len(list(graph_input.uses())) != len(list(pattern_input.uses())):
+            # Intermediate values in the pattern must have the same number of uses
+            # in the graph for a valid match. By design, this esnures patterns where
+            # intermediate values have extra uses in the graph do NOT match.
+            # However, pattern-input-values may have extra uses. This is because
+            # pattern-inputs will not be removed by pattern-replacement, but
+            # intermediate values will be removed.
+            if pattern_input.producer() is not None and len(list(graph_input.uses())) != len(
+                list(pattern_input.uses())
+            ):
```
justinchuby, line 309:
```suggestion
            if pattern_input.producer() is not None and len(graph_input.uses()) != len(
               (pattern_input.uses()
            ):
```

gramalingam: Oh, this was an unintentional change. I was experimenting if the generic-matcher could be fixed to handle the example, but this change is not sufficient. I will remove these lines as they are irrelevant to this PR.

justinchuby:
> when the number of output-nodes is large

Clarifying to understand better: is it referring to that of the non-match subgraphs or the target pattern? For example, if we have a simple pattern with small outputs and a graph with many outputs, will it still be costly?

gramalingam:
> > when the number of output-nodes is large
> 
> Clarifying to understand better: is it referring to that of the non-match subgraphs or the target pattern? For example, if we have a simple pattern with small outputs and a graph with many outputs, will it still be costly?

The complexity is sort of `(Number of graph nodes) ^ (Number of pattern-output-nodes)` ... here a pattern-output-node is a pattern-node that produces an output-value of the pattern (again, loosely speaking). In reality, `Number of graph nodes` is a worst-case estimate, it really is number of nodes with same optype as corresponding pattern-output-node.

So, if we are looking for a pattern `Z1 = Sin(X), Z2 = Cos(X), return Z1, Z2` both statements are pattern-output-nodes. If NS is the number of Sin nodes in graph, and NC is the number of Cos nodes in graph, the complexity is NSxNC.

onnxscript/rewriter/generic_pattern.py
gramalingam: Incorporated the suggested change, undid the unintentional change.

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -389,7 +395,8 @@ def apply_pattern(op, x, pos_ids, axis, **_):
         self.assertEqual(expected, [n.op_type for n in rewriten_model.graph.node])
         out = buffer.getvalue()
         # TODO(Rama): What is this assertion testing? Is it to check that `verbose` is working?
-        self.assertIn("[GenericPatternMatcher.match", out)
+        if self.matcher_algo == generic_pattern.GenericPatternMatcher:
```
justinchuby, line 398:
```suggestion
        if self.matcher_algo is generic_pattern.GenericPatternMatcher:
```

onnxscript/rewriter/pattern.py
```diff
@@ -665,21 +665,25 @@ def __init__(
...
+            if isinstance(value_pattern, Constant):
+                raise NotImplementedError(
+                    "Constant values are not allowed as graph pattern outputs."
+                )
+            if isinstance(value_pattern, NodeOutputPattern):
+                output_nodes.add(value_pattern.producer())
+        self.output_nodes: list[NodePattern] = list(output_nodes)
+
+    @property
+    def output_node(self) -> NodePattern:
```
justinchuby, line 683: Should this be a method instead? Properties can usually be assigned

```suggestion
    def output_node(self) -> NodePattern:
```

onnxscript/rewriter/pattern.py
```diff
@@ -706,18 +710,18 @@ def __reversed__(self) -> Iterator[NodePattern]:
 
     @property
```
justinchuby, line 711:
```suggestion
```

nit

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
             )
         return self._match_node(pattern_value.producer(), node)
 
-    def match(
+    def _init_match(self, verbose: int) -> None:
```
justinchuby, line 1036: Comment on when this method should be called?

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+        self._verbose = verbose
+        self._matched: dict[NodePattern, ir.Node] = {}
+        self._match: MatchResult = MatchResult()
+
+    def _get_output_values(self) -> list[ir.Value] | None:
```
justinchuby, line 1041: Add a brief docstring?

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
-        if pattern._output_node is None:
+
+        if not self._match_node(pattern.output_node, node):
+            return match
+
+        output_values = self._get_output_values()
+        if output_values is None:
+            return match
+        if not _valid_to_replace(match.nodes, output_values):
+            return match.fail("Matched nodes have other uses preventing replacement.")
```
justinchuby, line 1088: Print all uses for info?

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+                f"Number of node outputs mismatch: expected {pattern.num_outputs}, got {len(node.outputs)}."
             )
 
-        if self._match_node(pattern._output_node, node):
-            if not _valid_to_replace(match.nodes):
-                return match.fail("Matched nodes have other uses preventing replacement.")
+        match.outputs.extend(output_values)
+        return match
+
+    def _multi_match(self, candidate: dict[NodePattern, ir.Node]) -> MatchResult:
```
justinchuby, line 1098: I would add a docstring

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+        match.outputs.extend(output_values)
         return match
 
+    def match(
+        self,
+        model: ir.Model,
+        graph_or_function: ir.Graph | ir.Function,
+        node: ir.Node,
+        verbose: int = 0,
+    ) -> MatchResult:
```
justinchuby, line 1119: I would add a docstring

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+        if not _valid_to_replace(match.nodes, output_values):
+            return match.fail("Matched nodes have other uses preventing replacement.")
+
+        match.outputs.extend(output_values)
         return match
 
+    def match(
+        self,
+        model: ir.Model,
+        graph_or_function: ir.Graph | ir.Function,
```
justinchuby, line 1116: Just curious: why is both model and graph required as args? Do we need to take info from the model?

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+                op_to_nodes.setdefault(n.op_identifier(), []).append(n)
+            all_nodes = iter(graph_or_function)
+
+            def get_nodes(pattern_node):
+                id = pattern_node.op_identifier()
+                if id is None:
+                    return all_nodes
+                return op_to_nodes.get(id, [])
+
+            candidates = [iter([node])] + [get_nodes(pn) for pn in pattern_output_nodes[1:]]
```
justinchuby, line 1136: pn -> more descriptive name

onnxscript/rewriter/pattern.py
```diff
@@ -1029,37 +1033,118 @@ def _match_node_output(self, pattern_value: NodeOutputPattern, value: ir.Value)
...
+            match = None
+            for combination in itertools.product(*candidates):
+                candidate = dict(zip(pattern_output_nodes, combination))
```
justinchuby, line 1139: I would add a comment to describe the algo/logic of this loop (what `product` does etc.)

onnxscript/rewriter/pattern.py
justinchuby, line 1: I would add "the drawback ..." from the PR description to somewhere in this file

onnxscript/rewriter/pattern.py
gramalingam: Not sure whether we need both now. This may be a remnant leftover from earlier code due to (i) Reading and/or updating opset imports, when using Proto-based IR, and (b) Refactored methods coming out of the rewriter into a shared interface used by both pattern-matching implementations.

onnxscript/rewriter/pattern.py
gramalingam: IMHO it is acceptable when the scope of the identifier is this single line. The `pattern_output_nodes` clarifies what a `pn` is. A longer name has a good chance of breaking the single line into multiple, which would be worse IMO.

onnxscript/rewriter/pattern.py
gramalingam: Looking over the multiple implementations of this abstract method, I think we can eliminate both model and graph. While we use graph, I think we can get it from the node (in the new IR, unlike in proto). However, I think this does require some careful thinking, especially in dealing with control-flow sub-graphs. Maybe a candidate for a separate PR.

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -294,7 +300,7 @@ def apply_pattern(op, x, **_):
         rule = pattern.RewriteRule(
             match_pattern,
             apply_pattern,
-            matcher=generic_pattern.GenericPatternMatcher,
+            matcher=self.matcher_algo,
```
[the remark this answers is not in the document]
gramalingam: The matcher is optional, the constructor will automatically choose a matcher (based on number of output-nodes). But, here, for testing purpose, we want to specify which one to use. In general, users have to specify to choose a non-default one.

justinchuby (approved):

PR #1739, opened 2024-07-19
Cleanup new-IR based constant propagation

gramalingam:
Factor out some common logic between rewriter and constant-propagation into a utility function, and other minor cleanup.

### DIFF
--- a/onnxscript/ir/_convenience.py
+++ b/onnxscript/ir/_convenience.py
@@ -399,12 +399,13 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
 
 def replace_nodes_and_values(
     graph_or_function: _core.Graph | _core.Function,
+    /,
     insertion_point: _core.Node,
-    old_nodes: list[_core.Node],
-    new_nodes: list[_core.Node],
-    old_values: list[_core.Value],
-    new_values: list[_core.Value],
-):
+    old_nodes: Sequence[_core.Node],
+    new_nodes: Sequence[_core.Node],
+    old_values: Sequence[_core.Value],
+    new_values: Sequence[_core.Value],
+) -> None:
     """Replaces nodes and values in the graph or function.
 
     Args:
--- a/onnxscript/ir/_convenience.py
+++ b/onnxscript/ir/_convenience.py
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
                 continue
             values[value.name] = value
     return values
+
+
+def replace_nodes_and_values(
+    graph_or_function: _core.Graph | _core.Function,
+    insertion_point: _core.Node,
+    old_nodes: list[_core.Node],
+    new_nodes: list[_core.Node],
+    old_values: list[_core.Value],
+    new_values: list[_core.Value],
+):
+    """Replaces nodes and values in the graph or function.
+
+    Args:
+        graph_or_function: The graph or function to replace nodes and values in.
+        insertion_point: The node to insert the new nodes after.
+        old_nodes: The nodes to replace.
+        new_nodes: The nodes to replace with.
+        old_values: The values to replace.
+        new_values: The values to replace with.
+    """
+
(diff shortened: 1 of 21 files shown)


onnxscript/ir/_convenience.py
```diff
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
...
+
+
+def replace_nodes_and_values(
+    graph_or_function: _core.Graph | _core.Function,
+    insertion_point: _core.Node,
+    old_nodes: list[_core.Node],
+    new_nodes: list[_core.Node],
+    old_values: list[_core.Value],
+    new_values: list[_core.Value],
+):
```
justinchuby, line 407:
```suggestion
) -> None:
```

onnxscript/ir/_convenience.py
```diff
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
                 continue
             values[value.name] = value
     return values
+
+
+def replace_nodes_and_values(
+    graph_or_function: _core.Graph | _core.Function,
```
justinchuby, line 401:
```suggestion
    graph_or_function: _core.Graph | _core.Function,
    /,
```

onnxscript/ir/_convenience.py
```diff
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
                 continue
             values[value.name] = value
     return values
+
+
+def replace_nodes_and_values(
```
justinchuby, line 400: Helpful to have a test?

justinchuby (approved):

onnxscript/ir/_convenience.py
```diff
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
...
+    graph_or_function: _core.Graph | _core.Function,
+    insertion_point: _core.Node,
+    old_nodes: list[_core.Node],
+    new_nodes: list[_core.Node],
+    old_values: list[_core.Value],
+    new_values: list[_core.Value],
+):
+    """Replaces nodes and values in the graph or function.
```
justinchuby, line 408: Helpful to describe the invariance this function maintains?

onnxscript/ir/_convenience.py
```diff
@@ -395,3 +395,44 @@ def create_value_mapping(graph: _core.Graph) -> dict[str, _core.Value]:
...
     return values
+
+
+def replace_nodes_and_values(
+    graph_or_function: _core.Graph | _core.Function,
+    insertion_point: _core.Node,
+    old_nodes: list[_core.Node],
+    new_nodes: list[_core.Node],
+    old_values: list[_core.Value],
+    new_values: list[_core.Value],
```
justinchuby, line 406:
```suggestion
    old_nodes: Sequence[_core.Node],
    new_nodes: Sequence[_core.Node],
    old_values: Sequence[_core.Value],
    new_values: Sequence[_core.Value],
```

onnxscript/ir/_convenience.py
gramalingam: Done (but not pushed yet). But is there a specific principle being used here, to distinguish between the two classes of parameters? Except for the name, which is admittedly not great (we could do with a better alternative to graph_or_function ... how does `container` sound? or `graph_like` ?)

justinchuby: I like `graph_like`!

> is there a specific principle being used here, to distinguish between the two classes of parameters

not sure what you mean? could you say more?

justinchuby:
Is there a way to copy the metadata_props from old nodes as well?

onnxscript/ir/_convenience.py
gramalingam: I mean, why do you prefer that graph_or_function be positional, while the others are allowed to be positional or by keyword?

PR #1741, opened 2024-07-22
Fix weekly CI pipeline errors

justinchuby:
1. Fix type annotation for _add_attribute_to_torchscript_node and removed runtime type checking because int like inputs are not correctly recognized by beartype
2. Fix `test_tensor_proto_tensor_bfloat16` with latest onnx-weekly
3. pytorch-nightly errors will be fixed separately


shubhambhokare1 (approved):
Do we plan on fixing the failing dort ci's in a future PR?

justinchuby:
> Do we plan on fixing the failing dort ci's in a future PR?

I don't have plans for it yet. The dort ci failures should be the same as the torch-nightly failures, which are mostly tolerance issues.

PR #1742, opened 2024-07-22
[torchlib] Fix linspace and full

justinchuby:
- linspace: Add additional parameters
- full: Handle when size is `[]`

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3612,7 +3612,7 @@ def aten_from_file(
 @torch_op("aten::full", trace_only=True)
 def aten_full(
     size: Union[INT64, INT32],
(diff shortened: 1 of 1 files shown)


titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3606,30 +3611,34 @@ def aten_from_file(
 
 @torch_op("aten::full", trace_only=True)
 def aten_full(
-    size: INT64,
+    size: Union[INT64, INT32],
     fill_value: FLOAT,
```
shubhambhokare1, line 3615: Should fill_value be a TensorType instead? Given that fill_value can be an int or a bool as well?

shubhambhokare1 (approved):

PR #1750, opened 2024-07-23
Add Op (group_norm) | feat(torchlib)

titaiwangms:
Add group normalization with instance normalization simulation (reference on native group norm).

NOTE: This approach does not support the input shape that can't reshape to (0, num_group, -1)

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -593,6 +593,56 @@ def aten_glu_backward_jvp(
     raise NotImplementedError()
 
 
+@torch_op("aten::group_norm", trace_only=True)
+def aten_group_norm(
+    input: TFloat,
+    num_groups: int,
+    weight: Optional[TFloat] = None,
+    bias: Optional[TFloat] = None,
+    eps: float = 1e-05,
+    cudnn_enabled: bool = True,
+) -> TensorType:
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -593,6 +593,56 @@ def aten_glu_backward_jvp(
...
+    # Using the input weight and bias to do affine
+    # But need to unsqueeze to the target shape for broading cast easy
+    input_rank = Rank(input)
+    one = op.Constant(value_int=1)
+    axes_unsqueeze = op.Range(one, op.Sub(input_rank, one), one)
+    weight_full_shape = op.Unsqueeze(weight, axes_unsqueeze)
+    bias_full_shape = op.Unsqueeze(bias, axes_unsqueeze)
+    weight_full_shape = op.CastLike(weight_full_shape, norm)
+    norm_mul_weight = op.Mul(norm, weight_full_shape)
+    bias_full_shape = op.CastLike(bias_full_shape, norm_mul_weight)
```
titaiwangms, line 641: These CastLikes are not caught by `native_group_norm`, so they are not in `native_group_norm` implementation. Should we add them? I think group norm is optimized away when rewriter is applied.

justinchuby (approved):
Thank you!

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: Sure, thanks!

PR #1751, opened 2024-07-23
[torchlib] Remove adaptive_avg_pool implementation

justinchuby:
Remove adaptive_avg_pool implementation because our implementation using GlobalAveragePool is incorrect. We can rely on torch decomp instead.


titaiwangms (approved):
What ops do we get after decomposition? I recall 1D is correct, but 2d and 3d is not passing test_fx_op_consistency.py

justinchuby:
> What ops do we get after decomposition? I recall 1D is correct, but 2d and 2d is not passing test_fx_op_consistency.py

Good point. I can test that. In any case globalavgpool is incorrect so we need to either improve the implementation or fix decomp

PR #1754, opened 2024-07-25
[torchlib] Improve aten::fill

justinchuby:
I updated torch-onnx to handle empty `[]` inputs, so the isinstance check is not needed.


titaiwangms (approved):

titaiwangms:
Sounds like this might break torch.onnx.dynamo_export though?

justinchuby:
> Sounds like this might break torch.onnx.dynamo_export though?

Probably not? Because it used to be like this

titaiwangms:
> > Sounds like this might break torch.onnx.dynamo_export though?
> 
> Probably not? Because it used to be like this

Oh you added it for torch-onnx.

PR #1755, opened 2024-07-25
[torchlib] Fix aten::diagonal

justinchuby:
Turn aten::diagonal as trace only and fix its logic by explicitly converting python constants to onnx constants. This was needed because the exporter logic was not handling the type conversion correctly (yet)

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -2550,8 +2550,9 @@ def aten_diagonal(self: TReal, offset: int = 0, dim1: int = 0, dim2: int = 1) ->
     # This is because computing diagonal sum is on dim2 after transpose by perm
     axes = [self_rank - 2]
 
-    dim1_size = op.Shape(self, end=dim1, start=dim1 + 1)  # row
-    dim2_size = op.Shape(self, end=dim2, start=dim2 + 1)  # col
+    neg_1 = op.Constant(value_ints=[-1])
+    dim1_size = op.Reshape(op.Gather(op.Shape(self), dim1), neg_1)  # row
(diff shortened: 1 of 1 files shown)


titaiwangms (approved):

fatcat-z:
If possible, it'd better paste the error we saw into description so we know what kind of logic was fixed.

justinchuby:
> If possible, it'd better paste the error we saw into description so we know what kind of logic was fixed.

Done, thanks.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2550,19 +2550,10 @@ def aten_diagonal(self: TReal, offset: int = 0, dim1: int = 0, dim2: int = 1) ->
...
-
-
-@torch_op("aten::diagonal", private=True, traceable=True)
-def _aten_diagonal_onnx(
-    self: TTensor, offset: int, dim1: int, dim2: int, perm: Sequence[int], axes: Sequence[int]
-) -> TTensor:
-    neg_1 = op.Constant(value_ints=[-1])
-    dim1_size = op.Reshape(op.Gather(op.Shape(self), dim1), neg_1)  # row
-    dim2_size = op.Reshape(op.Gather(op.Shape(self), dim2), neg_1)  # col
+    dim1_size = op.Shape(self, end=dim1, start=dim1 + 1)  # row
```
gramalingam, line 2553: Can `dim1` be negative? Especially -1? Same for `dim2`.

justinchuby: ah it can. Thanks for pointing this out!

justinchuby: Reverted to use gather. Added a comment

justinchuby: Hmm but its handled in 
image

PR #1757, opened 2024-07-25
[torchlib] Add missing ops (im2col)

shubhambhokare1:


justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -658,16 +658,140 @@ def aten_huber_loss_backward(
...
+    #      [7., 8.]],
+    #     [[5., 6.],
+    #      [8., 9.]]]]]]
+    # Transpose dims 3 (depth) and 4 (rows), and then reshape to output shape (1, 1, 4, 4) to get:
+    #  [[[1., 2., 4., 5.],
+    #    [2., 3., 5., 6.],
+    #    [4., 5., 7., 8.],
+    #    [5., 6., 8., 9.]]]
+    output = op.Gather(padded_input, blocks_row_indices, axis=2)
+    output = op.Gather(output, blocks_col_indices, axis=4)
```
justinchuby, line 792: Possible to use slice, which is faster?

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -1396,6 +1520,7 @@ def aten_one_hot(self: TensorType, num_classes: int = -1) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::pad", trace_only=True)
```
justinchuby, line 1523: We can remove this

justinchuby (changes requested):
nit: blocking on the pad change, otherwise exporter will fail

onnxscript/function_libs/torch_lib/ops/nn.py
shubhambhokare1: I think we can use Slice, however the indices would need to be transformed to starts, ends format adding extra Reshape and Split nodes

justinchuby: I see. Then this lgtm. Thanks for explaining!

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -658,16 +658,141 @@ def aten_huber_loss_backward(
     raise NotImplementedError()
 
 
+def _get_im2col_indices_along_dim(
+    input_d: int,
+    kernel_size_d: int,
+    dilation_d: int,
+    padding_d: int,
+    stride_d: int,
+):
```
gramalingam, line 667: I see all the inputs here are expected to be known at export-time? That is, this is in trace-only mode? Why do we need to encode this logic using ONNX operators, instead of doing it directly in Python? (The ONNX encoding is useful if and when the operator becomes more dynamic, with some of these parameters being runtime-values. However, as it exists, this will become dependent on subsequent constant-folding for the current static case.)

onnxscript/function_libs/torch_lib/ops/nn.py
gramalingam: But the extra-operations can be done at export-time, is that correct? That is, they depend only on export-time values (torch parameters == onnx attributes), and not on run-time values. If so, there is no need to encode them using onnx ops, as it can be done in Python? In other words, using Slice should be doable in the trace-mode without any extra cost?

tests/function_libs/torch_lib/ops_test_data.py
```diff
@@ -283,6 +283,35 @@ def _grid_sample_input_wrangler(
     return args, kwargs
 
 
+def _im2col_input_wrangler(
+    args: list[Any], kwargs: dict[str, Any]
+) -> tuple[list[Any], dict[str, Any]]:
+    # Move kernel_size, dilation, padding and stride from args to kwargs
+    if len(args) == 5:
```
gramalingam, line 290: Is this guaranteed? Or, is it possible that the caller can mix args and kwargs?

justinchuby: This depends on how the test data is constructed and that tends to be relatively stable

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: Good point. At least input_d should be dynamic I think?

justinchuby (changes requested):
Blocking for dynamic input handling

onnxscript/function_libs/torch_lib/ops/nn.py
gramalingam: Ok. I guess it is just a matter of changing the type signature above for input_dy

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -658,16 +658,141 @@ def aten_huber_loss_backward(
...
+    # Input is always 4-D (N, C, H, W)
+    # Calculate indices of sliding blocks along spatial dimension
+    # Slide kernel over input each dim d:
+    # each dimension d ranges from 0 to input[d]+2xpadding[d]-dilation[d]x(kernel_size[d]-1)
+    # with steps = stride
+
+    blocks_d = op.Add(input_d, padding_d * 2)
+    blocks_d = op.Sub(blocks_d, (dilation_d * (kernel_size_d - 1)))
```
gramalingam, line 675: Why can't the above two be combined into one op? Only input_d is dynamic, right?

onnxscript/function_libs/torch_lib/ops/nn.py
gramalingam: I think this is worth thinking through.
* Slice is an operation that has a very regular access pattern that is easier to optimize and parallelize. But Gather is very irregular and random, harder to optimize and parallelize.
* The cost of operations on a large input tensor dominate overall cost, not cost of constant-time operations like Reshape.
* If we want to extract a million elements, creating the indices of these million elements seems potentially expensive, when it can be described using a slice-pattern with a few elements.

gramalingam: Aside: in the longer run, I wonder if we should create some utility to do kernel performance evaluation for onnxscript functions such as these? So that we can at least get some insight when comparing alternative implementations or comparing this to a pytorch implementation (eg. just using a cpu EP, etc.)?

justinchuby: kernel performance evaluation: Yes. I was thinking about this but didn't realize a good way to extract per op profile data from ort yet. If you can point me to how we can do that, it would be straightforward to create a utility to visualize it in model explorer.

gramalingam: But if the entire model consists of a single op-function call? I wasn't necessarily looking for something visual. Just knowing impl1 takes X time and impl2 takes Y time would be fine. The starting point would be a test-case for an op like im2col, we run its onnxscript impl exported to ORT as a model.

justinchuby: I wonder how correlatedly a tiny bench is with the e2e performance? Hopefully closely?

gramalingam: Good question. We will need to avoid overheads (like copying tensors, eg. due to conversion, etc.). And not count session creation (which should be easy). May be even warm up. Should be doable.

tests/function_libs/torch_lib/ops_test_data.py
shubhambhokare1: This is correct

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
gramalingam: Hi @shubhambhokare1 : I see this has been merged. I am concerned that the strategy used here might not be good, for reasons discussed above. Any thoughts about that? Thanks!

shubhambhokare1: Hi @gramalingam,

Agreed with the point about the case with a large number of elements, creating these indices and using gather might be inefficient. I think I must have missed this comment thread pre-merge. Slice might be a better option.
Will add a PR on top of this to remedy this, replacing the gathers ops with slice, I guess models using im2col should be unblocked for now. 

In regards to the second point, might be a good idea to create a single-op based evaluator for kernel performance. Will experiment and add that as part of the new PR.

PR #1761, opened 2024-07-26
Add op (batch_norm) | feat(torchlib)

titaiwangms:


titaiwangms:
torch.export.export does not use `aten::batch_norm`, but use `aten::native_batch_norm_legit` (https://github.com/pytorch/pytorch/pull/88697). We can revisit when we actually run into `batch_norm` missing op.

PR #1765, opened 2024-07-30
[torchlib] Fix more registrations

justinchuby:


titaiwangms (approved):
Do we not need to support size 0 input? I feel like it's in the test for a reason?

justinchuby:
> Do we not need to support size 0 input? I feel like it's in the test for a reason?

There are tests because the pytorch operator supports that, but in practice I don't see that being meaningful because it's just doing an expand. We can come back to it when it becomes an issue.

PR #1770, opened 2024-08-01
Track nodes added in pattern builder

gramalingam:
This fixes a couple of issues with the graph pattern builder by explicitly tracking the nodes added/created. This ensures that node ordering is exactly the same as what the user specifies (which helps with debugging and the verbose logs). In addition, we use a context manager to track the nodes added via the use of overloaded operators like + and *.

This also impacts how the "commuted" GraphPatterns are constructed. This has also been cleaned up (allowing commute to handle multiple output nodes as well).

Remove the unused "onnxop" imports left over after some refactoring a while back.

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -375,7 +375,7 @@ class MatchResult:
         self._matched_nodes.extend(other._matched_nodes)  # type: ignore[attr-defined]
 
 
-_pattern_builder: OpPatternBuilder = onnxop
+_pattern_builder: OpsetPatternBuilder = onnxop
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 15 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -523,36 +551,19 @@ def matches(self, node: ir.Node, match: MatchResult) -> MatchResult:
...
-                outputs,
-                self.allow_other_attributes,
-            )
-            for input in inputs
-        ]
+        copy = NodePattern(
+            self.domain, self.op, inputs, self.attributes, outputs, self.allow_other_attributes
+        )
+        node_map[self] = copy
+        return copy
```
justinchuby, line 566:
```suggestion
        copied = NodePattern(
            self.domain, self.op, inputs, self.attributes, outputs, self.allow_other_attributes
        )
        node_map[self] = copied
        return copied
```

Prefer a different name than copy as it is a built in module.

justinchuby (approved):
🎉

PR #1772, opened 2024-08-02
Refactor builder out as an utility

gramalingam:
Move the IR builder utility out as a separate utility.

TODO:
* Should this be moved into the `ir` folder?
* Should `Builder` be merged with the `Tape` class?
* Eventually, merge this with trace-mode onnxscript and expose it to end users.


justinchuby:
All sounds good!

justinchuby (approved):

PR #1773, opened 2024-08-04
Fix missing type in _add_attribute_to_torchscript_node for Deberta models

xadupre:


onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
```diff
@@ -459,8 +459,19 @@ def _add_attribute_to_torchscript_node(
...
-        raise TypeError(f"Unsupported sequence type '{type(value)}' for attribute '{key}'")
-    raise TypeError(f"Unsupported attribute type '{type(value)}' for attribute '{key}'")
+        raise TypeError(
+            f"Unsupported sequence type '{type(value)}' for attribute '{key}' in "
+            f"node={node!r}, value is {value!r}"
+        )
+    if "TensorProtoDataType" in str(type(value)):
+        # torch._C._onnx.TensorProtoDataType
+        itype = int(value)
+        return node.i_(key, itype)
```
justinchuby, line 469:
```suggestion
        return node.i_(key, int(value))
```

onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
```diff
@@ -459,8 +459,19 @@ def _add_attribute_to_torchscript_node(
             return node.fs_(key, list(value))  # type: ignore[arg-type]
         if isinstance(value[0], int):
             return node.is_(key, list(value))  # type: ignore[attr-defined]
-        raise TypeError(f"Unsupported sequence type '{type(value)}' for attribute '{key}'")
-    raise TypeError(f"Unsupported attribute type '{type(value)}' for attribute '{key}'")
+        raise TypeError(
+            f"Unsupported sequence type '{type(value)}' for attribute '{key}' in "
+            f"node={node!r}, value is {value!r}"
+        )
+    if "TensorProtoDataType" in str(type(value)):
```
justinchuby, line 466: I still don't understand why we would see `torch._C._onnx.TensorProtoDataType` here. Do you know what the caller is?

onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
xadupre: done

onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
xadupre: I did not find any issue with the caller. I think this issue can be fixed by inserting an exception when the container for the attribute is created and then you would know why this type is here. I did this fix to validate the fix works. The type is an int enum and I assume in many places it is considered as such except here. The best fix would be to catch this issue when this value enters the IR but you know better than me all the places where it happens.

justinchuby (approved):

PR #1777, opened 2024-08-05
Create a executable utility to call optimizer

justinchuby:
Usage:

```
python optimize.py model.onnx optimized_model.onnx
```

### DIFF
--- a/tools/optimize.py
+++ b/tools/optimize.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Utility for optimizing ONNX models.
+
+Usage:
+    python optimize.py model.onnx optimized_model.onnx
+"""
+
+import argparse
+import os
+
+import onnx
+import onnx.inliner
+
+import onnxscript
+
+
+def main(args) -> None:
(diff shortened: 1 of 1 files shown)


tools/optimize.py
```diff
@@ -0,0 +1,44 @@
...
+
+    model = onnx.load(path, load_external_data=False)
+    # Hack: Change the working directory to the model directory so the optimizer
+    # can load external data files with relative paths.
+    # TODO: Remove this hack by fixing the optimizer to handle external data files properly.
+    pwd = os.getcwd()
+    model_dir = os.path.dirname(path)
+    os.chdir(model_dir)
+    model = onnxscript.optimizer.optimize(model)
+    model = onnx.inliner.inline_local_functions(model)
```
titaiwangms, line 32: Is this necessary?

titaiwangms (approved):
Do we plan to use it anywhere?

tools/optimize.py
justinchuby: Since the current optimizer will specialize functions and retain them I wanted to flatten everything with this

justinchuby:
> Do we plan to use it anywhere?

Just a handy script. I am using it to understand the difference between torchscript models and the new onnx models

PR #1785, opened 2024-08-07
Fix none value for out_dtype in quantized_decomposed_dequantize_per_tensor

xadupre:
That fixes the dort unit test series.


justinchuby (approved):
Thanks!

PR #1786, opened 2024-08-07
Rename subfolder onnxruntime into onnxruntime_patterns

xadupre:


justinchuby:
How about making it private? We should expose individual patterns in a more structured way

gramalingam (approved):

PR #1788, opened 2024-08-07
[torchlib] Mark trig ops as traceable

justinchuby:
They are one operator simple functions.


titaiwangms (approved):
This will also be inlined in optimizer.

PR #1789, opened 2024-08-07
[torchlib] Fix _log_softmax

justinchuby:
Fix _log_softmax by moving the IsScalar call to the top so it can be eagerly evaluated.

Also specify the squeeze axis explicitly to improve compatibility with ORT: https://github.com/microsoft/onnxruntime/issues/21661

This should fix a runtime error in XGLMForCausalLM

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -89,7 +89,7 @@ def aten__log_softmax_half(
         self = op.Unsqueeze(self, op.Constant(value_ints=[0]))
     result = op.LogSoftmax(self, axis=dim)
     if self_is_scalar:
-        result = op.Squeeze(result, op.Constant(value_ints=[0]))
+        result = op.Squeeze(result)
     return result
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -82,11 +82,15 @@ def aten__log_softmax_half(
...
+    self_is_scalar = IsScalar(self)
     if half_to_float:
         self = op.Cast(self, to=FLOAT.dtype)
-
-    return aten__log_softmax(self, dim, half_to_float)
+    if self_is_scalar:
+        self = op.Unsqueeze(self, op.Constant(value_ints=[0]))
+    result = op.LogSoftmax(self, axis=dim)
+    if self_is_scalar:
+        result = op.Squeeze(result)
```
justinchuby, line 92:
```suggestion
        result = op.Squeeze(result, op.Constant(value_ints=[0]))
```

titaiwangms (approved):

PR #1795, opened 2024-08-08
[torchlib] Opportunistically implement prims ops

justinchuby:
Opportunistically implement prims ops since I have seen `prims::mul` being used in a model. No tests (yet) as `aten::mul` and `prims::mul` have different signatures.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/prims.py
+++ b/onnxscript/function_libs/torch_lib/ops/prims.py
@@ -22,28 +22,33 @@ from onnxscript.onnx_opset import opset18 as op
 from onnxscript.onnx_types import TensorType
 
 
-def prims_abs(self: TensorType) -> TensorType:
+@torch_op("prims::abs", traceable=True)
+def prims_abs(self: TTensor) -> TTensor:
     """abs(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.Abs(self)
 
 
+@torch_op("prims::acos", traceable=True)
 def prims_acos(self: TensorType) -> TensorType:
     """acos(Tensor self) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.Acos(self)
 
 
+@torch_op("prims::acosh", traceable=True)
 def prims_acosh(self: TensorType) -> TensorType:
     """acosh(Tensor self) -> Tensor"""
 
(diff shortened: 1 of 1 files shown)


titaiwangms:
I think it's https://github.com/pytorch/pytorch/blob/40767e84685fe9bba7a93f9cda35125a9a059943/torch/onnx/_internal/fx/decomposition_table.py#L103C14-L103C25

Being removed away

justinchuby:
> I think it's [pytorch/pytorch@`40767e8`/torch/onnx/_internal/fx/decomposition_table.py#L103C14-L103C25](https://github.com/pytorch/pytorch/blob/40767e84685fe9bba7a93f9cda35125a9a059943/torch/onnx/_internal/fx/decomposition_table.py#L103C14-L103C25)
> 
> Being removed away

I think that's fine. This just means we are getting more decomps for higher coverage. Thoughts?

titaiwangms:
It means we are allowing decomp_table to include unsupported ops -> prims ops.

justinchuby:
The decomp analysis for `mobilevit_s` where we saw this op:

## Decomposition comparison

Ops exist only in the ExportedProgram before decomposition: `['aten.adaptive_avg_pool2d.default', 'aten.conv2d.default', 'aten.dropout.default', 'aten.layer_norm.default', 'aten.linear.default', 'aten.scaled_dot_product_attention.default']`

Ops exist only in the ExportedProgram after decomposition: `['aten._scaled_dot_product_efficient_attention.default', 'aten._softmax.default', 'aten._to_copy.default', 'aten.addmm.default', 'aten.bmm.default', 'aten.convolution.default', 'aten.expand.default', 'aten.mean.dim', 'aten.native_layer_norm.default', 'aten.t.default', 'prims.mul.default']`

justinchuby:
I haven't seen this before from the same ops. Maybe pytorch updated.

titaiwangms:
Hmm I recall in default, exported program doesn't want prims as well: https://github.com/pytorch/pytorch/blob/40767e84685fe9bba7a93f9cda35125a9a059943/torch/_decomp/__init__.py#L258

justinchuby:
@titaiwangms In any case, having the implementation in torchlib doesn't hurt so I think we should merge this PR

titaiwangms (approved):

Issue #1696, opened 2024-06-22
[exporter] Create a pass to turn tensors into external tensors


justinchuby:
Create a pass to turn all tensors into external tensors in the IR. This can be seen as a pass with side effects. Needed to support efficient serialization in the exporter.

justinchuby:
Tensors need to align at 64kb: https://github.com/onnx/onnx/pull/6247

justinchuby:
Basically implementing https://onnx.ai/onnx/api/external_data_helper.html#convert-model-to-external-data

PR #1796, opened 2024-08-12
[IR] Pass to offload external tensors

shubhambhokare1:
Fix https://github.com/microsoft/onnxscript/issues/1696

### DIFF
--- a/onnxscript/ir/__init__.py
+++ b/onnxscript/ir/__init__.py
@@ -78,6 +78,7 @@ __all__ = [
 ]
 
 from onnxscript.ir import passes, serde, traversal
+from onnxscript.ir import _external_data
 from onnxscript.ir._convenience import tensor
 from onnxscript.ir._core import (
     Attr,
--- a/onnxscript/ir/__init__.py
+++ b/onnxscript/ir/__init__.py
@@ -70,6 +70,7 @@ __all__ = [
     # IR Tensor initializer
     "tensor",
     # Pass infrastructure
+    "_external_data",
     "passes",
     "traversal",
     # IO
@@ -77,8 +78,7 @@ __all__ = [
     "save",
 ]
 
-from onnxscript.ir import passes, serde, traversal
-from onnxscript.ir import _external_data
+from onnxscript.ir import _external_data, passes, serde, traversal
 from onnxscript.ir._convenience import tensor
 from onnxscript.ir._core import (
     Attr,
--- a/onnxscript/ir/__init__.py
+++ b/onnxscript/ir/__init__.py
@@ -70,7 +70,6 @@ __all__ = [
     # IR Tensor initializer
     "tensor",
     # Pass infrastructure
-    "_external_data",
     "passes",
     "traversal",
     # IO
@@ -78,7 +77,7 @@ __all__ = [
     "save",
 ]
 
-from onnxscript.ir import _external_data, passes, serde, traversal
+from onnxscript.ir import passes, serde, traversal
 from onnxscript.ir._convenience import tensor
 from onnxscript.ir._core import (
     Attr,
--- a/onnxscript/ir/_external_data.py
+++ b/onnxscript/ir/_external_data.py
@@ -19,7 +19,7 @@ class ExternalDataInfo:
     """
     A class that stores information about a tensor that is to be stored as external data.
 
-    Args:
+    Attributes:
         name: The name of the tensor that is to be stored as external data.
         offset: The offset is used to determine where exactly in the file the external data is written to.
         length: Stores the size of the tensor.
@@ -75,7 +75,7 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
 # Loading external data
 
 
-def load_external_tensor(mmap_buffer, external_tensor):
+def _load_external_tensor(mmap_buffer, external_tensor):
     """
     dt = np.dtype(external_tensor.dtype.numpy()).newbyteorder("<")
     if external_tensor.dtype in {_enums.DataType.INT4, _enums.DataType.UINT4}:
@@ -104,7 +104,7 @@ def load_external_tensor(mmap_buffer, external_tensor):
     return raw_data_np
 
 
-def check_external_data_file(model: _core.Model, file_path: str | os.PathLike) -> None:
+def _check_external_data_file(model: _core.Model, file_path: str | os.PathLike) -> None:
     """
     Check if file to which external data is to be written to is empty.
     If file already consists of external data, load external data.
@@ -130,7 +130,7 @@ def check_external_data_file(model: _core.Model, file_path: str | os.PathLike) -
                 assert external_tensor.offset is not None
                 data_file.seek(external_tensor.offset)
 
-                raw_data_np = load_external_tensor(raw_data, external_tensor)
+                raw_data_np = _load_external_tensor(raw_data, external_tensor)
                 tensor = _core.Tensor(
                     raw_data_np, name=external_tensor.name, dtype=external_tensor.dtype
                 )
@@ -168,7 +168,7 @@ def compute_new_offset(
     return current_offset
 
 
-def set_external_data(
+def _set_external_data(
     tensor: _protocols.TensorProtocol,
     current_offset: int,
     align_offset: bool = True,
@@ -194,7 +194,7 @@ def set_external_data(
     return external_data_info
 
 
-def save_external_data(
+def _save_external_data(
     external_data_info: list[tuple[_core.Value, ExternalDataInfo]],
     file_path: str | os.PathLike,
 ) -> None:
@@ -220,7 +220,7 @@ def save_external_data(
             data_file.write(raw_data)
 
 
-def store_as_external_tensors(
+def _store_as_external_tensors(
     model: _core.Model,
     external_data_info: list[tuple[_core.Value, ExternalDataInfo]],
     file_path: str | os.PathLike,
@@ -278,7 +278,7 @@ def to_external_data(
     # Check if file is empty. If not, load pre-existing external data.
     if os.stat(path).st_size != 0:
         if load_external_to_memory:
-            check_external_data_file(model, path)
+            _check_external_data_file(model, path)
         else:
             # If exisiting external tensors are not loaded to memory, copy the external data to a temporary location
             os.rename(path, parent_path + "/temp_" + file_name)
@@ -299,7 +299,7 @@ def to_external_data(
 
     current_offset = 0
     for value, tensor in tensors:
-        tensor_info = set_external_data(
+        tensor_info = _set_external_data(
             tensor,
             current_offset,
             align_offset=align_offset,
@@ -309,10 +309,10 @@ def to_external_data(
         external_data_info.append((value, tensor_info))
         current_offset = tensor_info.offset + tensor_info.length
 
-    save_external_data(external_data_info, path)
+    _save_external_data(external_data_info, path)
 
     # Convert initializers to ExternalTensors
-    model = store_as_external_tensors(model, external_data_info, path)
+    model = _store_as_external_tensors(model, external_data_info, path)
 
     # Append pre-exiting external tensors to new data file.
     return model
--- a/onnxscript/ir/_external_data.py
+++ b/onnxscript/ir/_external_data.py
@@ -150,7 +150,7 @@ def _set_external_data(
 
 
 def _save_external_data(
-    external_data_info: list[tuple[_core.Value, ExternalDataInfo]],
+    external_data_info: list[tuple[_protocols.TensorProtocol, ExternalDataInfo]],
     file_path: str | os.PathLike,
 ) -> None:
     """
@@ -161,15 +161,10 @@ def _save_external_data(
         file_path: Location to which external data is to be stored.
     """
     with open(file_path, "w+b") as data_file:
-        for value, tensor_info in external_data_info:
+        for tensor, tensor_info in external_data_info:
             current_offset = tensor_info.offset
-            assert value.const_value is not None
-            # If external tensor, retrieve raw_data from source_file
-            if isinstance(value.const_value, _core.ExternalTensor):
-                external_tensor = value.const_value
-                raw_data = external_tensor.tobytes()
-            else:
-                raw_data = value.const_value.tobytes()
+            assert tensor is not None
+            raw_data = tensor.tobytes()
             if current_offset is not None:
                 # Pad file to required offset if needed
                 file_size = data_file.tell()
@@ -179,9 +174,9 @@ def _save_external_data(
 
 
 def _convert_as_external_tensors(
-    external_data_info: list[tuple[_core.Value, ExternalDataInfo]],
+    external_data_info: list[tuple[_protocols.TensorProtocol, ExternalDataInfo]],
     file_path: str | os.PathLike,
-) -> None:
+) -> list[_core.ExternalTensor]:
     """
     Convert the tensors (stored within the values) written as external data to _core.ExternalTensor types.
 
@@ -189,8 +184,8 @@ def _convert_as_external_tensors(
         external_data_info: A collection of external data information stored for each tensor to be written as external data.
         file_path: Location to which external data is to be stored.
     """
-    for value, tensor_info in external_data_info:
-        tensor = value.const_value
+    external_tensors: list[_core.ExternalTensor] = []
+    for tensor, tensor_info in external_data_info:
         assert tensor is not None
         external_tensor = _core.ExternalTensor(
             file_path,
@@ -200,7 +195,42 @@ def _convert_as_external_tensors(
             shape=tensor.shape,  # type: ignore[arg-type]
             name=tensor.name,  # type: ignore[arg-type]
         )
-        value.const_value = external_tensor
+        external_tensors.append(external_tensor)
+    return external_tensors
+
+
+def convert_tensors_to_external(
+    tensors: list[_protocols.TensorProtocol],
+    abs_path: str | os.PathLike,
+) -> list[_core.ExternalTensor]:
+    """
+    This method takes in tensors to be converted to external tensors and returns the converted external tensors.
+
+    Args:
+        tensors: Tensors to be converted to external tensors
+        abs_path: Absolute path to which external data is to be stored.
+    """
+    external_data_info: list[tuple[_protocols.TensorProtocol, ExternalDataInfo]] = []
+    # Sort all tensors based on tensor sizes, in order to avoid unneccesarry alignment.
+    # All the smaller tensors are written earlier and alignment is peformed for the larger tensors.
+    sorted_indices = sorted(range(len(tensors)), key=lambda i: tensors[i].nbytes)
+    tensors = [tensors[i] for i in sorted_indices]
+
+    current_offset = 0
+    for tensor in tensors:
+        tensor_info = _set_external_data(tensor, current_offset)
+        external_data_info.append((tensor, tensor_info))
+        current_offset = tensor_info.offset + tensor_info.length
+    _save_external_data(external_data_info, abs_path)
+
+    # Convert initializers to ExternalTensors
+    external_tensors = _convert_as_external_tensors(external_data_info, abs_path)
+    # Sort external_tensors based on original key order
+    external_tensors = [
+        external_tensors[i]
+        for i in sorted(range(len(external_tensors)), key=lambda i: sorted_indices[i])
+    ]
+    return external_tensors
 
 
 def to_external_data(
@@ -242,24 +272,12 @@ def to_external_data(
     # Iterate through all the tensors, and extract the external data information such as
     # name, offset and length.
     # TODO: Currently attributes not handled, eventually try to use _all_tensors to include attrs
-    tensors: list[tuple[_core.Value, _protocols.TensorProtocol]] = []
+    tensors: list[_protocols.TensorProtocol] = []
     for value in model.graph.initializers.values():
         if value.const_value is not None:
-            tensors.append((value, value.const_value))
-    external_data_info: list[tuple[_core.Value, ExternalDataInfo]] = []
+            tensors.append(value.const_value)
 
-    # Sort all tensors based on tensor sizes, in order to avoid unneccesarry alignment.
-    # All the smaller tensors are written earlier and alignment is peformed for the larger tensors.
-    tensors = sorted(tensors, key=lambda x: x[1].nbytes)
-
-    current_offset = 0
-    for value, tensor in tensors:
-        tensor_info = _set_external_data(tensor, current_offset)
-        external_data_info.append((value, tensor_info))
-        current_offset = tensor_info.offset + tensor_info.length
-
-    _save_external_data(external_data_info, path)
-
-    # Convert initializers to ExternalTensors
-    _convert_as_external_tensors(external_data_info, path)
+    external_tensors = convert_tensors_to_external(tensors, path)
+    for value, external_tensor in zip(model.graph.initializers.values(), external_tensors):
+        value.const_value = external_tensor
     return model
--- a/onnxscript/ir/_external_data.py
+++ b/onnxscript/ir/_external_data.py
@@ -32,10 +32,9 @@ class _ExternalDataInfo:
         length: Stores the size of the tensor.
     """
 
-    def __init__(self, name: str | None, offset: int, length: int):
-        self.name = name
-        self.offset = offset
-        self.length = length
+    name: str | None
+    offset: int
+    length: int
 
 
 def _all_tensors(
@@ -84,13 +83,15 @@ def _load_external_data_file(
     base_path: str | os.PathLike,
     relative_path: str | os.PathLike,
 ) -> list[_protocols.TensorProtocol]:
-    """
-    Load all external data that is at relative_path into memory for the provided model.
+    """Load all external data that is at relative_path into memory for the provided model.
 
     Args:
         tensors: Tensors to be converted to external tensors. They can be external tensors themselves.
         base_path: Path of base directory.
         relative_path: Path to which external data is to be stored, relative to the ONNX file.
+
+    Returns:
+        A list of ir.Tensor values.
     """
     updated_tensors: list[_protocols.TensorProtocol] = []
     for tensor in tensors:
@@ -113,8 +114,7 @@ def _compute_new_offset(
     align_threshold: int = _ALIGN_THRESHOLD,
     allocation_granularity: int = _ALLOCATION_GRANULARITY,
 ) -> int:
-    """
-    Compute the offset to align the tensor data based on the curent offset.
+    """Compute the offset to align the tensor data based on the current offset.
 
     Args:
         current_offset: Current location in the file at which tensor data will be written to.
@@ -122,6 +122,9 @@ def _compute_new_offset(
         align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
         align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
         allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+
+    Returns:
+        The updated offset value.
     """
     if align_offset and tensor_size > align_threshold:
         alignment_factor = max(4096, allocation_granularity)
@@ -151,8 +154,7 @@ def _save_external_data(
     external_data_info: list[tuple[_protocols.TensorProtocol, _ExternalDataInfo]],
     file_path: str | os.PathLike,
 ) -> None:
-    """
-    Write tensor data to an external file according to information stored in ExternalDataInfo objects.
+    """Write tensor data to an external file according to information stored in ExternalDataInfo objects.
 
     Args:
         external_data_info: A collection of external data information stored for each tensor to be written as external data.
@@ -175,13 +177,15 @@ def _convert_as_external_tensors(
     base_path: str | os.PathLike,
     relative_path: str | os.PathLike,
 ) -> list[_core.ExternalTensor]:
-    """
-    Convert the tensors (stored within the values) written as external data to _core.ExternalTensor types.
+    """Convert the tensors (stored within the values) written as external data to _core.ExternalTensor types.
 
     Args:
         external_data_info: A collection of external data information stored for each tensor to be written as external data.
         base_path: Path of base directory.
         relative_path: Path to which external data is to be stored, relative to the ONNX file.
+
+    Returns:
+        A list of external tensors.
     """
     external_tensors: list[_core.ExternalTensor] = []
     for tensor, tensor_info in external_data_info:
@@ -205,14 +209,16 @@ def convert_tensors_to_external(
     relative_path: str | os.PathLike,
     load_external_to_memory: bool = False,
 ) -> list[_core.ExternalTensor]:
-    """
-    Convert a sequence of any TensorProtocol tensors to external tensors.
+    """Convert a sequence of any TensorProtocol tensors to external tensors.
 
     Args:
         tensors: Tensors to be converted to external tensors. They can be external tensors themselves.
         base_path: Path of base directory.
         relative_path: Path to which external data is to be stored, relative to the ONNX file.
         load_external_to_memory: If set to true, loads external tensors present in the same file path as destination path to memory.
+
+    Returns:
+        A list of external tensors derived from a list of input tensors.
     """
     path = os.path.join(base_path, relative_path)
     # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
@@ -273,14 +279,16 @@ def to_external_data(
     relative_path: str | os.PathLike,
     load_external_to_memory: bool = False,
 ) -> _core.Model:
-    """
-    Set all tensors with raw data as external data.
+    """Set all tensors with raw data as external data.
 
     Args:
         model: Model to process.
         base_path: Path of base directory.
         relative_path: Path to which external data is to be stored, relative to the ONNX file.
         load_external_to_memory: If set to true, loads external tensors present in the same file path as destination path to memory. Otherwise, the external tensors are appended to file.
+
+    Returns:
+        An ir.Model with all tensors with raw data converted to external tensors.
     """
 
     # Get all the tensors in the graph which are to be stored as external data.
--- a/onnxscript/ir/_external_data.py
+++ b/onnxscript/ir/_external_data.py
@@ -12,9 +12,27 @@ from typing import Iterator
 from onnxscript.ir import _core, _enums, _protocols, traversal
 
 
+class ExternalDataInfo:
+    """
+    A class that stores information about a tensor that is to be stored as external data.
+
+    Args:
+        name: The name of the tensor that is to be stored as external data.
+        offset: The offset is used to determine where exactly in the file the external data is written to.
+        length: Stores the size of the tensor.
+    """
+
+    def __init__(self, name: str, offset: int, length: int):
+        self.name = name
+        self.offset = offset
+        self.length = length
+
+
 def _all_tensors(
     graph: _core.Graph | _core.GraphView, include_attributes: bool = False
-) -> Iterator[_protocols.TensorProtocol]:
+) -> Iterator[
+    tuple[_protocols.TensorProtocol, _protocols.ValueProtocol | _protocols.AttributeProtocol]
+]:
     """Iterate over all tensors in the graph.
 
     Args:
@@ -27,7 +45,7 @@ def _all_tensors(
     # Yield all tensors in initializers
     for value in graph.initializers.values():
         if value.const_value is not None:
-            yield value.const_value
+            yield (value, value.const_value)
     if not include_attributes:
         return
     # Look at constant attributes in nodes
@@ -36,9 +54,9 @@ def _all_tensors(
             if isinstance(attr, _core.RefAttr):
                 continue
             if attr.type == _enums.AttributeType.TENSOR and attr.value is not None:
-                yield attr.value
+                yield (attr, attr.value)
             elif attr.type == _enums.AttributeType.TENSORS and attr.value is not None:
-                yield from attr.value
+                yield from (attr, attr.value)
 
 
 def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLike) -> None:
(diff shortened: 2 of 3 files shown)


onnxscript/ir/__init__.py
```diff
@@ -69,11 +69,12 @@
     # IR Tensor initializer
     "tensor",
     # Pass infrastructure
+    "_external_data",
```
justinchuby, line 72: A private module should not be exposed here. We can have a follow up to decide how we want to expose this

onnxscript/ir/__init__.py
```diff
@@ -69,11 +69,12 @@
...
     "passes",
     "traversal",
 ]
 
-from onnxscript.ir import passes, serde, traversal
+from onnxscript.ir import _external_data, passes, serde, traversal
```
justinchuby, line 77:
```suggestion
from onnxscript.ir import passes, serde, traversal
```

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+class ExternalDataInfo:
+    def __init__(self, name, offset, length):
+        self.name = name
+        self.offset = offset
+        self.length = length
+
+
+def save_external_data(
+    initializers,
+    file_path,
```
justinchuby, line 19: I would add typing annotations to all functions

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+    allocation_granularity: int = 65536,  # 64KB
+):
+    # Store external data such as name, offset and size
+    external_data_info = dict()
+
+    # Create file if it doesn't exist
+    if not os.path.isfile(file_path):
+        with open(file_path, "ab"):
+            pass
```
justinchuby, line 28: Is this neeeded?

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+
+def save_external_data(
+    initializers,
+    file_path,
+    allocation_granularity: int = 65536,  # 64KB
+):
+    # Store external data such as name, offset and size
+    external_data_info = dict()
+
+    # Create file if it doesn't exist
```
justinchuby, line 25: I think we should rather create the parent directory if it does not exsit. We can use pathlib for this

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+        self.length = length
+
+
+def save_external_data(
+    initializers,
+    file_path,
+    allocation_granularity: int = 65536,  # 64KB
+):
+    # Store external data such as name, offset and size
+    external_data_info = dict()
```
justinchuby, line 23: I wonder why ruff doesn't catch this
```suggestion
    external_data_info: dict[str, ???] = {}
```

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Create file if it doesn't exist
+    if not os.path.isfile(file_path):
+        with open(file_path, "ab"):
+            pass
+
+    with open(file_path, "r+b") as data_file:
+        current_offset = 0
+        for i_name, i_value in initializers.items():
```
justinchuby, line 32:
```suggestion
        for name, value in initializers.items():
```

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+            tensor_val = i_value
+            if isinstance(tensor_val, _core.Value):
+                tensor_val = i_value.const_value()  # pylint: disable=protected-access
```
justinchuby, line 35:
```suggestion
```

value is always _core.Value

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+            assert isinstance(tensor_val, _core.Tensor)
```
justinchuby, line 36: It can be any object conforming to ir.TensorProtocol, so a type checking is not needed.

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+            raw_data = tensor_val.tobytes()
+            tensor_size = tensor_val.size
+            # Convert each initializer to core.ExternalTensor
+            # Align tensors
+            alignment_factor = max(4096, allocation_granularity)
+            current_offset = (
+                (current_offset + alignment_factor - 1) // alignment_factor * alignment_factor
+            )
```
justinchuby, line 44: I would extract this logic to a helper function and create unit tests to harden it

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+                    data_file.write(b"\0" * (current_offset - file_size))
+                data_file.seek(current_offset)
+            data_file.write(raw_data)
+
+            # Store tensor external data
+            external_data_info[i_value.name] = ExternalDataInfo(
+                i_name,
+                current_offset,
+                tensor_size,
+            )
```
justinchuby, line 60: We can instead create a list[tuple[[Value, ExternalDataInfo]] list. This way it is then possible to operate on the const_value field of the Values directly.

I would also consider separating the logic for computing ExternalDataInfo from the logic of offloading to disk to make it easier to test

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+
+            # Update offset
+            current_offset += tensor_size
+    return external_data_info
+
+
+def convert_model_to_external_data(model: _core.Model, base_path: str, file_path: str = ""):
```
justinchuby, line 67: Always support os.PathLike in addition to str when an argument is a file path. `os.path.join` will handle both

```suggestion
def convert_model_to_external_data(model: _core.Model, base_path: str | os.PathLike, file_path: str = ""):
```

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+    file_path = os.path.join(base_path, file_path)
+    external_data_info = save_external_data(model.graph.initializers, file_path)
+
+    # Convert initializers to ExternalTensors
+    for i_name, i_value in model.graph.initializers.items():
+        assert i_name in external_data_info
+        tensor_info = external_data_info[i_name]
+        new_external_tensor = _core.ExternalTensor(
```
justinchuby, line 75:
```suggestion
        external_tensor = _core.ExternalTensor(
```

nit: conciseness

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+            current_offset += tensor_size
+    return external_data_info
+
+
+def convert_model_to_external_data(model: _core.Model, base_path: str, file_path: str = ""):
+    file_path = os.path.join(base_path, file_path)
+    external_data_info = save_external_data(model.graph.initializers, file_path)
+
+    # Convert initializers to ExternalTensors
+    for i_name, i_value in model.graph.initializers.items():
```
justinchuby, line 72:
```suggestion
    for name, value in model.graph.initializers.items():
```

onnxscript/ir/_external_data.py
justinchuby, line 67:
```suggestion
def convert_model_to_external_data(model: _core.Model, base_path: str, file_path: str = "") -> None:
```

nit: return types too, here and elsewhere

onnxscript/ir/_external_data_test.py
```diff
@@ -0,0 +1,88 @@
...
+            "",
+            "Op_1",
+            inputs=[node_0.outputs[0]],
+            num_outputs=1,
+            name="node_1",
+        )
+        graph = ir.Graph(
+            inputs=node_0.inputs,  # type: ignore
+            outputs=[node_1.outputs[0]],
+            initializers=[tensor1, tensor2],
```
justinchuby, line 59: initializers should be values:

```
ir.Value(..., const_value=tensor1)
```

onnxscript/ir/_external_data_test.py
```diff
@@ -0,0 +1,88 @@
...
+import tempfile
+import unittest
+
+import numpy as np
+
+from onnxscript import ir
+from onnxscript.ir import _core, _external_data
+
+
+class ExternalTensorTest(unittest.TestCase):
```
justinchuby, line 14: I would create unit tests for the offset logic, thanks!

onnxscript/ir/_external_data.py
```diff
@@ -0,0 +1,84 @@
...
+        self.name = name
+        self.offset = offset
+        self.length = length
+
+
+def save_external_data(
+    initializers,
+    file_path,
+    allocation_granularity: int = 65536,  # 64KB
+):
```
justinchuby, line 21: I would add a docstring and document what happens when the data file does exist (overwriten)

justinchuby:
For tensors smaller than a certain size, we can decide to not align them to save disks pace. The ONNX PR used 1mb which I think is reasonable.

shubhambhokare1:
> For tensors smaller than a certain size, we can/should decide to not align them to save disks pace. I would move all small tensors to the start of the data file so there is some reordering you may consider doing. The ONNX PR used 1mb which I think is reasonable.

Is this related to the implementation of align_threshiold?

justinchuby:
That's right. Additionally I would also consider what happens when some existing external tensors reference the data file being written to. Do we need to copy them out first? To a temp disk location?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    # Check if the file is empty
+    if os.stat(file_path).st_size == 0:
+        return
+    # If file is not empty, load external data
+    with open(file_path, "r+b") as data_file:
+        for value in model.graph.initializers.values():
+            if isinstance(value.const_value, _core.ExternalTensor):
+                external_tensor = value.const_value
+                external_tensor_path = (
+                    pathlib.Path(external_tensor._base_dir) / external_tensor._path
```
justinchuby, line 96: Always access the public property instead of the private one.
```suggestion
                    pathlib.Path(external_tensor.base_dir) / external_tensor._path
```

onnxscript/ir/_external_data.py
```diff
@@ -7,11 +7,30 @@
 __all__ = ["set_base_dir"]
 
 import os
+import pathlib
```
justinchuby, line 10: I think we can use os.makedirs() and others instead.

onnxscript/ir/_external_data.py
```diff
@@ -7,11 +7,30 @@
...
+import numpy as np
+
 from onnxscript.ir import _core, _enums, _protocols, traversal
 
 
+class ExternalDataInfo:
+    """
+    A class that stores information about a tensor that is to be stored as external data.
+
+    Args:
```
justinchuby, line 22:
```suggestion
    Attributes:
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+                external_tensor_path = (
+                    pathlib.Path(external_tensor._base_dir) / external_tensor._path
+                )
+                assert external_tensor_path == file_path
+                data_file.seek(external_tensor.offset)
+
+                # Read raw data from file based on offset and length
+                # TODO: Convert this into a more robust function
+                raw_data = data_file.read(external_tensor.length)
+                raw_data = np.frombuffer(raw_data, dtype=external_tensor.dtype.numpy())
```
justinchuby, line 104: So is this loading all initializers to disk? What happens when there is not enough memory?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        file_path: Path to which external data is to be stored.
+        align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+    """
+    file_path = pathlib.Path(base_path) / file_path
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    try:
+        file_path.mkdir(parents=True, exist_ok=True)
+    except:
```
justinchuby, line 247: I don't think the try except is needed?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        pass
+    # Check if file is empty. If not, load pre-existing external data.
```
justinchuby, line 249: Prob a better idea to not load the data if possible. I think we should be safe to assume there is more disk than there is memory

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    # name, offset and length.
+    # TODO: Currently attributes not handled, eventually try to use _all_tensors to include attrs
+    tensors: list[tuple[_core.Value, _protocols.TensorProtocol]] = []
+    for value in model.graph.initializers.values():
+        if value.const_value is not None:
+            tensors.append((value, value.const_value))
+    external_data_info: list[tuple[_core.Value, ExternalDataInfo]] = []
+
+    # Sort all tensors based on tensor sizes, in order to avoid unneccesarry alignment.
+    # All the smaller tensors are written earlier and alignment is peformed for the larger tensors.
```
justinchuby, line 263: Nice

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,213 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+            align_threshold=align_threshold,
+            allocation_granularity=allocation_granularity,
+        )
+        external_data_info.append((value, tensor_info))
+        current_offset = tensor_info.offset + tensor_info.length
+
+    save_external_data(external_data_info, file_path)
+
+    # Convert initializers to ExternalTensors
+    model = store_as_external_tensors(model, external_data_info, file_path)
```
justinchuby, line 281:
```suggestion
    model = _store_as_external_tensors(model, external_data_info, file_path)
```

I would make all internal functions private

onnxscript/ir/_external_data_test.py
```diff
@@ -1,12 +1,16 @@
...
 import unittest
 
 import onnx
 import onnx.external_data_helper
 
+import numpy as np
+
 from onnxscript import ir
-from onnxscript.ir import _external_data
+from onnxscript.ir import _core, _external_data
```
justinchuby, line 13:
```suggestion
from onnxscript.ir import _external_data
```

for the classes that is not from _external_data (aka not tested in this file) use `ir.xxx` directly so when users are using our tests as reference they can write the same thing and not have to import internal modules.

onnxscript/ir/_external_data_test.py
```diff
@@ -55,5 +59,167 @@ def test_set_base_dir_sets_base_dir_for_all_external_tensors(self):
...
+            current_offset,
+            tensor_size,
+        )
+        self.assertNotEqual(current_offset, new_offset_1)
+        self.assertNotEqual(current_offset, new_offset_2)
+        self.assertNotEqual(new_offset_1, new_offset_2)
+
+
+class ExternalTensorTest(unittest.TestCase):
+    """Test the memory mapped external tensor class."""
```
justinchuby, line 114: Is this needed/accurate? Are there new tests we should move to core test?

justinchuby:
For tests, I would make sure to cover these scenarios:
 
1. All tensors are internal tensors. Make sure some of them are not ir.Tensor. For example create a custom tensor class that implements TensorProtocol (it can just have a static tobytes field) etc. such that all TensorProtocol tensors can work in this pass
2. Mix external tensors and internal tensors. Make sure that all tensors are written
3. Write to an existing path that is referenced by the current external tensors (that has an offset !=0). Make sure that after writing all the external tensors contain data that is current and no data is corrupted.
4. Mix external tensors that reference different data paths. Check that the pass can produce correct results
5. Check that the tensors are sorted by their size.

onnxscript/ir/_external_data.py
shubhambhokare1: Think I will create a parameter that allows the user to choose whether to store in memory or copy to disk, with copying to disk being the default. Will implement the functionalities for both the paths as well.

onnxscript/ir/_external_data.py
shubhambhokare1: See comment above.

onnxscript/ir/_external_data_test.py
shubhambhokare1: I guess this is needed for now, given the model is built using IR classes as compared to the onnx helpers (in core). Will use this as a base unittest class for adding the tests you mentioned below.

justinchuby: There is one here: https://github.com/microsoft/onnxscript/blob/ce7711ea487a91baff2ae97a2db4e83b3ae442bb/onnxscript/ir/_core_test.py#L180.

Should they be combined?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+                assert external_tensor.offset is not None
+                data_file.seek(external_tensor.offset)
+
+                raw_data_np = _load_external_tensor(raw_data, external_tensor)
+                tensor = _core.Tensor(
+                    raw_data_np, name=external_tensor.name, dtype=external_tensor.dtype
+                )
+                value.const_value = tensor
+                assert (value.name is not None) and (value.name in model.graph.initializers)
+                model.graph.initializers[value.name] = value
```
justinchuby, line 126: We don't need to use mmap here because the external tensor class handles loading. To load into memory, simply 

1. Filter out all values with external tensor objects that you want to load into memory
2. Create a new tensor with `external_tensor.numpy()` as data.
3. do the same `value.const_value = new_tensor`
4. `model.graph.initializers[value.name] = value` is not needed as the value was not changed.

onnxscript/ir/_external_data.py
```diff
@@ -6,12 +6,31 @@
 
 __all__ = ["set_base_dir"]
 
+import mmap
```
justinchuby, line 9: We shouldn't require mmap in this file.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+        load_external_to_memory: If set to true, loads all pre-existing external tensors to memory. Otherwise, all new external tensors are appended to file.
+    """
+    path = os.path.join(base_path, file_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path, file_name = "/".join(path.split("/")[:-1]), path.split("/")[-1]
+    os.makedirs(parent_path, exist_ok=True)
+    os.makedirs(parent_path + "/temp", exist_ok=True)
+    # Check if file is empty. If not, load pre-existing external data.
```
justinchuby, line 276: The file may be not empty but non of the tensors reference the file. In this case it is save to just overwrite. I would consider the following:

1. For all tensors in the graph, check if any of the tensor is external tensor and refers to the file path
2. If so do this thing. When loading to memory, only load the filtered tensors, not all tensors.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    if os.stat(path).st_size != 0:
+        if load_external_to_memory:
+            _load_external_data_file(model, path)
+        else:
+            # If exisiting external tensors are not loaded to memory, copy the external data to a temporary location
+            os.rename(path, parent_path + "/temp/" + file_name)
```
justinchuby, line 282: The paths in the external tensors need to be updated too

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+                raw_data = value.const_value.tobytes()
+            if current_offset is not None:
+                # Pad file to required offset if needed
+                file_size = data_file.tell()
+                if current_offset > file_size:
+                    data_file.write(b"\0" * (current_offset - file_size))
+            data_file.write(raw_data)
+
+
+def _store_as_external_tensors(
```
justinchuby, line 220:
```suggestion
def _convert_to_external_tensors(
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+            file_path,
+            tensor_info.offset,
+            tensor_info.length,
+            tensor.dtype,  # type: ignore[arg-type]
+            shape=tensor.shape,  # type: ignore[arg-type]
+            name=tensor.name,  # type: ignore[arg-type]
+        )
+        value.const_value = external_tensor
+        assert value.name in model.graph.initializers
+        model.graph.initializers[value.name] = value
```
justinchuby, line 246: This is not needed?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+            if current_offset is not None:
+                # Pad file to required offset if needed
+                file_size = data_file.tell()
+                if current_offset > file_size:
+                    data_file.write(b"\0" * (current_offset - file_size))
+            data_file.write(raw_data)
+
+
+def _store_as_external_tensors(
+    model: _core.Model,
```
justinchuby, line 221:
```suggestion
```

looks like the model is not needed

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+        load_external_to_memory: If set to true, loads all pre-existing external tensors to memory. Otherwise, all new external tensors are appended to file.
+    """
+    path = os.path.join(base_path, file_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path, file_name = "/".join(path.split("/")[:-1]), path.split("/")[-1]
+    os.makedirs(parent_path, exist_ok=True)
+    os.makedirs(parent_path + "/temp", exist_ok=True)
```
justinchuby, line 275: Always use os.path.join to concatenate paths.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        base_path: Path of base directory.
+        file_path: Path to which external data is to be stored.
+        align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+        load_external_to_memory: If set to true, loads all pre-existing external tensors to memory. Otherwise, all new external tensors are appended to file.
+    """
+    path = os.path.join(base_path, file_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path, file_name = "/".join(path.split("/")[:-1]), path.split("/")[-1]
```
justinchuby, line 273: os.path.join

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    Call to set all tensors with raw data as external data.
+
+    Args:
+        model: Model to be converted.
+        base_path: Path of base directory.
+        file_path: Path to which external data is to be stored.
+        align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
+        load_external_to_memory: If set to true, loads all pre-existing external tensors to memory. Otherwise, all new external tensors are appended to file.
```
justinchuby, line 269:
```suggestion
        load_external_to_memory: If set to true, loads all pre-existing external tensors to memory. Otherwise, all new external tensors are appended to file.
```

we don't need to load "all" external tensors. Only the conflicting ones.

> Otherwise, all new external tensors are appended to file.

I would update to make accurate

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    return model
+
+
+def to_external_data(
+    model: _core.Model,
+    base_path: str | os.PathLike,
+    file_path: str = "",
+    align_offset: bool = True,
+    align_threshold: int = 1048576,  # 1MB,
+    allocation_granularity: int = 65536,  # 64KB
```
justinchuby, line 256:
```suggestion
```

I think for this function we can just hard code these values (I would move as internal global constants, to keep the interface simple We can add them as options when there is a need.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +70,245 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        value.const_value = external_tensor
+        assert value.name in model.graph.initializers
+        model.graph.initializers[value.name] = value
+    return model
+
+
+def to_external_data(
+    model: _core.Model,
+    base_path: str | os.PathLike,
+    file_path: str = "",
```
justinchuby, line 253: relative_path, etc. for accuracy? Otherwise file_path can create confusion with the other path.

Also clarify what `""` represents? Should we use `None`?

onnxscript/ir/_external_data.py
shubhambhokare1: Had done them in another function, but makes sense to put that here

onnxscript/ir/_external_data.py
shubhambhokare1: _load_external_data should take care of this logic

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    Args:
+        external_data_info: A collection of external data information stored for each tensor to be written as external data.
+        file_path: Location to which external data is to be stored.
+    """
+    with open(file_path, "w+b") as data_file:
+        for tensor, tensor_info in external_data_info:
+            current_offset = tensor_info.offset
+            assert tensor is not None
+            raw_data = tensor.tobytes()
+            if current_offset is not None:
```
justinchuby, line 168: Will current_offset ever be None?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    file_path: str | os.PathLike,
+) -> None:
+    """
+    Writes tensor data to an external file according to information stored in ExternalDataInfo object.
+
+    Args:
+        external_data_info: A collection of external data information stored for each tensor to be written as external data.
+        file_path: Location to which external data is to be stored.
+    """
+    with open(file_path, "w+b") as data_file:
```
justinchuby, line 163:
```suggestion
    with open(file_path, "wb") as data_file:
```

Use wb since we don't need to read the file

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+                # Pad file to required offset if needed
+                file_size = data_file.tell()
+                if current_offset > file_size:
+                    data_file.write(b"\0" * (current_offset - file_size))
+            data_file.write(raw_data)
+
+
+def _convert_as_external_tensors(
+    external_data_info: list[tuple[_protocols.TensorProtocol, ExternalDataInfo]],
+    file_path: str | os.PathLike,
```
justinchuby, line 178: Should we provide both the rel path and the base path? Otherwise the external tensor data cannot be accessed in the ir

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        external_tensors.append(external_tensor)
+    return external_tensors
+
+
+def convert_tensors_to_external(
+    tensors: list[_protocols.TensorProtocol],
+    abs_path: str | os.PathLike,
+) -> list[_core.ExternalTensor]:
+    """
+    This method takes in tensors to be converted to external tensors and returns the converted external tensors.
```
justinchuby, line 207:
```suggestion
    """Convert a sequence of any TensorProtocol tensors to external tensors.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+            tensor.dtype,  # type: ignore[arg-type]
+            shape=tensor.shape,  # type: ignore[arg-type]
+            name=tensor.name,  # type: ignore[arg-type]
+        )
+        external_tensors.append(external_tensor)
+    return external_tensors
+
+
+def convert_tensors_to_external(
+    tensors: list[_protocols.TensorProtocol],
```
justinchuby, line 203:
```suggestion
    tensors: Sequence[_protocols.TensorProtocol],
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    abs_path: str | os.PathLike,
+) -> list[_core.ExternalTensor]:
+    """
+    This method takes in tensors to be converted to external tensors and returns the converted external tensors.
+
+    Args:
+        tensors: Tensors to be converted to external tensors
```
justinchuby, line 210:
```suggestion
        tensors: Tensors to be converted to external tensors. They can be external tensors themselves.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        abs_path: Absolute path to which external data is to be stored.
```
justinchuby, line 211: We need a base path and a rel path. Otherwise the data written the tensor proto is incorrect.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    """
+    external_data_info: list[tuple[_protocols.TensorProtocol, ExternalDataInfo]] = []
+    # Sort all tensors based on tensor sizes, in order to avoid unneccesarry alignment.
+    # All the smaller tensors are written earlier and alignment is performed for the larger tensors.
+    sorted_indices = sorted(range(len(tensors)), key=lambda i: tensors[i].nbytes)
+    tensors = [tensors[i] for i in sorted_indices]
```
justinchuby, line 217:
```suggestion
    sorted_tensors = [tensors[i] for i in sorted_indices]
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        external_data_info.append((tensor, tensor_info))
+        current_offset = tensor_info.offset + tensor_info.length
+    _save_external_data(external_data_info, abs_path)
+
+    # Convert initializers to ExternalTensors
+    external_tensors = _convert_as_external_tensors(external_data_info, abs_path)
+    # Sort external_tensors based on original key order
+    external_tensors = [
+        external_tensors[i]
+        for i in sorted(range(len(external_tensors)), key=lambda i: sorted_indices[i])
```
justinchuby, line 231:
```suggestion
        for i in sorted_indices
```
maybe this?

justinchuby: Also consider the original:

if sorted_indices is `[2,0,1]`

for sorting `[0,1,2]`, we get keys 0->2, 1->0, 2->1, so the sorted array is `[1,2,0]`. Is this expected?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+
+
+def to_external_data(
+    model: _core.Model,
+    base_path: str | os.PathLike,
+    relative_path: str | os.PathLike,
+    load_external_to_memory: bool = False,
+) -> _core.Model:
+    """
+    Call to set all tensors with raw data as external data.
```
justinchuby, line 243:
```suggestion
    """Set all tensors as external tensor and offload their data to a file.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+
+    Args:
+        model: Model to be converted.
```
justinchuby, line 246:
```suggestion
        model: Model to process.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        base_path: Path of base directory.
+        relative_path: Path to which external data is to be stored.
```
justinchuby, line 248:
```suggestion
        relative_path: Path to which external data is to be stored, relative to the ONNX file.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        load_external_to_memory: If set to true, loads external tensors present in the same file path as destination path to memory. Otherwise, the external tensors are appended to file.
```
justinchuby, line 249: > the external tensors are appended to file.

Is this accurate?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    """
+    path = os.path.join(base_path, relative_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path = os.path.join(*path.split("/")[:-1])
```
justinchuby, line 253: Always avoid string manipulation on paths. On windows it's `\`

justinchuby: Consider `os.path.dirname`

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    tmp_path = os.path.join(base_path, "tmp")
+    os.makedirs(parent_path, exist_ok=True)
+    os.makedirs(tmp_path, exist_ok=True)
```
justinchuby, line 256: Move this line to the else branch of load_external_to_memory?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    Args:
+        model: Model to be converted.
+        base_path: Path of base directory.
+        relative_path: Path to which external data is to be stored.
+        load_external_to_memory: If set to true, loads external tensors present in the same file path as destination path to memory. Otherwise, the external tensors are appended to file.
+    """
+    path = os.path.join(base_path, relative_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path = os.path.join(*path.split("/")[:-1])
+    tmp_path = os.path.join(base_path, "tmp")
```
justinchuby, line 254: And this too: move to the else branch of load_external_to_memory?

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    os.makedirs(parent_path, exist_ok=True)
+    os.makedirs(tmp_path, exist_ok=True)
+    # Check if file is empty. If not, load pre-existing external data.
+    if os.stat(path).st_size != 0:
```
justinchuby, line 258: Check existence instead? Also, if none of the tensors from the model are referring to the file, we don't even need to bother renaming and all that. So I would consider checking if any tensors are using this file first, before doing any of this caching

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+        if load_external_to_memory:
+            _load_external_data_file(model, relative_path)
+        else:
+            # If exisiting external tensors are not loaded to memory, copy the external data to a temporary location
+            os.rename(path, os.path.join(tmp_path, relative_path))
+            for value in model.graph.initializers.values():
```
justinchuby, line 264: Caveat: If there is a non-initializer tensor that is referring to this file, that tensor is now invalid. This is a special case we are ok not handling right now but we should add a comment / FIXME here.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+                if (
+                    isinstance(value.const_value, _core.ExternalTensor)
+                    and value.const_value.path == relative_path
```
justinchuby, line 267: We need to first check if this tensor is even using the file.

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +75,209 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+    """
+    path = os.path.join(base_path, relative_path)
+    # Check if file path is valid, and create subsequent subdirectories within the path if they don't exist
+    parent_path = os.path.join(*path.split("/")[:-1])
+    tmp_path = os.path.join(base_path, "tmp")
+    os.makedirs(parent_path, exist_ok=True)
+    os.makedirs(tmp_path, exist_ok=True)
+    # Check if file is empty. If not, load pre-existing external data.
+    if os.stat(path).st_size != 0:
+        if load_external_to_memory:
```
justinchuby, line 259: Also, this logic needs to happen in `convert_tensors_to_external`. `convert_tensors_to_external` needs to be robust against all cases considered here. This function should just be a thin wrapper that prepares the list of tensors to `convert_tensors_to_external` and applies the results to the graph.

onnxscript/ir/_external_data.py
shubhambhokare1: This might be incorrect I think, 
Let's say originally, its `[a, b, c]` with the order being b>a>c
Sorted tensors are now `[c, a, b]` and sorted indices are `[2, 0, 1]

With the original logic, 
keys are `[1, 2, 0]`, so we get back `[a, b, c]`

With the proposed logic, 
keys are `[2, 0, 1]`, so we get back `[c, a, b]`

justinchuby: Thanks - I realized I was wrong with your explanation.

onnxscript/ir/_external_data.py
shubhambhokare1: What would be expected to happen in the scenario where none of the tensors reference the file however, the destination path the user is provided the same. Do we still overwrite/rename this file?

onnxscript/ir/_external_data.py
```diff
@@ -6,11 +6,37 @@
...
+    Attributes:
+        name: The name of the tensor that is to be stored as external data.
+        offset: The offset is used to determine where exactly in the file the external data is written to.
+        length: Stores the size of the tensor.
+    """
+
+    def __init__(self, name: str | None, offset: int, length: int):
+        self.name = name
+        self.offset = offset
+        self.length = length
```
justinchuby, line 38: Refer to https://docs.python.org/3/library/dataclasses.html for usage: we only need to list the attributes and their types

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +77,228 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
             tensor.base_dir = base_dir
+
+
+def _load_external_data_file(
+    tensors: Sequence[_protocols.TensorProtocol],
+    base_path: str | os.PathLike,
+    relative_path: str | os.PathLike,
+) -> list[_protocols.TensorProtocol]:
+    """
+    Load all external data that is at relative_path into memory for the provided model.
```
justinchuby, line 88:
```suggestion
    """Load all external data that is at relative_path into memory for the provided model.
```

please check and fix docstring styles

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +77,228 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+
+    Args:
+        tensors: Tensors to be converted to external tensors. They can be external tensors themselves.
+        base_path: Path of base directory.
+        relative_path: Path to which external data is to be stored, relative to the ONNX file.
```
justinchuby, line 93:
```suggestion
        relative_path: Path to which external data is to be stored, relative to the ONNX file.
        
        Returns:
            ...
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +77,228 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+
+def _compute_new_offset(
+    current_offset: int,
+    tensor_size: int,
+    align_offset: bool = _ALIGN_OFFSET,
+    align_threshold: int = _ALIGN_THRESHOLD,
+    allocation_granularity: int = _ALLOCATION_GRANULARITY,
+) -> int:
+    """
+    Compute the offset to align the tensor data based on the curent offset.
```
justinchuby, line 117:
```suggestion
    Compute the offset to align the tensor data based on the current offset.
```

onnxscript/ir/_external_data.py
```diff
@@ -51,3 +77,228 @@ def set_base_dir(graph: _core.Graph | _core.GraphView, base_dir: str | os.PathLi
...
+
+    Args:
+        current_offset: Current location in the file at which tensor data will be written to.
+        tensor_size: Size of the tensor data to be written to file.
+        align_offset: Offset will always be page aligned and alloction granularity aligned for mmap support. This is done by padding previous tensor data with zeros keeping same length. Tensor data will be aligned if > align_threshold
+        align_threshold: Alignment threshold for size of data. Having a low threshold will waste file space for small initializers. Only when tensor's data is > the page_align_threshold it will be force aligned.
+        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.
```
justinchuby, line 124:
```suggestion
        allocation_granularity: The allocation Granularity for mmap() support. Typically 64KB for Windows & 4KB for other OSes.

        Returns:
            ...
```

justinchuby (approved):
Thanks a lot! 🎉

PR #1800, opened 2024-08-13
Fix Op (scaled_dot_product_attention) | feat(torchlib)

### LINKED ISSUE #1799: [torchlib] Update `scaled_dot_product_attention` function signature
In https://github.com/pytorch/pytorch/pull/132689, the function is updated with an extra argument: enable_gqa

titaiwangms:
Fix #1799 

Add an extra argument: `enable_gqa` to unblock the export.
The real implementation: https://github.com/microsoft/onnxscript/issues/1802

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -1769,8 +1769,9 @@ def aten_scaled_dot_product_attention(
     dropout_p: float = 0.0,
     is_causal: bool = False,
     scale: Optional[float] = None,
+    enable_gqa: bool = False,
 ) -> TFloat:
-    """scaled_dot_product_attention(Tensor query, Tensor key, Tensor value, Tensor? attn_mask=None, float dropout_p=0.0, bool is_causal=False, *, float? scale=None) -> Tensor
+    """scaled_dot_product_attention(Tensor query, Tensor key, Tensor value, Tensor? attn_mask=None, float dropout_p=0.0, bool is_causal=False, *, float? scale=None, bool enable_gqa=False) -> Tensor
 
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2003,6 +2009,10 @@ def aten_scaled_dot_product_attention_bool_mask(
         is_causal and attn_mask is None
     ), "is_causal and attn_mask cannot be set at the same time"
 
+    assert (
+        not enable_gqa
+    ), "conversion of scaled_dot_product_attention not implemented if enable_gqa is True"
```
justinchuby, line 2014: Is it possible to implement? having partial implementation w/ assert is not recommended in torchlib

titaiwangms: I think GQA is now only accessible with contrib op though: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.GroupQueryAttention

justinchuby: Should be implementable as a function?

titaiwangms: I guess likely the ONNX framework (but not ORT) users should have their own custom ops if they want to run a model with this op.

justinchuby:
Ref: https://pytorch.org/docs/main/generated/torch.nn.functional.scaled_dot_product_attention.html

justinchuby (approved):
LGTM to unblock export

onnxscript/function_libs/torch_lib/ops/nn.py
titaiwangms: https://github.com/microsoft/onnxscript/issues/1802

Issue #1792, opened 2024-08-08
Allow optimizer to function when external data is not available


justinchuby:
Models with external data may not have their data available when being passed into the optimizer. We should still optimize what we can without raising an error.

justinchuby:
cc @gramalingam

gramalingam:
I took a quick look at the IR-based constant-folder/optimizer. I think it should work, as long as the `numpy()` method returns None or value.const_value is None. We can change it to use whatever appropriate methods the IR exposes to access such external tensors, allowing for possibility that the value may not be available.

On a different note: I just remembered that ir.Value.const_value is used for initializers that are graph inputs as well. This could break a few things. The optimizer etc. assume `.const_value` has a value only for constants, but graph inputs that are initializers should NOT be treated as constants.

gramalingam:
I do see an error message `Data of TensorProto ( tensor name: model/model/decoder/Constant_1_attr::value) should be stored in model_model_decoder_Constant_1_attr__value but it doesn't exist or is not accessible.` for ir-based optimization of BlenderbotForConditionalGeneration.

justinchuby:
> I took a quick look at the IR-based constant-folder/optimizer. I think it should work, as long as the `numpy()` method returns None or value.const_value is None. We can change it to use whatever appropriate methods the IR exposes to access such external tensors, allowing for possibility that the value may not be available.
> 
> On a different note: I just remembered that ir.Value.const_value is used for initializers that are graph inputs as well. This could break a few things. The optimizer etc. assume `.const_value` has a value only for constants, but graph inputs that are initializers should NOT be treated as constants.

Right. We shouldn’t assume only Constants have const_value. Any value can have constant values.

PR #1801, opened 2024-08-13
[IR] Implement save/load functions in IR and handle external data properly

justinchuby:
Implement efficient save/load and handle loading external data properly in the IR.

Before this change, when a ModelProto containing external data is converted to IR, the external tensor objects will load the data from a path relative to the working directory, not the ONNX file. This is because we do not store the onnx file path and thus have no way to look for the external data file.

With the change, a `base_dir` property is added to ExternalTensor that we can set, in a separate pass when the directory is available, so the object has full information to find the data file on disk. The base_dir is not serialized to the proto to maintain a relative path in the "location" field in TensorProto.

https://github.com/microsoft/onnxscript/issues/1701, https://github.com/microsoft/onnxscript/issues/1792

Example:

```
>>> m.graph.initializers["model.model.decoder.layers.2.encoder_attn.v_proj.weight"].const_value.display()
ExternalTensor<FLOAT,[512,512]>(path='model.onnx.data', 
name='model.model.decoder.layers.2.encoder_attn.v_proj.weight', offset=245864448, length=1048576, 
base_dir='/home/justinchu/dev/ONNXConverter/docker/dump_bash_bench/BlenderbotSmallForConditionalGeneration-torch
-onnx-detailed-cpu-')

Min: -0.08586505800485611, Max: 0.09103105217218399, NaN count: 0, Inf count: 0
Sparsity (abs<1e-06): 0.00
Histogram:
   11504 ┼
   10226 ┤                                  ╭───────╮
    8948 ┤                                ╭─╯       ╰─╮
    7670 ┤                              ╭─╯           ╰─╮
    6392 ┤                            ╭─╯               ╰─╮
    5113 ┤                          ╭─╯                   ╰─╮
    3835 ┤                        ╭─╯                       ╰─╮
    2557 ┤                     ╭──╯                           ╰─╮
    1279 ┤                ╭────╯                                ╰────╮
       1 ┼────────────────╯                                          ╰───────────────────
    -0.0859  -0.0682  -0.0505  -0.0306  -0.0129  0.0070  0.0225  0.0402  0.0557  0.0733  0.0910
```

### DIFF
--- a/onnxscript/ir/_io.py
+++ b/onnxscript/ir/_io.py
@@ -4,6 +4,8 @@
 
 from __future__ import annotations
 
+__all__ = ["load", "save"]
+
 import os
 from typing import Iterator
 
--- a/onnxscript/ir/_io.py
+++ b/onnxscript/ir/_io.py
@@ -5,10 +5,34 @@
 from __future__ import annotations
 
 import os
+from typing import Iterator
 
 import onnx
 
-from onnxscript.ir import _core, serde
+from onnxscript.ir import _core, _enums, _protocols, serde, traversal
+
+
+def _all_tensors(
+    graph: _core.Graph | _core.GraphView, include_constants: bool = False
+) -> Iterator[_protocols.TensorProtocol]:
+    """Iterate over all tensors in the graph."""
+
+    # Yield all tensors in initializers
+    for value in graph.initializers.values():
+        if value.const_value is not None:
+            yield value.const_value
+    if not include_constants:
+        return
+    # Look at constant attributes in nodes
+    for node in traversal.RecursiveGraphIterator(graph):
+        for attr in node.attributes.values():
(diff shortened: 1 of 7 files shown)


onnxscript/ir/_io.py
```diff
@@ -0,0 +1,75 @@
...
+
+import os
+from typing import Iterator
+
+import onnx
+
+from onnxscript.ir import _core, _enums, _protocols, serde, traversal
+
+
+def _all_tensors(
```
justinchuby, line 17: This will be refactored out as a utility later

titaiwangms (approved):

titaiwangms:
Is it only the path change that makes it more efficient?

justinchuby:
> Is it only the path change that makes it more efficient?

I added PR description

PR #1807, opened 2024-08-14
[torchlib] Update stack implementations

justinchuby:
Set aten_stack to be traced only
Remove hstack and vstack as the implementations are slow but they are decomposed in pytorch

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -8910,7 +8910,6 @@ def aten_vstack(tensors: Sequence[TTensor]) -> TTensor:
     raise NotImplementedError()
 
 
-
 @torch_op(
     (
         "aten::where.Scalar",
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3921,27 +3921,12 @@ def aten_hspmm(mat1: TensorType, mat2: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::hstack")
```
titaiwangms, line 3924: Maybe just delete the registration and keep the code? Put a TODO or NOTE saying that this is slow so we rely on PyTorch decomposition. Just in case we will need them, and this could save the efforts?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: ok will do

justinchuby: Done

PR #1811, opened 2024-08-15
Rewrite rules implementation for LLaMA-2/ LLaMA-3

lainey1570:
## Summary
This PR introduces the implementation of LLaMA 3 and LLaMA 2 rewrite rules for the MLP and LLaMAAttention layers in transformers. The rules are designed to work with transformer versions 4.39 to 4.42, and they handle the optimization and fusion operations.

## Key Changes
MLP RewriteRule:

A new rewrite rule for optimizing the LLaMA MLP layer (LlamaMLP) in transformer versions 4.39 to 4.42.
The optimization includes handling different input sizes (5 or 6) and performing matrix multiplication and activation operations to produce an optimized output.

GQA Llama RewriteRule:

Introduces a rewrite rule for the LLaMAAttention layer as well as the first attention (LlamaAttention) with support for specified number of inputs.
Two methods are implemented for handling 2D and 4D cache configurations during the Group Query Attention (GQA) process, enabling optimized matrix multiplication and attention operations.


initializer.py
```diff
@@ -0,0 +1,231 @@
+
```
justinchuby, line 1: I suggest excluding this file for now. We can focus on the rewriter rules for this PR.

lainey1570: Got it

justinchuby:
Congrats on your first PR! 🎉 For autofix-able lint errors, you can follow https://github.com/microsoft/onnxscript#coding-style to run the autofix.

PR #1813, opened 2024-08-15
[rewriter] Create the Dropout->Identity rules

### LINKED ISSUE #1776: [Pattern] Dropout(p=0) -> Identity
image

justinchuby:
Fix #1776


onnxscript/rewriter/no_op_test.py
```diff
@@ -177,6 +177,23 @@ def test_div_one_should_become_no_op_with_initializer(
         """
         )
 
+    @parameterized.parameterized.expand(
+        [
+            ("dropout zero", "ratio=0"),
+            ("dropout inference", "inference=1"),
+        ]
+    )
+    def test_dropout_zero_or_inference_no_op_with_initializer(self, _, attribute: str):
```
justinchuby, line 186: Why doesn't it work

gramalingam: Which one? Also: is the attribute `training_mode` or `inference`?

onnxscript/rewriter/no_op.py
```diff
@@ -32,6 +41,8 @@ def identity(op, x):
 add_0_rule = pattern.RewriteRule(add_0, identity)
 sub_0_rule = pattern.RewriteRule(sub_0, identity)
 div_by_1_rule = pattern.RewriteRule(div_by_1, identity)
+dropout_zero_rule = pattern.RewriteRule(dropout_zero, identity)
```
justinchuby, line 44: Specify don't care fields

onnxscript/rewriter/no_op_test.py
```diff
@@ -177,6 +177,23 @@ def test_div_one_should_become_no_op_with_initializer(
         """
         )
 
+    @parameterized.parameterized.expand(
+        [
+            ("dropout zero", "ratio=0"),
+            ("dropout inference", "inference=1"),
```
justinchuby, line 183:
```suggestion
            ("dropout inference", "training_mode=1"),
```

onnxscript/rewriter/no_op_test.py
justinchuby: Thanks! Corrected the attribute

justinchuby:
@gramalingam could you remind me which option will enable matching any unspecified attributes?

onnxscript/rewriter/no_op.py
```diff
@@ -23,6 +23,14 @@ def div_by_1(op, x):
     return x / 1
 
 
+def dropout_zero(op, x):
+    return op.Dropout(x, ratio=0.0, _allow_other_attributes=True)
```
justinchuby, line 27: @gramalingam will `ratio=0` match `ratio=0.0`?

gramalingam: Hmm, it's a good question what it *should* do. Currently, I think this test will be performed as [here](https://github.com/microsoft/onnxscript/blob/74ae4cc0efac8a33869a4a0dfd520dd42702c726/onnxscript/rewriter/pattern.py#L118) ... via `attr.value == 0`, so I guess it will succeed regardless of whether the attribute is an int or float? If we add a type-check in future, it may not.

gramalingam (approved):

onnxscript/rewriter/no_op.py
justinchuby: Sounds good thanks! I think it should match both like it currently does.

gramalingam: BTW, I checked and noticed that the the allow_other_attributes' default value is True, so it doesn't have to be specified. (Maybe the original issue was unrelated to this, it should have worked even without this flag.)

justinchuby: Thanks, updated

Issue #1819, opened 2024-08-19
[torchlib] Remove aten.linear


justinchuby:
Un-register https://github.com/microsoft/onnxscript/blob/acda045a65e9773dd2b2810f5b3fbf5a7dcbc4a0/onnxscript/function_libs/torch_lib/ops/nn.py#L825-L833 as it is decomposed by pytorch to Gemm and is required for model accuracy.

justinchuby:
cc @shubhambhokare1 @titaiwangms @fatcat-z

PR #1821, opened 2024-08-20
Unregister aten_linear

titaiwangms:
Fix https://github.com/microsoft/onnxscript/issues/1819

Instead, we rely on PyTorch decomposition to use aten::addmm (Gemm) for accuracy.


justinchuby (approved):
Thanks!

PR #1823, opened 2024-08-23
[IR] Fix external tensor path handling

justinchuby:
Previously the external tensors are deserialized with the base path and the relative location combined. This is in accurate because we lose information on which is the actual "location" that should be written to the proto.

With this change, I added a new `location` parameter/attribute to represent the ONNX proto relative "location", and update the `path` attribute to be computed by joining the base path with the relative location. As such users can always access the `tensor.path` attribute to get to the data file. Updated `serde` to make use of this attribute.

@shubhambhokare1 please also update the offload pass in #1796 to make use of the `path` attribute.

## BC breaking

The first parameter of the `ExternalTensor` initializer is renamed to `location`.

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -473,7 +473,8 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
     The :attr:`location` must be a relative path conforming to the ONNX
     specification. Given the correct :attr:`base_dir`, the :attr:`path` is computed
     to be the full path to the data file. Users should expect that the :attr:`path`
-    always leads to the correct file.
+    always leads to the correct file. At initialization, paths are not checked.
+    It is the user's responsibility to ensure the paths are valid and accessible.
 
     Attributes:
         location: The location of the data file. It is the path relative to the base directory.
@@ -530,7 +531,7 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
             metadata_props: The metadata properties.
             base_dir: The base directory for the external data. It is used to resolve relative paths.
         """
-        # Do not verify the location by default. This is because the location field
+        # NOTE: Do not verify the location by default. This is because the location field
         # in the tensor proto can be anything and we would like deserialization from
         # proto to IR to not fail.
         self._location = location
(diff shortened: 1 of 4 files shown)


onnxscript/ir/_core.py
```diff
@@ -470,11 +470,14 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
     To obtain an array, call :meth:`numpy`. To obtain the bytes,
     call :meth:`tobytes`.
 
-    The :attr:`path` can be a relative path or an absolute path.
-    Serializers should handle the path correctly to conform with the ONNX spec.
+    The :attr:`location` must be a relative directory conforming to the ONNX
```
justinchuby, line 473:
```suggestion
    The :attr:`location` must be a relative path conforming to the ONNX
```

onnxscript/ir/_core.py
```diff
@@ -470,11 +470,14 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
...
-    The :attr:`path` can be a relative path or an absolute path.
-    Serializers should handle the path correctly to conform with the ONNX spec.
+    The :attr:`location` must be a relative path conforming to the ONNX
+    specification. Given the correct :attr:`base_dir`, the :attr:`path` is computed
+    to be the full path to the data file. Users should expect that the :attr:`path`
+    always leads to the correct file.
 
     Attributes:
-        path: The path to the data file. This can be a relative path or an absolute path.
+        path: The path to the data file.
```
shubhambhokare1, line 479: This line no longer needed?

shubhambhokare1 (approved):

onnxscript/ir/_core.py
justinchuby: It is still relevant because path is one of the attributes of this class (it is computed)

onnxscript/ir/serde.py
```diff
@@ -767,10 +767,11 @@ def deserialize_tensor(
     if proto.data_location == onnx.TensorProto.EXTERNAL:
         external_info = onnx.external_data_helper.ExternalDataInfo(proto)
         return _core.ExternalTensor(
-            path=os.path.join(base_path, external_info.location),
+            location=external_info.location,
```
justinchuby, line 770:
```suggestion
            external_info.location,
```

PR #1824, opened 2024-08-23
[BE] Remove the external_tensor helper from onnxscript namespace

justinchuby:
It is not intended to be an onnxscript feature


justinchuby:
The translator is using this

PR #1825, opened 2024-08-23
Trace Op (aten::addmm) | feat(torchlib)

titaiwangms:
addmm is used a lot, and it's not traced yet. We trace it for better debugging and graph experience.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -238,7 +238,7 @@ def aten_addcmul(
     return op.Add(self, op.Mul(op.Mul(value, tensor1), tensor2))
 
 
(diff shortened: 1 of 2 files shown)


justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -238,7 +238,7 @@ def aten_addcmul(
     return op.Add(self, op.Mul(op.Mul(value, tensor1), tensor2))
 
 
-@torch_op("aten::addmm")
+@torch_op("aten::addmm", traceable=True)
```
justinchuby, line 241: For now, we can make it trace only and explicitly cast beta and alpha with

alpha=float(alpha), beta=float(beta)

Issue #1427, opened 2024-04-23
[IR] Implement topological sorting on `Graph`


justinchuby:
Implement the `Graph.sort()` method.

PR #1828, opened 2024-08-26
[IR] Implement topological sorting on Graph

yichen-li-ucla:
A Priority Queue + Breadth First Search (PQ+BFS) stable topological sort.

1. Parent node is be always be ahead of its child node.
2. Nodes should keep original position as possible.

This algorithm handles nested subgraphs.

Fixes https://github.com/microsoft/onnxscript/issues/1427

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1980,53 +1980,73 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def sort(self) -> None:
         """Perform a topological sort of this graph and all subgraphs in O(#nodes + #values) time.
 
-        This method sorts the nodes that each node appears AFTER all its parents, while maintaining the original order as much as possible.
+        This method sorts the nodes that each node appears AFTER all its predecessors,
+        while maintaining the original order as much as possible.
 
         Raises:
             ValueError: If the graph contains a cycle, making topological sorting impossible.
         """
-        # 1. Initialization
-        nodes = list(
-            onnxscript.ir.traversal.RecursiveGraphIterator(self)
-        )  # traverse all nodes
+        # Obtain all nodes from the graph and its subgraphs for sorting
+        nodes = list(onnxscript.ir.traversal.RecursiveGraphIterator(self))
+        # Store the sorted nodes of each subgraph
         sorted_nodes_by_graph: dict[Graph, list[Node]] = {
-            graph: []  # type: ignore[misc]
-            for graph in {node.graph for node in nodes}  # type: ignore[misc]
-        }  # store the sorted nodes of each subgraph
-        node_depth: dict[Node, int] = dict.fromkeys(nodes, 0)  # number of direct children
-        node_parents: dict[Node, list[Node]] = {node: [] for node in nodes}  # direct parents
-        node_index: dict[Node, int] = {
-            node: -i for i, node in enumerate(nodes)
-        }  # NEGATIVE original order
+            graph: [] for graph in {node.graph for node in nodes if node.graph is not None}
+        }
+        # TODO: Explain why we need to store direct predecessors and children and why
+        # we only need to store the direct ones
 
-        # a helper function to add a parent of a node, and increment the depth of the parent
-        def add_parent(child: Node, parent: Node | None) -> None:
-            if parent:
-                node_parents[child].append(parent)
-                node_depth[parent] += 1
+        # The depth of a node is defined as the number of direct children it has
+        node_depth: dict[Node, int] = dict.fromkeys(nodes, 0)
+        # Direct predecessors of a node
+        node_predecessors: dict[Node, list[Node]] = {node: [] for node in nodes}
+        # Store the negative index of the nodes because heapq is a min heap and we
+        # want to pop the node with largest index value first, effectively turning
+        # it to a max heap
+        neg_node_index: dict[Node, int] = {node: -i for i, node in enumerate(nodes)}
 
+        def add_predecessor(child: Node, predecessor: Node | None) -> None:
+            """Add a predecessor of a node, and increment the depth of the predecessor."""
+            if predecessor is None:
+                return
+            node_predecessors[child].append(predecessor)
+            node_depth[predecessor] += 1
+
+        # 1. Build the direct predecessors of each node and the depth of each node
+        # for sorting topolocally using Kahn's algorithm.
+        # Note that when a node contains graph attributes (aka. has subgraphs),
+        # we consider all nodes in the subgraphs *predecessors* of this node. This
+        # way we ensure the implicit dependencies of the subgraphs are captured
+        # as predecessors of the node.
         for node in nodes:
-            # All producers of input values are considered as direct parents.
+            # All producers of input values are considered as direct predecessors.
             for input_value in node.inputs:
                 if input_value is None:
                     continue
-                parent_node = input_value.producer()
-                add_parent(node, parent_node)
-            # All nodes in attribute graphs are considered as direct parents.
+                predecessor_node = input_value.producer()
+                add_predecessor(node, predecessor_node)
+            # All nodes in attribute graphs are considered as direct predecessors.
             for attr in node.attributes.values():
                 if not isinstance(attr, Attr):
                     continue
+                # A nice thing about this algorithm is that we only need to record
+                # direct predecessors. This continues to be true even with subgraphs.
+                # When a node in a subgraph (a) contains its own subgraphs (b), the
+                # node in subgraphs (b) are guranteed to appear before the node
+                # in (a).
                 if attr.type == _enums.AttributeType.GRAPH:
-                    for parent_node in attr.value:
-                        add_parent(node, parent_node)
+                    for predecessor_node in attr.value:
+                        add_predecessor(node, predecessor_node)
                 elif attr.type == _enums.AttributeType.GRAPHS:
                     for attribute_graph in attr.value:
-                        for parent_node in attribute_graph:
-                            add_parent(node, parent_node)
-        # 2. Priority Queue: Track nodes with zero direct children in a priority queue, using NEGATIVE original index for ordering.
-        # This ensures nodes appearing LATER in the original order are processed EARLIER. We get REVERSED topological order of each subgraph.
+                        for predecessor_node in attribute_graph:
+                            add_predecessor(node, predecessor_node)
+
+        # 2. Priority Queue: Track nodes with zero direct children in a priority queue,
+        # using NEGATIVE original index for ordering.
+        # This ensures nodes appearing LATER in the original order are processed EARLIER.
+        # We get REVERSED topological order of each subgraph.
         priority_queue: list[tuple[int, Node]] = [
-            (node_index[node], node) for node in nodes if node_depth[node] == 0
+            (neg_node_index[node], node) for node in nodes if node_depth[node] == 0
         ]
         heapq.heapify(priority_queue)
 
@@ -2035,21 +2055,25 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
         while priority_queue:
             # Pop the node with the most negative index and add it to the sorted nodes by subgraph.
             _, current_node = heapq.heappop(priority_queue)
-            sorted_nodes_by_graph[current_node.graph].append(current_node)  # type: ignore[index]
+            assert current_node.graph is not None
+            sorted_nodes_by_graph[current_node.graph].append(current_node)
             num_of_sorted_nodes += 1
-            # Decrement the depth of its parents. If any parent node has zero direct children, push it into the queue.
-            for parent_node in node_parents[current_node]:
-                node_depth[parent_node] -= 1
-                if node_depth[parent_node] == 0:
-                    heapq.heappush(priority_queue, (node_index[parent_node], parent_node))
+            # Decrement the depth of its predecessors. If any predecessor node has zero direct children, push it into the queue.
+            for predecessor_node in node_predecessors[current_node]:
+                node_depth[predecessor_node] -= 1
+                if node_depth[predecessor_node] == 0:
+                    heapq.heappush(
+                        priority_queue, (neg_node_index[predecessor_node], predecessor_node)
+                    )
 
         # 4. Cycle Check: Ensure all nodes are processed. If not, raise a ValueError indicating a cycle.
         if num_of_sorted_nodes != len(nodes):
             raise ValueError("Graph contains a cycle, topological sort is not possible.")
 
         # 5. Reverse: Reverse the sorted nodes of each subgraph to get the topological order.
-        for subgraph, sorted_nodes in sorted_nodes_by_graph.items():
-            subgraph.extend(reversed(sorted_nodes))
+        for graph, sorted_nodes in sorted_nodes_by_graph.items():
+            # The graph container ensures all the nodes are unique so we can safely extend
+            graph.extend(reversed(sorted_nodes))
 
     # End of mutation methods
 
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1980,8 +1980,7 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def sort(self) -> None:
         """Perform a topological sort of this graph and all subgraphs in O(#nodes + #values) time.
 
-        This method sorts the nodes that each node appears AFTER all its predecessors,
-        while maintaining the original order as much as possible.
+        This sort is stable. It preserves the original order as much as possible.
 
         Raises:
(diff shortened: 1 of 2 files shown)


yichen-li-ucla:
@microsoft-github-policy-service agree [company="{UCLA}"]

yichen-li-ucla:

justinchuby:
Thank you for your contribution!

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
         self._nodes.insert_before(node, new_nodes)
 
     def sort(self) -> None:
-        """Topologically sort the nodes in the graph."""
-        raise NotImplementedError("Not implemented yet")
-
+        # make sure the graph is a DAG
+        import heapq
```
justinchuby, line 1981: I would move the import to the top of the file. See https://gist.github.com/justinchuby/9085242a53158f2fd7ae7aa650e55ee3

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        stk = []
```
justinchuby, line 1982: I would spell words out and avoid abbreviations:

readability: Avoid abbreviations - they add cognitive load to readers.

> Do not use abbreviations that are ambiguous or unfamiliar to readers outside the project, and do not abbreviate by deleting letters within a word.

Good read: https://testing.googleblog.com/2024/03/whats-in-name.html

https://google.github.io/styleguide/pyguide.html#316-naming
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        nodes = [node for node in self._nodes if isinstance(node, Node)] # screen nodes
```
justinchuby, line 1983: nodes is always a list of Node so the insinstance check is unnecessary.

```suggestion
```

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        depth = {node: 0 for node in nodes} # topological depth
+        adjacency_list =  {node: [] for node in nodes} # parent list
+        ind = {node: i for i, node in enumerate(nodes)} # index (priority)
```
justinchuby, line 1986: Could you add a short comment to explain the algorithm and what these variables are used for?

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+            if depth[node] == 0:
+                heapq.heappush(pq, (ind[node], node))
+        while pq: # use PQ+BFS, archive the leaf with largest index, then check all its parents if they become leaves
+            _, node = heapq.heappop(pq)
+            stk.append(node)
+            for neighbor in adjacency_list[node]:
+                depth[neighbor] -= 1
+                if depth[neighbor] == 0:
+                    heapq.heappush(pq, (ind[neighbor], neighbor))
+        res = _linked_list.DoublyLinkedSet() # reconstruct doubly linked set
```
justinchuby, line 2004: Is the stack a sorted list of nodes? If so I would just call

```python
self.extend(reversed(stk))
```

The linked list implementation ensures uniqueness as a property.

This way we don't need to initialize a new linked list for it.

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        depth = {node: 0 for node in nodes} # topological depth
+        adjacency_list =  {node: [] for node in nodes} # parent list
+        ind = {node: i for i, node in enumerate(nodes)} # index (priority)
+        for node in nodes:
+            for input_value in node.inputs:
+                p = input_value.producer()
+                if isinstance(p, Node):
+                    adjacency_list[node].append(p)
+                    depth[p] += 1 # parent gain 1 depth with each child
+        pq = [] # priority queue
```
justinchuby, line 1993: I would add a comment on how the priority queue is used and why it is needed

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
         self._nodes.insert_before(node, new_nodes)
 
     def sort(self) -> None:
```
justinchuby, line 1979: I would add a comment on what guarantees this method provides and add a breif docstring. Thanks! https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

onnxscript/ir/_core.py
```diff
@@ -1977,9 +1977,86 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+    def sort_DFS(self) -> None:
+        # DFS
+        visited = set()
+        stack = []
+        nodes = [node for node in self._nodes if isinstance(node, Node)]
+        adjacency_list = {node: [] for node in nodes}
+        for node in nodes:
+            for input_value in node.inputs:
+                if isinstance(input_value.producer(), Node):
+                    adjacency_list[input_value.producer()].append(node)
```
justinchuby, line 2018: Redundant?

onnxscript/ir/_core.py
yichen-li-ucla: OK, but make sure all nodes are valid and belong to graph!

onnxscript/ir/_core.py
```diff
@@ -1533,7 +1534,7 @@ def _add_usage(self, use: Node, index: int) -> None:
 
         This is an internal method. It should only be called by the Node class.
         """
-        self._uses[(use, index)] = None
+        self._uses[use, index] = None
```
justinchuby, line 1537: Could you revert this change? Thanks! I will disable the check in https://github.com/microsoft/onnxscript/pull/1831 instead

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,54 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        """
+        # Perform a topological sort of the graph using Kahn's algorithm plus a priority queue.
+        # 1. **Initialization**: Create dictionaries to track each node's depth, parent nodes, and original index.
+        # 2. **Queue Initialization**: Track all nodes with zero child nodes in a priority queue, using their original index for ordering.
+        # 3. **Topological Sort**:
+        #   - While the priority queue is not empty:
+        #     - Pop the node with the largest original index and add it to the revesed sorted list.
+        #     - Decrement the depth of its parent nodes. If any parent node's has zero child node, push it into the queue.
+        # 4. **Cycle Check**: Ensure all nodes are processed. If not, raise a `ValueError` indicating a cycle.
+        # 5. **Finishing**: Reverse the sorted list to get the topological order.
```
justinchuby, line 1997:
```suggestion
        # 1. Initialization: Create dictionaries to track each node's depth, parent nodes, and original index.
        # 2. Queue Initialization: Track all nodes with zero direct child nodes in a priority queue, using their original index for ordering.
        # 3. Topological Sort:
        #   - While the priority queue is not empty:
        #     - Pop the node with the largest original index and add it to the revesed sorted list.
        #     - Decrement the depth of its parent nodes. If any parent node's has zero child node, push it into the queue.
        # 4. Cycle Check: Ensure all nodes are processed. If not, raise a `ValueError` indicating a cycle.
        # 5. Finishing: Reverse the sorted list to get the topological order.
```

nit: formatting

Also: It is more important to describe _why_ something is done over what is done. For example, in 

> Queue Initialization: Track all nodes with zero direct child nodes in a priority queue, using their original index for ordering.

I would update to explain why the priority queue is needed. Same for other items

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,54 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+            (node_index[node], node) for node in self._nodes if node_depth[node] == 0
+        ]
+        heapq.heapify(priority_queue)
+        reversed_sorted_nodes: list[Node] = []
+        # Topological sort
+        while priority_queue:
+            _, current_node = heapq.heappop(priority_queue)
+            reversed_sorted_nodes.append(current_node)
+            for parent_node in node_parents[current_node]:
+                node_depth[parent_node] -= 1
```
justinchuby, line 2022: I would note here, as a comment, that we only need to update the direct parents and so the complexity is small.

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,54 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+                if node_depth[parent_node] == 0:
+                    heapq.heappush(priority_queue, (node_index[parent_node], parent_node))
+        # Cycle check
+        if len(reversed_sorted_nodes) != len(self._nodes):
+            raise ValueError("Graph contains a cycle, topological sort is not possible.")
```
justinchuby, line 2027: Could you add a test case for this error? I would create an indirect cycle like {a -> b -> c -> a} and a self loop {a -> a}

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,54 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        # 3. **Topological Sort**:
+        #   - While the priority queue is not empty:
+        #     - Pop the node with the largest original index and add it to the revesed sorted list.
+        #     - Decrement the depth of its parent nodes. If any parent node's has zero child node, push it into the queue.
+        # 4. **Cycle Check**: Ensure all nodes are processed. If not, raise a `ValueError` indicating a cycle.
+        # 5. **Finishing**: Reverse the sorted list to get the topological order.
+        # Initialization
+        node_depth: dict[Node, int] = dict.fromkeys(self._nodes, 0)
+        node_parents: dict[Node, list[Node]] = {node: [] for node in self._nodes}
+        node_index: dict[Node, int] = {node: i for i, node in enumerate(self._nodes)}
```
justinchuby, line 2001: When a node like `If` has attributes that contains graphs, it may have an implicit dependency on previous nodes that is not seen in the inputs

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,54 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        ]
+        heapq.heapify(priority_queue)
+        reversed_sorted_nodes: list[Node] = []
+        # Topological sort
+        while priority_queue:
+            _, current_node = heapq.heappop(priority_queue)
+            reversed_sorted_nodes.append(current_node)
+            for parent_node in node_parents[current_node]:
+                node_depth[parent_node] -= 1
+                if node_depth[parent_node] == 0:
```
justinchuby, line 2023: This branch is not fully covered by tests. Could you create a test case with more nodes so `node_depth[parent_node]` is not 0 the first time we see it?

onnxscript/ir/_core_test.py
```diff
@@ -797,7 +797,105 @@ def test_remove_safe_removes_uses_of_removed_nodes(self):
...
+            (node3.outputs[0], node4.outputs[0]),
+            nodes=(node1, node2, node3, node4),
+            opset_imports={"": 1},
+        )
+        graph.sort()
+        sorted_nodes = tuple(graph)
+        self.assertLess(sorted_nodes.index(node1), sorted_nodes.index(node3))
+        self.assertLess(sorted_nodes.index(node2), sorted_nodes.index(node3))
+
+    def test_topological_sort_keep_original_1(self):
```
justinchuby, line 862: Prefer a more descriptive name than `1` and `2`. Consider the following for naming tests:

I would include the expected behavior as part of the test method name. To make what a test is testing clear: `test_<method or function name>_<expected behavior>_[when_<condition>]`.

e.g. `test_method_x_raises_error_when_dims_is_not_a_sequence`

justinchuby:
One this is done. We can also move it to passes and share it across Graph and Function

onnxscript/ir/_core.py
yichen-li-ucla: - [ ] Please do it for me... Not really good at wordings.

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,63 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+        # 3. Topological Sort:
+        #   - While the priority queue is not empty:
+        #     - Pop the node with the largest index and add it to the sorted nodes by graph.
+        #     - Decrement the depth of its parent nodes. If any parent node has zero direct children, push it into the queue.
+        # 4. Cycle Check: Ensure all nodes are processed. If not, raise a ValueError indicating a cycle.
+        # 5. Finishing: Reverse the sorted nodes in each subgraph to get the topological order.
+        
+        # Initialize
+        nodes = list(onnxscript.ir.traversal.RecursiveGraphIterator(self))
+        sorted_nodes_by_graph: dict[Graph, list[Node]] = {graph: [] for graph in {node.graph for node in nodes}}
```
justinchuby, line 2002: Just checking. Was the implicit dependencies from subgraphs handled? This will affect the topological ordering.

yichen-li-ucla:
@justinchuby

onnxscript/ir/_core_test.py
```diff
@@ -797,6 +797,154 @@ def test_remove_safe_removes_uses_of_removed_nodes(self):
...
+            (node3.outputs[0], node4.outputs[0]),
+            nodes=(node4, node3, node2, node1),
+            opset_imports={"": 1},
+        )
+        graph.sort()
+        sorted_nodes = tuple(graph)
+        expected_order = (node2, node4, node1, node3)
+        self.assertEqual(sorted_nodes, expected_order)
+
+    def test_topological_sort_shared_child(self):
```
justinchuby, line 846: nit: in all names, child -> successor / parent -> predecessor for clarity

onnxscript/ir/_core_test.py
```diff
@@ -797,6 +797,154 @@ def test_remove_safe_removes_uses_of_removed_nodes(self):
...
+        # edges=[(1,3),(2,3)]
```
justinchuby, line 847: I would create a test to verify that 

`(node3, node2, node1)` yeilds `(node2, node1, node3)`

and 

`(node3, node1, node2)` yeilds `(node1, node2, node3)`

to verify the stability claim.

Also test `(node1, node3, node2)` and `(node2, node3, node1)` 

as well as `(node1, node2, node3)` (the sorted case)

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,78 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
         self._nodes.insert_before(node, new_nodes)
 
     def sort(self) -> None:
-        """Topologically sort the nodes in the graph."""
-        raise NotImplementedError("Not implemented yet")
+        """Perform a topological sort of this graph and all subgraphs in O(#nodes + #values) time.
```
justinchuby, line 1981: Could you include the complexity guarantee in the docstring? Thanks!

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,102 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
...
+
+        This method sorts the nodes that each node appears AFTER all its predecessors,
+        while maintaining the original order as much as possible.
```
justinchuby, line 1984:
```suggestion
        This sort is stable. It preserves the original order as much as possible.
```

onnxscript/ir/_core.py
```diff
@@ -1977,8 +1978,101 @@ def insert_before(self, node: Node, new_nodes: Iterable[Node] | Node, /) -> None
         self._nodes.insert_before(node, new_nodes)
 
     def sort(self) -> None:
-        """Topologically sort the nodes in the graph."""
-        raise NotImplementedError("Not implemented yet")
+        """Perform a topological sort of this graph and all subgraphs in O(#nodes + #values) time.
+
+        This sort is stable. It preserves the original order as much as possible.
+
+        Raises:
```
justinchuby, line 1985:
```suggestion
        This sort is stable. It preserves the original order as much as possible.

        Referece: https://github.com/madelson/MedallionTopologicalSort#stable-sort

        Raises:
```

justinchuby (approved):
Thank you! This is impressive contribution

PR #1832, opened 2024-08-27
[API] Create stable APIs for PyTorch 2.5

### LINKED ISSUE #1827: [RFC] Create APIs for pytorch
PyTorch calls onnxscript apis and replies on stability of the apis because a single pytorch version can stay for a long time. I propose that we create apis dedicated to a specific version of PyTorch like below. As we evolve our logic we are more free to change user apis like `optimize` without being constrained by existing PyTorch usage. This also allows use to evolve the logic for these apis even after a PyTorch version is released.

```python
onnxscript._framework_apis.torch_2_5.optimize()
onnxscript._framework_apis.torch_2_5.convert_version()
onnxscript._framework_apis.torch_2_5.check_model()
onnxscript._framework_apis.torch_2_5.offload_external_data()
onnxscript._framework_apis.torch_2_5.get_torchlib_ops()
```

cc @gramalingam @xadupre @titaiwangms @shubhambhokare1

justinchuby:
Create stable APIs for PyTorch 2.5 so that it does not need to use any internal ONNX Script APIs. Created APIs are

```
"check_model",
"convert_version",
"get_torchlib_ops",
"optimize",
"save_model_with_external_data",
```

In pytorch, it is expected to write:

```python
import onnxscript._framework_apis.torch_2_5
```

Fixes #1827

### DIFF
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -67,6 +67,7 @@ def convert_version(model: ir.Model, target_version: int) -> ir.Model:
     # return ir.serde.deserialize_model(proto)
     # TODO(justinchuby): This function needs to be carefully implemented
     # to handle large models. For now, we just return the model.
+    del target_version  # Unused
     return model
 
 
@@ -123,7 +124,9 @@ def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike
 
 def get_torchlib_ops() -> list[_OnnxFunctionMeta]:
     # Trigger op registration
-    from onnxscript.function_libs.torch_lib import ops
+    from onnxscript.function_libs.torch_lib import (  # pylint: disable=import-outside-toplevel
+        ops,
+    )
 
     del ops  # Unused
 
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -8,9 +8,9 @@ __all__ = [
     "check_model",
     "convert_version",
     "get_torchlib_ops",
-    "OnnxFunctionMeta",
     "optimize",
     "save_model_with_external_data",
+    "OnnxFunctionMeta",
 ]
 
 import dataclasses
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -1,3 +1,5 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
 """Stable APIs for PyTorch 2.5."""
 
 from __future__ import annotations
@@ -17,9 +19,9 @@ import pathlib
 from typing import Callable
 
 import onnx
+
 from onnxscript import ir
 from onnxscript.ir import _external_data
-
 
 # Internal flag. Will go away.
 _TORCH_ONNX_OFFLOAD_EXTERNAL_DATA_WITH_IR = (
@@ -61,12 +63,9 @@ def convert_version(model: ir.Model, target_version: int) -> ir.Model:
 
 def check_model(model: ir.Model) -> None:
     """Check the model."""
-    pass
 
 
-def save_model_with_external_data(
-    model: ir.Model, model_path: str | os.PathLike
-) -> None:
+def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike) -> None:
     """Save the model with external data. The model is unchanged after saving."""
 
     if _TORCH_ONNX_OFFLOAD_EXTERNAL_DATA_WITH_IR:
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -102,7 +102,7 @@ def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike
 
 
 def get_torchlib_ops() -> list[OnnxFunctionMeta]:
-    from onnxscript.function_libs.torch_lib import (
+    from onnxscript.function_libs.torch_lib import (  # pylint: disable=import-outside-toplevel
         registration as torchlib_registration,
     )
 
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -73,12 +73,18 @@ def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike
     if _TORCH_ONNX_SAVE_EXTERNAL_DATA_WITH_IR:
         initializer_values = tuple(model.graph.initializers.values())
         tensors = [v.const_value for v in initializer_values]
+        for tensor in tensors:
+            if tensor is None:
+                raise ValueError(
+                    "The model contains uninitialized initializer values. "
+                    "Please make sure all initializer values are initialized."
+                )
         destination_path = pathlib.Path(model_path)
         base_dir = destination_path.parent
         data_path = f"{destination_path.name}.data"
 
         external_tensors = _external_data.convert_tensors_to_external(
-            tensors, base_dir, data_path
+            tensors, base_dir, data_path  # type: ignore[arg-type]
         )
 
         # Replace the initializer values with external tensors and save the model
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -18,6 +18,13 @@ from typing import Callable
 
 import onnx
 from onnxscript import ir
+from onnxscript.ir import _external_data
+
+
+# Internal flag. Will go away.
+_TORCH_ONNX_OFFLOAD_EXTERNAL_DATA_WITH_IR = (
+    os.getenv("TORCH_ONNX_OFFLOAD_EXTERNAL_DATA_WITH_IR") == "1"
+)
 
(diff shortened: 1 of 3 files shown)


onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,147 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class OnnxFunctionMeta:
+    """A wrapper of onnx-script function with additional metadata.
+
+    qualified_name: The qualified name of the aten operator.
+    function: The onnx-script function.
+    domain: The domain of the function.
+    name: The name of the function.
+    is_complex: Whether the function is a complex function.
```
gramalingam, line 40: I guess this is a pre-existing concept ... does it mean that all inputs/outputs of the aten function are complex? Or, that some input/output is complex?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def check_model(model: ir.Model) -> None:
+    """Check the model."""
+
+    del model  # Unused yet
+
+
+def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike) -> None:
+    """Save the model with external data. The model is unchanged after saving."""
```
gramalingam, line 71: What about tensors stored as attribute values (eg., in Constant ops)? Would be helpful to make them external too. As well as the size-based filters to store only tensors larger than small constants/shapes externally?

justinchuby: In practice the constant ops do not carry large buffers or weights (like they used to in the old exporter) in PyTorch, so I think it is easier and more robust to just store the initializers. I think if a user wants to customize the behavior they should load the onnx model back using our tools to have finer control. Right now the options are direct - store the weights with the model, or store the weights externally. What do you think?

onnxscript/_framework_apis/torch_2_5.py
justinchuby: It can mean either, but for our purposes we know what to expect for a particular op. Usually they are both complex because PyTorch will promote the types to be the same for use.

onnxscript/_framework_apis/torch_2_5.py
gramalingam: Easier, but not necessarily more robust? IIUC, you are suggesting using "initalizers" as a proxy/approximation for large tensors, and "constant ops" as a proxy/approximation for small tensors ... but that may not always be true? If small initializers (eg., shapes) get externalized, shape-inference may be impacted, because it works without loading external tensors. On the other hand, optimizers/rewriters might produce Constant ops instead of initializers because it is easier? (Also, because we don't have initializers inside functions, and only constant ops, but this may become less of a concern if and when we get to inlining all functions.)

gramalingam: I am fine with adjusting the implementation later on, as long as we are fine with the API here. I think we may be better off making all tensors (initializers and node-attributes) of size above a cutoff limit external, leaving the rest internal. But I am fine if the cutoff limit is hard-coded in this logic for now ... only if we want to expose this to users, does it get a bit more complicated.

justinchuby: https://github.com/microsoft/onnxscript/issues/1835

gramalingam (approved):

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
...
+    function: Callable
+    domain: str
+    name: str
+    is_complex: bool = False
+
+
+def optimize(model: ir.Model) -> ir.Model:
+    """Optimize the model."""
+
+    # TODO(justinchuby): Use the optimizer
```
shubhambhokare1, line 53: Just for clarification, what is the purpose of this API. Will it consist of something other than calling optimize.optimize()

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
...
+
+from __future__ import annotations
+
+__all__ = [
+    "check_model",
+    "convert_version",
+    "get_torchlib_ops",
+    "optimize",
+    "save_model_with_external_data",
+    "OnnxFunctionMeta",
```
titaiwangms, line 13: Do we need to expose it?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Stable APIs for PyTorch 2.5."""
```
titaiwangms, line 3: So if we move forward to 2.6, we generate a new file like this one?

onnxscript/_framework_apis/torch_2_5.py
titaiwangms, line 53: Do we plan to release it within 2.5?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def optimize(model: ir.Model) -> ir.Model:
+    """Optimize the model."""
+
+    # TODO(justinchuby): Use the optimizer
+    return model
+
+
+def convert_version(model: ir.Model, target_version: int) -> ir.Model:
```
titaiwangms, line 57: I didn't know we have this already! Is this stable? Should we return informative message once it fails?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
...
+    """Convert the model to the specified ONNX opset version."""
+    proto = ir.serde.serialize_model(model)
+    proto = onnx.version_converter.convert_version(proto, target_version)
+    return ir.serde.deserialize_model(proto)
+
+
+def check_model(model: ir.Model) -> None:
```
titaiwangms, line 64: Do we want a different validation than onnx.checker?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -0,0 +1,148 @@
...
+    """Check the model."""
+
+    del model  # Unused yet
+
+
+def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike) -> None:
+    """Save the model with external data. The model is unchanged after saving."""
```
titaiwangms, line 71: For my education, when we do save model with external data, do we still expect the data in the model? This says the saved model still keeps the data, while the data is also saved externally.

titaiwangms (approved):

onnxscript/_framework_apis/torch_2_5.py
justinchuby: Yes. I plan to release the APIs for pt2.5. We can then take our time implementing the apis

onnxscript/_framework_apis/torch_2_5.py
justinchuby: We don't really have an implementation for IR yet. The idea for this api is that we can fill in the implementation afterward.

onnxscript/_framework_apis/torch_2_5.py
justinchuby: Yes. This way we can call save() multiple times.

onnxscript/_framework_apis/torch_2_5.py
justinchuby: Yes in the potential future.

onnxscript/_framework_apis/torch_2_5.py
justinchuby: Likely, yes. It may simply be aliasing the existing functions

onnxscript/_framework_apis/torch_2_5.py
justinchuby: We don't have to. I will make private

onnxscript/_framework_apis/torch_2_5.py
titaiwangms: Should we put in onnx.checker for now?

onnxscript/_framework_apis/torch_2_5.py
titaiwangms: I thought we only save model with external data when the model is too large to be in one file?

onnxscript/_framework_apis/torch_2_5.py
justinchuby: onnx.checker will fail when the model is too large. It may segfault some times. We need to be more thoughtful using it. For now I am just going to do nothing and declare success. I think that's appropriate for this PR.

onnxscript/_framework_apis/torch_2_5.py
justinchuby: This is controlled by the ONNXProgram.save() api. It will decide when to call this function. When the model is saved, the ir.Model in the onnx program should not be altered, and hence we ensure this in this function.

titaiwangms: But if the model is too large, does it still save the very same size of model to one file? Would that hit protobuf limitation?

justinchuby: When the model is too large. It will save to external tensors.

PR #1834, opened 2024-08-29
[torchlib] [ci] Fix failing external data tests for Windows

shubhambhokare1:
Fix failing external data tests for Windows due to file permission errors

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -671,7 +671,7 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
         length = self._length or self.nbytes
         return self.raw[offset : offset + length]
 
-    def cleanup(self) -> None:
+    def release(self) -> None:
         """Delete all references to the memory buffer and close the memory-mapped file."""
         self._array = None
         if self.raw is not None:
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -671,18 +671,11 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
         length = self._length or self.nbytes
         return self.raw[offset : offset + length]
 
-    def copy(self) -> np.ndarray:
-        """Creates a copy of the numpy array representing the tensor.
-
-        This will also delete all references to the memory buffer and close the memory-mapped file.
-        """
-        _copy = self.numpy().copy()
-        # Delete references
+    def cleanup(self) -> None:
+        """Delete all references to the memory buffer and close the memory-mapped file."""
         self._array = None
-        # Close mmap file
         assert self.raw is not None
         self.raw.close()
-        return _copy
 
     @property
     def metadata_props(self) -> dict[str, str]:
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -680,6 +680,7 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
         # Delete references
         self._array = None
         # Close mmap file
+        assert self.raw is not None
         self.raw.close()
         return _copy
 
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -674,8 +674,9 @@ class ExternalTensor(TensorBase, _protocols.TensorProtocol):  # pylint: disable=
     def cleanup(self) -> None:
         """Delete all references to the memory buffer and close the memory-mapped file."""
         self._array = None
-        assert self.raw is not None
-        self.raw.close()
+        if self.raw is not None:
+            self.raw.close()
+            self.raw = None
 
     @property
     def metadata_props(self) -> dict[str, str]:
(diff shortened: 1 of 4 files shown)


onnxscript/ir/_external_data.py
```diff
@@ -170,6 +170,7 @@ def _save_external_data(
             if current_offset > file_size:
                 data_file.write(b"\0" * (current_offset - file_size))
             data_file.write(raw_data)
+        data_file.close()
```
justinchuby, line 173: The open context manager should handle closing the file

onnxscript/ir/_external_data.py
```diff
@@ -270,6 +270,12 @@ def convert_tensors_to_external(
         external_tensors[i]
         for i in sorted(range(len(external_tensors)), key=lambda i: sorted_indices[i])
     ]
+
+    # Clean-up temporary file if it is created
+    tmp_path = os.path.join(base_path, "tmp", relative_path)
+    if os.path.exists(tmp_path):
+        os.remove(tmp_path)
+
```
justinchuby, line 278: Consider set a flag at 243 and do this only when the flag is set? Otherwise if there already exists a file named tep*** this will remove it unexpectedly.

gramalingam: Agree. Also, can't we use `tempfile`?

shubhambhokare1: Had avoided using tempfile to avoid copy data between files, so a renaming should be sufficient here?

onnxscript/ir/_core.py
```diff
@@ -670,6 +670,12 @@ def tobytes(self) -> bytes:
         length = self._length or self.nbytes
         return self.raw[offset : offset + length]
 
+    def cleanup(self) -> None:
+        """Delete all references to the memory buffer and close the memory-mapped file."""
+        self._array = None
+        assert self.raw is not None
+        self.raw.close()
```
justinchuby, line 677:
```suggestion
    def release(self) -> None:
        """Delete all references to the memory buffer and close the memory-mapped file."""
        self._array = None
        if self.raw is not None:
            self.raw.close()
            self.raw = None
```

onnxscript/ir/_external_data_test.py
```diff
@@ -115,7 +116,10 @@ class OffloadExternalTensorTest(unittest.TestCase):
 
     def setUp(self):
         # File paths
-        self.temp_dir = tempfile.TemporaryDirectory()  # pylint: disable=consider-using-with
+        if sys.version_info == 3 and sys.version_info >= 10:
```
justinchuby, line 119: Is this right? `sys.version_info == 3` and `sys.version_info >= 10` cannot be both true?

onnxscript/ir/_core.py
justinchuby: Could you add a test to verify that calling numpy() etc. on a released tensor will still work (e.g. we can release and load again at will)? Thanks!

onnxscript/ir/_core.py
```diff
@@ -673,8 +673,9 @@ def tobytes(self) -> bytes:
     def cleanup(self) -> None:
```
justinchuby, line 673: Consider calling it release()

justinchuby:
Could you add a test to verify that calling numpy() etc. on a released tensor will still work (e.g. we can release and load again at will)? Thanks!

onnxscript/ir/_core_test.py
```diff
@@ -244,6 +244,25 @@ def test_initialize(self):
         # Ensure repeated reads are consistent
         np.testing.assert_equal(tensor, self.data)
 
+    def test_release(self):
```
justinchuby, line 247:
```suggestion
    def test_release_does_not_invalidate_tensor(self):
```

onnxscript/ir/_core_test.py
```diff
@@ -244,6 +244,25 @@ def test_initialize(self):
...
+            base_dir=self.base_path,
+            name="input",
+            shape=_core.Shape(external_tensor.dims),
+        )
+        self.assertEqual(tensor.dtype, ir.DataType.FLOAT)
+        self.assertEqual(tensor.tobytes(), self.data.tobytes())
+        # Release tensor
+        tensor.release()
+        self.assertEqual(tensor.raw, None)
+        self.assertEqual(tensor.tobytes(), self.data.tobytes())
```
justinchuby, line 264:
```suggestion
        # Tensor can be re-loaded after release
        self.assertEqual(tensor.tobytes(), self.data.tobytes())
```

justinchuby (approved):
Thank you!

justinchuby:
Windows crashes: Windows fatal exception: access violation

https://github.com/microsoft/onnxscript/actions/runs/10693258903/job/29643192895?pr=1834

shubhambhokare1:
> Windows crashes: Windows fatal exception: access violation
> 
> https://github.com/microsoft/onnxscript/actions/runs/10693258903/job/29643192895?pr=1834

Failures seem to be unrelated.

justinchuby (approved):

PR #1836, opened 2024-08-30
[torchlib] Trace some activation functions

justinchuby:
Trace commonly used activation functions and fix elu

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -398,8 +398,8 @@ def aten_elu(
 ) -> TFloat:
     """elu(Tensor self, Scalar alpha=1, Scalar scale=1, Scalar input_scale=1) -> Tensor"""
 
-    input_scale = op.CastLike(self, input_scale)
-    scale = op.CastLike(self, scale)
+    input_scale = op.CastLike(input_scale, self)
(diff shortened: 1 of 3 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -398,9 +398,8 @@ def aten_elu(
 ) -> TFloat:
     """elu(Tensor self, Scalar alpha=1, Scalar scale=1, Scalar input_scale=1) -> Tensor"""
 
-    # del scale
-    # del input_scale
-    return op.Elu(self, alpha=alpha)
+    self = self * input_scale
```
titaiwangms, line 401: We don't need to specify op for this operation?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
gramalingam: Also: was this an omission-bug? (Just that the PR description didn't say this was also fixing errors.)

gramalingam (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: We don't but updated to trace. thanks.

PR #1838, opened 2024-08-30
Introduce OpSignature to IR

justinchuby:
Introduce OpSignature accessible from the `.op_signature` property of all OpLike objects (traced function, onnx function and op). The OpSignature class leverages the IR to represent the signature of an operator, preserving ordering of all inputs and provides easy to work with type representations.

The PR also deprecates the ParamSchema class and properties.

Fixes https://github.com/microsoft/onnxscript/issues/1697

The next PR will replace param_schemas usage.


onnxscript/ir/_schemas.py
```diff
@@ -0,0 +1,544 @@
...
+def _get_type_from_str(
+    type_str: str,
+) -> ir.TensorType | ir.SequenceType | ir.OptionalType:
+    """Converter a type_str from ONNX Opschema to ir.TypeProtocol.
+
+    A type str has the form of "tensor(float)" or composite type like "seq(tensor(float))".
+    """
+
+    # TODO: Upstream this to IR
+
```
justinchuby, line 156:
```suggestion
    """
```

onnxscript/ir/_schemas.py
```diff
@@ -0,0 +1,548 @@
+# Copyright (c) Microsoft Corporation.
```
titaiwangms, line 1: Should we put a comment saying this should align to torch _schema.py until we move it to here?

titaiwangms: Actually, we only have the test for this file, which means that torch's _schema.py should be following this one.

titaiwangms (approved):

justinchuby:
When the signature property is defined, PyTorch onnx exporter will not generate a new signature, and instead use what is here. However, since it is doing isinstance checks on the signature the exporter itself defines, we get isinstance errors there.

PR #1853, opened 2024-09-04
Revert the test of aten::_scaled_dot_product_efficient_attention

titaiwangms:
Revert it back to the original test. We can revisit and secure the test when we have spare time.


tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -1339,7 +1338,7 @@ def sample_inputs__scaled_dot_product_efficient_attention(
                 make(shape_q),
                 make(shape_kv),
                 make(shape_kv),
-                attn_bias=make(shape_attn_bias),
+                attn_bias=None,
```
justinchuby, line 1341: Add a comment?

justinchuby (approved):

PR #1856, opened 2024-09-06
[torchlib] Fix reduce functions handing of axes for cuda kernels

shubhambhokare1:
In the ORT Cuda kernel for all reduction ops there consists of a parameter `allow_mutli_axes` which is set as true for all ops excluding `argmax` and `argmin`. When this parameter is set, the following check is mandatory for `input_count=2`,

```
ORT_ENFORCE(axes_tensor != nullptr, "Axes input is null");
```
Although technically input_count should be `=1` in most cases, input_count is `=2`, making the axes a required input and the check is triggered. I suspect could be something to do with handling of optional inputs for cuda kernela but will need to be investigated further. As a workaround (passing an empty tensor for axes) this should suffice.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -8798,7 +8798,7 @@ def _aten_var_onnx(self: TReal, correction: float, keepdim: bool = False) -> TRe
     mean = op.ReduceMean(self, empty_axes, keepdims=keepdim)
     sub_mean = op.Sub(self, mean)
     sqr_mean = op.Mul(sub_mean, sub_mean)
-    var = op.ReduceMean(sqr_mean, empty, keepdims=keepdim)
+    var = op.ReduceMean(sqr_mean, empty_axes, keepdims=keepdim)
     # Adjust var according to correction value
     if correction > 0.0:
         self_shape = op.Shape(self)
@@ -8884,7 +8884,7 @@ def _aten_var_mean_onnx(
     mean = op.ReduceMean(self, empty_axes, keepdims=keepdim)
     sub_mean = op.Sub(self, mean)
     sqr_mean = op.Mul(sub_mean, sub_mean)
-    var = op.ReduceMean(sqr_mean, empty, keepdims=keepdim)
+    var = op.ReduceMean(sqr_mean, empty_axes, keepdims=keepdim)
     # Adjust var according to correction value
     if correction > 0.0:
         self_shape = op.Shape(self)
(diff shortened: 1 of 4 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -345,7 +345,8 @@ def aten_all(self: TTensor) -> BOOL:
     else:
         self_bool = op.Cast(self, to=BOOL.dtype)
         self_int = op.Cast(self_bool, to=INT64.dtype)
-        all_true = op.ReduceMin(self_int, keepdims=False)
+        empty_axes = op.Shape(self, start=0, end=0)
+        all_true = op.ReduceMin(self_int, empty_axes, keepdims=False)
```
justinchuby, line 349: is empty_axes just `[]`?

shubhambhokare1: That what I had initially, but this error is raised,
` onnxscript.converter.TranslationError: ERROR: Cannot convert an empty list to tensor
E   at: Function 'elseGraph_5', line 12
E           any_true = op.ReduceMax(self_int, axes=[], keepdims=False)`

gramalingam: We can't use `[]` because the translator cannot figure out the intended type of the tensor.

Nit: I wonder if `op.Constant(value_ints = [])` (IF IT WORKS) will be more natural then using Shape op.

gramalingam: On an unrelated note: why is it cast to `INT64` in line 347? Wouldn't `INT32` or `INT16` work just as well, while being more efficient?

shubhambhokare1: > op.Constant(value_ints = []

Yup, this will work, and better alternative to the Shape op.

justinchuby:
It could be that we in the opset definition passes None as input to the op (as a Python default value), this None is then translated to `""` in the graph.

gramalingam (approved):
Minor nit (unrelated to this change): some of the logic for boolean-reduction (any/all) looks like it is repeated in several places. Having a utility function might consolidate the duplication.

xadupre:
I'm not sure I fully understood this issue. ``ReduceMax(x)`` is ok for onnxruntime so this fix should not be needed. Can we have the original error? What does torch_script do?

xadupre:
I checked with torch_script, ReduceMax((bool)) is replaced by ReduceSum((int)x) > 0. onnxruntime has a bug. It should be fixed there.

xadupre:
I checked that ``ReduceMax("X")`` succeeds and ``ReduceMax("X","")`` fails. My guess is onnxscript is automatically adding an empty input for the axes.

shubhambhokare1:
> I checked that `ReduceMax("X")` succeeds and `ReduceMax("X","")` fails. My guess is onnxscript is automatically adding an empty input for the axes.

Yep, I think the core issue lies here. And resolving this should also solve the Squeeze issue.

justinchuby:
Now that we have fixed the exporter, I think we can safely close this PR. Please reopen if I am wrong.

PR #1859, opened 2024-09-09
Fix #1858 for Python 3.8 support

### LINKED ISSUE #1858: Latest version of onnxscript does not work for Python 3.8 `TypeError: 'type' object is not subscriptable`
```python
onnxscript/optimizer/_inliner.py:22:
    CallStack = list[CallSiteId]
```

`list[CallSiteId]` type annotation is only supported for Python 3.9+.

A fix would be to replace this with `typing.List[CallSiteId]`

kevalmorabia97:
Details in https://github.com/microsoft/onnxscript/issues/1858

### DIFF
--- a/onnxscript/optimizer/_inliner.py
+++ b/onnxscript/optimizer/_inliner.py
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 from collections import defaultdict
-from typing import Iterable, Sequence, Tuple
+from typing import Iterable, List, Sequence, Tuple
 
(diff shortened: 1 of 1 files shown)


xadupre (approved):

kevalmorabia97:
@xadupre thanks for merging this. Can we have a new wheel release as well?

xadupre:
You can look at https://pypi.org/project/onnxscript/#history. You can use development releases.

Issue #1749, opened 2024-07-22
[torchlib] Improve/reimplement index_put


justinchuby:
1. Handle None as inputs
2. Turn the function into `trace_only`
3. Fix failing CI tests

Ref:

- https://github.com/pytorch/pytorch/issues/131349
- https://github.com/pytorch/pytorch/issues/119363

Thakor-Yashpal:
Can I work on this issue?

justinchuby:
First fix in https://github.com/microsoft/onnxscript/pull/2061. Follow ups need to handle more dynamic shapes.

justinchuby:
https://github.com/microsoft/onnxscript/pull/2712

PR #1864, opened 2024-09-11
Add unit test to investigate torch issues (scaled_dot_product_attention, index_put)

xadupre:
See https://github.com/pytorch/pytorch/issues/135615, https://github.com/pytorch/pytorch/issues/135233.


justinchuby:
Do you have plans to merge or is this for investigation only? Marking as draft for now

justinchuby:
We do need to rewrite our index put implementation. #1749

xadupre:
I don't have time to implement a fix this week but anybody doing it should check the with unit tests I made and decide whether or not they should be kept.

PR #1865, opened 2024-09-12
Fixes for IR optimizer

gramalingam:
A few fixes needed to make the IR based optimizer work for the HF benchmark.

Removed the changes relating to SymbolicTensor hash issue, and what's remaining are pure bug fixes.


onnxscript/ir/_core.py
```diff
@@ -421,7 +421,12 @@ def numpy(self) -> np.ndarray:
         if isinstance(self._raw, np.ndarray):
             return self._raw
         # We do not cache the value to save memory
-        return self.__array__()
+        try:
+            result = self.__array__()
+        except Exception:
+            self._raw = self._raw.detach()
```
justinchuby, line 427: Can be reverted after today's nightly

onnxscript/ir/_core.py
```diff
@@ -1676,7 +1681,7 @@ def Input(
 
 
 def _check_node_safe_to_remove(
-    node: Node, to_remove: AbstractSet[Node], graph_outputs: AbstractSet[Value]
+    node: Node, to_remove: AbstractSet[Node], graph_outputs: AbstractSet[int]
```
justinchuby, line 1684: I still think we should assume Values as hashable. This is required by the protocol (should make it explicit in text. cannot inherit Hashable because that requires explicit definition of `__hash__`.

gramalingam: Yes, it would be convenient to use dictionaries with IR components as keys (or sets of them) ... if we can assume that, I will undo all related changes.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -77,15 +77,15 @@ class Replacement:
 
 class OptimizerState:
     def __init__(self):
-        self._sym_value_map: dict[ir.Value, Any] = {}
+        pass
```
gramalingam, line 80: Even the changes here can be undone if ir.Value is usable as dictionary key. That is preferable to use the `.meta` option ... because it makes it easy to discard temporary dictionaries once they are no longer needed ... info stored in `.meta` will persist unless we have some mechanism to explicitly free it.

onnxscript/ir/_core.py
justinchuby: Yes. Please assume that. I will fix any violations if you see any

justinchuby (approved):

onnxscript/optimizer/_inliner.py
```diff
@@ -142,10 +147,11 @@ def clone_graph(self, graph: ir.Graph) -> ir.Graph:
         input_values = [self.clone_value(v) for v in graph.inputs]
         nodes = [self.clone_node(node) for node in graph]
         initializers = [self.clone_value(init) for init in graph.initializers.values()]
+        output_values = [self.clone_value(v) for v in graph.outputs]
```
justinchuby, line 150: So this will simply retrieve the values from the value pool (instead of actual cloning). Is that the correct understanding? If so a comment will be helpful I think.

onnxscript/optimizer/_inliner.py
```diff
@@ -142,10 +147,11 @@ def clone_graph(self, graph: ir.Graph) -> ir.Graph:
...
 
         return ir.Graph(
             input_values,  # type: ignore
-            graph.outputs,
+            output_values,  # type: ignore
```
justinchuby, line 154: Curious why # type: ignore is needed?

onnxscript/optimizer/_inliner.py
gramalingam: Yes. Real cloning happens at places which introduce a symbol (which are graph inputs, and outputs of nodes). At places which just reference (consume) a symbol, we just retrieve the value.

onnxscript/optimizer/_inliner.py
gramalingam: Because of a limitation in the type annotation. Consider a function `f(x, y)`. We support/allow the actual-value of `y` to be omitted (an optional input) in a call to f. When we inline such a call, inside the body of the function, if we have a call to `someop(x, y)`, we will replace `y` by `None` (essentially, omitting the presumably optional input of someop). So, the dictionary used to do the substitution maps names to `ir.Value | None`. However, in some contexts, (such as the output value of the graph), we don't expect the value to be `None` (as per ONNX requirement).

I suppose it may be a good idea to actually check for this condition, and throw an appropriate exception. In a correct model, it shouldn't happen.

PR #1867, opened 2024-09-16
[torchlib] Unregister stft, var, var_mean, std, std_mean

justinchuby:
Following https://github.com/pytorch/pytorch/pull/136153, we remove stft, var, var_mean, std, std_mean ops. They were never used even before because the ops were always decomposed.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8054,7 +8046,8 @@ def _aten_stft_onnx(
     return result
 
 
-@torch_op("aten::stft", trace_only=True)
+# aten::stft is decomposed. The implementation is not complete and is left for
```
xadupre, line 8049: I think we can remove if it is not complete.

PR #1872, opened 2024-09-19
Bug fixes in split-sequence optimization

gramalingam:
A couple of bugs in the optimization for split-sequence:

* Handle the case where there is only one split-value (as the op-builder returns a single IR value instead of a list of IR values in this case).
* Use 1D constant [axis] instead of scalar axis in Squeeze op.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -448,7 +448,7 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
         return None
     if keepdims == 0:
         # squeeze the split dimension if keepdims is 0
-        axis_val = op.Constant(value_int=axis, _outputs=[f"{output.name}_axis"])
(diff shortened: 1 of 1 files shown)


justinchuby (approved):

onnxscript/optimizer/_constant_folding.py
```diff
@@ -439,12 +439,16 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     else:
         return None
 
+    # If Split returns a single value, we need to wrap it into a list.
+    if isinstance(split_values, ir.Value):
+        split_values = [split_values]
```
justinchuby, line 444: Looks like this line was not covered

gramalingam: Will add test case later. It's triggered by a model in the benchmark. Will merge it to enable benchmarking. It's a weird example, though, and worth investigating more ... because a split of a tensor into one value is a no-op or identity. A candidate for optimization, either via rewrite or avoid generating it in the first-place.

PR #1874, opened 2024-09-23
[torchlib] Implement upsample_nearest{nd}.vec

justinchuby:


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2645,6 +2645,18 @@ def aten_upsample_nearest3d(
...
+@torch_op("aten::upsample_nearest3d.vec", trace_only=True)
+def aten_upsample_nearest3d_vec(
+    input: TReal,
+    output_size: INT64,
+    scale_factors: Optional[Sequence[float]] = None,
+) -> TReal:
+    """upsample_nearest3d.vec(Tensor input, SymInt[]? output_size, float[]? scale_factors) -> Tensor"""
+
+    # TODO: Use scale_factors?
+    return _aten_upsample_output_size(self, size, "nearest", "asymmetric")
```
justinchuby, line 2657:
```suggestion
    return _aten_upsample_output_size(input, output_size, "nearest", "asymmetric")
```

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2645,6 +2645,18 @@ def aten_upsample_nearest3d(
     return _aten_upsample_output_size(self, size, "nearest", "asymmetric")
 
 
+@torch_op("aten::upsample_nearest3d.vec", trace_only=True)
+def aten_upsample_nearest3d_vec(
```
gramalingam, line 2649: Is this meant to be same as function above, differing only in how parameter scales is specified?

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2645,6 +2645,18 @@ def aten_upsample_nearest3d(
...
+    input: TReal,
+    output_size: INT64,
```
gramalingam, line 2651: nit: should we use same name for parameter in previous and this function?

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2645,6 +2645,18 @@ def aten_upsample_nearest3d(
...
+    scale_factors: Optional[Sequence[float]] = None,
+) -> TReal:
+    """upsample_nearest3d.vec(Tensor input, SymInt[]? output_size, float[]? scale_factors) -> Tensor"""
+
+    # TODO: Use scale_factors?
```
gramalingam, line 2656: Does this TODO also apply to previous function?

gramalingam (approved):

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: I am following the function signature: https://github.com/pytorch/pytorch/blob/52c917b0bab712ef103d8e0f8eba687819e15070/aten/src/ATen/native/native_functions.yaml#L12570

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: I think the previous function will ignore the params, based on Xiaowu's tests IIRC

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2645,6 +2645,18 @@ def aten_upsample_nearest3d(
...
+    return _aten_upsample_output_size(input, output_size, "nearest", "asymmetric")
```
gramalingam, line 2657: For AHNet, we get a failure because output_size is missing here: is it supposed to be an optional input? The calling node in exported model looks as below:
```
%"upsample_nearest3d_7"<FLOAT,[1,2,128,128,128]> ⬅️ pkg.onnxscript.torch_lib::_aten_upsample_output_size(%"conv3d_121") {mode=nearest, 
coordinate_transformation_mode=asymmetric}
```

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2602,18 +2595,21 @@ def aten_upsample_nearest1d_backward(
 @torch_op("aten::upsample_nearest2d", trace_only=True)
 def aten_upsample_nearest2d(
     self: TReal,
-    size: INT64,
+    output_size: INT64,
```
gramalingam, line 2598: Is this supposed to be optional?

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2602,18 +2595,21 @@ def aten_upsample_nearest1d_backward(
...
-    return _aten_upsample_output_size(self, size, "nearest", "asymmetric")
+    if scales_h is not None and scales_w is not None:
+        return _aten_upsample_scales(
+            self,
+            [scales_h, scales_w],
+            "nearest",
+            "asymmetric",
+        )
+    else:
+        return _aten_upsample_output_size(self, output_size, "nearest", "asymmetric")
```
gramalingam, line 2612: Can we assert output_size is not None here?

gramalingam: Or, in the callee above, I guess that's probably simpler

justinchuby: Done

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: Not optional, as seen in the signature `"upsample_nearest2d(Tensor self, SymInt[2] output_size, float? scales_h=None, float? scales_w=None) -> Tensor`. I think the op was not very well defined which is why they are moving to use the `.vec` variant.

onnxscript/function_libs/torch_lib/ops/nn.py
justinchuby: Fixed

PR #1880, opened 2024-09-24
[torchlib] Fix upsample_output_size

justinchuby:
Add a line to cast inputs to INT64 because when output_size is passed in as a list of integers, the graph builder in PyTorch exporter may fail to determine the output type. We cast it to INT64 to ensure the output

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2365,7 +2365,8 @@ def _aten_upsample_output_size(
     starts = op.Constant(value_ints=[0])
     ends = op.Constant(value_ints=[2])
     batch_channel = op.Slice(self_shape, starts, ends)
-    # When output_size is passed in as a list of integers, op.Concat may fail
+    # When output_size is passed in as a list of integers, the torch.onnx
+    # graph builder when handling op.Concat may fail
     # to determine the output type. We cast it to INT64 to ensure the output
     output_size = op.Cast(output_size, to=INT64.dtype)
     output_size = op.Concat(batch_channel, output_size, axis=0)
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2361,15 +2361,12 @@ def _aten_upsample_output_size(
     mode: str,
     coordinate_transformation_mode: str,
 ) -> TReal:
-    self_shape = op.Shape(self)
-    starts = op.Constant(value_ints=[0])
-    ends = op.Constant(value_ints=[2])
-    batch_channel = op.Slice(self_shape, starts, ends)
+    batch_channels = op.Shape(self, end=2, start=0)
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2361,11 +2361,13 @@ def _aten_upsample_output_size(
     mode: str,
     coordinate_transformation_mode: str,
 ) -> TReal:
-    self_shape = op.Shape(self)
-    starts = op.Constant(value_ints=[0])
-    ends = op.Constant(value_ints=[2])
-    batch_channel = op.Slice(self_shape, starts, ends)
-    output_size = op.Concat(batch_channel, output_size, axis=0)
+    batch_and_channel = op.Shape(self, end=2, start=0)
+    # When output_size is passed in as a list of integers, the torch.onnx
```
gramalingam, line 2365: What is a list of integers mapped to (when treated as an ONNX value)? If the cast below works, then it must be mapped to a tensor of some type. So, INT32 is a possibility. But my memory is that some parts of onnxscript at least treat ints as INT64 (since that is the common use, eg. in shapes, as indices, etc.). That seems a preferable interpretation/conversion for a list of ints. Which would make the cast unnecessary. Maybe worth digging into at some point.

justinchuby: I can explain more later: this is in the graph building logic in the exporter.

gramalingam: Is [this](https://github.com/pytorch/pytorch/blob/a8ed873ba2aa13437a336863ae9d73c235798acc/torch/onnx/_internal/exporter/_building.py#L224) the logic? If so, this code should make a difference only if output_size includes a floating value. I guess that is what might be happening, perhaps somewhere 2.0 was used instead of 2? If so, maybe the above comment is worth updating. It is used when floating values are present in output_size.

gramalingam: Actually, we don't need a runtime Cast. We can just do `output_sizes = [int(x) for x in output_sizes]`. If desired, we can also add a check that the int-conversion is lossless ...

justinchuby: Output size can be dynamic because it can contain symints. Maybe I need to revisit how we handle lists with mixed dynamic and python values

gramalingam: My conjecture was that Concat has a type schema, which means that when we have a python list as the 2nd argument, the type schema plays a factor in how the 2nd argument is converted into an onnx value (specifically the type of the first argument and second are expected to be the same). This is further complicated by the fact that concat is variadic. But these issues do not exist when the same list is used with a Cast (as in your fix here). Maybe that is why the list is translated fine here, but not when used with Concat.

In any case, let me merge this fix in for now, to let us continue our evaluation.

gramalingam (approved):

PR #1892, opened 2024-10-08
A few fixes relating to constant propagation

gramalingam:
Fixes a few different issues. Helps resolve an issue relating to ir-based optimization for the Blender model in the benchmark.

* Move the utility for evaluating `Constant` op into the IR, and make `const_value` automatically perform the related computation.
* Eliminate the dependence on the reference-implementation for evaluation of Constant op.
* There are still a couple of issues relating to the use of reference-implementation (eg., when we have tensor-valued attributes in external-data format, and the use of float16) which will need to be addressed separately, but the above bypasses this issue for Constant op (and the Blender model).
* Make the optimizer robust to external-data-tensors whose files are not available.


onnxscript/optimizer/_constant_folding.py
```diff
@@ -168,7 +172,11 @@ def _get_numpy_value(val: ir.Value | None) -> np.ndarray | None:
         return None
     const_value = val.const_value
     if const_value is not None:
-        return const_value.numpy()
+        try:
+            return const_value.numpy()
+        except FileNotFoundError:
```
gramalingam, line 177: Should this exception handling be in the implementation of the `numpy()` function? Wasn't sure how it might impact all other callers if we changed the signature of `numpy()` to return None sometimes. So, leaving it in its current form for now.

onnxscript/ir/_core.py
```diff
@@ -1621,6 +1650,10 @@ def const_value(
         The value can be backed by different raw data types, such as numpy arrays.
         The only guarantee is that it conforms TensorProtocol.
         """
+        if self._const_value != None:
+            return self._const_value
+        # On-demand computation of constant value. Currently limited to outputs of Constant nodes.
+        self._compute_const_value()
```
justinchuby, line 1656: I am hesitant to include this because const_value is a user set field (not computed). Maybe we can have a new api called (`propagate_constants()`) or something like that on the value, to compute the constants?

onnxscript/optimizer/_constant_folding.py
justinchuby: numpy() should always return a numpy array or error. I think this is appropriate.

onnxscript/ir/_core.py
gramalingam: Looks like it's a bit tricky with some tradeoffs:
* The case where the above maybe tricky/unexpected is if the user explicitly sets this field to `None` (but it is the output of a Constant node). If the user sets the field to any non-None value, the behavior is still as expected. Setting the value to None is not expected or used right now, I think, especially for the output of a Constant node. We could check for it in the setter too, to disallow it.
* I think I can move the call to _compute_const_value to the Node(...) constructor. I think that may be a good choice.
* I prefer that users don't have to remember to call propagate() before looking up the constant-value.
* Most of the approaches still have some risk if nodes are explicitly modified. Eg., I could change the op_type from `Constant` to something else. I guess, for now, we can leave it as the user-responsibility to clear the constant-value if they make mutations that affect the const-value.

justinchuby: The main concern I have for computing const_value is the potential overhead for accessing this field, since it is a property, we have conveyed an expectation of O(1) access. The same for the Node constructor - it would be good to stay lightweight imo. 

I wonder if there is a way to do it more elegantly? For example, would it be a good idea to allow something like `propagate()` to return a constant tensor (and set const_value as well)? This way when users expect computed values they will just call propagate() (or `compute()` or some properly named method) without having to do the two step access?

This way passes can still use the const_value field to store constants for their own usage, and when users need a computed value they can call this new method to obtain it.

gramalingam: Just to clarify the tradeoffs above: calling `_compute_const_value` lazily (as in the current PR) is a bit more robust than calling it earlier (when the value is first created as the output of a specific node). Eg., it would work better if the user is allowed to create a node with given inputs and given number of outputs, and the user sets the domain/op_type etc. for the node subsequently. That's not an idiom we use or want to encourage, but it is doable. But I am not sure if it will make a difference in our current implementation.

gramalingam: The overhead is still O(1). I don't think the cost significant enough to really matter. I think we can add a boolean field to ensure that this overhead happens only the first time it is called. (I did think about it, but didn't bother since it just eliminates a string comparison more or less, but it can be done).

We can change it to *not* be a property if that will help. But I am concerned about exposing two different APIs (const_value and propagate) which will increase the likelihood that somebody ends up using const_value instead of propagate, which will be a bigger concern.

onnxscript/ir/_core.py
```diff
@@ -1612,6 +1612,35 @@ def shape(self, value: Shape | None) -> None:
             return
         raise TypeError(f"Expected value to be a Shape or None, got '{type(value)}'")
 
+    def _compute_const_value(self) -> None:
+        """Evaluates Constant nodes and sets the const_value attribute of the Value."""
+        node = self.producer()
```
justinchuby, line 1617: From our discussion: I would add a strong clause saying this should be limited to Constant nodes only and maintain O(1) access

onnxscript/ir/_core.py
```diff
@@ -1612,6 +1612,35 @@ def shape(self, value: Shape | None) -> None:
...
+        if node is None:
+            return
+        if node.op_type != "Constant" or node.domain not in {"", "ai.onnx"}:
+            return
+        attr_name, attr_value = next(iter(node.attributes.items()))
```
justinchuby, line 1622: Will this fail if attributes is empty?

onnxscript/ir/_core.py
```diff
@@ -1612,6 +1612,35 @@ def shape(self, value: Shape | None) -> None:
...
+            const_value = typing.cast(_protocols.TensorProtocol, attr_value.value)
+        else:
+            return
+
+        self.const_value = const_value
+        self.shape = const_value.shape  # type: ignore
+        self.dtype = const_value.dtype
+
     @property
     def const_value(
```
justinchuby, line 1645: Possible to construct some invalid const nodes in test and ensure that getting const_value does not fail?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -604,6 +612,12 @@ def process_node(self, node: ir.Node):
         for i, value in enumerate(node.inputs):
             sym_value = self._state.get_sym_value(value)
             if isinstance(sym_value, ir.Value):
+                logger.debug(
+                    "Node [%s]: Replacing input %s with %s",
+                    node.name,
+                    value.name,
```
justinchuby, line 618: Looks like value may be None?

justinchuby (approved):
lgtm with tests added, thanks!

onnxscript/ir/_core.py
gramalingam: @justinchuby  I have removed this change. Now, it is the responsibility of any traversal/mutator to ensure that the const_value is set/updated as needed. Which is less than ideal, but hopefully it works fine for now, will need to try it on the entire benchmark.

PR #1896, opened 2024-10-08
[API] Create stable APIs for PyTorch 2.6

justinchuby:
- optimize is turned on. It will be controlled by an option in PyTorch
- Remove the `_TORCH_ONNX_SAVE_EXTERNAL_DATA_WITH_IR` flag

### DIFF
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -17,16 +17,9 @@ import os
 import pathlib
 from typing import Callable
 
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):
Could you remind me again where to control the version of api when users try to export?

justinchuby:
Each pytorch release will use a version that is design for that release. We are responsible to update the pytorch code to use the correct version. For example, we should go in now and change pytorch to use the 2.6 api.

Issue #1299, opened 2024-03-13
ConvTranspose has the wrong bias shape


yuanyao-nv:
In some exported model with ConvTranspose I'm seeing the wrong shape for bias:
![image](https://github.com/microsoft/onnxscript/assets/99693700/e5b1c830-493e-4757-9111-8eeddcdaa478)
![image](https://github.com/microsoft/onnxscript/assets/99693700/a0ec9ae3-aaef-4bfb-8f46-c03a40ae9078)
where `B` should have shape (128) instead of (256) according to the ONNX spec.

Looking at the onnxscript code, the problem seems to be that since 
`_aten_convolution_onnx()` is used for both Conv and ConvTranspose, the logic for filling in bias should check the 0th dimension of weight if it's Conv and the 1st dimension of weight if it's ConvTranspose:
```
    if bias is None:
        weight_dim_0 = op.Shape(weight, start=0, end=1)
        bias_shape = op.Expand(weight_dim_0, op.Constant(value_ints=[1]))
        zero = op.CastLike(0.0, input)
        bias = op.Expand(zero, bias_shape)
```
I think the value of `groups` is also needed since the shape of weight is (M x C/group x kH x kW) for Conv and (C x M/group x kH x kW) for ConvTranspose?

justinchuby:
@fatcat-z would you be able to help take a look? Thanks!

gramalingam:
Hi, wondering about this issue ... any progress? Thanks!

fatcat-z:
> Hi, wondering about this issue ... any progress? Thanks!

Just finished another issue and start working on it now.

fatcat-z:
@yuanyao-nv ,
The function _aten_convolution_onnx was designed to serve those convolution ops, like aten::convolution, aten::conv1d, aten::2d, aten::3d. It is not used by any conv_transpose op(conv_transpose1d/conv_transpose2d/conv_transpose3d).

Based on this, the logic of filling bias will be as same as Conv op. We call op.ConvTranspose op in _aten_convolution_onnx() because the parameter 'transposed' of aten::convolution op. If transposed it True, then we can call op.ConvTranspose() op to get the same results as aten::convolution op. And it has been covered by our tests and proved it is true.

```
    if transposed:
        result = op.ConvTranspose(
            input,
            weight,
            bias,
            strides=strides,
            pads=pads,
            group=groups,
            dilations=dilations,
            output_padding=output_padding,
        )
    else:
        result = op.Conv(
[...]

PR #1901, opened 2024-10-09
[torchlib] Fix wrong bias shape of ConvTranspose

yuanyao-nv:
Previously, the bias shape of ConvTranpose was wrong since, unlike Conv, it should using the 1st dimension of weight shape and not the 0th. See described in https://github.com/microsoft/onnxscript/issues/1299

In other words,
```
    if bias is None:
        weight_dim_0 = op.Shape(weight, start=0, end=1)
        bias_shape = op.Expand(weight_dim_0, op.Constant(value_ints=[1]))
        zero = op.CastLike(0.0, input)
        bias = op.Expand(zero, bias_shape)
```
should be changed to something like:
```
weight_dim_0 = op.Shape(weight, start=1, end=2) if transposed else op.Shape(weight, start=0, end=1)
```

However, I think it's more efficient to just eliminate bias altogether if it's not provided instead of filling it with zeros, since the ONNX spec allows bias to be absent.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -2030,12 +2030,6 @@ def aten_convolution(
         stride = (stride, stride)
     strides = list(stride)
 
-    if bias is None:
-        weight_dim_0 = op.Shape(weight, start=0, end=1)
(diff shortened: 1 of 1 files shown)


justinchuby:
Amazing, thanks!

justinchuby (approved):

PR #1903, opened 2024-10-11
Use input size limits for constant folding

gramalingam:
Add input size limits for constant folding. Helps avoid excessive time in optimizer in some edge cases. (The edge cases, where we have non-trivial ops applied to large tensors, are not relevant for the exporter itself. They may be of potential interest for optimization in other settings, but that can be done by user taking explicit steps.)

Still to be done: how do we specify these values from the benchmarking code? For now, the default values will be quite useful, but experimenting with these values from the benchmarking code will need a way to control these option values.

### DIFF
--- a/onnxscript/optimizer/__init__.py
+++ b/onnxscript/optimizer/__init__.py
@@ -111,17 +111,33 @@ def optimize(
     return model
 
 
+_DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT = (
+    _constant_folding._DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT
+)
+
+_DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT = (
+    _constant_folding._DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT
+)
+
+
 def optimize_ir(
     model: ir.Model,
     num_iterations: int = 2,
     *,
     onnx_shape_inference: bool = True,
     stop_if_no_change: bool = True,
+    input_size_limit: int = _DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT,
+    output_size_limit: int = _DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT,
 ) -> None:
     del stop_if_no_change  # Looks like rewriter doesn't support this yet.
     _inliner.inline(model)
     for _ in range(num_iterations):
-        _constant_folding.fold_constants(model, onnx_shape_inference=onnx_shape_inference)
+        _constant_folding.fold_constants(
+            model,
+            onnx_shape_inference=onnx_shape_inference,
+            input_size_limit=input_size_limit,
+            output_size_limit=output_size_limit,
+        )
         rewriter.rewrite(model, pattern_rewrite_rules=_DEFAULT_REWRITE_RULES)
     remove_unused_nodes(model)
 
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -696,6 +703,14 @@ def process_node(self, node: ir.Node):
         if any(x is None for x in input_values):
             return None
 
+        if any(input.size > self._input_size_limit for input in input_values):
+            if logger.isEnabledFor(logging.DEBUG):
+                input_sizes = [input.size for input in input_values]
+                logger.debug(
+                    f"Skipping constant folding for op {node.op_type} due to large input size: {input_sizes}"
```
justinchuby, line 710: Use %s for logs

onnxscript/optimizer/__init__.py
```diff
@@ -111,17 +111,33 @@ def optimize(
...
+
+
 def optimize_ir(
     model: ir.Model,
     num_iterations: int = 2,
     *,
     onnx_shape_inference: bool = True,
     stop_if_no_change: bool = True,
+    input_size_limit: int = _DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT,
+    output_size_limit: int = _DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT,
```
justinchuby, line 130: Helpful to document these options?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -696,6 +703,14 @@ def process_node(self, node: ir.Node):
         if any(x is None for x in input_values):
             return None
 
+        if any(input.size > self._input_size_limit for input in input_values):
```
justinchuby, line 706: Looks like there might be some None cases to handle?

onnxscript/optimizer/_constant_folding.py
gramalingam: Does that still apply when the code is already protected by `isEnabledFor` ?

onnxscript/optimizer/_constant_folding.py
gramalingam: Already filtered out in the line before.

onnxscript/optimizer/_constant_folding.py
gramalingam: IIUC, the main value is in avoiding unnecessary formatting. If we are enabled for it, then nothing is achieved by delaying the formatting.

justinchuby: That's true. Just easier to check for than having to make an exception for the lint rule I suppose?

onnxscript/optimizer/__init__.py
```diff
@@ -111,17 +111,47 @@ def optimize(
...
+        model: The model to be optimized
+        num_iterations: Number of times the optimization loop is repeated
+        onnx_shape_inference: Applies node-level shape-inference as part of optimization
+        input_size_limit: Will not apply constant folding to ops with any input of size
+            greater than this. Does not apply to special ops like Shape() and Size().
+        output_size_limit: Will not rewrite any foldable-op into a Constant op if the size
+            of the output tensor is greater than this.
+        stop_if_no_change: Not supported currently (has no effect). Meant to stop the
+            outer optimization loop if no change is detected in one iteration.
+    """
```
justinchuby, line 145:
```suggestion
    """Optimizes a model.

    Args:
        model: The model to be optimized.
        num_iterations: Number of times the optimization loop is repeated.
        onnx_shape_inference: Applies node-level shape-inference as part of optimization
        input_size_limit: Will not apply constant folding to ops with any input of size
            greater than this. Does not apply to special ops like Shape() and Size().
        output_size_limit: Will not rewrite any foldable-op into a Constant op if the size
            of the output tensor is greater than this.
        stop_if_no_change: Not supported currently (has no effect). Meant to stop the
            outer optimization loop if no change is detected in one iteration.
    """
```

nit: minor formatting

justinchuby (approved):

PR #1905, opened 2024-10-14
[IR] Do not serialize the trailing outputs that have empty names

justinchuby:


onnxscript/ir/serde.py
```diff
@@ -1288,8 +1293,15 @@ def serialize_node_into(node_proto: onnx.NodeProto, from_: _protocols.NodeProtoc
             node_proto.input.append("")
         else:
             node_proto.input.append(input_.name)
-    for output in from_.outputs:
+    # Do not include the trailing outputs that have empty names
+    non_empty_outputs_reversed = []
+    for output in reversed(from_.outputs):
+        if not output.name:
+            continue
```
gramalingam, line 1300: Logic doesn't look right. Isn't this skipping all empty names, instead of just trailing ones?

Also: may be it would be more efficient to just find the index position of last non-empty string (in a reverse traversal) and use a forward enumeration traversal that exits at that index, rather than constructing a new list. (But may be doesn't matter much, whatever looks simple might be fine.)

justinchuby: Good catch thanks - will update; adding a test

gramalingam (approved):

PR #1908, opened 2024-10-15
[IR] Support float4e2m1

justinchuby:
Support the float4e2m1 dtype from IRv11 (which is not yet released). This allows our tests to pass in the weekly-onnx CI. We use the ml_dtypes.float4_e2m1fn type for numpy conversion. Since ml_dtypes.float4_e2m1fn is only available in the latest ml_dtypes release which has dropped support for python 3.8, I used a conditional logic to build the numpy dtype mapping table.

### DIFF
--- a/onnxscript/ir/_enums_test.py
+++ b/onnxscript/ir/_enums_test.py
@@ -32,6 +32,8 @@ class DataTypeTest(unittest.TestCase):
         self.assertEqual(_enums.DataType.FLOAT8E5M2FNUZ, onnx.TensorProto.FLOAT8E5M2FNUZ)
         self.assertEqual(_enums.DataType.UINT4, onnx.TensorProto.UINT4)
         self.assertEqual(_enums.DataType.INT4, onnx.TensorProto.INT4)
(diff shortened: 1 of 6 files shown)


onnxscript/ir/_enums_test.py
```diff
@@ -32,6 +32,8 @@ def test_enums_are_the_same_as_spec(self):
         self.assertEqual(_enums.DataType.FLOAT8E5M2FNUZ, onnx.TensorProto.FLOAT8E5M2FNUZ)
         self.assertEqual(_enums.DataType.UINT4, onnx.TensorProto.UINT4)
         self.assertEqual(_enums.DataType.INT4, onnx.TensorProto.INT4)
+        if hasattr(onnx.TensorProto, "FLOAT4E2M1"):
```
xadupre, line 35: I guess the new functions are already tested somewhere else.

titaiwangms (approved):

onnxscript/ir/_enums_test.py
justinchuby: Added more tests

PR #1912, opened 2024-10-22
A couple of extensions to rewriter

gramalingam:
A couple of extensions to the rewriter, motivated by fusion optimization experimentation with SmoLLM.

* Support list of constants in match-pattern.
* One multi-output scenario is easy to handle with the single-output pattern-matcher (eg. defining a fusion rule for SkipNormalization): namely when the extra outputs are intermediate values used in the computation of the first value. Extend algorithm to handle this scenario using the efficient single-output matching-algorithm.

An example for the second point is the following pattern:
```py
def skip_norm_pattern(op, input, skip, gamma, epsilon, stash_type):
    skip_sum = op.Add(input, skip)
    normalized = op.SimplifiedLayerNormalization(
        skip_sum,
        gamma,
        axis=-1,
        epsilon=epsilon,
        stash_type=stash_type,
        _domain="com.microsoft")
    return normalized, skip_sum
```
If we successfully find a match for `normalized` (which transitively finds a match for all of the pattern subgraph that leads up to `normalized`), we have also found a successful match for `skip_sum`, so no need for a multi-output match.

(Will add test-cases later, as I work through the fusion optimizations I am experimenting with.)

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -963,8 +963,9 @@ class SimplePatternMatcher(PatternMatcher):
         pattern_constant_value = pattern_constant._value
 
         if isinstance(pattern_constant_value, list):
-            if constant_value_numpy.shape != (len(pattern_constant_value),):
-                return self.fail(f"Value has mismatching shape, expecting ({self.value},).")
+            expected_shape = (len(pattern_constant_value),)
+            if constant_value_numpy.shape != expected_shape:
+                return self.fail(f"Value has mismatching shape, expecting {expected_shape}.")
             if not all(
                 math.isclose(
                     constant_value_numpy.item(i),
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -601,7 +601,10 @@ class Constant(ValuePattern):
     """Represents a pattern that matches against a scalar constant value."""
 
     def __init__(
-        self, value: int | float | list[int] | list[float], rel_tol: float = 1e-5, abs_tol: float = 1e-8
+        self,
+        value: int | float | list[int] | list[float],
+        rel_tol: float = 1e-5,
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -282,12 +282,11 @@ def _to_value_pattern(
         return x
     if isinstance(x, (int, float)):
         return Constant(x)
-    # TODO(rama): support lists of int/float
-    # if isinstance(x, list):
-    #     if all(isinstance(i, (int, float)) for i in x):
-    #         return Constant(x)
-    #     raise ValueError("Only lists of int/float can be used as a ValuePattern")
-    # TODO(titaiwang): Could this be wrapped Constant?
+    if isinstance(x, list):
```
justinchuby, line 285: Can we handle sequences too, or are lists special here? (e.g. tuples are not expected?)

onnxscript/rewriter/pattern.py
```diff
@@ -664,6 +684,15 @@ def visit(value_patterns: Sequence[ValuePattern | None]) -> None:
     return node_patterns
 
 
+def _add_backward_slice(node: NodePattern, backward_slice: set[NodePattern]) -> None:
```
justinchuby, line 687: Helpful to add a docstring?

justinchuby (approved):
I am still a little confused by the second case. Would it be helpful to make the example more concrete in the PR description? Thanks!

gramalingam:
> I am still a little confused by the second case. Would it be helpful to make the example more concrete in the PR description? Thanks!

Added an example

onnxscript/rewriter/pattern.py
gramalingam: done

onnxscript/rewriter/pattern.py
gramalingam: done

PR #1781, opened 2024-08-05
[torchlib] Fix aten::arange

justinchuby:
Change the function signature and simplify logic to handle scalars only. Some values can be eagerly evaluated.

PR #1913, opened 2024-10-22
[torchlib] Fix aten::arange to support dynamic shapes

titaiwangms:
Runinng SmolLM_1_7b with dynamic shapes discovered that arange.start was not dynamic anymore, since https://github.com/microsoft/onnxscript/pull/1781/files.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -632,7 +632,42 @@ def aten_arange_start_step(
     """arange.start_step(Scalar start, Scalar end, Scalar step=1, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
 
     if dtype == -1:
-        result = op.Range(start, end, step)
+        # TODO: Because this is a trace_only function, the inputs are not promoted to
+        # Tensor until it hits ONNX ops. However, if it's dynamic, it should be
+        # Tensor at this point.
+        # https://github.com/microsoft/onnxscript/issues/1914
+        if isinstance(start, (int, float)):
+            start_is_int = isinstance(start, int)
+        else:
+            start_is_int = start.dtype in {
+                INT16.dtype,
+                INT32.dtype,
+                INT64.dtype,
+            }
+        if isinstance(end, (int, float)):
+            end_is_int = isinstance(end, int)
+        else:
+            end_is_int = end.dtype in {
+                INT16.dtype,
+                INT32.dtype,
+                INT64.dtype,
+            }
+        if isinstance(step, (int, float)):
+            step_is_int = isinstance(step, int)
+        else:
+            step_is_int = step.dtype in {
+                INT16.dtype,
+                INT32.dtype,
+                INT64.dtype,
+            }
+        if start_is_int and end_is_int and step_is_int:
+            result = op.Range(start, end, step)
+        else:
+            # to float
+            start = op.Cast(start, to=FLOAT.dtype)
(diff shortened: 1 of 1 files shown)


justinchuby (approved):
Thanks, also interesting - does this mean we need to update https://github.com/pytorch/pytorch/blob/10f16cc7daf89a73c1177dfbd0bb0fbb0fc8683b/aten/src/ATen/native/native_functions.yaml#L780 ? Should we create an issue for propose a fix?

justinchuby:
Should we also fix the rest of the arange overloads?

titaiwangms:
> Should we also fix the rest of the arange overloads?

Sure, I can update the whole family. And record the question that does Scalar cover sym_int and sym_float in Meta doc.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -599,66 +595,65 @@ def aten_arange_start(
...
+    end: TRealUnlessFloat16OrInt8,
+    step: TRealUnlessFloat16OrInt8,
+) -> Tuple[FLOAT, FLOAT, FLOAT]:
+    zero = op.Cast(0.0, to=FLOAT.dtype)
+    start = op.Cast(start, to=FLOAT.dtype)
+    end = op.Cast(end, to=FLOAT.dtype)
+    step = op.Cast(step, to=FLOAT.dtype)
+
+    start = op.Where(op.Less(start, zero), op.Ceil(start), start)
+    start = op.Where(op.Less(step, zero), op.Floor(start), start)
```
titaiwangms, line 617: @justinchuby This is unavoidably brought back. Were you trying to solve a performance issue with https://github.com/microsoft/onnxscript/pull/1781/files perhaps?

xadupre: I wonder why it is needed to fix start paremeter?

xadupre:
Is it possible to add a unit test to make sure there is no regression about this?

titaiwangms:
> Is it possible to add a unit test to make sure there is no regression about this?

I will merge this first to unblock the model, and then come up with tests as follow-up.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: I remember there were test cases that would break without this logic

titaiwangms: I am reverting the whole changes from [https://github.com/microsoft/onnxscript/pull/1781](https://github.com/microsoft/onnxscript/pull/1781/files)

justinchuby:
The old arange logic had issues with inconsistent types. Please make sure that's resolved. I am happy if you fix one and take time to ensure the rest is correct in a separate PR, if that allows faster progress

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -599,66 +595,63 @@ def aten_arange_start(
...
 
 
 @torch_op("aten::arange.start_step", trace_only=True)
 def aten_arange_start_step(
-    start: float,
-    end: float,
-    step: float = 1.0,
+    start: TRealUnlessFloat16OrInt8,
+    end: TRealUnlessFloat16OrInt8,
+    step: TRealUnlessFloat16OrInt8 = 1.0,
```
justinchuby, line 626: does this default work?

titaiwangms: Add issue https://github.com/microsoft/onnxscript/issues/1914 to follow up on TracedONNXFunction extension.

Issue #1914, opened 2024-10-23
[torchlib] TracedONNXFunction promotes python constants to tensors in evaluator function call


titaiwangms:
Currently, TracedONNXFunction does not support prmoting python constant to tensors when inputs and attributes enters the function. The challenge arises when dynamic shapes is involved, since dynamic shapes could potentially turn python constants to torch.SymInt (torch.fx.Node), which makes writing torchlib more complicated. The solution would be let the well-defined (the function has op signature) TracedONNXFunction be able to promote python constants to tensors before entering the function, while the function without good annotations staying fully tracing.

PR #1915, opened 2024-10-23
[torchlib] Use traced function param schema to process inputs

titaiwangms:
The firs step of https://github.com/microsoft/onnxscript/issues/1914, this is setting up onnxscript CI to test whether traced_only function has enough information to process inputs to tensors.


onnxscript/function_libs/torch_lib/graph_building/_graph_building_ir.py
justinchuby, line 1: graph building ir is not being used anywhere (I don't think)

titaiwangms:
Moved to #1916

PR #1916, opened 2024-10-23
[torchlib] Use traced function param schema to process inputs

titaiwangms:
The firs step of https://github.com/microsoft/onnxscript/issues/1914, this is setting up onnxscript CI to test whether traced_only function has enough information to process inputs to tensors.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -422,8 +422,10 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
         if function.traceable:
             inputs = self._graph.preprocess_inputs(inputs)
             inputs = _wrap_torch_value_to_tensor(inputs)  # type: ignore[assignment]
+            # return to args order, as it's traced onnx function
+            args = param_manipulation.return_to_args_order(param_schemas, inputs, attributes)
             # Trace the function call instead of adding the function as a node
-            return function.function(*inputs, **attributes)
+            return function.function(*args)
         return self._graph.add_function_call(function, inputs, attributes)
 
 
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -422,8 +422,6 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
         if function.traceable:
             inputs = self._graph.preprocess_inputs(inputs)
             inputs = _wrap_torch_value_to_tensor(inputs)  # type: ignore[assignment]
-            # return to args order, as it's traced onnx function
-            args = param_manipulation.return_to_args_order(param_schemas, inputs, attributes)
             # Trace the function call instead of adding the function as a node
             return function.function(*args)
         return self._graph.add_function_call(function, inputs, attributes)
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -421,6 +421,7 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
                 attributes[name] = (value,)
         if function.traceable:
             inputs = self._graph.preprocess_inputs(inputs)
+            inputs = _wrap_torch_value_to_tensor(inputs)  # type: ignore[assignment]
             # Trace the function call instead of adding the function as a node
             return function.function(*inputs, **attributes)
         return self._graph.add_function_call(function, inputs, attributes)
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -422,10 +422,12 @@ class TorchScriptTracingEvaluator(evaluator.Evaluator):
         if function.traceable:
             inputs = self._graph.preprocess_inputs(inputs)
             inputs = _wrap_torch_value_to_tensor(inputs)  # type: ignore[assignment]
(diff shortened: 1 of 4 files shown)


onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
```diff
@@ -730,14 +731,7 @@ def _add_constant_to_graph(self, constant) -> torch.Value:
...
 
-    @runtime_typing.checked
-    def _add_torchscript_op_call(
-        self,
-        name: str,
-        onnx_inputs: Sequence[ValidInputType],
-        onnx_attributes: Mapping[str, ValidArgumentType],
-        n_outputs: int,
-    ) -> Union[TorchScriptTensor, Tuple[TorchScriptTensor, ...]]:
+    def preprocess_inputs(self, onnx_inputs: Sequence[ValidInputType]) -> List[torch.Value]:
```
justinchuby, line 734:
```suggestion
    def _preprocess_inputs(self, onnx_inputs: Sequence[ValidInputType]) -> List[torch.Value]:
```

justinchuby (approved):

onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
titaiwangms: This will trigger `protect-access` to line:

```python
inputs = self._graph.preprocess_inputs(inputs)
```

justinchuby: I see. then keeping the original is ok

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5752,14 +5752,9 @@ def aten_nansum(
...
-        dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
-
-    if IsScalar(start):
-        start = op.Reshape(start, op.Constant(value_ints=[-1]))
-
-    if IsScalar(length):
-        length = op.Reshape(length, op.Constant(value_ints=[-1]))
+    dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
+    start = op.Reshape(start, op.Constant(value_ints=[-1]))
+    length = op.Reshape(length, op.Constant(value_ints=[-1]))
```
titaiwangms, line 5757: @justinchuby here,

I can't figure out why IsScalar failing to recognize these as scalars, which leads to ort errors: here is the part of the model:

```
E   Model:
E   <
E      ir_version: 8,
E      opset_import: ["pkg.onnxscript.torch_lib.common" : 1, "" : 18, "pkg.onnxscript.torch_lib.common" : 1],
E      producer_name: "pytorch",
E      producer_version: "2.6.0"
E   >
E   main_graph (int64[5,5,5] input_0, int64 input_2) => (int64[5,5,2] _val_9) 
E      <bool _val_4, bool _val_7, int64 _val_2, int64 _val_3, int64[1] _val_5, int64[1] _val_6, int64[1] _val_8>
E   {
E      _val_2 = Constant <value: tensor = int64 {2}> ()
E      _val_3 = Constant <value: tensor = int64 {2}> ()
E      _val_4 = pkg.onnxscript.torch_lib.common.IsScalar (_val_2)
E      _val_5 = Constant <value_ints: ints = [-1]> ()
E      _val_6 = Reshape <allowzero: int = 0> (input_2, _val_5)
E      _val_7 = pkg.onnxscript.torch_lib.common.IsScalar (_val_3)
E      _val_8 = Add (_val_6, _val_3)
E      _val_9 = Slice (input_0, _val_6, _val_8, _val_2)
E   }
E   <
E     domain: "pkg.onnxscript.torch_lib.common",
E     opset_import: ["" : 18]
E   >
E   IsScalar (input) => (return_val)
E   {
E      tmp = Shape (input)
E      tmp_0 = Size (tmp)
E      tmp_1 = Constant <value_int: int = 0> ()
E      return_val = Equal (tmp_0, tmp_1)
E   }
E   <
E     domain: "pkg.onnxscript.torch_lib.common",
E     opset_import: ["" : 18]
E   >
E   Rank (input) => (return_val)
E   {
E      tmp = Shape (input)
E      return_val = Size (tmp)
E   }
E   <
E     domain: "pkg.onnxscript.torch_lib.common",
E     opset_import: ["" : 18]
E   >
E   IsScalar (input) => (return_val)
E   {
E      tmp = Shape (input)
E      tmp_0 = Size (tmp)
E      tmp_1 = Constant <value_int: int = 0> ()
E      return_val = Equal (tmp_0, tmp_1)
E   }
```

Is _val_2 and _val_3 not scalar in ONNX spec?

justinchuby: Is scalar only works in a function, or when the input rank is known. Here _val_3 is an intermediate, and its rank is not propagated.

titaiwangms: In that case, we need to remember to set shape and type after promoting python constants to tensors.

justinchuby: Doing shape propagation in the export is possible, but I wonder if there is a more general way to do it. Having onnxscript to handle that is of course ideal, but that is going to be a lot of work.

titaiwangms: I thought we used the shape and type from fx graph during conversion?

justinchuby: Yes but that’s just the torch tensors. Constant nodes created from Python values are more ambiguous. As they are created by calling onnxscript Constant

titaiwangms: I see. How about we only annotates int/float/bool atm? https://github.com/microsoft/onnxscript/pull/1918 (This is just taking torchscript as an example, but I believe IR can also do this.)

PR #1918, opened 2024-10-28
Set 0d shape to python constant tensors

titaiwangms:
From #1916, it is found that the undefined shape disables the functionality of IsScalar op: https://github.com/microsoft/onnxscript/blob/561a6006ff48b744c7a03d80994fd74ea06be5f7/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py#L357 This PR annotates python scalar to be 0d tensor.

### DIFF
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -739,7 +739,7 @@ class TorchScriptGraph:
         )[0]
         value.setDebugName(_rename_intermediate_value(value.debugName()))
         if shape is not None:
-            value.setType(value.type().with_sizes(shape))
+            value.setType(value.type().with_sizes(shape))  # type: ignore[arg-type]
         return value
 
     def preprocess_inputs(self, onnx_inputs: Sequence[ValidInputType]) -> List[torch.Value]:
--- a/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
+++ b/onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
@@ -704,13 +704,17 @@ class TorchScriptGraph:
             value.setDebugName(_rename_intermediate_value(value.debugName()))
             return value
 
+        shape: list[int] | None = None
         if isinstance(constant, bool):
             # Be sure to put bool before int, because bool is a subclass of int
             constant_tensor = torch.tensor(constant, dtype=torch.bool)
+            shape = []
         elif isinstance(constant, float):
             constant_tensor = torch.tensor(constant, dtype=torch.float)
+            shape = []
         elif isinstance(constant, int):
             constant_tensor = torch.tensor(constant, dtype=torch.int64)
+            shape = []
         elif isinstance(constant, (tuple, list)) and all(
             isinstance(val, int) for val in constant
         ):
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5752,9 +5752,14 @@ def aten_nansum(
 def aten_narrow(self: TTensor, dim: INT64, start: INT64, length: INT64) -> TTensor:
     """narrow(Tensor(a) self, int dim, SymInt start, SymInt length) -> Tensor(a)"""
 
-    dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
-    start = op.Reshape(start, op.Constant(value_ints=[-1]))
-    length = op.Reshape(length, op.Constant(value_ints=[-1]))
+    if IsScalar(dim):
+        dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
```
gramalingam, line 5756: I suggest changing the `-1` to `1` ... since it is going to have the shape `[1]`. Ensures better shape inference with all of the various shape-inference implementations around.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5752,9 +5752,14 @@ def aten_nansum(
...
+
+    if IsScalar(start):
+        start = op.Reshape(start, op.Constant(value_ints=[-1]))
```
gramalingam, line 5759: Same as above, here and below

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5752,9 +5752,14 @@ def aten_nansum(
 def aten_narrow(self: TTensor, dim: INT64, start: INT64, length: INT64) -> TTensor:
```
justinchuby, line 5752: Dim is always int right? So it is always a scalar.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5752,9 +5752,14 @@ def aten_nansum(
...
     """narrow(Tensor(a) self, int dim, SymInt start, SymInt length) -> Tensor(a)"""
 
-    dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
-    start = op.Reshape(start, op.Constant(value_ints=[-1]))
-    length = op.Reshape(length, op.Constant(value_ints=[-1]))
+    if IsScalar(dim):
+        dim = op.Reshape(dim, op.Constant(value_ints=[-1]))
+
+    if IsScalar(start):
```
justinchuby, line 5758: Same with start and lengh. If they are symint, they are always scalars.

onnxscript/function_libs/torch_lib/graph_building/_graph_building_torch.py
```diff
@@ -734,6 +738,8 @@ def _add_constant_to_graph(self, constant) -> torch.Value:
             attributes=dict(value=constant_tensor),
         )[0]
         value.setDebugName(_rename_intermediate_value(value.debugName()))
+        if shape is not None:
```
justinchuby, line 741: I suggest not touching this file as it is deprecated.

justinchuby (commented):
I would suggest doing this differently. The shape info may not be needed?

titaiwangms:
> I would suggest doing this differently. The shape info may not be needed?

This may be low priority. It all starts https://github.com/microsoft/onnxscript/pull/1916 that I re-wrote aten_narrow to adapt the change that I moved YracedONNXFunction to promote python constants to tensors before function call. With doing that, IsScalar loses its functionnality because we didn't assign shape to tensors. That's why I create this PR.

titaiwangms:
Leave it as it is for now. For torchlib, we keep the implementation, and CI tests will be deprecated soon with [OpSignature PR](https://github.com/microsoft/onnxscript/pull/1838).

PR #1921, opened 2024-10-29
[Rewriter Rules] Fix Expand-Identity rule

justinchuby:
ExpandIdentity may fail if x_shape is None. This change fixes it.

- Also created parameterized tests

### DIFF
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -71,6 +71,7 @@ class ExpandIdentity(orp.RewriteRuleAsClass):
             # Shape is not a constant and cannot be guessed.
             return False
         if (x_shape := x.shape) is None:
+            # We don't know the shape of the input
             return False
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -70,8 +70,9 @@ def check(cls, context, x, shape) -> bool:
         if shape.const_value is None:
             # Shape is not a constant and cannot be guessed.
             return False
-        shape_x = x.shape
-        return shape_x.dims == tuple(shape.const_value.numpy().tolist())
+        if (x_shape := x.shape) is None:
```
justinchuby, line 73:
```suggestion
        if (x_shape := x.shape) is None:
            # We don't know the shape of the input
```

titaiwangms (approved):

justinchuby:
There are more errors in other rules

xadupre (approved):

PR #1922, opened 2024-10-29
Update opgen for it to be runnable

justinchuby:
Previously the opgen module assumes that it is installed as a package and can be found at script import. Since we are now just running it as a standalone script, I updated to imports so that it can be run directly. Added a README.


titaiwangms (approved):
Sorry, I am not really following here. It was a bug? Could you write some in description.

justinchuby:
updated, thanks

PR #1924, opened 2024-10-29
Use IR types to define onnx_types

justinchuby:
- Use IR types to define onnx_types so that it is not dependent on onnx package version.
- Also add INT4 and UINT4 types.
- Make some helper functions private.


justinchuby:
@gramalingam for approval

gramalingam (approved):

Issue #1491, opened 2024-05-02
[Rewriter] Why doesn't it match?


justinchuby:
Mechanisms for examining and understanding why a pattern was not matched.

justinchuby:
Track node name and list all the passes that does not match the node.

PR #1925, opened 2024-10-30
[rewriter] Remove redundant op.Slice and op.ScatterND

titaiwangms:
Fixes https://github.com/microsoft/onnx-converters-private/issues/270

It is observed that ExportedProgram could generates aten::slice.Tensor and aten::slice_scatter.default that slice nothing:

Screenshot 2024-10-29 112157

The slices would result in redundant op.Slice ops in ONNX graph that does nothing, and op.ScatterND that basically replaces the whole input to updates, which takes a lot of time in inference.

This rule set recognizes the redundant slices by checking if the following requirements are met:
(1) starts = 0
(2) ends >= inputs[dim].shape or ends == _INT64_MAX 
(3) steps == 1

This rule set recognizes the redundant scatterND by checking if the following requirements are met:
(1) indices has the same length as the first dim of input
(2) indices is from 0 to input.shape[0]
(3)input has the same shape as updates

Benchmark on ghostnet_100 (the original speed up was 0.0256):
|| Stat       | Speedup | Increase  | Med    |
|-------|---------------|------------|-----------|-----------|
|      Suite  | Model Name  | onnx_dynamo | onnx_dynamo | onnx_dynamo |
| Timm  | ghostnet_100  | 1.1599     | 15.986%   | 1.1580    |

### DIFF
--- a/onnxscript/rewriter/collapse_slices.py
+++ b/onnxscript/rewriter/collapse_slices.py
@@ -11,7 +11,7 @@ from onnxscript.rewriter import pattern
 logger = logging.getLogger(__name__)
 
 
-def check_if_redundant_slice(
+def _check_if_redundant_slice(
     context,
     data: ir.Value,
     starts: ir.Value,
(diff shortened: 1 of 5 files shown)


onnxscript/rewriter/collapse_slices.py
```diff
@@ -0,0 +1,114 @@
...
+    updates: ir.Value,
+    **_,
+):
+    """If the indices is the same length as the first dim of data, and the shape of updates is equal to data, we can simply swap the whole value."""
+    del context  # Reserved for future extensions
+
+    # To validate data can be replaced directly by updates, we need to check the following:
+    # 1. they have the same shape
+    if data.shape is None:
+        logger.info("The value 'data' shape is not statically known.")
```
justinchuby, line 80: side: At some point we need to pump this info in some debug run so we can show users why something is not matched. #1491

onnxscript/rewriter/collapse_slices.py
justinchuby, line 1: The file should prob be private?

onnxscript/rewriter/collapse_slices.py
```diff
@@ -0,0 +1,114 @@
...
+        logger.info("The value 'start', 'end', 'axis', 'step' is not statically known.")
+        return False
+    if steps_const.numpy().item() != 1:
+        return False
+    # starts is 0
+    if starts_const.numpy().item() != 0:
+        return False
+    # In case data.shape is not statically known, we still can tell the slice is redundant if ends is sys.maxsize
+    if ends_const.numpy().item() == sys.maxsize:
+        return True
```
justinchuby, line 41: Same for these

onnxscript/rewriter/collapse_slices.py
```diff
@@ -0,0 +1,114 @@
...
+    if starts_const is None or ends_const is None or axes_const is None or steps_const is None:
+        logger.info("The value 'start', 'end', 'axis', 'step' is not statically known.")
+        return False
+    if steps_const.numpy().item() != 1:
+        return False
+    # starts is 0
+    if starts_const.numpy().item() != 0:
+        return False
+    # In case data.shape is not statically known, we still can tell the slice is redundant if ends is sys.maxsize
+    if ends_const.numpy().item() == sys.maxsize:
```
justinchuby, line 40: sys.maxsize may not be INT64_MAX, so it is better to hard code this number

onnxscript/rewriter/collapse_slices.py
```diff
@@ -0,0 +1,114 @@
...
+        return True
+    if data.shape is None:
+        logger.info("The value 'data' shape is not statically known.")
+        return False
+    if ends_const.numpy().item() < data.shape[axes_const.numpy().item()]:
```
justinchuby, line 45: `numpy().item()` calls may need to be more robust

onnxscript/rewriter/collapse_slices.py
```diff
@@ -0,0 +1,114 @@
...
+def _identity_to_updates(op, data, indices, updates, **_):
+    return op.Identity(updates)
+
+
+def _potential_redundant_slice(op, data, starts, ends, axes, steps):
+    return op.Slice(data, starts, ends, axes, steps)
+
+
+def _potential_redundant_scatternd(op, data, indices, updates):
+    return op.ScatterND(data, indices, updates)
```
justinchuby, line 64: Create a docstring for each of these to describe when they are applicable? For example it isn't immediately clear to me when `_identity_to_updates` is applicable. (and when it is not, for that matter)

onnxscript/rewriter/collapse_slices_test.py
```diff
@@ -0,0 +1,78 @@
...
+            [
+                onnx.helper.make_tensor(
+                    "indices",
+                    onnx.TensorProto.FLOAT16,
+                    indices.shape,
+                    indices,
+                ),
+            ]
+        )
+        model = ir.serde.deserialize_model(model_proto)
```
justinchuby, line 76: Prefer doing this in the IR model I would suggest

onnxscript/rewriter/collapse_slices_test.py
```diff
@@ -0,0 +1,78 @@
...
+        count = collapse_slices.rules.apply_to_model(model)
+        self.assertEqual(count, 1)
```
justinchuby, line 78: I think we should also test the resulting model to see that it is expected. Running numerical checks is also helpful if possible.

justinchuby: On this note we should develop a better testing infra for the rewriter rules, so we can in the future call

test_rewrite_rule(model, rule)

to test the standard things.

onnxscript/rewriter/collapse_slices.py
titaiwangms: This is consumed in optimizer/__init__.py, like other rules.

justinchuby: That's ok. As long as it is not user facing it should be private. We can do a batch update separately too

titaiwangms: https://github.com/microsoft/onnxscript/issues/1927

onnxscript/rewriter/collapse_slices.py
titaiwangms: Done

onnxscript/rewriter/collapse_slices_test.py
titaiwangms: Added a general test for numeric.

onnxscript/rewriter/testing.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Any
+
+import numpy as np
+import onnx
+import onnxruntime as ort
+
+
+def assert_numerically_equal(
+    original_model_proto: onnx.ModelProto,
+    the_rewritten_model_proto: onnx.ModelProto,
```
justinchuby, line 13:
```suggestion
    rewritten_model_proto: onnx.ModelProto,
```

onnxscript/rewriter/testing.py
```diff
@@ -0,0 +1,68 @@
...
+    args: tuple[Any, ...],
+    rtol: float = 1,
+    atol: float = 1e-3,
+):
+    """Assert that the two models are numerically equal.
+
+    Args:
+        original_model_proto (onnx.ModelProto): The original model proto.
+        the_rewritten_model_proto (onnx.ModelProto): The rewritten by the rules model proto.
```
justinchuby, line 22:
```suggestion
        original_model_proto: The original model proto.
        rewritten_model_proto: The rewritten by the rules model proto.
```

onnxscript/rewriter/collapse_slices_test.py
```diff
@@ -0,0 +1,100 @@
...
+        original_model_proto = ir.serde.serialize_model(model)
+
+        count = collapse_slices.rules.apply_to_model(model)
+        self.assertEqual(count, 1)
+        self.assertEqual(len(model.graph), 1)
+        self.assertIn("Identity", [node.op_type for node in model.graph])
+
+        input = np.random.rand(112, 16, 512).astype(np.float32)
+        testing.assert_numerically_equal(
+            original_model_proto, ir.serde.serialize_model(model), (input, input)
```
justinchuby, line 99: Add support for taking ir.Model, so callers do not have to serialize?

justinchuby (approved):

onnxscript/rewriter/collapse_slices_test.py
titaiwangms: Done

PR #1926, opened 2024-10-31
[WIP] Add framework for version converter API

shubhambhokare1:


onnxscript/version_converter/version_converter_test.py
```diff
@@ -0,0 +1,138 @@
...
+from __future__ import annotations
+
+import unittest
+
+import onnx.checker
+import onnx.parser
+import onnx.shape_inference
+
+from onnxscript import ir
+from onnxscript.version_converter import convert_version
```
justinchuby, line 12: nit: import modules only

onnxscript/version_converter/version_converter_test.py
```diff
@@ -0,0 +1,138 @@
...
+                reshape_y = Reshape (input_x, shape_b)
+                gridsample = GridSample <mode = "bilinear"> (reshape_x, reshape_y)
+                shape_c = Constant<value: tensor = int64[4] {4, 1024, 1024}>()
+                output = Reshape (gridsample, shape_c)
+            }
+        """
+        )
+        model = ir.serde.deserialize_model(model_proto)
+        self.assertEqual(model.graph._nodes[4].op_type, "GridSample")
+        self.assertEqual(model.graph._nodes[4]._attributes['mode'].value, 'bilinear')
```
justinchuby, line 90:
```suggestion
        self.assertEqual(model.graph[4].attributes['mode'].value, 'bilinear')
```

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+from __future__ import annotations
+
+from onnxscript import ir
+
+### Adapters
+
+# Compatibility Adapter
+def adapter_compatible(op: ir.Node, target_opset):
```
gramalingam, line 10: nit: I recommend `node` instead of `op`. We frequently use `op` for building a node via the `op.MatMul(x, y)` syntax.

gramalingam: Furthermore, the general interface for a version-converter would likely need an op/node builder as an input, and that would be best called `op`. We would need that when creating new nodes as part of the version-conversion (eg., even in the example below to create a `Constant` node from an attribute value.

onnxscript/version_converter/_adapter_lib.py
```diff
@@ -0,0 +1,75 @@
...
+from onnxscript.version_converter.version_converter import (
+    adapter_compatible,
+    adapter_axis_attr_to_input,
+    adapter_gridsample_19_20,
+    adapter_groupnormalization_20_21,
+)
+
+
+_ADAPTERS_18_19 = {
+    "Equal": adapter_compatible,
```
gramalingam, line 14: Minor nit: I think it would be better to not have to register an adapter for compatible extensions, since it is basically an identity operation.

I can see a value in explicitly documenting that a particular opset-update is backwards-compatible (to catch the case where we forget to register an adapter). We could do that using a separate set of all compatible ops.

onnxscript/version_converter/version_converter_test.py
```diff
@@ -0,0 +1,138 @@
...
+                reshape_y = Reshape (input_y, shape_b)
+                matmul = MatMul (reshape_x, reshape_y)
+                shape_c = Constant<value: tensor = int64[4] {1, 4, 512, 64}>()
+                output = Reshape (matmul, shape_c)
+            }
+        """
+        )
+        model = ir.serde.deserialize_model(model_proto)
+        target_version = 19
+        new_model = convert_version(model, target_version=target_version)
```
gramalingam, line 34: Where is `convert_version` defined? I couldn`t find it. Anyway: I don't think it should return a model, since it will update the input model destructively.

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,44 @@
...
+    op.version = target_opset
```
gramalingam, line 11: Setting the version on a node need not be part of the adapter logic. It should be in the converter logic that calls the adapter, since this logic is the same for all adapters. No point in duplicating this line in every adapter. That would also make a compatible adapter an identity function, and we don't even need to register one for compatible extensions,

onnxscript/version_converter/_adapter_lib.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+_ADAPTERS_20_21 = {
+    "Cast": adapter_compatible,
+    "CastLike": adapter_compatible,
+    "Constant": adapter_compatible,
+    "ConstantOfShape": adapter_compatible,
+    "DequantizeLinear": adapter_compatible,
+    "Flatten": adapter_compatible,
+    "GroupNormalization": [adapter_compatible, adapter_groupnormalization_20_21],
```
gramalingam, line 50: We shouldn't need a list of adapters here.

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+# Adapter to convert axis from an attribute to a node input
+def adapter_axis_attr_to_input(op: ir.Node, target_opset):
```
gramalingam, line 15: I suggest parameterizing this over the attribute name to make it reusable as below:
```py
def change_attr_to_input(attr_name: str) -> Adapter:
    def adapt (node: ir.Node):
        ...
    return adapt
```
This will let us register `change_attr_to_input("axis")` as an adapter, but also use it for other attributes as needed

gramalingam:
One of the main design question relates to the "adapter" signature: what form should it take? Essentially it is a function that takes a single node as a parameter, and modifies it in some form. The changes are typically a simple mutation of a node along with potentially other changes (such as the insertion of extra nodes). 

For now, I think it might be fine to follow the pattern used in the optimizer and rewriter, which are based on node-transformers that, given an input node, return a sequence of replacement nodes or None (if no replacement is required). This allows a simple loop over all nodes in the graph that transforms each node in sequence. This can be generalized later if necessary.

onnxscript/version_converter/_adapter_lib.py
```diff
@@ -0,0 +1,162 @@
...
+def adapter_attr_to_input(attr_name: str):
+    def adapt(node: ir.Node):
+        assert attr_name in node._attributes
+        _attr = node._attributes.pop(attr_name)
+        # Create an ir.Value obj with properties of the axis attribute
+        attr_as_input = ir.Value(
+            name=_attr.name,
+        )
+        # Add the ir.Value as inputs to the node and graph
+        node._inputs = node._inputs + (attr_as_input,)
```
justinchuby, line 22: note: avoid using and modifying private fields. To change inputs to a node, always initialize a new node to replace the current one.

onnxscript/version_converter/version_converter_test.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+        nodes = model.graph._nodes
+
+        self.assertEqual(nodes[0].op_type, "Constant")
+        self.assertEqual(nodes[0].version, 19)
+        self.assertEqual(nodes[1].op_type, "Reshape")
+        self.assertEqual(nodes[1].version, 19)
+        self.assertEqual(nodes[4].op_type, "GridSample")
+        self.assertEqual(nodes[4].version, 20)
+        self.assertEqual(model.graph._nodes[4]._attributes["mode"].value, "cubic")
```
justinchuby, line 179: note: Avoid accessing internal fields

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,55 @@
...
+        # Updating each node based on the correct adapter [ver->ver+1]
+        for opset_version in range(self.model_version, target_version):
+            for node in graph:
+                adapter_set = pick_adapter_set(opset_version)
+                if node.op_type in adapter_set:
+                    adapter_func = adapter_set[node.op_type]
+                    if adapter_func is not None:
+                        adapter_func(node)
+                    node.version = opset_version + 1
+                # Apply custom adapters
```
gramalingam, line 39: I would not bother with custom adapters to begin with. I don't see a need for it right now. But there are other features (will comment in other places as needed) that are far more important. It's better to not expose extra functionality for now (until we see a good reason to have it).

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, model: ir.Model) -> None:
+        self._functions = model.functions
+        self.model_version = model.opset_imports.get("")
+        self.custom_adapters = {}
+
+    def _register_custom_adapters(self, custom_adapter_list) -> None:
+        self.custom_adapters = custom_adapter_list
+
+    def graph_version_convert(self, graph: ir.Graph, target_version: int) -> None:
+        if self.model_version == target_version:
```
gramalingam, line 19: (Extension) I think we will need to soon support the case where the incoming model has nodes with different opset versions. At that point, such checks should happen at the node level, not at the model level.

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # Iterate through all nodes in the graph
+        # Set node.version as the model_version
+        # TODO: Do this internally in IR?
+        for node in graph:
+            node.version = self.model_version
+
+        # Iterate from current model version -> target version
+        # Updating each node based on the correct adapter [ver->ver+1]
+        for opset_version in range(self.model_version, target_version):
```
gramalingam, line 31: May be better to explicitly check for target_version being > or < current version. (It's ok to focus on up-conversion in first version, but may be better to have an error/warning message if we run into down-conversion when it is unimplemented.)

gramalingam: Actually, this could be bundled into a check after `pick_adapter_set` to more generally handle the case when we don't have an adapter set (for either 23 to 24 or for 18 to 17).

onnxscript/version_converter/_adapter_lib.py
gramalingam: Further, we can't just create a value as above ... we need to create a constant value with the given value. For now, the simplest way is to create a new `Constant` node. (Orthogonally to this, we should extend the "builder" API we currently use to create initializers as well, but that's a separate issue, for now Constant nodes should be fine.)

justinchuby:
Questions that are related to the design doc that comes to mind

1. How do we ensure down conversion can be supported in the future
2. What invariants do we preserve in the nodes; an opset version may or may not be associated with an `ir.Node`. How are both cases handled
3. How does the design relate to the rewriter? Conceptually version conversion is a model rewriting process. How do we plan to maintain a consistent dev/user experience when authoring and debugging subgraph replacement logic?
4. When version conversion fails, how should it fail? Succeed partially, abort, etc.? What are the guarantees/invariants of the model state when conversion is not possible?

onnxscript/version_converter/version_converter.py
shubhambhokare1: Removed

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+
+@dataclasses.dataclass
+class Replacement:
+    """A replacement for a node in the graph."""
+
+    new_outputs: Sequence[ir.Value]
+    new_nodes: Sequence[ir.Node]
+
+
+# A partial-evaluator function takes a node, a RewriterContext, OptimizerState and returns
```
gramalingam, line 140: Comments need to be updated to current usage

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+
+@dataclasses.dataclass
+class AdapterVersionChecker:
+    """A class that represents a version checker for a particular op.
+
+    It is applicable for a specific version upgrade (orignal_version -> target_version) of the op.
+    """
+
+    node_version: int | None
+    upgrade_version: int | None
```
gramalingam, line 157: minor nit: I am not sure if we need two ints. In future, we will likely add down-conversion, but I think we just need up or down conversion by one? Eg., 18 to 19 as well as 18 to 17. So, probably, a version-number (18) and a boolean (to indicate if this is for up-conversion or down-conversion) might be enough.

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+    def __init__(self):
+        self.op_adapters: dict[tuple[str, str], list[AdapterVersionChecker]] = {}
+
+    def lookup_adapters(
+        self, domain: str, opname: str, original_version: int, target_version: int
+    ):
+        adapter_list = self.op_adapters.get((domain, opname), [])
+        return [
+            adapter.function
+            for adapter in adapter_list
```
gramalingam, line 182: I wonder if we really need to support a list of adapters (eg., for a given op from 18 to 19)? I understand that the constant-folding optimizer does this, but wondering if we need it here.

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+    """A class that represents a version checker for a particular op.
+
+    It is applicable for a specific version upgrade (orignal_version -> target_version) of the op.
+    """
+
+    node_version: int | None
+    upgrade_version: int | None
+    function: AdapterFunction
+
+    def valid_for(self, opname: str, original_version: int, target_version: int) -> bool:
```
gramalingam, line 160: I don't understand this function. Why is it not checking against self.node_version and self.upgrade_version here?

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+    "QuantizeLinear",
+    "Reshape",
+    "Scan",
+    "Shape",
+    "Size",
+}
+
+
+_ADAPTERS_19_20 = {
+    "DFT",
```
gramalingam, line 38: Ideally, we should not have to add this here and have the separate `@register` down below, just one will do. I think you should be able to drop these here for anything registered down below. (See other related comments below.)

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+# a Replacement for the node or None (if no replacement is needed). It may also return just
+# the ir.Value or ir.Values to replace the output values of the node, when the new nodes
+# can be inferred from the RewriterContext used to build the new nodes.
+
+ReturnValue = Union[Replacement, Sequence[ir.Value], ir.Value, None]
+AdapterFunction = Callable[[ir.Node, orp.RewriterContext], ReturnValue]
+
+
+@dataclasses.dataclass
+class AdapterVersionChecker:
```
gramalingam, line 150: nit: `VersionAdapter` is a better name. This represents not just a checker, but the adapter, along with version metadata.

justinchuby:
An op can stay the same for many versions. For example, Acosh-9 is the same all the way to 21. Would it make sense to mark the adapter for it to support the "base to next" version (e.g. 9->22) so that it is clear on the range of opsets it supports? IMO this is a clearer notation than what the ONNX version converter uses, which is `21->22` and can be confusing, because in this case Acosh-21 is implicitly defined by Acosh-9.

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+
+## Op-specific adapters
+
+# Opset 19 -> 20
+
+
+@register("DFT", node_version=19, upgrade_version=20)
+def dft_19_20(node: ir.Node, op):
+    input = node.inputs[0]
+    inverse = _get_int_attribute(node, "inverse", 0)
```
gramalingam, line 263: There is a complication here, relating to handling of functions. In functions, we can have attribute-parameters, which mean we don't know the value of the attribute. The version-conversion, as written, will be wrong in such cases.

For now, I think it is okay to inline all functions (see relevant comment below) and ignore this issue.

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+            if tuple((opset_version, opset_version + 1)) not in _ADAPTER_SETS:
+                logger.warning(
+                    "Conversion from opset: %s to target opset: %s not currently supported.",
+                    opset_version,
+                    opset_version + 1,
+                )
+                return
+
+            self.visit_graph(model.graph, opset_version)
+            for function in model.functions.values():
```
gramalingam, line 420: As explained in earlier comment, I think it is better to skip functions (they won't be handled correctly). Instead, we should call the inliner to inline all functions as the first step of version_conversion (probably in the caller, before line 426 below).

onnxscript/version_converter/version_converter.py
```diff
@@ -1,55 +1,427 @@
...
+                self.visit_function(function, opset_version)
+
+
+def version_convert(model: ir.Model, target_version: int) -> None:
     """Convert the model to the specified ONNX opset version."""
-    version_converter = _VersionConverter(model)
-    version_converter._register_custom_adapters(custom_adapters)
-    version_converter.graph_version_convert(model.graph, target_version)
+    version_converter = _VersionConverter(target_version=target_version)
```
gramalingam, line 426: As explained earlier: let us call the inliner here.

onnxscript/version_converter/version_converter.py
justinchuby: > they won't be handled correctly

Curious why that would be the case?

onnxscript/version_converter/version_converter.py
justinchuby: Would it be possible to just promote the attribute to inputs and use them in the upgraded op?

onnxscript/version_converter/version_converter.py
shubhambhokare1: Removed

onnxscript/version_converter/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+from __future__ import annotations
+
+__all__ = [
+    # Functions
+    "convert_version",
+    "inline",
```
justinchuby, line 8:
```suggestion
```

inline is not part of the converter

onnxscript/version_converter/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+]
+
+from onnxscript import ir
+from onnxscript.optimizer._inliner import inline
```
justinchuby, line 12: Import modules only

```suggestion
from onnxscript.optimizer import _inliner
```

onnxscript/version_converter/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from onnxscript.version_converter.version_converter import version_convert
```
justinchuby, line 13: nit: Perfer calling it convert_version. Prefer using a verb as a function name

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+# Licensed under the MIT License.
+"""Convert the model to the specified ONNX opset version."""
+
+from __future__ import annotations
+
+import dataclasses
+import logging
+from typing import Callable, Sequence, Union
+
+import onnxscript.ir._convenience as _convenience
```
justinchuby, line 11:
```suggestion
import onnxscript.ir._convenience as ir_convenience
```

conventional naming

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+    """A replacement for a node in the graph."""
+
+    new_outputs: Sequence[ir.Value]
+    new_nodes: Sequence[ir.Node]
+
+
+# A version-adapter function takes a node, a RewriterContext and returns
+# a Replacement for the node or None (if no replacement is needed).
+
+ReturnValue = Union[Replacement, Sequence[ir.Value], ir.Value, None]
```
justinchuby, line 32: When would a function return a Replacement over a Value?

justinchuby, line 32:
```suggestion
_ReturnValue = Union[Replacement, Sequence[ir.Value], ir.Value, None]
```

make private. Everything not in `__all__` should be private

onnxscript/version_converter/version_converter.py
justinchuby, line 1: This module can be private

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+    def __init__(self):
+        self.op_adapters: dict[tuple[str, str, int, bool], VersionAdapter] = {}
+
+    def lookup_adapters(
+        self,
+        domain: str,
+        opname: str,
+        original_version: int,
+        up_conversion: bool = True,
+    ) -> Union[AdapterFunction, None]:
```
justinchuby, line 61:
```suggestion
    ) -> AdapterFunction | None:
```

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+        adapter = self.op_adapters.get((domain, opname, original_version, up_conversion), None)
```
justinchuby, line 62:
```suggestion
        adapter = self.op_adapters.get((domain, opname, original_version, up_conversion))
```

default is None

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+        if adapter is not None:
+            return adapter.function
+        else:
+            return None
```
justinchuby, line 66:
```suggestion
        return None
```

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def register(
```
justinchuby, line 68: Add a docstring?

justinchuby, line 68: Decorate with functools.wraps so the typing information is preserved

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+        self, opname: str, domain: str = "", node_version=None, up_conversion=True
+    ) -> Callable[[AdapterFunction], AdapterFunction]:
+        def decorator(function: AdapterFunction) -> AdapterFunction:
```
justinchuby, line 71: Decorate with functools.wraps so the typing information is preserved

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+register = registry.register
+
+
+def _get_input(node: ir.Node, index: int) -> ir.Value | None:
+    if index < len(node.inputs):
+        return node.inputs[index]
+    return None
+
+
+def _get_output(node: ir.Node, index: int) -> ir.Value | None:
```
justinchuby, line 91: Was this used?

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+    return None
+
+
+class _VersionConverter:
+    opset_imports: dict[str, int]
+
+    def __init__(self, target_version: int):
+        self.target_version = target_version
+
+    def process_node(self, node: ir.Node, opset_version, up_conversion: bool = True):
```
justinchuby, line 218:
```suggestion
    def process_node(self, node: ir.Node, opset_version: int, up_conversion: bool = True):
```

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+        if node.domain not in self.opset_imports:
+            return None
```
justinchuby, line 220: Create a test to cover this line?

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+        adapter = registry.lookup_adapters(
+            node.domain, node.op_type, opset_version, up_conversion
+        )
+        if adapter is None:
+            return None
+        context = orp.RewriterContext()
+        output = adapter(node, context)
+        if output is not None:
+            if isinstance(output, Replacement):
+                return output
```
justinchuby, line 230: Create a test to cover this line?

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+            if isinstance(output, ir.Value):
+                output = [output]
+            return Replacement(output, context.nodes)
+
+    def replace_node(self, node: ir.Node, replacement, root: ir.Graph | ir.Function):
```
justinchuby, line 235:
```suggestion
    def replace_node(self, node: ir.Node, replacement, root: ir.Graph | ir.Function) -> None:
```

onnxscript/version_converter/version_converter.py
```diff
@@ -0,0 +1,310 @@
...
+# Opset 20 -> 21
+
+
+@register("GroupNormalization", node_version=20, up_conversion=True)
+def groupnormalization_20_21(node: ir.Node, op):
+    x = _get_input(node, 0)
+    scale = _get_input(node, 1)
+    bias = _get_input(node, 2)
+    if x is None or scale is None or bias is None:
+        return None
```
justinchuby, line 169: Some branches are missing coverage

onnxscript/version_converter/version_converter.py
shubhambhokare1: No, removed for now

onnxscript/version_converter/version_converter.py
shubhambhokare1: Would be for ops with control flow, but we dont seem to have any adapters using that, so removing for now.

onnxscript/version_converter/version_converter.py
shubhambhokare1: Will cover all error handling cases with a future PR

onnxscript/version_converter/_version_converter.py
```diff
@@ -19,6 +19,10 @@
 CURRENT_MAX_ONNX_OPSET = 23
 
 
+class VersionConverterError(Exception):
```
justinchuby, line 22: Prefer a RuntimeError. It is easier to catch

```suggestion
class VersionConverterError(RuntimeError):
```

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -17,10 +17,12 @@
 import pathlib
 from typing import Callable
 
-from onnxscript import ir, optimizer
+from onnxscript import ir, optimizer, version_converter
 from onnxscript.function_libs.torch_lib import registration
 from onnxscript.ir import _external_data
 
+ENABLE_VERSION_CONVERTER = 0
```
justinchuby, line 24: Prefer allowing an env var for this. See the `optimize` api

onnxscript/version_converter/_version_converter.py
```diff
@@ -0,0 +1,337 @@
...
+                    logger.warning(
+                        "Skipping version conversion for node %s due to exception: %s",
+                        node.op_type,
+                        e,
+                    )
+        return None
+
+    def visit_model(self, model: ir.Model) -> None:
+        self.opset_imports = model.opset_imports
+        for opset_import in self.opset_imports:
```
gramalingam, line 310: nit: you don't need an iteration. You can just use `get`:
```py
   model_version = model.opset_imports.get("")
   if model_version is None:
      model_version = model.opset_imports.get("ai.onnx")
      if model_version is None:
         return
```

onnxscript/version_converter/_version_converter.py
```diff
@@ -0,0 +1,337 @@
...
+    def visit_graph(self, graph: ir.Graph) -> None:
+        for node in graph:
+            if node.version is None:
+                node.version = self.model_version
+            # Iterate each node from current node version -> target version
+            # and updating node based on the correct adapter
+            # Up-conversion [ver->ver+1] or down-conversion [ver->ver-1]
+            for opset_version in range(node.version, self.target_version):
+                up_conversion = True
+                if self.target_version < opset_version:
```
gramalingam, line 288: Better to move the check outside the for loop (compare node.version and self.target_version).

onnxscript/version_converter/_version_converter.py
```diff
@@ -0,0 +1,337 @@
...
+                    up_conversion = False
+                if up_conversion is True and opset_version == CURRENT_MAX_ONNX_OPSET:
```
gramalingam, line 290: This check can also be done outside the loop.

onnxscript/version_converter/_version_converter.py
```diff
@@ -0,0 +1,337 @@
...
+    def _upgrade_version(self, node: ir.Node, opset_version: int, up_conversion: bool) -> None:
+        if up_conversion is True:
+            node.version = opset_version + 1
+        else:
+            node.version = opset_version - 1
+
+    def process_node(
+        self, node: ir.Node, opset_version: int, up_conversion: bool = True
+    ) -> Replacement | None:
+        if node.domain not in self.opset_imports:
```
gramalingam, line 232: This check should be `if node.domain not in {"", "ai.onnx"}`

onnxscript/version_converter/_version_converter.py
```diff
@@ -0,0 +1,337 @@
...
+    def lookup_adapters(
+        self,
+        domain: str,
+        opname: str,
+        original_version: int,
+        up_conversion: bool = True,
+    ) -> AdapterFunction | None:
+        adapter = self.op_adapters.get((domain, opname, original_version, up_conversion))
+        if adapter is not None:
+            return adapter.function
```
gramalingam, line 69: This is the only place where the adapter is used, and we just return adapter.function. So, do we really need the `VersionAdapter` class? The map can just store the "function" directly, which will also get rid the extra 3 lines here.

justinchuby:
Minor but should be quick change for consistency

```diff
- TODO (shubhambhokare1) :
+ TODO(shubhambhokare1):
```

justinchuby (approved):
Very nice to have a version converter. Thanks! 🎉

Issue #1499, opened 2024-05-05
[IR] Tensor todos


justinchuby:
- [x] INT4 etc. for external tensors
- [x] Implement dlpack on all tensors, add it to the protocol
- [ ] ~Make ArrayComparible non-runtime-checkable~
- [ ] Create an example of the external tensor
- [x] #1548
- [x] Fix typos on the tensor documentation
- [x] Implement compatible adapters for torch tensors, jax.Array, Tensorflow tensor, ORT Value (https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.OrtValue), cupy, openvino, safetensors

PR #1930, opened 2024-11-01
[IR] Create tensor adapters

justinchuby:
Create the new public module `ir.tensor_adapters` to implement `ir.TensorProtocol` for popular frameworks so users do not have to do that again.

Next PRs will include safetensors support.

#1499

### DIFF
--- a/onnxscript/ir/tensor_adapters.py
+++ b/onnxscript/ir/tensor_adapters.py
@@ -1,0 +1,87 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Compatible adapters implementing the TensorProtocol interface for various framework tensor types."""
+
+# NOTE: DO NOT import any framework-specific modules here in the global namespace.
+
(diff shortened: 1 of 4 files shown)


onnxscript/ir/tensor_adapters.py
```diff
@@ -0,0 +1,117 @@
...
+
+Example::
+    import torch
+    from onnxscript import ir
+
+    # Create a PyTorch tensor
+    torch_tensor = torch.tensor([1, 2, 3])
+
+    # Wrap the PyTorch tensor in a TorchTensor object
+    ir_tensor = ir.tensor_adapters.TorchTensor(torch_tensor)
```
xadupre, line 18: Should we have one function which selects the corresponding adapters based on the input type?

justinchuby: We could - that can be a higher level function or be built into the `ir.tensor` function

gramalingam (approved):

PR #1931, opened 2024-11-04
Fix CI

titaiwangms:
The from future import annotation in onnxscript/rewriter/testing.py was not set


titaiwangms:
@justinchuby

justinchuby (approved):
Thanks!

PR #1933, opened 2024-11-05
[IR] Support safetensors in tensor adapters

justinchuby:
Load safetensors as numpy arrays. It is actually incredibly easy and we are able to reuse the `ir.Tensor` implementation because the numpy array loaded by safetensors is memory mapped. We can use it as a normal numpy array without memory constraints.

## TODO

- Add tests
- Verify bfloat16/fp8 support


justinchuby:
Supported in https://github.com/justinchuby/onnx-safetensors instead

PR #1937, opened 2024-11-08
A couple of optimizer and rewriter extensions

gramalingam:
A few extensions motivated by ongoing transformer fusion optimizations:

Pattern matching:
* Extend pattern-matching pattern to allow specifying that extra-inputs are allowed.

Optimizations:
* Concat (x) can be replaced by Identity(x)
* Redundant cast optimization was missing in core optimizer (though present as a llama rewrite rule).
* Dropout optimizations moved into core optimizer (from rewrite rule; rewrite rule has an issue, use of attribute instead of input, and it seemed better to move it into core optimizer).

In general, for optimizations involving a single node, the core optimizer is a better place (at least, as long as they are generic, and not backend-specific) than rewrite rules. It is more efficient.

* Fix input/output size limit of constant-folding to be number of bytes. (It is currently inconsistent, as number of bytes for one and number of elements for another).

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -255,8 +255,8 @@ class OpPatternBuilder:
             inputs,
             attributes,
             _outputs,
-            _allow_other_attributes,
-            _allow_other_inputs,
+            allow_other_attributes=_allow_other_attributes,
+            allow_other_inputs=_allow_other_inputs,
         )
         self.pattern_builder.add_node(node_pattern)
         output_values = node_pattern.outputs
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -557,7 +569,13 @@ def clone(self, node_map: dict[NodePattern, NodePattern], swap: bool) -> NodePat
...
         outputs = [value.name for value in self.outputs]
         copied = NodePattern(
-            self.domain, self.op, inputs, self.attributes, outputs, self.allow_other_attributes
+            self.domain,
+            self.op,
+            inputs,
+            self.attributes,
+            outputs,
+            self.allow_other_attributes,
+            self.allow_other_inputs,
```
justinchuby, line 578: I would call by keyword because these are all boolean options

onnxscript/rewriter/pattern.py
```diff
@@ -472,15 +479,20 @@ def __init__(
         attributes: dict[str, AttrPattern],
         outputs: Sequence[str | None],
         allow_other_attributes: bool | None,
+        _allow_other_inputs: bool | None,
```
justinchuby, line 482: Is there a significance in prepending `_` for this arg?

gramalingam: No, thanks for catching this.

onnxscript/rewriter/pattern.py
```diff
@@ -225,6 +225,7 @@ def __call__(
         _version: int | None = None,
```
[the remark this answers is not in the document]
gramalingam: But doesn't the `*args` make these keyword-only?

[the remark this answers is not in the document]
justinchuby: It does. I missed that sorry

onnxscript/rewriter/pattern.py
```diff
@@ -472,15 +479,20 @@ def __init__(
         attributes: dict[str, AttrPattern],
         outputs: Sequence[str | None],
```
justinchuby, line 480:
```suggestion
        outputs: Sequence[str | None],
        *,
```

Consider making these options kw only for robustness

PR #1940, opened 2024-11-12
[IR] Create getters for the Attr class

justinchuby:
Create well typed getters for the Attr class

Previously, users need to self assert the values to be the correct type. From this PR on they can do

```py
a = attr.as_int()
b = a + 1
```

etc. with a being a well typed object


titaiwangms (approved):
PyTorch side needs an update too

PR #1941, opened 2024-11-13
[IR] Implement register_initializer

justinchuby:


onnxscript/ir/_core.py
```diff
@@ -1824,6 +1824,42 @@ def outputs(self) -> list[Value]:
...
+
+        Raises:
+            ValueError: If a value of the same name that is not this value
+                is already registered.
+            ValueError: If the value does not have a name.
+            ValueError: If the initializer is produced by a node.
+            ValueError: If the value does not have its ``.const_value`` set.
+        """
+        if value.name in self._initializers:
+            if self._initializers[value.name] is not value:
```
gramalingam, line 1845: * Not necessarily in this PR, but would be nice to have a method to reuse initializers, especially for common duplicated constants, like 0, 1 or even shapes like (batchsize, seqlen, hiddensize).
* On a related note, for very small tensors, we could replace the `is` comparison above with a value-equality check. May not be that important if we add a separate utility to reuse initializers as above, but if not this could serve so approximately.

onnxscript/ir/_core.py
```diff
@@ -1824,6 +1824,42 @@ def outputs(self) -> list[Value]:
...
+                raise ValueError(
+                    f"Initializer '{value.name}' is already registered, but"
+                    " it is not the same object: existing={self._initializers[value.name]!r},"
+                    f" new={value!r}"
+                )
+        if not value.name:
+            raise ValueError(f"Initializer must have a name: {value!r}")
+        if value.producer() is not None:
```
gramalingam, line 1853: I wonder if we should support promotion of a `Constant` node to an initializer (later on perhaps; could be an option in this method or a separate one).

gramalingam (approved):

titaiwangms (approved):
Should we update this usage to exporter in terms of robustness?

onnxscript/ir/_core.py
justinchuby: I think that can be a graph pass?

onnxscript/ir/_core.py
justinchuby: I think this can be a graph pass too

justinchuby:
> Should we update this usage to exporter in terms of robustness?

Sure, we can do that

onnxscript/ir/_core.py
justinchuby: https://github.com/microsoft/onnxscript/issues/1942

onnxscript/ir/_core.py
justinchuby: https://github.com/microsoft/onnxscript/issues/1943

PR #1944, opened 2024-11-14
Handle input initializers correctly in constant folding

gramalingam:
Values that are both inputs and initializers of a model/graph should not be treated as constants (and cannot be used for constant-folding). Unfortunately, the single `const_value` field is class Value is used both to indicate constant-values of proper constants as well as initializer values of initializers. Ideally, the IR should provide an easy way to distinguish this at the value level (with either an extra boolean flag to indicate the value is an input-value or by using distinct fields for "initializer_value" and "const_value".

Meanwhile, this PR introduces a workaround to handle the main issue.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -137,6 +137,7 @@ class Replacement:
 class OptimizerState:
     def __init__(self):
         self._sym_value_map: dict[ir.Value, Any] = {}
+        self._initializer_inputs: list[set[ir.Value]] = []
 
     def get_sym_value(self, value: ir.Value | None) -> Any:
         if value is None:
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -817,9 +834,16 @@ def visit_node(self, node: ir.Node, root: ir.Graph | ir.Function):
             self.replace_node(node, replacement, root)
 
     def visit_graph(self, graph: ir.Graph) -> None:
+        self._state.push_initializer_inputs()
+        for input in graph.inputs:
+            if input.const_value is not None:
+                self._state.add_initializer_input(input)
```
justinchuby, line 840: Possible to check the graph.initializers.values() field for initializers?

gramalingam: I don't see the value in doing that. This is more efficient. And it should be equivalent currently. Further, the real goal here is to stop treating `.const_value` as a constant-value if it is a graph input (so this condition reflects exactly the intended purpose). (Actually, we can omit the if-condition completely, but this is an optimization to track just inputs with a const-value instead of all inputs.)

justinchuby: Oh I see. initializers will have non-input values. Makes sense

onnxscript/optimizer/_constant_folding.py
```diff
@@ -817,9 +834,16 @@ def visit_node(self, node: ir.Node, root: ir.Graph | ir.Function):
             self.replace_node(node, replacement, root)
 
     def visit_graph(self, graph: ir.Graph) -> None:
+        self._state.push_initializer_inputs()
```
justinchuby, line 837: Helpful to add the PR description here as a comment?

justinchuby (approved):

PR #1945, opened 2024-11-14
A couple of bug fixes

gramalingam:
Fixes a couple of bugs that show up in GPT2 optimization.

### DIFF
--- a/onnxscript/optimizer/_inliner.py
+++ b/onnxscript/optimizer/_inliner.py
@@ -236,10 +236,9 @@ class _Inliner:
 
         # Identify call-stack for node, used to generate unique names.
         call_stack = self.node_context.get(node, [])
+        new_call_stack = [*call_stack, call_site_id]
 
-        cloner = _CopyReplace(
-            self, attributes, value_map, node.metadata_props, call_stack + [call_site_id]
-        )
+        cloner = _CopyReplace(self, attributes, value_map, node.metadata_props, new_call_stack)
 
         # iterate over the nodes in the function, creating a copy of each node
         # and replacing inputs with the corresponding values in the value map.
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_inliner.py
```diff
@@ -236,9 +236,8 @@ def _instantiate_call(self, node: ir.Node, call_site_id: CallSiteId) -> NodeRepl
 
         # Identify call-stack for node, used to generate unique names.
         call_stack = self.node_context.get(node, [])
-        call_stack.append(call_site_id)
 
-        cloner = _CopyReplace(self, attributes, value_map, node.metadata_props, call_stack)
+        cloner = _CopyReplace(self, attributes, value_map, node.metadata_props, call_stack + [call_site_id])
```
justinchuby, line 240:
```suggestion
        cloner = _CopyReplace(self, attributes, value_map, node.metadata_props, [*call_stack, call_site_id])
```

justinchuby (approved):

onnxscript/optimizer/_inliner.py
gramalingam: Is there a way to suppress/ignore this suggestion from ruff? It doesn't make sense to me. I don't see how the replacement can be more efficient? But it is clearly less readable (my personal opinion).

gramalingam: I went ahead and changed it so that this can be merged, but it would be good to know if there is a mechanism to override and ignore such a warning. I think efficiency argument might hold true for more complicated concatenations involving more than a single concetnation, but don't see it making a difference in a single concatenation like here.

justinchuby: To ignore, you can use `# noqa: RUF005`  or whatever ruff error code it is.

Personally I like the unpack syntax better. It seems clear especially when we are unpacking multiple Sequences. It also make it immediately clear that the argument is a list, instead of the reader having to know that `call_stack` is a list and thus `list + list` is support. This is also useful when, say, `call_stack` is an arbitrary sequence and may not have `+` defined and we can use the same syntax without having to ensure `call_stack` being a list.

Efficiency wise, this saves an additional list construction on `[call_site_id]`, although practically the overhead should be trivial.

PR #1938, opened 2024-11-09
[DRAFT] First version of fusion optimizations for transformers

gramalingam:
* Introduce fusion rules for SdpaAttention, RMS Normalization, Skip Normalization, Rotary Embedding, and Multi Head Attention
* Replace Expand by Identity when applicable (in core optimization)
* Cleanup Dropout Identity replacement in the case when Dropout has mask output
* Make repeated (redundant) call to inliner efficient

Still TODO: 
* Multi Head Attention requires extra validation conditions
* Need to cleanup use of "local" sub-patterns

PR #1947, opened 2024-11-15
A couple of optimizations and refinements

gramalingam:
Extract the independent optimization/refinements from [the fusion PR](https://github.com/microsoft/onnxscript/pull/1938) as a separate PR, ready to be reviewed/merged. (The fusion work is still WIP.)

* Replace Expand by Identity when applicable (in core optimization)
* Cleanup Dropout Identity replacement in the case when Dropout has mask output
* Make repeated (redundant) call to inliner efficient

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -13,6 +13,7 @@ from typing import Any, Callable, Iterable, Sequence, Union
 
 import numpy as np
 import onnx
+import onnx.helper
 import onnx.reference.ops
 
 import onnxscript.ir as ir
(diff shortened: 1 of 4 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -13,6 +13,7 @@
 
 import numpy as np
 import onnx
+import onnx.helper
```
justinchuby, line 16:
```suggestion
```

ditto

justinchuby:
```
>>> ir.tensor([True])
Tensor<BOOL,[1]>(array([ True]), name=None)
```

This works for me?

also vscode doesn't seem to complain:

image

gramalingam:
> ```
> >>> ir.tensor([True])
> Tensor<BOOL,[1]>(array([ True]), name=None)
> ```
> 
> This works for me?
> 
> also vscode doesn't seem to complain:
> 
> image

Oh, sorry. I had a typo, I did `ir.Tensor(...)` instead of `ir.tensor(...)` and that complained that an array-compatible object was required :-(

onnxscript/optimizer/_constant_folding.py
```diff
@@ -448,25 +448,54 @@ def concat(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
...
-        # TODO: handle this case. But unlikely to be needed in practice.
-        return None
+
+    def optimized_dropout():
+        input = node.inputs[0]
+        output = op.Identity(input)
+        if len(node.outputs) == 1:
+            return output
+        else:
+            true_tensor = ir.tensor([True])
```
justinchuby, line 458: Does it need to be 1d? Or is 0d better? No big deal though

justinchuby (approved):

PR #1948, opened 2024-11-15
Improve handling of trailing optional inputs in pattern matching

gramalingam:
Pattern-variables are allowed in places where an optional input is expected. If the input is present, the pattern-variable will be bound to the corresponding ir.Value. Otherwise, the pattern-variable will be bound to None.

This PR extends this convention to handle trailing optional inputs better: specifically, `SomeOp(x)` and `SomeOp(x, None)` are treated equivalently during the pattern-match. If the pattern is `SomeOp(pattern_x, pattern_y)`, then `pattern_y` will be bound to None in both cases, and the pattern-match will succeed.

Side note: A consequence of allowing pattern-variables to be bound to None is that the corresponding guard/rewrite function will have to handle the case where the pattern-variable is None correctly. This is not an issue for correctly-typed models (where we can assume that a non-optional input will never be None). However, to protect against incorrectly-typed models, we may end up having to add tedious `x is not None` checks before accessing x's fields/attributes. It may be useful to perhaps allow the user to flip the convention for a given rule (as to whether pattern-variables are allowed to be bound to None). Or, ensure a model is type-valid before applying optimizations. (Or, a quick hack might be to catch these exceptions in the rewrite-engine and skip that particular rewrite, which is probably a good idea anyway.)

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -1040,14 +1040,19 @@ class SimplePatternMatcher(PatternMatcher):
 
         self._matched[pattern_node] = node
 
-        if len(node.inputs) > len(pattern_node.inputs) and not pattern_node.allow_other_inputs:
-            return self.fail(
-                f"Number of inputs ({len(node.inputs)}) is more than expected ({len(pattern_node.inputs)})"
-            )
+        if len(node.inputs) > len(pattern_node.inputs):
+            if pattern_node.allow_other_inputs:
+                # Ignore extraneous inputs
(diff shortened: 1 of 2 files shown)


gramalingam:
@justinchuby re. our discussion

justinchuby:
I think most of the time a user would want Non-Nones as default. I agree having to check for None for every input can be tedious. Given this I think specifically declaring something can be None is more intentional and less error prone.

gramalingam:
Created https://github.com/microsoft/onnxscript/pull/2460 which handles this via explicit user annotation.

PR #1949, opened 2024-11-15
[rewriter] Fix slices pattern

titaiwangms:
The pattern did not cover the dynamic shapes case, so it leads to "'<' not supported between instances of 'int' and 'SymbolicDim'" when input is dynamic.

### DIFF
--- a/onnxscript/rewriter/collapse_slices.py
+++ b/onnxscript/rewriter/collapse_slices.py
@@ -56,7 +56,7 @@ def _check_if_redundant_slice(
     # In case data.shape is not statically known, we still can tell the slice is redundant if ends is sys.maxsize
     if ends_const.numpy().item() == _INT64_MAX:
         return True
-    if data.shape is None:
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/collapse_slices.py
```diff
@@ -56,7 +56,7 @@ def _check_if_redundant_slice(
     # In case data.shape is not statically known, we still can tell the slice is redundant if ends is sys.maxsize
     if ends_const.numpy().item() == _INT64_MAX:
         return True
-    if data.shape is None:
+    if data.shape is None or not isinstance(data.shape[axes_const.numpy().item()], int):
```
justinchuby, line 59: Created https://github.com/microsoft/onnxscript/pull/1952

gramalingam (approved):

onnxscript/rewriter/collapse_slices.py
justinchuby, line 59:
```suggestion
    if data.shape is None or data.shape.is_dynamic(axes_const.numpy().item()):
```

PR #1952, opened 2024-11-15
[IR] Implement methods to check dynamism on Shape

justinchuby:


onnxscript/ir/_core.py
```diff
@@ -997,6 +997,13 @@ def __eq__(self, other: object) -> bool:
     def __ne__(self, other: object) -> bool:
         return not self.__eq__(other)
 
+    def is_static(self, dim: int) -> bool:
+        # Raise if index error
```
justinchuby, line 1001: Add docs

onnxscript/ir/_core_test.py
```diff
@@ -637,6 +637,34 @@ def test_set_denotation_is_still_possible_when_shape_is_frozen(self):
...
+
+    def test_is_static_raises_when_index_out_of_range(self):
+        shape = _core.Shape([42])
+        with self.assertRaises(IndexError):
+            shape.is_static(1)
+
+    def test_is_symbolic(self):
+        dim_from_numpy = np.array([42]).shape[0]
+        np_int = np.int32(42)
+        shape = _core.Shape([42, "any string", dim_from_numpy, np_int])
```
justinchuby, line 657: Add support for np int

titaiwangms (approved):
Do you think a method return whether the Shape is static or not would be another useful method? (When users do not want to go through all dims one by one.)

justinchuby:
> Do you think a method return whether the Shape is static or not would be another useful method? (When users do not want to go through all dims one by one.)

Sure! Basically we can make `shape.is_static()` return this info

justinchuby:
> > Do you think a method return whether the Shape is static or not would be another useful method? (When users do not want to go through all dims one by one.)
> 
> Sure! Basically we can make `shape.is_static()` return this info

Done

PR #1971, opened 2024-12-01
[torchlib] torch.where(x) - default overload - is actually not implemented

Bludator:
As the title said the overload is not implemented in [`aten_where`](https://github.com/microsoft/onnxscript/blob/99cf79fd4ab150e3726b36fb3e9104304e203200/onnxscript/function_libs/torch_lib/ops/core.py#L8871C1-L8874C44). It should be decomposed into `nonzero` function by pytorch. 

Now it throws error as there is not enough parameters. Minimal reproducible example:
```python
import torch

class Model(torch.nn.Module):
	def forward(self, x):
		return torch.where(x)

torch.onnx.export(Model(), (torch.tensor([0, 1, 2, 0, 3]),), dynamo=True)
```
```
<class 'ValueError'>: Required parameter 'self' is not provided. Signature: pkg.onnxscript.torch_lib::aten_where(condition: T_condition, self: TTensor, other: TTensor) -> (TTensor) where T_condition=BOOL, TTensor=INT8 | FLOAT16 | INT16 | INT32 | UINT8 | FLOAT | BOOL | COMPLEX128 | BFLOAT16 | COMPLEX64 | DOUBLE | INT64. Args: (SymbolicTensor('x', type=Tensor(INT64), shape=[5], producer=None, index=None),). Kwargs: {}.
```
As for the tests I would have thought it is handled by the [ops_test.py](https://github.com/microsoft/onnxscript/tree/main/tests/function_libs/torch_lib) but apparently it is not.

---
As a side note, the `pylint` is somehow broken for this file (at least).

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -8861,7 +8861,6 @@ def aten_vstack(tensors: Sequence[TTensor]) -> TTensor:
 
 @torch_op(
     (
-        "aten::where",
         "aten::where.Scalar",
         "aten::where.ScalarSelf",
         "aten::where.ScalarOther",


titaiwangms:
I hit NotImplementedError though. It's not decomposed.

titaiwangms:
Based on this PR: https://github.com/pytorch/pytorch/issues/21798
The real fix is to implement: https://github.com/microsoft/onnxscript/blob/99cf79fd4ab150e3726b36fb3e9104304e203200/onnxscript/function_libs/torch_lib/ops/core.py#L6355 and register `aten::where` under it.

Something like:
```python
# aten::where.default is the same as aten::nonzero_numpy
# https://github.com/pytorch/pytorch/issues/21798
@torch_op("aten::where", trace_only=True)
def aten_nonzero_numpy(self: TensorType) -> TensorType:
    """nonzero_numpy(Tensor self) -> Tensor[]"""
    nonzero = aten_nonzero(self)
    return aten_unbind(nonzero, dim=1)
```

And enable its test in ops_test.py.

Bludator:
The odd thing is it works for me by just removing it. Let me just install fresh venv and try again

Bludator:
OK, on today's nightly build of torch the decompositions does not work but on the stable version and the nightly from last week it have worked... 😮‍💨

Bludator:
You are right, on nightly build of torch the decompositions does not work but on the stable version it do... 😮‍💨

I think that it is better if it is decomposed by the pytorch, as there should be decomposition of where(x) to the nonzero in any way.
Something like:
``` python
if a is None and b is None:
    return torch.nonzero(pred, as_tuple=True)
```
[here, which is actually where the NotImplementedError originates](https://github.com/pytorch/pytorch/blob/9125e9119cee131d7f839c105d79e8036b44e92b/torch/_refs/__init__.py#L1971)

Bludator:
@titaiwangms with [this PR](https://github.com/pytorch/pytorch/issues/141770) it works without any other change. I have no idea why as it seems to me there is no decomposition that would handle it.

But I will probably make PR from my previous comment in any way.

titaiwangms (approved):
I was just looking at this again. Nice!

titaiwangms:
> @titaiwangms with [this PR](https://github.com/pytorch/pytorch/issues/141770) it works without any other change. I have no idea why as it seems to me there is no decomposition that would handle it but it works.
> 
> But I would probably make PR from my previous comment in any way.

`aten.where.default` was overwritten by the second decomp table leading to a different result.

PR #1972, opened 2024-12-03
A couple of ir utilities

gramalingam:
Extracting out some IR utilities from the ongoing fusion draft version, which are useful for debugging and use in other (constant-folding and rewriter) optimizations.

### DIFF
--- a/onnxscript/rewriter/_ir_utils.py
+++ b/onnxscript/rewriter/_ir_utils.py
@@ -2,12 +2,58 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
+import numpy as np
+
 import onnxscript.ir as ir
 from onnxscript.optimizer import basic_constant_propagation
 
+
+def display_slice(x: ir.Value | ir.Node, backward: bool = True, depth_limit: int = 5):
+    """Display the subgraph computing a given value or node upto a certain depth."""
+    slice = []
+    def visit(node: ir.Node, depth):
+        if node in slice:
+            return
+        slice.append(node)
+        if (depth < depth_limit):
+            if backward:
+                for inp in node.inputs:
+                    if inp is not None and inp.producer() is not None:
+                        visit(inp.producer(), depth + 1)
+            else:
+                for out in node.outputs:
+                    for consumer, _ in out.uses():
+                        visit(consumer, depth + 1)
+    if isinstance(x, ir.Node):
+        visit(x, 0)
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/_ir_utils.py
```diff
@@ -2,12 +2,70 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
+import numpy as np
+
 import onnxscript.ir as ir
 from onnxscript.optimizer import basic_constant_propagation
 
 
+def display_slice(x: ir.Value | ir.Node, backward: bool = True, depth_limit: int = 5):
```
shubhambhokare1, line 11: nit: -> None

onnxscript/rewriter/_ir_utils.py
```diff
@@ -2,12 +2,70 @@
...
+    if const_value is not None:
+        try:
+            return const_value.numpy()
+        except FileNotFoundError:
+            # External data is not available.
+            return None
+    return None
+
+
+def get_singleton_value(val: ir.Value | None):
```
shubhambhokare1, line 66: nit: -> None

onnxscript/rewriter/_ir_utils.py
```diff
@@ -2,12 +2,70 @@
...
+
+
 def get_const_value(value: ir.Value) -> ir.TensorProtocol | None:
     node = value.producer()
     if node is not None:
         basic_constant_propagation([node])
     return value.const_value
+
+
+def get_numpy_value(val: ir.Value | None) -> np.ndarray | None:
```
shubhambhokare1, line 53: nit: add function string/ or is this just a helper for get_singleton_value

shubhambhokare1 (commented):
Should we add a couple of tests demonstrating this functionality of display_slice?

onnxscript/rewriter/_ir_utils.py
gramalingam: Done

onnxscript/rewriter/_ir_utils.py
gramalingam: This can return any type of value stored in a tensor or None. Typically should be `int | float | str | None`, but I am not sure about float16/bfloat16 or other such types. Do we have a type to represent all tensor-element types? If so, I can use that. Leaving it as is for now.

shubhambhokare1 (approved):
LGTM! nit: Fix lintrunner warnings

PR #1974, opened 2024-12-09
RMS Normalization and Skip RMS Normalization fusion optimizations

gramalingam:
Implements RMS Normalization and Skip RMS Normalization fusion optimizations (for use of onnxruntime custom fused ops for these).


xadupre (approved):

xadupre:
It looks good. At some point, we have a page in the documentation which lists all the possible optimization and how to trigger them. Even for us, this page should prove useful.

PR #1978, opened 2024-12-14
Add missing close parenthesis in conversion to script

gramalingam:
The tool for converting onnx model to onnxscript code has a missing parenthesis when generating external data tensor descriptor.


titaiwangms (approved):
Do we need a test?

PR #1981, opened 2024-12-18
Add rotary embedding fusion rule (part 1)

gramalingam:
Initial version of fusion for rotary embedding. 

Limitations: currently addresses only non-interleaved and full rotation.

Other:
* Add support for rewriting rules where the matched nodes are not removed. Useful in cases where matched nodes include some shared nodes.
* Add optimization to eliminate redundant Reshape (helps simplify pattern).


onnxscript/rewriter/onnxruntime/xformers/rotary_embedding.py
```diff
@@ -0,0 +1,64 @@
...
+# Note: This targets the new op being proposed to ONNX. This version does not exist in ORT yet.
+# so it can't be tested by running against ORT. See cos_sin_cache.py for a transformation that
+# rewrites the pattern into one that can be run against ORT.
+
+
+def _rotate_half_pattern(op, x, start1, end1, start2, end2):
+    # Slice(input, starts, ends, axes, steps)
+    x1 = op.Slice(x, start1, end1, [3], [1])
+    x2 = op.Slice(x, start2, end2, [3], [1])
+    minus_x2 = op.Neg(x2)
```
shubhambhokare1, line 21: Although this logic is correct and makes sense, this doesn't match the function logic in the op definition. Is it correct to assume that the pattern logic should mimic the onnx function in the op schema?

Currently in the op schema, this pattern would look like after x1, x2 (which uses split instead of slice for non-interleaved case):
```
real = cos * x1 - sin * x2
imag = sin * x1 + cos * x2
rotated_x = op.Concat(real, imag)
```
So the concat happens after the multiplication

onnxscript/rewriter/onnxruntime/xformers/rotary_embedding.py
```diff
@@ -0,0 +1,64 @@
...
+    rotated_x = op.Concat(minus_x2, x1, axis=-1)
+    return rotated_x
+
+
+class RotaryEmbeddingFusion(pattern.RewriteRuleClassBase):
+    def pattern(self, op, x, cos, sin, start1, end1, start2, end2):
+        return x * cos + _rotate_half_pattern(op, x, start1, end1, start2, end2) * sin
+
+    def check(self, op, x, start1, end1, start2, end2, **_):
+        # x needs to be a 4D tensor with known last dimension size (== head_size) and known second dimension (num_heads)
```
shubhambhokare1, line 31: According to schema, x can be a 3D tensor as well. And num_heads are necessary to be known in cases with 3D tensor.

onnxscript/rewriter/onnxruntime/xformers/rotary_embedding.py
```diff
@@ -0,0 +1,64 @@
...
+            _ir_utils.is_singleton_value(start1, 0)
+            and _ir_utils.is_singleton_value(end1, half_head_size)
+            and _ir_utils.is_singleton_value(start2, half_head_size)
+            and _ir_utils.is_singleton_value(end2, lambda x: x >= head_size)
+        )
+
+    def rewrite(self, op, x, cos, sin, **_):
+        num_heads = x.shape[1]
+        return op.RotaryEmbedding(
+            x, cos, sin, interleaved=0, num_heads=num_heads, _domain="ai.onnxruntime.fusion"
```
shubhambhokare1, line 52: Just curious about the domain here

gramalingam: Good question. I am currently using these to split the fusion optimization into multiple stages. We may need to clean this up finally. For now, we also need to target the existing RotaryEmbedding op in ORT (which is what we can test against also). Eventually, we can target the new proposed RotaryEmbedding op ... so we may also need to support some variations in the fusion optimization (depending on target ORT/ONNX versions).

onnxscript/rewriter/onnxruntime/xformers/rotary_embedding.py
gramalingam: That means the optimization is safe and correct (in this regard). To generalize and allow a 3D also here, we would need to guarantee that the entire-fusion is guaranteed to be semantically correct ... it is not enough to know that the RotaryEmbedding op permits 3D inputs.

What do you think about the correctness of this fusion optimization? Do you think it is fine to generalize and allow 3D here?

onnxscript/rewriter/onnxruntime/xformers/rotary_embedding.py
gramalingam: It does not have to match the function logic in the op definition. But it has to match the function graph produced by the ONNX exporter from the logic defined in the source (eg., the transformers implementation). 

But what we have to guarantee or ensure is that replacing this logic by the pattern in `rewrite` is fine: that they will both produce the same values.

gramalingam: Specifically, it is more important to match the source logic like [this transformer code](https://github.com/huggingface/transformers/blob/42865860ec6dc135972d9555753cb7ee17f51fb4/src/transformers/models/llama/modeling_llama.py#L147)

onnxscript/rewriter/onnxruntime/xformers/__init__.py
```diff
@@ -1,3 +1,15 @@
...
+from onnxscript.rewriter.onnxruntime.xformers.rms_normalization import fuse_rms_normalization
+from onnxscript.rewriter.onnxruntime.xformers.rotary_embedding import fuse_rotary_embedding
+from onnxscript.rewriter.onnxruntime.xformers.skip_normalization import fuse_normalization
+
+__all__ = [
+    "fuse_rms_normalization",
+    "fuse_normalization",
+    "fuse_rotary_embedding",
+    "fuse_cos_sin_cache",
+]
```
justinchuby, line 15:
```suggestion
from __future__ import annotations

__all__ = [
    "fuse_rms_normalization",
    "fuse_normalization",
    "fuse_rotary_embedding",
    "fuse_cos_sin_cache",
]

from onnxscript.rewriter.onnxruntime.xformers.cos_sin_cache import fuse_cos_sin_cache
from onnxscript.rewriter.onnxruntime.xformers.rms_normalization import fuse_rms_normalization
from onnxscript.rewriter.onnxruntime.xformers.rotary_embedding import fuse_rotary_embedding
from onnxscript.rewriter.onnxruntime.xformers.skip_normalization import fuse_normalization
```

nit: all should be right after future imports

onnxscript/rewriter/onnxruntime/xformers/cos_sin_cache.py
```diff
@@ -0,0 +1,102 @@
...
+        return op.RotaryEmbedding(
+            x,
+            cos_4d,
+            sin_4d,
+            interleaved=interleaved,
+            num_heads=num_heads,
+            _domain="ai.onnxruntime.fusion",
+        )
+
+    def check(self, context, inv_freq, position_ids, **_):
```
justinchuby, line 64:
```suggestion
    def check(self, context, inv_freq, position_ids, **_) -> bool:
```

onnxscript/rewriter/pattern.py
```diff
@@ -1197,6 +1201,7 @@ def match(
         graph_or_function: ir.Graph | ir.Function,
         node: ir.Node,
         verbose: int = 0,
+        remove_nodes: bool = True,
```
justinchuby, line 1204: Helpful to document new args?

justinchuby (approved):
Just some nits, thanks!

onnxscript/rewriter/pattern.py
gramalingam: Good point. Will do this separately, there are likely to be overlapping (merge) conflicts from other changes I have been making concurrently to extend pattern matcher.

PR #1986, opened 2024-12-27
First version of attention fusion

gramalingam:
First version of attention fusion.

Limitations:
* Targetting only static shapes for now. Dynamic shapes will alter the pattern.
* Targetting only MHA and the new onnx Attention op.


justinchuby (approved):

justinchuby:
The mypy checks can be ignored here: https://github.com/microsoft/onnxscript/blob/347aa066181af888f649fea963d6ab2c3c9044f4/.lintrunner.toml#L51-L53

PR #1987, opened 2024-12-27
Extend onnx to script converter

gramalingam:
Extend onnx to script converter to suppress initializers in generated script, and replace them with randomly generated weights. This allows a compact onnxscript representation of models with large initializers, especially when we don't care about the exact weights. This is useful for generating source-based test-cases from models.

Eg., this allows us to generate test-cases such as the one [here](https://github.com/microsoft/onnxscript/blob/ca11a20625fb3b170c23798e03b4366d17d6bc1c/onnxscript/rewriter/onnxruntime/xformers/_smollm_1layer.py#L18).

### DIFF
--- a/onnxscript/backend/onnx_export.py
+++ b/onnxscript/backend/onnx_export.py
@@ -730,21 +730,21 @@ class Exporter:
         indented_initializers_as_params = "\n".join(
            f"{_SINGLE_INDENT}{_SINGLE_INDENT}{x}," for x in init_names
         )
-        return """
+        return f"""
 def make_model(
-{initializers_as_params}):
+{initializers_as_params}
 ):
-    {script}
+{script}
 
-    model = {script_function_name}.to_model_proto()
-    return model
+{_SINGLE_INDENT}model = {script_function_name}.to_model_proto()
+{_SINGLE_INDENT}return model
 
 def make_model_with_random_weights():
 {random_initializer_values}
-   model = make_model(
+{_SINGLE_INDENT}model = make_model(
 {indented_initializers_as_params}
-   )
-   return model 
+{_SINGLE_INDENT})
+{_SINGLE_INDENT}return model 
 """
 
     def _import_onnx_types(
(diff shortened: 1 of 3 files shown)


tools/onnx2external.py
```diff
@@ -0,0 +1,31 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
```
justinchuby, line 4:
```suggestion
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
```

justinchuby (approved):

tools/onnx2external.py
```diff
@@ -0,0 +1,31 @@
...
+import argparse
+import os
+
+import onnx
+import onnx.external_data_helper
+
+
+def convert2external(input_file_name: str) -> None:
+    dir_name = os.path.dirname(input_file_name)
+    base_name, suffix = os.path.splitext(os.path.basename(input_file_name))
```
justinchuby, line 15:
```suggestion
    base_name, _suffix = os.path.splitext(os.path.basename(input_file_name))
```

tools/onnx2external.py
```diff
@@ -0,0 +1,31 @@
+# -------------------------------------------------------------------------
```
justinchuby, line 1: Helpful to add a docstring for a brief description of this script?

onnxscript/backend/onnx_export.py
```diff
@@ -684,15 +694,61 @@ def _translate_graph(self, model: onnx.ModelProto, function_name: Optional[str])
...
+
+{__}model = {script_function_name}.to_model_proto()
+{__}return model
+
+def make_model_with_random_weights():
+{random_initializer_values}
+{__}model = make_model(
+{indented_initializers_as_params}
+{__})
+{__}return model
```
justinchuby, line 750:
```suggestion
{__}return model
```

Issue #1984, opened 2024-12-26
informing about bad positional argument in node constructor


leshabirukov:
[rewr.txt](https://github.com/user-attachments/files/18253201/rewr.txt)

When you mistakingly do 

```
  def lr_pattern( op, x ):
      return op.LeakyRelu( x, 1.0)
```
  
instead of

```
  def lr_pattern_good( op, x ):
      return op.LeakyRelu( x, alpha=1.0)
```

You will get such error text:

`  AttributeError: 'float' object has no attribute '_add_usage'`

Expected behavior: something like

`  TypeError: LeakyRelu() takes 2 positional arguments but 3 were given`

Possible patch: [missed_keyword.patch](https://github.com/user-attachments/files/18253239/missed_keyword.patch)

justinchuby:
Thanks! Will fix

leshabirukov:
I have a question:

is `opset??.Operation` supposed to allow non-Value wrapped arguments? I see here

https://github.com/microsoft/onnxscript/blob/3c75013e3957768fe7679a6152d16ebee3284a9f/onnxscript/rewriter/pattern.py#L250
https://github.com/microsoft/onnxscript/blob/3c75013e3957768fe7679a6152d16ebee3284a9f/onnxscript/rewriter/pattern.py#L290

`OpPatternBuilder` try to wrap float as Constant, returning Value(None). But it ends with this topic problem. 

By the way, traceback:

```
Matched: Identity
Traceback (most recent call last):
  File "C:\Users\lesha\Documents\GitHub\onnxscript\.hypothesis\letsbegin.py", line 33, in <module>
    apply_to_file("with_id.onnx", [rule_to_lr, ])
  File "C:\Users\lesha\Documents\GitHub\onnxscript\.hypothesis\letsbegin.py", line 23, in apply_to_file
    rule[0].apply_to_model(ir_model)
  File "C:\Users\lesha\Documents\GitHub\onnxscript\onnxscript\rewriter\pattern.py", line 1348, in apply_to_model
    return RewriteRuleSet([self], commute=commute).apply_to_model(model, verbose=verbose)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lesha\Documents\GitHub\onnxscript\onnxscript\rewriter\pattern.py", line 1503, in apply_to_model
[...]

PR #1991, opened 2024-12-30
fix: informing about bad positional argument in node constructor #1984

leshabirukov:
Patch, proposed in  
https://github.com/microsoft/onnxscript/issues/1984


justinchuby (changes requested):
Thanks for the proposal! I would consider creating the check outside of the IR. Here, a lingering `None` value as the last input is entirely valid.

leshabirukov:
About `None`, - this part must allow None, if my understanding correct:  
 
    if self._inputs **and self._inputs[-1]** and not isinstance(self._inputs[-1], (Value)):

(actually, it allows 0.0, and that's bad, needs to change that)

    if self._inputs and self._inputs[-1] is not None and not isinstance(self._inputs[-1], (Value)):

About moving check outside, will research, probably after 01.01 :)

justinchuby:
I see, you are right. Although this check should still be moved outside for performance reasons. Feel free to propose something, or I can take a look when I get a chance.

PR #1992, opened 2024-12-31
[IR] Improve from_proto/to_proto typing with overloads

justinchuby:
- Use typing.overload to annotate the from_proto method for accurate type hinting. With this change we can recommend users to use `ir.from/to_proto` over the `ir.serde.(de)serialize*` methods and still keep mypy happy. This simplifies the serialization apis for users.
- Create deserialize_tensor_shape to deserialize tensor shapes.


onnxscript/ir/serde.py
```diff
@@ -151,24 +171,47 @@ def from_proto(
...
-    | _protocols.ValueProtocol
-    | _protocols.AttributeProtocol
-    | _protocols.ReferenceAttributeProtocol
-    | _protocols.TensorProtocol
-    | _protocols.TypeProtocol
-    | _protocols.GraphViewProtocol
-    | _protocols.FunctionProtocol,
-) -> Any:
+@typing.overload
+def to_proto(ir_object: _protocols.ModelProtocol) -> onnx.ModelProto: ...
```
gramalingam, line 193: Is testing instances of a protocol efficient? Or, is that a concern?

justinchuby: I think sit's ok in this case, because it is just going to be a onetime cost when user calls this function.

gramalingam (approved):

PR #1994, opened 2025-01-02
[torchlib] Squeeze sym_size

justinchuby:
Add Squeeze to aten_sym_size because the output is expected to be 0d.


gramalingam (approved):
Thanks!

PR #1995, opened 2025-01-03
[torchlib] Implement window functions

justinchuby:
- BlackmanWindow
- Hann
- Hamming

Fixes https://github.com/pytorch/pytorch/issues/142458


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1495,10 +1495,16 @@ def aten_bitwise_xor(self: TInt, other: TInt) -> TInt:
...
+    window_length: int,
+    dtype: int = -1,
+    layout: str = "",
+    device: str = "",
+    pin_memory: bool = False,
+) -> TensorType:
     """blackman_window(int window_length, *, ScalarType? dtype=None, Layout? layout=None, Device? device=None, bool? pin_memory=None) -> Tensor"""
 
-    raise NotImplementedError()
+    return op.BlackmanWindow(window_length)
```
titaiwangms, line 1507: Can `output_datatype` from the onnx op represent `dtype`?

titaiwangms (approved):
Prob not easy, but maybe we need a test?

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Good catch, thanks!

justinchuby:
> Prob not easy, but maybe we need a test?

Done. Thanks!

PR #2000, opened 2025-01-07
Delete onnxscript/diagnostics/infra

justinchuby:
The SARIF diagnostics in ONNX Script was unused. We have no future plans to use it. So we remove it here, with appreciation of initial efforts by @fatcat-z and @BowenBao


titaiwangms (approved):

titaiwangms:
We might need to deprecate SARIF with dynamo_export. They are basically a bundle: https://github.com/pytorch/pytorch/blob/main/torch/onnx/_internal/fx/fx_onnx_interpreter.py

diagnostic is everywhere in it.

titaiwangms:
Oh never mind. This is in onnxscript.

justinchuby:
that's right

PR #2001, opened 2025-01-08
A couple of extensions to rewriter

gramalingam:
Extends the rewriter with a couple of features:

* A debugging mode to perform the pattern matching (without any graph modifications) and to report instances that get the best score for a match (even if incomplete). Helps quickly identify causes for mismatches when we expect a match.

* Rewrite-rules can now specify a pre/post visitor method called before applying it to a graph/function. This is useful for rules that need to create "cached" values that are reused across multiple instances of the pattern.


onnxscript/rewriter/pattern.py
```diff
@@ -8,6 +8,8 @@
 import inspect
 import itertools
 import math
+from collections import defaultdict
+from enum import IntEnum
```
justinchuby, line 12:
```suggestion
import enum
```

I would import modules only in general

onnxscript/rewriter/pattern.py
```diff
@@ -328,13 +330,15 @@ def __init__(self) -> None:
         self.outputs: list[ir.Value] = []
         # For a failed match, _reason is a string that describes the reason for the failure.
         self._reason: str = ""
+        self._failure_node: ir.Node | None = None
```
justinchuby, line 333: I would add a comment on what this field stores

onnxscript/rewriter/pattern.py
```diff
@@ -1349,15 +1379,19 @@ def try_rewrite(
...
         return None
 
     def apply_to_model(
-        self, model: ir.Model, *, commute: bool = False, verbose: int | None = None
+        self, model: ir.Model, *, commute: bool = False, verbose: int | None = None, debug: bool = False
     ):
         # A convenience method to apply the rule to a model. We use a RewriteRuleSet to
         # handle commutative rules.
-        return RewriteRuleSet([self], commute=commute).apply_to_model(model, verbose=verbose)
+        return RewriteRuleSet([self], commute=commute).apply_to_model(model, verbose=verbose, debug=debug)
```
justinchuby, line 1394: Would it be a good idea to call this mode `dry_run`? Or does `debug` do more than that?

onnxscript/rewriter/pattern.py
```diff
@@ -1484,16 +1526,34 @@ def _apply_to_graph_or_function(
...
+        tracer: MatchingTracer | None = None,
     ) -> int:
+        """
+        Apply the rewrite rules to the given graph or function.
+
+        Args:
+            model (ir.Model): The model to which the rewrite rules are applied.
+            graph_or_function (ir.Graph | ir.Function): The graph or function to which the rewrite rules are applied.
+            verbose (int | None, optional): The verbosity level. Defaults to None.
+            tracer (MatchingTracer | None, optional): The tracer for debugging. Defaults to None.
```
justinchuby, line 1540:
```suggestion
            model: The model to which the rewrite rules are applied.
            graph_or_function: The graph or function to which the rewrite rules are applied.
            verbose: The verbosity level. Defaults to None.
            tracer: The tracer for debugging. Defaults to None.
```

Since we have type annotations already, we don't need to duplicate the info in docstring

onnxscript/rewriter/pattern.py
```diff
@@ -1510,17 +1570,116 @@ def _apply_to_graph_or_function(
...
+                rule.graph_post_visitor()
 
         return count
 
-    def apply_to_model(self, model: ir.Model, verbose: int | None = None) -> int:
+    def apply_to_model(
+        self, model: ir.Model, *, verbose: int | None = None, debug: bool = False
+    ) -> int:
+        """
+        Apply the rewrite rules in the set to the model.
```
justinchuby, line 1582:
```suggestion
        """Apply the rewrite rules in the set to the model.
```

onnxscript/rewriter/pattern.py
gramalingam: I called it `traceonly` originally. Thought `debug` would be easier for people ...

gramalingam: Anyway, will leave it as such for now. This may be a bit premature to decide. I think allowing users to pass in their own optional "tracer" object (that decides what information to track/log/print) may be better instead of a boolean.

onnxscript/rewriter/pattern.py
gramalingam: done

justinchuby (approved):

onnxscript/rewriter/_ir_utils.py
```diff
@@ -11,6 +11,21 @@
 from onnxscript.optimizer import basic_constant_propagation
 
 
+def display_nodes(nodes: list[ir.Node]) -> None:
```
justinchuby, line 14:
```suggestion
def display_nodes(nodes: Sequence[ir.Node]) -> None:
```

PR #2003, opened 2025-01-08
Optimizer extensions

gramalingam:
Extend the optimizer to enable optimizations (such as elimination of redundant Expand/Reshape) when symbolic dimensions are present. This requires propagating symbolic shape values (tensors that carry shape information that is not completely known at compile time) through the optimizer. These optimizations also help fusion optimizations (otherwise, we need more pattern variations and more complex patterns).

Handles symbolic shape propagation through Abs, Gather, Concat and uses them in Reshape/Expand. Abs shows up in Expand translation because Torch allows -1 for "no expansion" while ONNX uses 1, but this is not necessary if the input is a symbolic shape where every value is guaranteed to be non-negative.

Also fix node-level shape-inference to refine shape by merging best info from pre-existing shape and inferred shape.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -16,7 +16,6 @@ import onnx
 import onnx.reference.ops
 
 import onnxscript.ir as ir
-import onnxscript.ir._convenience as _convenience
 import onnxscript.rewriter.pattern as orp
 import onnxscript.utils.utils as utils
 
@@ -282,10 +281,12 @@ def _get_numpy_value(
         if size_limit is not None and const_value.size > size_limit:
             return None
         try:
-            return const_value.numpy()
+            array = const_value.numpy()
         except FileNotFoundError:
             # External data is not available.
             return None
+        assert isinstance(array, np.ndarray)
+        return array
     return None
 
 
@@ -295,14 +296,7 @@ def _get_bool_value(val: ir.Value | None) -> bool | None:
     value = _get_numpy_value(val)
     if value is None:
         return None
-    # TODO: cleanup following checks, which seem redundant. But need to also ensure
-    # the invariant when setting the value (and also use clearly defined representation
-    # types in evaluators, such a reference-evaluator).
-    if isinstance(value, bool):
-        return value
-    if isinstance(value, np.bool_):
-        return bool(value)
-    if isinstance(value, np.ndarray) and value.size == 1 and value.dtype == bool:
+    if value.size == 1 and value.dtype == bool:
         return value.item(0)
     return None
 
@@ -870,10 +864,6 @@ class ConstantFolder:
                 )
 
     def new_constant(self, irvalue: ir.Value, value):
-        # TODO(rama): Why do we need the conversion below?
-        if isinstance(value, (int, float, np.ScalarType)):
-            value = np.array(value)
-
         if not isinstance(value, np.ndarray):
             # ONNX does not have a way to represent non-tensor constants, eg. a sequence.
             # So, a constant-value of type sequence is not folded, but it can be used
@@ -885,7 +875,9 @@ class ConstantFolder:
             )
             return None
 
-        irvalue.const_value = _convenience.tensor(value)
+        tensor = ir.tensor(value)
+        tensor.name = irvalue.name
+        irvalue.const_value = tensor
 
         if value.nbytes > self._output_size_limit:
             logger.info(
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -133,6 +133,18 @@ class Replacement:
...
 
+# The optimizer tracks an optional symbolic value for each value in the model.
+# The symbolic value attached to a value X can be:
+# - another IR value Y (indicating that X is equal to Y)
+# - a list of IR values [Y1, Y2, ...] (indicating that X is a sequence of values Y1, Y2, ...)
+# - a Shape object (indicating that X is a shape value)
+# A Shape object as a symbolic value indicates that the corresponding value is
+# 1-D (or 0-D) tensor of INT64 values. The values in this object may be constants
+# or symbolic dimension values (like "batch_size", "sequence_length", etc.).
+# Currently, we assume that symbolic dimensions are also guranteed to be non-negative.
```
justinchuby, line 144:
```suggestion
# Currently, we assume that symbolic dimensions are also guaranteed to be non-negative.
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -235,11 +259,27 @@ def decorator(function: PartialEvaluatorFunction) -> PartialEvaluatorFunction:
...
+    # (which represents an unknown dimension value). Thus, two shapes such
+    # as (Batch, 1024) and (Batch, 1024) are considered equal, but (None, 1024)
+    # and (None, 1024) are not considered equal.
+    if any(isinstance(dim, ir.SymbolicDim) and dim.value is None for dim in shape1):
+        return False
+    return shape1.dims == shape2.dims
+
+
+def _get_numpy_value(
+    val: ir.Value | None, dtype: ir.DataType | None = None, size_limit: int | None = None
```
justinchuby, line 273: Seems helpful to document the behavior of `dtype`; intuitively it can also mean casting but here it's used as a condition

onnxscript/optimizer/_constant_folding.py
```diff
@@ -300,6 +340,54 @@ def _get_int_attribute(node: ir.Node, name: str, default: int | None = None) ->
...
 
 
+@register("Abs")
+def abs(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
+    """Replace an Abs node by Identity when applicable.
+
+    Currently, addresses Abs applied to symbolic shapes.
+    """
+    input = _get_input(node, 0)
+    input_sym_value = state.get_shape_value(input)
```
justinchuby, line 350: Do we need to ensure that input is the output of Shape? Or it this guaranteed / unrelated?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -459,6 +551,19 @@ def concat(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     inputs = node.inputs
     if len(inputs) == 1:
         return op.Identity(inputs[0])
+    # Track value of tensors that carry a shape value:
+    output = node.outputs[0]
+    if output is None:
+        return None
```
justinchuby, line 557: When would output be None?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -658,6 +766,27 @@ def sequence_at(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     return None
 
 
+def _merge_shapes(shape1: ir.Shape, shape2: ir.Shape) -> ir.Shape:
```
justinchuby, line 769:
```suggestion
def _merge_shapes(shape1: ir.Shape | None, shape2: ir.Shape | None) -> ir.Shape | None:
```

justinchuby (approved):

onnxscript/optimizer/_constant_folding.py
gramalingam: Only in the case of erroneous models.

onnxscript/optimizer/_constant_folding.py
gramalingam: This is relying on the invariant maintained by the optimizer that any "symbolic" name present in the returned value is a real dimension-value obtained via a Shape op.

gramalingam: If we need to generalize this in the future, we will need to explicitly add another boolean flag to track whether a symbolic name is known to be non-negative.

onnxscript/optimizer/_constant_folding.py
gramalingam: Done. The dtype is not critical, main goal was to add size limit to avoid triggering conversion of large tensors into numpy array, like when optimizing a Gather.  But a dtype filter can also help avoid some unnecessary conversions.

PR #2004, opened 2025-01-09
Doc script const 1

leshabirukov:
I believe, new code snippet annotation have more sense.


docs/tutorial/index.md
```diff
@@ -123,7 +123,7 @@ subsequently modified, this modification has no effect on the attribute-value
 or the ONNX function/model created. This may potentially cause the behavior
 of eager-mode execution to be inconsistent with the ONNX construct generated.
 
-Thus, the example shown above is equivalent to the following:
+Thus, second `script_const` definition in the following code have no effect:
```
titaiwangms, line 126:
```suggestion
Thus, the second `script_const` definition in the following code has no effect:
```

However, I still like the origin. cc the authors @justinchuby @gramalingam @shubhambhokare1

justinchuby: I can see why this change may be helpful: the "example shown above" is a little far away from this code. I think rewording the paragraph a little more may be helpful

gramalingam: Maybe extend `has no effect` to `has no effect on the subsequent on the subsequent call "tensor_attr.to_function_proto()"`.

gramalingam (approved):

PR #2010, opened 2025-01-15
[torchlib] Implement type promotion

justinchuby:
Implement PyTorch type promotion for torchlib functions. Reference https://github.com/pytorch/pytorch/blob/bdd942efd76e74baa5dd0a262f7c843ddfe2e11b/torch/_prims_common/__init__.py#L1160


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -258,12 +263,13 @@ def aten_addmv(
 
 @torch_op("aten::addr", traceable=True)
```
justinchuby, line 264: traceonly etc.

titaiwangms:
Should we merge a basic version before moving the whole thing to PyTorch so that we don't need to start from scratch?

justinchuby:
> Should we merge a basic version before moving the whole thing to PyTorch so that we don't need to start from scratch?

Sure! We can do that

onnxscript/function_libs/torch_lib/_type_promotion.py
```diff
@@ -0,0 +1,68 @@
...
+            return ordered_datatypes[idx + 1][0]
+        if a in dtypes:
+            return b
+        if b in dtypes:
+            return a
+
+    raise ValueError(f"Unexpected data types: {a}, {b}")
+
+
+def promote_types(op, values: Sequence[ir.Value]) -> Sequence[ir.Value]:
```
xadupre, line 42: If it is used only for binary operators, I would add it to the function name just to make sure it is not called in any other operator.

PR #2018, opened 2025-01-17
Fix index_put with boolean index

xadupre:
Related: #1749


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4297,36 +4297,8 @@ def aten_index_put_bool(
...
-            result = op.ScatterElements(zeros, new_ind_t, values)
-            # FIXME: type promotion
-            result = op.CastLike(result, self)
-            result = op.Add(result, self)
-        else:
-            result = op.ScatterElements(self, new_ind_t, values)
-
-    return result
+    # accumulate should be always False, True does not make sense but an assert would be great
+    return op.Where(index, values, self)
```
gramalingam, line 4301: This is much simpler. One problem is that this requires `values` to be broadcastable to `self`. But that may not always be the case. Eg.:
```py
import torch

x = torch.Tensor([0, 1, 2, 3])
print(x)

first_half = x < 2
print(first_half)

y = torch.Tensor([99, 98])
x[first_half] = y

print(x)
```

gramalingam: The above produces the output:
```
tensor([0., 1., 2., 3.])
tensor([ True,  True, False, False])
tensor([99., 98.,  2.,  3.])
```

gramalingam: It should work in the case we have, where the RHS (values) is a scalar

gramalingam: So, it is definitely better than the existing version. So, it is fine to merge it, but I think we may need a more complete solution at some point.

xadupre: What do you mean? When indices are boolean, I don't think we need more. Operator where takes care of the broadcast. When indices are integer, that's another story.

gramalingam: The example above uses indices as booleans: I don't think the translation will work for that example.

tests/function_libs/torch_lib/e2e_test.py
justinchuby, line 1: We can remove this file for now. It would be nice if we can reduce the test in size to keep it minimum. I would add this test in pytorch and mark it as xfail. Later on when we bump onnxscript in pytorch ci we will be able to enable it. Currently the op should be tested by https://github.com/microsoft/onnxscript/blob/e7d199e0d6be53636e69eab24735781e0287d5a6/tests/function_libs/torch_lib/ops_test_data.py#L852-L859

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby, line 4301: This line is currently not covered 

https://app.codecov.io/gh/microsoft/onnxscript/pull/2018?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=microsoft

I would tweak https://github.com/microsoft/onnxscript/blob/e7d199e0d6be53636e69eab24735781e0287d5a6/tests/function_libs/torch_lib/ops_test_data.py#L852-L859 and ensure that the tests run as expected (maybe the skip was wrong)

justinchuby: Right, it was skipped entirely: https://github.com/microsoft/onnxscript/blob/e7d199e0d6be53636e69eab24735781e0287d5a6/tests/function_libs/torch_lib/ops_test_data.py#L860

We should remove this line

justinchuby: Moved to #1749

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4295,37 +4295,10 @@ def aten_index_put_bool(
...
-            # FIXME: type promotion
-            result = op.CastLike(result, self)
-            result = op.Add(result, self)
-        else:
-            result = op.ScatterElements(self, new_ind_t, values)
-
-    return result
+    # TODO: Support indices with more than 1 elements
+    index = indices[0]
+    # accumulate should be always False, True does not make sense but an assert would be great
```
justinchuby, line 4300:
```suggestion
    # accumulate should be always False, True does not make sense but an assert would be great
        index = op.Transpose(index)
```

I don't know why it works, but it works

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4295,37 +4295,17 @@ def aten_index_put_bool(
...
+    # Reshape indices so it can be properly broadcasted
+    lself, lindex = len(self.shape), len(index.shape)
+    if lself > lindex:
+        shape = op.Shape(index)
+        append = op.Constant(value_ints=[1 for _ in range(lself - lindex)])
+        new_shape = op.Concat(shape, append, axis=0)
+        index = op.Reshape(index, new_shape)
+    return op.Where(index, values, self)
```
justinchuby, line 4308:
```suggestion
    self_rank = len(self.shape)
    index_rank = len(index.shape)
    if self_rank > index_rank:
        index_shape = op.Shape(index)
        padding = op.Constant(value_ints=[1 for _ in range(self_rank - index_rank)])
        padded_shape = op.Concat(index_shape, padding, axis=0)
        index = op.Reshape(index, padded_shape)
    return op.Where(index, values, self)
```

tests/function_libs/torch_lib/aten_ops_test.py
```diff
@@ -0,0 +1,83 @@
...
+import onnxruntime
+import torch
+
+
+class TestOnnxExportAten(unittest.TestCase):
+    def test_aten_index_put_mask_bool_fixed_broadcast_2d(self):
+        class Model(torch.nn.Module):
+            def forward(self, x, values):
+                x = x.clone()
+                mask = torch.tensor([True, False, True, True, False]).to(torch.bool)
```
justinchuby, line 14: I would add the tests in to https://github.com/microsoft/onnxscript/blob/a04ebfdae03250469adc3b2e8bd47f5ede256cb1/tests/function_libs/torch_lib/extra_opinfo.py#L2412-L2418. You can use slice_scatter as an example. Tests that calls ort etc. and has a dependency on torch.export are harder to maintain and should be moved to torch

justinchuby: The goal of the unit tests we have for the torchlib is to ensure the ops we have behaves exactly as the aten counterparts. It's much easier to create comprehensive tests with OpInfo by throwing in a lot of test cases very compactly. They are more self-contained because they target the operator logic directly to test for numerical accuracy. By calling torch.export, we aren't really sure what operator we are testing against.

Tests calling `export` essentially go through the whole loop for the exporter infrastructure and is thus more end-to-end. They have their own place, but I think are more fitting if we put them in the pytorch repo.

I don't think it's a good idea to have them here because (1) the tests are dependent on the exporter implementation, which is dependent heavily on PyTorch versions, so they don't necessarily test our very implementation of index_put and (2) they are not testing the ops directly, but are rather testing the whole export flow (not small/tight enough).

Issue #2016, opened 2025-01-16
[Rewriter] Create a way to create an initializer in the rewrite pattern


justinchuby:
Allow this usage in the rewrite pattern

```py
def replacement(x):
    tensor = ir.tensor(..., name="new_name")
    return op.Add(op.Initializer(tensor), x)
```

```
def Initializer(tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
    # When name is None, the value name will be taken from the tensor
    # Otherwise the name is specified in name
    # Optionally we can also generate a name if nothing is specified, but I am not sure if that's error-prone
    ...
```

cc @gramalingam

gramalingam:
Completed by #2019

PR #2019, opened 2025-01-17
Rama/new init

gramalingam:
Add support for creating new initializers in rewrite rules. The same can serve as the basis for creating new initializers in onnxscript (eager mode), but that is a separate issue to be tackled separately.

Addresses https://github.com/microsoft/onnxscript/issues/2016

### DIFF
--- a/onnxscript/ir/_tape.py
+++ b/onnxscript/ir/_tape.py
@@ -18,6 +18,7 @@ class Tape(Iterable[ir.Node]):
 
     def __init__(self) -> None:
         self._nodes: list[ir.Node] = []
+        self._initializers: list[ir.Value] = []
 
     def __iter__(self) -> Iterator[ir.Node]:
         return iter(self._nodes)
@@ -25,6 +26,10 @@ class Tape(Iterable[ir.Node]):
     @property
     def nodes(self) -> Sequence[ir.Node]:
         return tuple(self._nodes)
+
+    @property
+    def initializers(self) -> Sequence[ir.Value]:
+        return tuple(self._initializers)
 
     def op(
         self,
@@ -60,6 +65,10 @@ class Tape(Iterable[ir.Node]):
 
         return node.outputs
 
+    def initializer(self, name: str, tensor: ir.TensorProtocol) -> ir.Value:
+        value = ir.Value(name=name, shape=tensor.shape, type=ir.TensorType(tensor.dtype), const_value=tensor)
+        self._initializers.append(value)
+        return value
 
 # A type representing the domains/versions used in creating nodes in IR.
 UsedOpsets = List[Tuple[str, Optional[int]]]
--- a/onnxscript/ir/_tape.py
+++ b/onnxscript/ir/_tape.py
@@ -65,7 +65,10 @@ class Tape(Iterable[ir.Node]):
 
         return node.outputs
 
-    def initializer(self, name: str, tensor: ir.TensorProtocol) -> ir.Value:
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
+        name = name or tensor.name
+        if name is None:
+            raise ValueError("Name must be provided for initializer.")
         value = ir.Value(name=name, shape=tensor.shape, type=ir.TensorType(tensor.dtype), const_value=tensor)
         self._initializers.append(value)
         return value
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -119,7 +119,11 @@ class ReferenceEvaluator:
         evaluator = self.get_evaluator(domain, op, version)
         if evaluator is None:
             return None
-        return evaluator(*args, **kwargs)
+        try:
+            return evaluator(*args, **kwargs)
+        except Exception as e:
+            logger.debug("Evaluation failed: %s", e)
+            return None
 
 
 _reference_evaluator = ReferenceEvaluator()
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -900,6 +900,7 @@ class ReplacementSubgraph:
     match: MatchResult
     new_outputs: Sequence[ir.Value]
     new_nodes: Sequence[ir.Node]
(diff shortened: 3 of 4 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -119,7 +119,11 @@ def evaluate(self, domain: str, op: str, version: int, *args, **kwargs) -> Any:
         evaluator = self.get_evaluator(domain, op, version)
         if evaluator is None:
             return None
-        return evaluator(*args, **kwargs)
+        try:
```
gramalingam, line 122: This is unrelated to current PR ... but useful anyway.

onnxscript/rewriter/pattern.py
```diff
@@ -1566,6 +1567,12 @@ def _apply_to_graph_or_function(
                 if delta is None or tracer is not None:
                     continue
                 assert isinstance(delta, ReplacementSubgraph)
+                if delta.new_initializers and isinstance(graph_or_function, ir.Function):
+                    # TODO(rama): Can't add initializers to functions. But currently this is not
+                    # an issue, as we apply inlining before applying rewrite rules.
+                    continue
```
justinchuby, line 1573: Should we add a warning?

onnxscript/rewriter/pattern.py
```diff
@@ -1566,6 +1567,12 @@ def _apply_to_graph_or_function(
...
+                for initializer in delta.new_initializers:
+                    graph_or_function.initializers[initializer.name] = initializer
```
justinchuby, line 1575: Should we warn about overwriting initializers?

onnxscript/ir/_tape.py
```diff
@@ -60,6 +65,13 @@ def op_multi_output(
 
         return node.outputs
 
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
+        name = name or tensor.name
+        if name is None:
+            raise ValueError("Name must be provided for initializer.")
+        value = ir.Value(name=name, shape=tensor.shape, type=ir.TensorType(tensor.dtype), const_value=tensor)
```
justinchuby, line 72: We can add some asserts (not None) here and there to appease the type checker and ensure robustness

onnxscript/optimizer/_constant_folding.py
```diff
@@ -119,7 +119,11 @@ def evaluate(self, domain: str, op: str, version: int, *args, **kwargs) -> Any:
...
+            return evaluator(*args, **kwargs)
+        except Exception as e:
+            logger.debug("Evaluation failed: %s", e)
```
justinchuby, line 125:
```suggestion
            logger.warning("Evaluation failed: %s", e)
```

maybe an warning is appropriate

onnxscript/ir/_tape.py
```diff
@@ -60,6 +65,13 @@ def op_multi_output(
 
         return node.outputs
 
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
+        name = name or tensor.name
+        if name is None:
+            raise ValueError("Name must be provided for initializer.")
```
justinchuby, line 71: I would cover this line with test too

onnxscript/rewriter/pattern.py
gramalingam: Good question. I guess you mean if there is already an initializer with that name? This is also related to naming authority issues ... we could rename them also, but, for now, may be failing would be fine.

onnxscript/rewriter/pattern.py
gramalingam: I guess. An alternative would be to convert initializers into "Constant" nodes. Ok, let me do something simple for now.

onnxscript/rewriter/pattern.py
justinchuby: I think there are instances when we want to keep/reuse the original name. Maybe a warning is fine for now? I don't have a strong opinion

onnxscript/ir/_tape.py
```diff
@@ -60,6 +65,13 @@
...
+        value = ir.Value(name=name, shape=tensor.shape, type=ir.TensorType(tensor.dtype), const_value=tensor)
```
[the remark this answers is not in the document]
gramalingam: @justinchuby : there seems to be a mismatch here. What is the suggested approach? Maybe a quick fix is fine here for now.
* Could ignore it, assume it is going to be a Shape
* Create a new Shape from the ShapeProtocol
* [Longer term] Should the Value constructor accept a ShapeProtocol? It seems to accept a TypeProtocol.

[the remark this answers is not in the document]
justinchuby: I think we can ignore it for now. the [longer term] option seems reasonable to me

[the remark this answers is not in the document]
gramalingam: oh, I ended up creating a copy meanwhile. I guess that should be fine too.

onnxscript/ir/_tape.py
```diff
@@ -69,7 +69,8 @@ def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.
         name = name or tensor.name
         if name is None:
             raise ValueError("Name must be provided for initializer.")
-        value = ir.Value(name=name, shape=tensor.shape, type=ir.TensorType(tensor.dtype), const_value=tensor)
+        shape = ir.Shape((d if isinstance(d, int) else d.value) for d in tensor.shape.dims)
```
justinchuby, line 72: We can do `shape=ir.Shape(tensor.shape)`

onnxscript/ir/_tape.py
```diff
@@ -60,6 +65,14 @@ def op_multi_output(
 
         return node.outputs
 
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
+        name = name or tensor.name
+        if name is None:
+            raise ValueError("Name must be provided for initializer.")
+        shape = ir.Shape((d if isinstance(d, int) else d.value) for d in tensor.shape.dims)
```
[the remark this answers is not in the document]
justinchuby: NVM

PR #2020, opened 2025-01-17
[IR] Improve external data handling

justinchuby:
1. Add an `external_data` option to `ir.save`. This will save initializers as external tensors. It is robust against data loss when overwriting, and is idempotent when the current model does not contain external tensors already referencing the same path.
2. Expose `ir.external_data` module as a public module users can use to manipulate external data.
    3. It defines the following methods
		```py
		[
		    "set_base_dir",
		    "unload_from_model",
		    "load_to_model",
		    "convert_tensors_to_external",
		    "convert_tensors_from_external",
		]
		```
        I renamed `to_external_data` to `unload_from_model` for clarity. **Reviewers please let me know if the naming sounds good.**
1. Support setting a threshold `size_threshold_bytes` to control which tensors are offloaded.
1. Simplified torch_apis logic by leveraging to updated `ir.save` method.
1. Updated the to_external_data function to always load data to memory, iff the tensor references an external data file that is being written to. This simplifies the logic and avoids creating and managing temporary files.
2. Implemented a polyfill of the `zip()` function's strict mode to support Python<=3.9 

> [!NOTE]
> We **do not** need to add external data options to `ir.load`. The external data is always loaded lazily in the IR. If users want to transfer the data to memory at loading, they can use `ir.external_data.load_to_model()`.

## Example usage

```py
ir.save(model, "model.onnx", external_data="model.onnx.data")
# Can save many times
ir.save(model, "model_copy.onnx", external_data="model_copy.onnx.data")
```


onnxscript/ir/_external_data.py
```diff
@@ -268,56 +253,57 @@ def convert_tensors_to_external(
...
     model: _core.Model,
-    base_path: str | os.PathLike,
+    base_dir: str | os.PathLike,
     relative_path: str | os.PathLike,
-    load_external_to_memory: bool = False,
 ) -> _core.Model:
-    """Set all tensors with raw data as external data.
+    """Set all tensors with raw data as external data, into a single data file.
+
+    Exsiting external tensors are loaded to memory if they are referring to the
```
justinchuby, line 274:
```suggestion
    Existing external tensors are loaded to memory if they are referring to the
```

onnxscript/ir/_io.py
```diff
@@ -35,16 +35,72 @@ def load(path: str | os.PathLike, format: str | None = None) -> _core.Model:
...
+            and :param:`modify_model` is set to False.
     """
-    proto = serde.serialize_model(model)
-    onnx.save(proto, path, format=format)
-    # TODO(justinchuby): Handle external data when the relative path has changed
-    # TODO(justinchuby): Handle off loading external data to disk when saving
+    if external_data is not None:
+        if os.path.isabs(external_data):
+            raise ValueError(
+                f"The external data path must be a relative to the ONNX file path, not '{external_data}'."
```
justinchuby, line 72:
```suggestion
                f"The external data path must be relative to the ONNX file path, not '{external_data}'."
```

onnxscript/ir/_io.py
```diff
@@ -35,16 +35,75 @@ def load(path: str | os.PathLike, format: str | None = None) -> _core.Model:
     return model
 
 
-def save(model: _core.Model, path: str | os.PathLike, format: str | None = None) -> None:
+def save(
+    model: _core.Model,
+    path: str | os.PathLike,
+    format: str | None = None,
+    external_data: str | os.PathLike | None = None,
```
shubhambhokare1, line 42: nit suggestion: maybe name it as external_data_path to clarify it is a path to file?

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -68,32 +67,17 @@ def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike
     """Save the model with external data. The model is unchanged after saving."""
 
     # TODO(#1835): Decide if we want to externalize large attributes as well
-    initializer_values = tuple(model.graph.initializers.values())
-    tensors = [v.const_value for v in initializer_values]
+    tensors = [v.const_value for v in model.graph.initializers.values()]
     for tensor in tensors:
```
shubhambhokare1, line 71: nit: if this is the only use of tensors here, we can just rewrite as
`
for v in model.graph.initializers.values():
if v.const_value is None:
`

justinchuby: Done, thanks!

onnxscript/ir/_io.py
justinchuby: Since the option is intended to provided by keyword, I find `external_data` to be more concise and as clear. Let me know if you think differently?

gramalingam:
I have a question (the same one discussed earlier too): is it possible to ensure that small tensors (typically scalar constants, and shapes) stay internal to the model?

This is a question that impacts all ONNX users, so it is worth making a decision based on overall requirements (as opposed to just the IR perspective). The clear need is to have large tensors external (typically weights). Scalars and shapes: not so much. Further, no point in incurring any of the overheads of external tensors for them (not worth memory-mapping, not worth lazy loading, etc.) ONNX shape inference specifically does not load any external tensors during inference, and will be impacted if shapes are stored externally. Updating ONNX shape inference to do this will be a somewhat non-trivial task.

Another related question is the role of "Constant()" nodes vs. initializers ... here, there is no issue with using initializers in favor of Constant nodes (except inside Functions, which don't support them, but that's another issue worth discussing separately). Eg., initializers seem to be handled better by Netron (and a few other tools) than Constant nodes.

justinchuby:
> is it possible to ensure that small tensors (typically scalar constants, and shapes) stay internal to the model?

That makes sense. We can have a size limit like the onnx.save function does

onnxscript/ir/_polyfill.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Polyfill for Python builtin functions."""
+
+import sys
+from typing import Any, Sequence
+
+if sys.version_info >= (3, 10):
+    zip = zip
```
justinchuby, line 9:
```suggestion
    zip = zip  # pylint: disable=self-assigning-variable
```

onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -68,32 +67,16 @@ def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike
     """Save the model with external data. The model is unchanged after saving."""
 
     # TODO(#1835): Decide if we want to externalize large attributes as well
-    initializer_values = tuple(model.graph.initializers.values())
-    tensors = [v.const_value for v in initializer_values]
-    for tensor in tensors:
-        if tensor is None:
+    for value in model.graph.initializers.values():
+        if value.const_value is None:
```
xadupre, line 71: Maybe it is worth catching this when the value is added to the model not when it is saved. That way, the user has more info about when he did something wrong.

justinchuby: Sounds good. For that we will change pytorch exporter to add the message

onnxscript/ir/external_data.py
```diff
@@ -0,0 +1,393 @@
...
+    for tensor in sorted_tensors:
+        external_info = _compute_external_data_info(tensor, current_offset)
+        external_data_infos.append(external_info)
+        current_offset = external_info.offset + external_info.length
+    _write_external_data(sorted_tensors, external_data_infos, path)
+
+    # Create external tensor objects
+    external_tensors: list[_core.ExternalTensor] = [
+        _create_external_tensor(tensor, external_info, base_dir, relative_path)
+        for tensor, external_info in zip(sorted_tensors, external_data_infos, strict=True)
```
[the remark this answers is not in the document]
justinchuby: Polyfilled.

onnxscript/ir/_io.py
justinchuby: The parameter does not exist in ONNX standard API. It is controlled by a boolean flag and a location parameter. I consider the whole combination a little confusing (`location` itself is a strange naming for the api) so I deviated from it entirely.

onnxscript/_framework_apis/torch_2_5.py
gramalingam: I guess this is an internal invariant? I am guessing the only reason it can fail is if some user explicitly sets this value to be None via Value's setter method?

gramalingam (approved):

PR #2021, opened 2025-01-18
[torchlib] Register aten.linear and use matmul to simplify graph

justinchuby:
Use matmul when the input is not rank 2 to avoid decomp to addmm.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -831,7 +831,7 @@ def aten_linear(input: TFloat, weight: TFloat, bias: TFloat | None = None) -> TF
 
     if len(input.shape) == 2:
         # Use Gemm for the rank 2 input
-        return op.Gemm(input, weight, transB=True)
+        return op.Gemm(input, weight, bias, transB=True)
     weight_transposed = op.Transpose(weight, perm=[1, 0])
     mul = op.MatMul(input, weight_transposed)
     if bias is None:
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):

justinchuby:
```py
import torch


class TestModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = torch.nn.Linear(10, 10)

    def forward(self, x):
        return self.linear(x)


model = TestModule()
ep = torch.onnx.export(model, (torch.randn(1, 10),), dynamo=True, verify=True)
print(ep)
ep = torch.onnx.export(model, (torch.randn(1, 12, 15, 10),), dynamo=True, verify=True)
print(ep)
```

titaiwangms:
https://github.com/microsoft/onnxscript/pull/1821/files also disabled linear_bias though, do we want that back as well?

justinchuby:
Dynamic axes works too

```py
model = TestModule()
ep = torch.onnx.export(model, (torch.randn(2, 10),), dynamic_shapes=(
    {0: torch.export.Dim.DYNAMIC},
), dynamo=True, verify=True)
print(ep)
ep = torch.onnx.export(model, (torch.randn(2, 12, 15, 10),), dynamo=True, dynamic_shapes=(
    {0: torch.export.Dim.DYNAMIC},
), verify=True)
print(ep)
```

justinchuby:
> [#1821 (files)](https://github.com/microsoft/onnxscript/pull/1821/files) also disabled linear_bias though, do we want that back as well?

I combined the implementation.

Issue #2023, opened 2025-01-20
Python 3.9 - TypeError: unsupported operand type(s) for |: 'TypeVar' and 'NoneType'


kevalmorabia97:
It seems onnxscript should be supported for Python 3.8+ but the `|` operator for type annotation was only added in Python 3.10 so onnxscript may not really work for Python < 3.10. Is that correct?

I started seeing this error in `onnxscript==0.1.0.dev20250120` which was not the case in `onnxscript==0.1.0.dev20250118` when doing `torch.onnx.dynamo_export`.

```python
    torch.onnx.dynamo_export(
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/torch/onnx/__init__.py:517: in dynamo_export
    return dynamo_export(
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/torch/onnx/_internal/_exporter_legacy.py:1206: in dynamo_export
    else ResolvedExportOptions(export_options, model=model)
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/torch/onnx/_internal/_exporter_legacy.py:360: in __init__
    self.onnx_registry = resolve(options.onnx_registry, OnnxRegistry())
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/torch/onnx/_internal/_exporter_legacy.py:121: in __init__
    self._initiate_registry_from_torchlib()
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/torch/onnx/_internal/_exporter_legacy.py:139: in _initiate_registry_from_torchlib
    for meta in onnxscript_apis.get_torchlib_ops():
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/onnxscript/_framework_apis/torch_2_5.py:119: in get_torchlib_ops
    domain=overload_func.function_ir.domain,
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/onnxscript/values.py:657: in function_ir
    return converter.translate_function_signature(func_ast)
/tmp/-modelopt-tox/py39-torch25-unit/lib/python3.9/site-packages/onnxscript/converter.py:1470: in translate_function_signature
    return self._translate_function_signature_common(fn)
[...]

Issue #2011, opened 2025-01-15
Eager mode for onnxscript


justinchuby:
1. Create an eager graph building mode
2. The mode needs to be eager, aka the graph is built on the fly, instead of built lazily (?)
3. Types for ONNX Script functions should pass mypy checks. This is important for adoption in other code bases because we don't want our users to disable their checks for onnx script
4. We can create overloaded methods to handle numpy (concrete value) inputs. The inputs can also be ir.Tensor. Users should annotate the functions to take ir.Value or a derivative of this class (aka. onnxscript Tensor)

PR #2024, opened 2025-01-20
[torchlib] Use Optional in type annotations for linear

justinchuby:
Fix https://github.com/microsoft/onnxscript/issues/2023

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -826,7 +826,7 @@ def aten_leaky_relu_backward(
 
 
 @torch_op("aten::linear", trace_only=True)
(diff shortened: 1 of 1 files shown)


kevalmorabia97:
This repository does use `|` operator instead of `Union` for type annotations quite a lot in most files which is not supported for Python <3.10. Does that mean onnxscript should not be used for python <3.10?

titaiwangms (approved):

justinchuby:
This is an issue when we want to evaluate the function signature to get the type. This is why I think it maybe better to decorate the function with type information explicitly to avoid having to parse python types and bridge different python versions (https://github.com/microsoft/onnxscript/issues/2011)

justinchuby:
And yes as long as we don't evaluate the annotations, `__future__ import annotations` turns all annotations into strings so nothing hurts.

PR #2025, opened 2025-01-20
Refine constant folding size limit heuristic

gramalingam:
Refine the size-limit heuristics used to control constant-folding. This refinement allows some common cases to be handled automatically, such as Transpose(weight) which is typically generated by the exporter. The refinement looks at the increase in model-size that would be caused replacing a node by a constant, by accounting for inputs of the node that would be eliminated as a result of the replacement.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -874,7 +874,8 @@ class ConstantFolder:
                     e,
                 )
 
-    def new_constant(self, irvalue: ir.Value, value):
+    def new_constant(self, node: ir.Node, value):
+        irvalue = node.outputs[0]
         if not isinstance(value, np.ndarray):
             # ONNX does not have a way to represent non-tensor constants, eg. a sequence.
             # So, a constant-value of type sequence is not folded, but it can be used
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -891,12 +892,22 @@ def new_constant(self, irvalue: ir.Value, value):
...
 
         if value.nbytes > self._output_size_limit:
-            logger.info(
-                "Skip storing constant folded nvalue %s due to large size %s.",
-                irvalue.name,
-                value.nbytes,
-            )
-            return None
+            # Handle examples like Transpose(weight) to be folded even if the size is large,
+            # as long as weight has no other uses. This won't increase model size.
```
justinchuby, line 896: Do we consider the computation cost for different ops?

shubhambhokare1 (approved):
LGTM, needs rebase

onnxscript/optimizer/_constant_folding.py
justinchuby: I understand the primary reasons for us to avoid processing large tensors are (1) compute cost and (2) potential increment in data size. This seems to address (2) but (1) can still be an issue, which can causes optimization to take a long time to finish

gramalingam: Discussed offline. Currently, users can use input-size-limits to prevent processing large tensors. Several refinements/extensions are possible for this, but that's future work.

PR #2037, opened 2025-01-24
[torchlib] Fix reflection pad

justinchuby:


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -1517,37 +1519,16 @@ def aten_reflection_pad1d_backward(
...
-    odd_elements = op.Slice(torch_paddings, starts, ends, zero, steps)
-    # paddings[-1::-2]
-    starts = neg_1
-    ends = op.Sub(starts, size_d)
-    even_elements = op.Slice(torch_paddings, starts, ends, zero, steps)
-    # paddings[-2::-2] + paddings[-1::-2]
-    onnx_padding = op.Concat(odd_elements, even_elements, axis=0)
+    rank = len(self.shape)
+    paddings = list(padding) + [0] * (rank * 2 - len(padding))
+    paddings = paddings[-2::-2] + paddings[-1::-2]
```
justinchuby, line 1529: Why reversed?

titaiwangms (approved):
Do we need a new reflection_pad test to guard the missing corner case?

justinchuby:
> Do we need a new reflection_pad test to guard the missing corner case?

Could you remind me which was the missing case?

titaiwangms:
> pytorch/pytorch#144382

Aren't we refactoring because it has this corner case?

justinchuby:
Let me test

justinchuby:
Thanks for the reminder. It's fixed. Would be helpful to create some cases. I will do that

PR #2046, opened 2025-01-28
Fix misleading annotation in the documentation

xadupre:
The function does not seem to work inplace in all cases.

### DIFF
--- a/onnxscript/optimizer/__init__.py
+++ b/onnxscript/optimizer/__init__.py
@@ -24,7 +24,7 @@ def optimize(model: ir.Model | onnx.ModelProto, *args, **kwargs) -> ir.Model | o
         return legacy_optimizer.optimize(model, *args, **kwargs)
 
 
-def fold_constants(model: ir.Model | onnx.ModelProto, *args, **kwargs) -> bool
+def fold_constants(model: ir.Model | onnx.ModelProto, *args, **kwargs) -> bool:
     if isinstance(model, ir.Model):
         return constant_folding.fold_constants(model, *args, **kwargs)
     else:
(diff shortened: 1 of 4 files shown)


onnxscript/optimizer/__init__.py
```diff
@@ -15,14 +15,14 @@
 fold_constants_ir = constant_folding.fold_constants
 
 
-def optimize(model: ir.Model | onnx.ModelProto, *args, **kwargs):
+def optimize(model: ir.Model | onnx.ModelProto, *args, **kwargs) -> ir.Model | onnx.ModelProto:
```
justinchuby, line 18: We should make this a TypeVar 

# define TModel as a typevar then

```suggestion
def optimize(model: TModel, *args, **kwargs) -> TModel:
```

etc.

xadupre: I just did it for the documentation because without it, I assumed the modifications were done inplace which is not always the case. For that only purpose, I feel like keeping both types gives better information than a TypeVar.

justinchuby: Using union confuses the type checker in user code because then it will not know which is the return type without an isinstance assertion. That's going to be a inconvenience to users. Right now since we don't annotate at all, the type checker is able to infer the return type

gramalingam: Maybe we need to clean this up. It stems from our transition from proto to IR, which also simultaneously switched from copy-and-optimize to optimize-in-place. One issue to resolve is how much backward-compatibility we need here. Ideally, we should migrate away from proto-based optimization (not sure if there is any usage of this remaining), and have just in-place IR modification.

justinchuby: I think we may be able to annotate with only IR and hide the usage of the proto input. At some point after some deprecation period we can remove the logic entirely.

justinchuby (approved):

justinchuby:
I updated the function signature to hide the proto related types. Users are discouraged to send in proto types but the logic is still there for now.

PR #2050, opened 2025-02-03
[torchlib] Fix various implementations

justinchuby:
Fix implementations according to https://github.com/pytorch/pytorch/pull/146224. Removed Eager Mode tests since we only care about the graph constructed. Changed all traceable ops to trace_only. Removed usage of the IsScalar function.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -832,7 +832,7 @@ def aten_as_strided(
 ) -> TTensor:
     """as_strided(Tensor(a) self, SymInt[] size, SymInt[] stride, SymInt? storage_offset=None) -> Tensor(a)"""
 
-    rank = len(stride)
+    rank = len(stride.shape)
     return _aten_as_strided_onnx(self, size, stride, storage_offset, rank)
 
 
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -80,7 +80,7 @@ def aten__log_softmax_half(
 ) -> FLOAT:
     """_log_softmax(Tensor self, int dim, bool half_to_float) -> Tensor"""
 
-    self_is_scalar = IsScalar(self)
+    self_is_scalar = len(self.shape) == 0
     if half_to_float:
         self = op.Cast(self, to=FLOAT.dtype)
     if self_is_scalar:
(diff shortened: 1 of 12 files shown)


onnxscript/function_libs/torch_lib/ops/special.py
```diff
@@ -219,7 +217,7 @@ def aten_special_log_ndtr(self: TensorType) -> TensorType:
 def aten_special_log_softmax(self: TFloat, dim: int, dtype: int = -1) -> TFloat:
     """special_log_softmax(Tensor self, int dim, *, ScalarType? dtype=None) -> Tensor"""
 
-    self_is_scalar = IsScalar(self)
+    self_is_scalar = len(self.shape) == 0
```
titaiwangms, line 220: Does removing IsScalar potentially make the output axes static?

onnxscript/function_libs/torch_lib/ops/prims.py
```diff
@@ -837,7 +837,7 @@ def prims_view_of(a: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("prims::where", traceable=True)
```
titaiwangms, line 840: `traceable` is now fully retired? Or we still rely it on somewhere? If not, we can delete the parameter.

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/special.py
justinchuby: It does not. This is because the rank of a tensor in pytorch is always fixed.

onnxscript/function_libs/torch_lib/ops/prims.py
justinchuby: Yes. We can delete the param

justinchuby: Done

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1714,12 +1699,12 @@ def aten_clamp_max(self: TReal, max_: TReal) -> TReal:
...
 
 
-@torch_op(("aten::clamp_min", "aten::clamp_min.Tensor"), traceable=True)
+@torch_op(("aten::clamp_min", "aten::clamp_min.Tensor"), trace_only=True)
 def aten_clamp_min(self: TReal, min_: TReal) -> TReal:
     """clamp_min(Tensor self, Tensor min) -> Tensor"""
 
     # This implementation does not intent to handle when self is an empty tensor
-    min_rank = Rank(min_)
+    min_rank = len(min_.shape)
```
rrc-sony, line 1707: This breaks exporting scripts that use clamp(min=1e-12) (or any other float number) instead of a tensor, since floating point numbers don't have a .shape.
I expect the same issue appears in clamp_max but haven't tried to reproduce.

justinchuby: Thanks for reporting. I will create a patch today

PR #2053, opened 2025-02-05
Expose opset22

justinchuby:


titaiwangms (approved):
I forgot where we consume these?

PR #2057, opened 2025-02-14
[torchlib] Simplify aten_trunc implementation

justinchuby:
Simplify aten_trunc implementation according to https://github.com/onnx/onnx/issues/4588#issuecomment-2658170591

Thanks @fdwr


fdwr:
@justinchuby I have a few more decompositions [in this table](https://github.com/webmachinelearning/webnn/issues/817), but `trunc` is probably the only pertinent one left, as PyTorch appears to support 4 rounding modes:
- **r**ound **t**oward **z**ero (torch.trunc)
- **r**ound **h**alves to **e**ven (torch.round)
- **r**ound **u**p (torch.ceil)
- **r**ound **d**own (torch.floor).

(I just wandered across your old issue today while writing the WebNN issue and researching whether ONNX and PyTorch supported rounding away from zero - they does not)

titaiwangms (approved):

PR #2059, opened 2025-02-14
Bump version to 0.3.0

justinchuby:


titaiwangms:
Why do we bump so quick?

gramalingam:
> Why do we bump so quick?

+1 ... we will hit 1.0 quickly at this rate. I am referring to the delta in the version number (not the frequency of updated versions).

justinchuby:
My current thinking is that we can follow semantic versioning https://semver.org/ more closely, given that the versioning isn't as big for branding onnxscript as for onnx and pytorch. (aka we can bump major versions with less constraints).

The minor version happens when new apis/features are introduced, in this case opset22. We don't need to hit 1.0 if we don't want to. I.e. it will become 0.11, 0.12 etc. The patch versions are reserved for simple bug fixes that doesn't impact apis etc.

justinchuby:
Also for bug fixes we can branch out, change version numbers, and cherry pick like how onnx and pytorch do it. The VERISON number in the main branch is always bumped to the next minor version number after release.

justinchuby:
@titaiwangms ptal

gramalingam (approved):

PR #2061, opened 2025-02-17
[torchlib] Fix index_put: handle None cases

AyoubMDL:
This PR introduces support for `None` indices in the `index_put` function. If an index is None, it acts as a full slice (`:`).
### index_put Logic:
1. Construct index grid that contains all the indices to be updated
2. Reshapes the update values to match the computed indices.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -14,6 +14,8 @@ from __future__ import annotations
 import math
 from typing import Any, Optional, Sequence, Tuple, Union
 
+import numpy as np
+
 from onnxscript import (
     BFLOAT16,
     BOOL,
@@ -4303,17 +4305,21 @@ def aten_index_put(
         while len(reshape_list) > len(values_shape) and 1 in reshape_list:
             reshape_list.remove(1)
 
+        # Or add ones until the rank of reshape_list matches values_shape.
+        while len(reshape_list) < len(values_shape):
+            reshape_list.append(1)
+
         # Now ensure each dimension is broadcastable:
         # This is mandatory when mixing basic and advanced indexing
         # Example: data((10, 3, 4)), indices([[0, 1], :, [0, 1]]) values(2, 3)
         # the reshape list should be : [[2, 1], [1, 3], [2, 1]]
         for i, r in enumerate(reshape_list):
-            if r != 1 and r != values_shape[i]:
-                one_index = reshape_list.index(1)
+            if r not in (1, values_shape[i]):
+                value_index = values_shape.index(r)
                 # Swap elements
                 # For the example above the current reshape list is [1, 2] for last dim,
                 # to make it broadcastable, we swap the elements
-                reshape_list[one_index], reshape_list[i] = reshape_list[i], 1
+                reshape_list[value_index], reshape_list[i] = r, 1
 
         return reshape_list
 
@@ -4322,8 +4328,8 @@ def aten_index_put(
     if len(indices) < self_rank:
         indices = list(indices) + [None] * (self_rank - len(indices))
 
-    # Get values shape (we use .numpy to make it hashable)
-    values_shape = values.shape.numpy()
+    # Get values shape
+    values_shape = tuple(values.shape)
 
     index_vectors = []
     for i in range(self_rank):
@@ -4333,7 +4339,15 @@ def aten_index_put(
             reshape_update = self.shape[i]
         else:
             idx = indices[i]
-            reshape_update = indices[i].shape[0]
+            reshape_update = np.prod(idx.shape).item()
+            # when Index is more than 1D, flatten it and also the values shape
+            # Example: self shape: (10, 3), indices[i] shape: (2, 4), values shape: (2, 4, 3)
+            # Indices -> (2*4,) and values shape (2*4, 32)
+            if len(idx.shape) > 1:
+                values_shape = (reshape_update,) + values_shape[len(idx.shape) :]
+
+            # Flatten index (always working with 1D index in each dim)
+            idx = op.Reshape(idx, [-1])
 
         # Create a reshape pattern: one value per index dimension,
         # with the current dimension set to the update size.
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -4329,7 +4329,7 @@ def aten_index_put(
     for i in range(self_rank):
         if indices[i] is None:
             # For a full slice along dim i, create a range index [0, self.shape[i]).
-            idx = op.Range(0, op.Shape(self, start=i, end=i + 1), 1)
+            idx = op.Range(0, self.shape[i], 1)
             reshape_update = self.shape[i]
         else:
             idx = indices[i]
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -14,8 +14,6 @@ from __future__ import annotations
 import math
 from typing import Any, Optional, Sequence, Tuple, Union
 
-import numpy as np
-
 from onnxscript import (
     BFLOAT16,
     BOOL,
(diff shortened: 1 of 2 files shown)


AyoubMDL:
@justinchuby PR is ready for review

justinchuby:
Thank you! The failling test is unrelated

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4259,15 +4260,51 @@ def aten_index_put(
...
     """
+    # Pad indices with None so It has the same rank as self
+    self_rank = len(self.shape)
+    if len(indices) < self_rank:
+        indices = list(indices) + [None] * (self_rank - len(indices))
+
+    values_shape = values.shape.numpy()
+    # Pad values_shape with 1 so It has the same rank as self
+    if len(values_shape) < self_rank:
+        values_shape = (1,) * (self_rank - len(values_shape)) + values_shape
```
justinchuby, line 4271:
```suggestion
        values_shape = [*((1,) * (self_rank - len(values_shape))), *values_shape]
```

tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -790,20 +790,27 @@ def sample_inputs_index_put(op_info, device, dtype, requires_grad, **kwargs):
...
+    make_arg = functools.partial(
+        torch_testing.make_tensor, device=device, dtype=dtype, requires_grad=requires_grad
     )
-    yield opinfo_core.SampleInput(data, indices, values)
+
+    data = make_arg((10, 3, 4))
+
+    cases = [
+        # Case 1: Full slices in dims 0 and 2, tensor index in dim 1
+        ([None, torch.arange(2, device=device), None], (10, 2, 4)),
```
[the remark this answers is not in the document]
AyoubMDL: I actually wanted to test that but It does behave differently, Here's an example:
If I do torch.arange to specify the indices it will behave like this:
```python
import torch
data = torch.randn((1, 8, 4))
updates = torch.randn((1, 2, 3))
data[:, [0, 1], [1, 2, 3]] = updates
```
This results in an error: `shape mismatch: indexing tensors could not be broadcast together with shapes [2], [3]`

However, the following approach works:
```python
import torch
data = torch.randn((1, 8, 4))
updates = torch.randn((1, 2, 3))
data[:, 0:2, 1:4] = updates
```
The issue is that I don't know how to specify indices in the unit test in this way.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1647,7 +1647,8 @@ def aten_chunk(self: TTensor, chunks: int, dim: int = 0) -> Sequence[TTensor]:
     dim_size = op.Gather(self_shape, dim, axis=0)
     # Compute size/chunk to get the number of data in one chunk
     num_per_chunk = op.Div(dim_size, chunks)
-    num_per_chunk = op.Cast(op.Mod(dim_size, chunks) > 0, to=INT64.dtype) + num_per_chunk  # type: ignore[operator]
+    num_per_chunk = op.Cast(op.Mod(dim_size, chunks) > 0, to=INT64.dtype) + \
```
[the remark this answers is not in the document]
AyoubMDL: oops (it was the linter)

tests/function_libs/torch_lib/extra_opinfo.py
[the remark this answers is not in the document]
AyoubMDL: Added the cases, tell me if other cases are needed

AyoubMDL:
Force pushed as I changed a little bit the logic. Also added more test cases.

justinchuby:
Could you take a look at the falling CI, and we can merge this PR? Thanks!

AyoubMDL:
> Could you take a look at the falling CI, and we can merge this PR? Thanks!

it was missing a case with multidimensional index (added a test for that case)

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4303,17 +4305,21 @@ def _make_reshape_list_broadcastable(reshape_list, values_shape):
         while len(reshape_list) > len(values_shape) and 1 in reshape_list:
             reshape_list.remove(1)
 
+        # Or add ones until the rank of reshape_list matches values_shape.
+        while len(reshape_list) < len(values_shape):
```
justinchuby, line 4309: Possible to compute the difference and extend directly to avoid a while loop?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4298,14 +4300,82 @@ def aten_index_put(
...
+
+    index_vectors = []
+    for i in range(self_rank):
+        if indices[i] is None:
+            # For a full slice along dim i, create a range index [0, self.shape[i]).
+            idx = op.Range(start=0, limit=self.shape[i], delta=1)
+            reshape_update = self.shape[i]
+        else:
+            idx = indices[i]
+            reshape_update = np.prod(idx.shape).item()
```
justinchuby, line 4342: We can use math.prod instead
```suggestion
            reshape_update = math.prod(idx.shape)
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4298,14 +4300,82 @@ def aten_index_put(
...
+        indices = list(indices) + [None] * (self_rank - len(indices))
+
+    # Get values shape
+    values_shape = tuple(values.shape)
+
+    index_vectors = []
+    for i in range(self_rank):
+        if indices[i] is None:
+            # For a full slice along dim i, create a range index [0, self.shape[i]).
+            idx = op.Range(start=0, limit=self.shape[i], delta=1)
```
justinchuby, line 4338:
```suggestion
            idx = op.Range(0, op.Shape(self, i, i+1), 1)
```

Do we want this to be dynamic? If so

AyoubMDL: Sure

onnxscript/function_libs/torch_lib/ops/core.py
AyoubMDL: Actually, I don't need that part as I shrink the values_shape before (so always len(reshape_list) >= len(values_shape))

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4298,14 +4298,78 @@ def aten_index_put(
     <https://github.com/pytorch/pytorch/blob/main/torch/onnx/symbolic_opset11.py#L212>`_.
     """
 
-    # TODO(justinchuby): Handle when indicies has more than one element
-    index = indices[0]
-    new_index = op.Unsqueeze(index, [-1])
+    def _make_reshape_list_broadcastable(reshape_list, values_shape):
```
justinchuby, line 4301: Could you annotate the type of this function? I am a little lost on which one is expected contain dynamic values and which one doesn't.

In general, if x is a dynamic value (tensor), we can only take the rank of it by `len(x.shape)` but not individual shapes via `x.shape` (due to implementation constraints). To keep the dim dynamic we need `op.Shape(x, start=, end=)`. I see a few places where `shape` is assume to contain static (int) dimensions. Could you fix those places? Thanks!

onnxscript/function_libs/torch_lib/ops/core.py
AyoubMDL: Not sure why the test fails [here](https://github.com/microsoft/onnxscript/actions/runs/13590681526/job/37998239141?pr=2061) and passes on other setups. I think because Shape returns a tensor and not a scalar which what Range expect. Maybe adding a Gather would solve that.

justinchuby:
For range, I think we need to Squeeze the tensor to remove its dimension.

onnxscript/function_libs/torch_lib/ops/core.py
AyoubMDL: Should I rewrite the logic of this function assuming that values are dynamic ?

AyoubMDL: Ok, I'll convert this PR to draft while working on it

AyoubMDL:
@justinchuby I wasn’t able to make this work on symbolic tensors. Please feel free to close the draft or continue from here if helpful.

justinchuby:
Thanks. Could you share the issues you have seen? I can take it from here

justinchuby (approved):

justinchuby:
Will merge as-is and create follow up improvements

AyoubMDL:
> Thanks. Could you share the issues you have seen? I can take it from here

- I didn't know how to do the part where I construct the reshape_list using only onnx operations.
- I also tried to adapt the implementation from pytorch https://github.com/pytorch/pytorch/blob/6c3492b49118d9407db39fc5b134c0f96739b831/torch/onnx/symbolic_opset11.py#L206 but there logic doesn't work when I want to handle None indices.

PR #2065, opened 2025-02-19
Enable extraction of rewritten subgraph as model-local function

gramalingam:
Enable extraction of rewritten subgraph as model-local function.

This will enable multi-step rewrite optimizations: eg., map subgraph G1 to new-op1, and then map subgraph G2 containing new-op1 to new-op2, and then inlining can replace any remaining new-op1 (that was not rewritten) by original G1.


onnxscript/rewriter/pattern.py
```diff
@@ -1591,6 +1673,37 @@ def _apply_to_graph_or_function(
...
+                    used_domains: set[str] = set(node.domain for node in original_nodes)
+                    parent_opset_imports = graph_or_function.opset_imports
+                    used_opset_imports = {
+                        k: v for k, v in parent_opset_imports.items() if k in used_domains
+                    }
+
+                    graph = ir.Graph(
+                        inputs, outputs, nodes=nodes, opset_imports=used_opset_imports
+                    )
+                    f = ir.Function(domain, name, overload, graph=graph, attributes={})
```
justinchuby, line 1705:
```suggestion
                    f = ir.Function(domain, name, overload, graph=graph, attributes=())
```

onnxscript/rewriter/pattern.py
```diff
@@ -1591,6 +1673,37 @@ def _apply_to_graph_or_function(
                 # is sufficient for patterns with a single output-node "node", which can serve as the
                 # insertion-point.
                 onnxscript.optimizer.basic_constant_propagation(delta.new_nodes)
+                if rule.as_function:
+                    # Create a function out of a copy of the matched nodes
+                    if len(delta.new_nodes) != 1:
+                        raise ValueError(
+                            "as_function=True is only supported for patterns with a single replacement node."
```
justinchuby, line 1680: Curious why the limitation?

onnxscript/rewriter/pattern.py
```diff
@@ -1529,6 +1532,85 @@ def rewrite(self, op, *args, **kwargs):
...
+            return None
+        if value not in value_map:
+            raise ValueError(f"Value {value} not found in value_map.")
+        return value_map[value]
+
+    def copy_attr_value(attr: ir.Attr | ir.RefAttr) -> ir.Attr | ir.RefAttr:
+        if not isinstance(attr, ir.Attr):
+            raise ValueError("RefAttr not supported.")
+        if attr.type in {ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS}:
+            raise ValueError("Graph attributes not supported.")
```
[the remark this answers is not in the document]
gramalingam: Not yet. It's technically more complicated (even the underlying rewriting, ignoring the current extention) and there are no use-cases currently.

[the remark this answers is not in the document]
justinchuby: Maybe raise an NotImplement error?

onnxscript/rewriter/pattern.py
```diff
@@ -1529,6 +1532,85 @@ def rewrite(self, op, *args, **kwargs):
...
+        new_node = ir.Node(
+            node.domain,
+            node.op_type,
+            new_inputs,
+            new_attributes,
+            overload=node.overload,
+            num_outputs=len(node.outputs),
+            graph=None,
+            name=node.name,
+            doc_string=node.doc_string,  # type: ignore
```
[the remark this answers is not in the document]
gramalingam: Good question: copied the node copying code from inliner, not sure now.

onnxscript/rewriter/pattern.py
justinchuby: I think some version of this note would be helpful as comments for future readers

justinchuby (approved):

onnxscript/rewriter/pattern.py
```diff
@@ -1529,6 +1532,91 @@ def rewrite(self, op, *args, **kwargs):
...
+            raise NotImplementedError("RefAttr not supported.")
+        if attr.type in {ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS}:
+            # No need to support this currently, as rewriting control-flow constructs
+            # is not used and has several challenges.
+            raise NotImplementedError("Graph attributes not supported.")
+        return ir.Attr(attr.name, attr.type, attr.value, doc_string=attr.doc_string)
+
+    def copy_node(node: ir.Node) -> ir.Node:
+        new_inputs = [copy_value(v) for v in node.inputs]
+        new_attributes = [copy_attr_value(v) for v in node.attributes.values()]
```
justinchuby, line 1578: Is this needed? attributes should be immutable as long as they don't have a subgraph in them

onnxscript/rewriter/pattern.py
```diff
@@ -1529,6 +1532,91 @@ def rewrite(self, op, *args, **kwargs):
...
+    def copy_attr_value(attr: ir.Attr | ir.RefAttr) -> ir.Attr | ir.RefAttr:
+        if not isinstance(attr, ir.Attr):
+            # No need to support this currently, as rewriting inside a function is
+            # not used, as it has several challenges.
+            raise NotImplementedError("RefAttr not supported.")
+        if attr.type in {ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS}:
+            # No need to support this currently, as rewriting control-flow constructs
+            # is not used and has several challenges.
+            raise NotImplementedError("Graph attributes not supported.")
+        return ir.Attr(attr.name, attr.type, attr.value, doc_string=attr.doc_string)
```
justinchuby, line 1574:
```suggestion
        return attr
```

maybe just this?

onnxscript/rewriter/pattern.py
gramalingam: I see. If that is guaranteed, there should be no need to copy. (But if we change the classes to support mutation in future, it might become an issue.) I can remove the copy for now.

PR #2069, opened 2025-02-21
[torchlib] Update operator:pow implementation

justinchuby:
Register it to aten_pow instead because exponent may not be a tensor when self is SymInt.

Fixes https://github.com/pytorch/pytorch/issues/147606

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6652,7 +6652,7 @@ def aten_positive(self: TensorType) -> TensorType:
 
 
 @torch_op(
-    ("aten::pow.Tensor_Tensor", "aten::pow.Tensor_Scalar"),
+    ("aten::pow.Tensor_Tensor", "aten::pow.Tensor_Scalar", "_operator::pow"),
     trace_only=True,
 )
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6651,18 +6651,15 @@ def aten_positive(self: TensorType) -> TensorType:
 
 
 @torch_op(
-    ("aten::pow.Tensor_Tensor", "aten::pow.Tensor_Scalar"),
+    ("aten::pow.Tensor_Tensor", "aten::pow.Tensor_Scalar", "_operator::pow"),
```
titaiwangms, line 6654: Should we also update the annotation if exponent may not be a tensor?

justinchuby: Sure

titaiwangms: https://github.com/pytorch/pytorch/blob/af1072ffb6ecbc5c98f9de48a50797112dfe0544/aten/src/ATen/native/native_functions.yaml#L10256

justinchuby: Done. PTAL

titaiwangms (approved):

PR #2077, opened 2025-02-26
Add a couple of variants of patterns in ORT fusions

gramalingam:
Add a couple of variants of patterns in ORT fusions (motivated by Phi4)


onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -9,22 +9,30 @@
 
 
 class SDPA(pattern.RewriteRuleClassBase):
-    def __init__(self, name: str, *, use_mask: bool, pre_scale: bool):
+    def __init__(self, name: str, *, use_mask: bool, pre_scale: bool, use_mul: bool):
```
justinchuby, line 12: Possible to document the options?

onnxscript/rewriter/pattern.py
```diff
@@ -1830,7 +1830,7 @@ def report(self) -> None:
             print(f"Rule: {rule}")
             print(f"Best score: {matches[0].score()}")
             for match in matches:
-                print(f"Status: {match.status}")
+                print(f"Status: {match.status.name}")
```
justinchuby, line 1833: We can make status a String Enum

justinchuby (approved):

onnxscript/rewriter/pattern.py
gramalingam: It is also used as an int in calculating a "score" (for the "best" match).

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Let me do that later. I would first like to get to the stage where we can successfully complete fusion for a core set of LLM models. These options are used in short patterns and, I hope, the patterns clarify the intention. (These are not intended to be used outside the file.)

PR #2083, opened 2025-02-28
Squeeze Reshape Identity optimization

gramalingam:
A recent fix to the translation of pytorch symints introduces a Squeeze=>Reshape pattern that can be optimized away. This PR introduces a rewrite-rule to do this optimization.

TODO (in a separate PR): for now, this optimization needs to be explicitly invoked. This should be done by default. (But there are several other such optimizations that need to be collected and included in the default-rule list.)

### DIFF
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -12,6 +12,26 @@ import onnxscript.rewriter.no_op as no_op
 import onnxscript.rewriter.pattern as orp
 
 
+class SqueezeReshape(orp.RewriteRuleClassBase):
+    """Replaces ``Reshape(Squeeze(x), [-1]])`` with ``Identity(x)`` for 1D x.
+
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -12,6 +12,26 @@
...
+    """Replaces ``Reshape(Squeeze(x), [-1]])`` with ``Identity(x)`` for 1D x.
+
+    This pattern arises from the translation of pytorch symints.
+    """
+
+    def __init__(self):
+        super().__init__("SqueezeReshape1d", remove_nodes=False)
+
+    def pattern(self, op, x):
+        return op.Reshape(op.Squeeze(x), [-1])
```
justinchuby, line 25: Does it work if reshape have multiple users?

gramalingam: Yes. It is not different from any other rewrite-rule in that regard.

PR #2088, opened 2025-03-04
Pass match object to conditional function.

rezaasjd:
Match object can be used to single out a replacement to a pattern that was matched if the user intends to.
This would be helpful since node.name is not passed to replacer functions and as a result user cannot have logic based on the node names that were part of the match.


justinchuby:
@rezaasjd could you share a concrete use case?

PR #2089, opened 2025-03-05
Add Op (Slice - complex) | feat torchlib

titaiwangms:
Fix https://github.com/pytorch/pytorch/issues/147896

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7698,6 +7698,19 @@ def aten_sinh(self: TFloat) -> TFloat:
     return op.Sinh(self)
 
 
+@torch_op(("aten::slice.Tensor"), trace_only=True, complex=True)
+def aten_slice_complex(
+    self: TTensor,
+    dim: int = 0,
+    start: Optional[INT64] = None,
+    end: Optional[INT64] = None,
+    step: Optional[INT64] = None,
+) -> TTensor:
+    """slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)"""
+
+    return aten_slice(self, dim, start, end, step)
+
+
 @torch_op(("aten::slice.Tensor"), trace_only=True)
 def aten_slice(
     self: TTensor,
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7707,7 +7707,9 @@ def aten_slice_complex(
     step: Optional[INT64] = None,
 ) -> TTensor:
     """slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)"""
-
+    if dim < 0:
+        # Account for the complex dimension in ONNX
+        dim = dim - 1
     return aten_slice(self, dim, start, end, step)
 
 
--- a/tests/function_libs/torch_lib/ops_test_data.py
+++ b/tests/function_libs/torch_lib/ops_test_data.py
@@ -2048,6 +2048,7 @@ TESTED_TORCHLIB_OPS: tuple[TorchLibOpInfo, ...] = (
     ),
     TorchLibOpInfo("ops.aten.slice_scatter", core_ops.aten_slice_scatter),
     TorchLibOpInfo("slice", core_ops.aten_slice),
+    TorchLibOpInfo("slice", core_ops.aten_slice_complex, complex=True),
     TorchLibOpInfo(
         "sum",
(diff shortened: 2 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7698,6 +7698,19 @@ def aten_sinh(self: TFloat) -> TFloat:
...
+def aten_slice_complex(
+    self: TTensor,
+    dim: int = 0,
+    start: Optional[INT64] = None,
+    end: Optional[INT64] = None,
+    step: Optional[INT64] = None,
+) -> TTensor:
+    """slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)"""
+
+    return aten_slice(self, dim, start, end, step)
```
xadupre, line 7711: Should we turn negative axes into positive? I don't remember how complex are represented, real with one more dimension or...

justinchuby: That’s a good point

justinchuby (approved):
Do you think negative axes need to be handled like Xavier suggested? Tests didn’t error which was a little unexpected

titaiwangms:
> Do you think negative axes need to be handled like Xavier suggested? Tests didn’t error which was a little unexpected

In this example,

```python
import torch

class ComplexSliceModel(torch.nn.Module):
    def forward(self, x):
        # Convert input to a complex tensor
        x_complex = x.to(torch.complex64)
        # Apply a slice operation on the complex tensor
        return x_complex[:, :2]

model = ComplexSliceModel()
dummy_input = torch.randn(3, 4)

# Verify the model works as expected
print("Model output:", model(dummy_input))

# This call fails due to the slice op on a complex tensor.
torch.onnx.export(model, dummy_input, "complex_slice.onnx", dynamo=True)
```
```
    class GraphModule(torch.nn.Module):
        def forward(self, x: "f32[3, 4]"):
             # File: /home/titaiwang/pytorch/test_slice.py:6 in forward, code: x_complex = x.to(torch.complex64)
            _to_copy: "c64[3, 4]" = torch.ops.aten._to_copy.default(x, dtype = torch.complex64)
            _assert_tensor_metadata = torch.ops.aten._assert_tensor_metadata.default(x, None, None, torch.float32);  x = _assert_tensor_metadata = None
            
             # File: /home/titaiwang/pytorch/test_slice.py:8 in forward, code: return x_complex[:, :2]
            slice_1: "c64[3, 4]" = torch.ops.aten.slice.Tensor(_to_copy, 0, 0, 9223372036854775807);  _to_copy = None
            slice_2: "c64[3, 2]" = torch.ops.aten.slice.Tensor(slice_1, 1, 0, 2);  slice_1 = None
            return (slice_2,)
            
Graph signature: ExportGraphSignature(input_specs=[InputSpec(kind=<InputKind.USER_INPUT: 1>, arg=TensorArgument(name='x'), target=None, persistent=None)], output_specs=[OutputSpec(kind=<OutputKind.USER_OUTPUT: 1>, arg=TensorArgument(name='slice_2'), target=None)])
Range constraints: {}
```

It works because ExportedProgram use two slice ops to slice real and image respectively. I think it means the dim is respected?

justinchuby:
Looks like this is slicing two dimensions. What happens if you did `x_complex[:, :-1]`?

justinchuby:
What happens if you did

`torch.ops.aten.slice.Tensor(slice_1, -1, 0, 2)` ?

Granted this may be unlikely.

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: Done

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7698,6 +7698,21 @@ def aten_sinh(self: TFloat) -> TFloat:
...
+    self: TTensor,
+    dim: int = 0,
+    start: Optional[INT64] = None,
+    end: Optional[INT64] = None,
+    step: Optional[INT64] = None,
+) -> TTensor:
+    """slice.Tensor(Tensor(a) self, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor(a)"""
+    if dim < 0:
+        # Account for the complex dimension in ONNX
+        dim = dim - 1
```
justinchuby, line 7712: I think this could be a little confusing for users when they examine the graph. What do you think if we did `dim += len(self.shape)`?

titaiwangms: OK I can create a follow up

titaiwangms: https://github.com/microsoft/onnxscript/pull/2094

PR #2092, opened 2025-03-07
Generalize MHA pattern

gramalingam:
Generalize the MHA pattern (motivated by the Phi models). Specifically, we remove the initial MatMuls from the pattern (as being unnecessary). Phi uses packed MatMul (Q, K, and V are multiplied using a single MatMul and then sliced).

However, this is not sufficient yet, since Phi also uses partial rotary-embedding, which is not yet supported by the RotaryEmbedding pattern. I will separately work on the extension to the RotaryEmbedding pattern to handle partial embedding.


onnxscript/rewriter/_ir_utils.py
```diff
@@ -124,3 +124,17 @@ def has_rank(value: ir.Value | None, rank: int) -> bool:
...
+
+def get_dim(value: ir.Value | None, dim: int) -> ir.SymbolicDim | int | None:
+    """Returns the value of the given dimension, or None if it is not statically known."""
+    if value is None:
+        return None
+    shape = value.shape
+    if shape is None:
+        return None
+    if dim < 0:
+        dim += shape.rank()
```
justinchuby, line 137: Can this be absorbed into `return shape[dim]`?

onnxscript/rewriter/ort_fusions/_test_utils.py
```diff
@@ -21,6 +22,9 @@ def _save(model, modelpath):
         io.save(model, modelpath)
 
 
+ort_version = packaging.version.Version(onnxruntime.__version__)
```
justinchuby, line 25: Is this used? If so

```suggestion
_ORT_VERSION = packaging.version.Version(onnxruntime.__version__)
```

onnxscript/rewriter/_ir_utils.py
gramalingam: You mean that this is already handled by `shape[dim]` ? I suppose so, but we would need to adjust the condition in the next line to `dim < -shape.rank()` (to avoid an exception in the indexing). I find this as an easy way to remember the condition to check (whether it is < or <= etc.)

justinchuby (approved):

onnxscript/rewriter/_ir_utils.py
justinchuby: I see!

PR #2095, opened 2025-03-10
Fusion for partial rotary embedding

gramalingam:
Add a fusion rule for recognizing partial rotary embedding, along with test case.


onnxscript/rewriter/ort_fusions/_rotary_embedding_models.py
justinchuby, line 1: Is this file for test only? If so I would consider moving it to tests/ so it doesn’t get included in the package

gramalingam: What about the actual test files (all the *_test.py)? Should we be moving them too? This is logically a part of the _test.py, except that it is shared by several. (Also, are you suggesting a new `tests` folder, I don't see a pre-existing one.)

justinchuby: https://github.com/microsoft/onnxscript/tree/main/tests#readme I would leave unit tests with the source; integration tests (that tests bigger units, across multiple methods or with shared components etc.) can be in tests/.

gramalingam: So, I would leave it be here. These are unit tests.

justinchuby (approved):

PR #2106, opened 2025-03-13
A couple of extensions to MHA fusion

gramalingam:
A couple of extensions to MHA fusion:

* One deals with variations in positions-ids. The challenge is to verify that the position-ids used in the two RotaryEmbedding are the same. In some models, they are the same value (by reference). In some models, there is some duplication of the code in computing the 2D position-id from 1D position-id. If we had a common-sub-expression identification/elimination, that would help. For now, just handling it in the pattern itself.

* The second deals with variations in how the last two axes of key are transposed. Some models reshape the input tensor to 3D and do the transpose, while some directly transpose a 4D tensor.


onnxscript/rewriter/ort_fusions/mha.py
```diff
@@ -220,9 +241,10 @@ def rewrite(
         )
 
 
-_rule1 = MultiHeadAttention.rule()
+_mha_4d_transpose = MultiHeadAttention.rule("MHA_3D_Transpose", transpose_4d=True)
```
justinchuby, line 244:
```suggestion
_mha_4d_transpose = MultiHeadAttention.rule("MHA_4D_Transpose", transpose_4d=True)
```

justinchuby (approved):

onnxscript/rewriter/ort_fusions/mha.py
gramalingam: Good catch, thanks!

PR #2108, opened 2025-03-15
Fix Op (convolution) | add nd support to convolution

titaiwangms:
Screenshot 2025-03-14 171007

Ran into a case that aten.convolution.default takes 2D image with [0] as padding, which broke our assumption of it comes with the same rank of nd image.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2074,16 +2074,30 @@ def aten_convolution(
...
+    # 3D: rank=5
+    rank = len(input.shape)
+
+    rank_to_image_d = {3: 1, 4: 2, 5: 3}
+    image_d = rank_to_image_d[rank]
+
     if not isinstance(padding, Sequence):
-        padding = (padding, padding)
+        padding = [padding] * image_d
+    elif len(padding) != image_d:
```
justinchuby, line 2087: Assert len(padding) == 1 in this case?

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: I used len(padding) > 0. I think this way makes more sense?

justinchuby: What happens when len(padding) == 2, for example, when imsage_d == 3; is it a valid case? Down below we take padding[0] which implies all elements after the first element will be ignored.

titaiwangms: Yeah, I take the first one and expand it for the users, because we don't know what aten op would get.

justinchuby: Do we want to raise an explicit error when len is not 1 then, because otherwise there will be silent errors?

titaiwangms: Done. Change it to only support len == 1

justinchuby (approved):

PR #2109, opened 2025-03-17
Turn constant folder and dce into passes

justinchuby:
Turn constant folder and dce into passes to allow them to be used as individual passes in the future.

@gramalingam

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -797,8 +797,7 @@ def _merge_shapes(shape1: ir.Shape | None, shape2: ir.Shape | None) -> ir.Shape 
     return ir.Shape([merge_dims(dim1, dim2) for dim1, dim2 in zip(shape1, shape2)])
 
 
-class ConstantFolder:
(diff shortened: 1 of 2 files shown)


gramalingam (approved):
This looks fine, thanks.

What would be interesting (but is not clear yet to me) is how we use passes effectively and whether that requires other abstractions. For example, I see that the PassManager allows execution of a sequence of passes (iteratively). Would we need the ability to turn a sequence of passes (may be even an iteration N times) into a pass (which can itself then be used as one pass among another sequence of passes)? Essentially a "Sequence of Passes" as a "Pass"? 

Of course, that leads to other ways of composing passes (eg. "pass1 OR pass2": in ORT fusion context, there is the possibility that we try GQA fusion only if MHA fusion fails). Of course, this may end up encoding a full programming-language within passes ... but that may be a bit too much right now, and it seems easier to use Python to express compositions for right now.

justinchuby:
> Essentially a "Sequence of Passes" as a "Pass"?

Sure! PassManager itself can be used as a pass. Something that is specialized for that will work too.

PR #2115, opened 2025-03-20
Cleanup ort transformer fusions

gramalingam:
Cleanup ort transformer-fusions.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/_core.py
+++ b/onnxscript/rewriter/ort_fusions/_core.py
@@ -2,12 +2,8 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
-import os
-import tempfile
-
-import onnx
-
 import onnxscript.ir as ir
+from onnxscript.ir.passes.common import shape_inference
 from onnxscript.optimizer import optimize, remove_unused_nodes
 from onnxscript.rewriter import rewrite
 from onnxscript.rewriter.llama_rule_sets import llama_p0_rule_set
@@ -18,9 +14,13 @@ from onnxscript.rewriter.ort_fusions import (
     softmax,
 )
 from onnxscript.rewriter.ort_fusions.cos_sin_cache import fuse_cos_sin_cache
+from onnxscript.rewriter.ort_fusions.gelu import fuse_gelu
 from onnxscript.rewriter.ort_fusions.mha import fuse_mha
 from onnxscript.rewriter.ort_fusions.rms_normalization import fuse_rms_normalization
-from onnxscript.rewriter.ort_fusions.rotary_embedding import fuse_rotary_embedding
+from onnxscript.rewriter.ort_fusions.rotary_embedding import (
+    fuse_partial_rotary_embedding,
+    fuse_rotary_embedding,
+)
 from onnxscript.rewriter.ort_fusions.sdpa import fuse_sdpa
 from onnxscript.rewriter.ort_fusions.skip_normalization import fuse_normalization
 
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -26,9 +32,32 @@
...
+
+# Preliminary optimizations before applying the transformer fusions.
+# TODO: There are some potential redundancies below. Can be targeted for optimization
+# once we have robust fusion.
+def _pre_optimize(model):
+    optimize(model)
+    # TODO: Do we need this dependence on ONNX's partial-data-propagation? There are some
+    # extra shape-propagation and partial-data-propagation rules in ONNX that are not yet
+    # incorporated in our optimizer.
+    with tempfile.TemporaryDirectory() as tmpdir:
```
justinchuby, line 46: Is it possible to instead run shape inference on a model without initializers, then re-assemble it into an ir Model? This way we don’t need to write to disk

justinchuby: We can create a shape inference pass that does this

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -26,9 +32,32 @@
     *fused_matmul_rule_sets.fused_matmul_rule_sets(),
 ]
 
+_extra_opt_rules = llama_p0_rule_set()
```
shubhambhokare1, line 35: maybe we can make this a different structure if we want to apply multiple rule sets in the future?

For example,
`_extra_opt_rules = [llama_p0_rule_set()]`

and then loop over them,
`
for rule in _extra_opt_rules:
      rule.apply_to_model(model)
`

gramalingam: Not sure I understand. I think a few things to be done are: what's called llama_p0_rule_set is actually generic, we should be adding them to the list of generic optimizations, so we won't even need this here. It should become part of the default optimizer anyway.

We may want, as a convenience, the ability to merge multiple rule sets (define a new ruleset as the union of other rulesets).

onnxscript/rewriter/ort_fusions/_core.py
justinchuby: https://github.com/microsoft/onnxscript/pull/2117

justinchuby:
I implemented https://github.com/microsoft/onnxscript/pull/2117 so we don't need to write to disk hopefully

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -26,9 +32,32 @@
...
+def _pre_optimize(model):
+    optimize(model)
+    # TODO: Do we need this dependence on ONNX's partial-data-propagation? There are some
+    # extra shape-propagation and partial-data-propagation rules in ONNX that are not yet
+    # incorporated in our optimizer.
+    with tempfile.TemporaryDirectory() as tmpdir:
+        infile = os.path.join(tmpdir, "in.onnx")
+        outfile = os.path.join(tmpdir, "out.onnx")
+        ir.save(model, infile)
+        onnx.shape_inference.infer_shapes_path(infile, outfile, True, True, True)
```
shubhambhokare1, line 50: #2117 Merged, so we can use the shape inference call from there

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -26,9 +32,32 @@
...
+        model = ir.load(outfile)
     optimize(model)
+    # TODO: The extra optimization rules should be included in the base optimization
+    _extra_opt_rules.apply_to_model(model)
```
justinchuby, line 54:
```suggestion
```

https://github.com/microsoft/onnxscript/pull/2124

shubhambhokare1 (approved):

PR #2116, opened 2025-03-20
Add SDPA fusion unit test case

gramalingam:
Add SDPA fusion unit test case

### DIFF
--- a/onnxscript/rewriter/ort_fusions/sdpa_test.py
+++ b/onnxscript/rewriter/ort_fusions/sdpa_test.py
@@ -17,18 +17,18 @@ from onnxscript.onnx_opset import opset18 as op
 from onnxscript.onnx_types import FLOAT
 from onnxscript.rewriter.ort_fusions.sdpa import fuse_sdpa
 
-batchsize = 2
-num_heads = 4
-seq_len = 8
-headsize = 128
-scale_factor = math.sqrt(headsize)
-sqrt_scale_factor = math.sqrt(scale_factor)
+B = 2  # batch size
+N = 4  # number of heads
+S = 8  # sequence length
+H = 128  # head size
+SCALE_FACTOR = math.sqrt(H)
+SQRT_SCALE_FACTOR = math.sqrt(SCALE_FACTOR)
 
 
 @script()
 def _masked_pre_div_sdpa_script(query, key, value, mask):
     key_transposed = op.Transpose(key, perm=[0, 1, 3, 2])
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/sdpa_test.py
```diff
@@ -0,0 +1,90 @@
...
+import numpy
+
+import onnxscript.ir as ir
+import onnxscript.optimizer
+from onnxscript import script
+from onnxscript.onnx_opset import opset18 as op
+from onnxscript.onnx_types import FLOAT
+from onnxscript.rewriter.ort_fusions.sdpa import fuse_sdpa
+
+batchsize = 2
```
justinchuby, line 20:
```suggestion
BATCH_SIZE = 2
```
here and below: all caps for constants?

onnxscript/rewriter/ort_fusions/sdpa_test.py
```diff
@@ -0,0 +1,90 @@
...
+    masked_attn_score = op.Add(attn_score, mask)
+    attn_weight = op.Softmax(masked_attn_score, axis=-1)
+    attn_output = op.MatMul(attn_weight, value)
+    return attn_output
+
+
+class _MaskedPreDivSDPATestCase:
+    def get_onnx_model(self):
+        if not hasattr(self, "_onnx_model"):
+            qkv_type = FLOAT[batchsize, num_heads, seq_len, headsize]
```
shubhambhokare1, line 44: Maybe we can add a test model for 3d inputs? I'm curious to see how the check function behaves for 4d/3d inputs

onnxscript/rewriter/ort_fusions/sdpa_test.py
gramalingam: I am tempted to use B, S, N, H ... which is, hopefully, not cryptic in this context (with expansion in comments). Looks better for the shapes down below anyway.

onnxscript/rewriter/ort_fusions/sdpa_test.py
gramalingam: Sure, maybe you can add it as part of your extension? First, the checks will need to be added to the rule (as you can see in the comments there, they are currently missing.)

shubhambhokare1: Sounds good will do.

shubhambhokare1 (approved):

PR #2117, opened 2025-03-21
[IR] Create a shape inference pass using onnx shape inference

justinchuby:
It handles large models by removing the initializers before sending the model to onnx shape inference.

### DIFF
--- a/onnxscript/ir/passes/common/shape_inference.py
+++ b/onnxscript/ir/passes/common/shape_inference.py
@@ -4,7 +4,10 @@
 
 from __future__ import annotations
 
-__all__ = ["ShapeInferencePass"]
+__all__ = [
+    "ShapeInferencePass",
+    "infer_shapes",
+]
 
 import logging
 
@@ -13,6 +16,10 @@ import onnx
 from onnxscript import ir
 
 logger = logging.getLogger(__name__)
+
+# Temporarily remove initializers larger than this size to keep model size down
+# for the onnx.shape_inference call because it needs to serialize the model
+_BIG_TENSOR_SIZE_LIMIT = 1000  # 1KB
 
 
 class ShapeInferencePass(ir.passes.PassBase):
@@ -54,7 +61,10 @@ class ShapeInferencePass(ir.passes.PassBase):
                 initializer.dtype = initializer.const_value.dtype
             if initializer not in model.graph.inputs:
                 model.graph.inputs.append(initializer)
-            initializer.const_value = None
+            if initializer.const_value.nbytes > _BIG_TENSOR_SIZE_LIMIT:
+                # Temporarily remove the initializer value to reduce model size
+                # for onnx.shape_inference
+                initializer.const_value = None
         model.graph.initializers.clear()
 
         # Perform shape inference
@@ -96,3 +106,26 @@ class ShapeInferencePass(ir.passes.PassBase):
         # Even though modified, we know the pass will not change the model if we ran it again.
         # So set modified to False
         return ir.passes.PassResult(inferred_model, modified=False)
+
+
+def infer_shapes(
+    model: ir.Model,
+    *,
+    check_type: bool = True,
+    strict_mode: bool = True,
+    data_prop: bool = True,
+) -> ir.Model:
+    """Perform shape inference on the model.
+
+    Args:
+        model: The model to perform shape inference on.
+        check_type: If True, check the types of the inputs and outputs.
+        strict_mode: If True, use strict mode for shape inference.
+        data_prop: If True, use data propagation for shape inference.
+
+    Returns:
+        The model with shape inference applied.
+    """
+    return ShapeInferencePass(
+        check_type=check_type, strict_mode=strict_mode, data_prop=data_prop
+    )(model).model
--- a/onnxscript/ir/passes/common/shape_inference.py
+++ b/onnxscript/ir/passes/common/shape_inference.py
@@ -74,7 +74,7 @@ class ShapeInferencePass(ir.passes.PassBase):
             model.graph.inputs.clear()
             model.graph.inputs.extend(inputs)
 
-        # Restore the original initializer values for the new (inferred) model
+        # Add the original initializer tensors to the new (inferred) model
         for new_input in inferred_model.graph.inputs:
             # Assign the tensors back to the initializers
             if new_input.name in initializer_names:
@@ -82,9 +82,9 @@ class ShapeInferencePass(ir.passes.PassBase):
                 inferred_model.graph.register_initializer(new_input)
 
         # Remove the inputs that were added
-        inputs = inferred_model.graph.inputs[:original_inputs_len]
+        new_inputs = inferred_model.graph.inputs[:original_inputs_len]
         inferred_model.graph.inputs.clear()
-        inferred_model.graph.inputs.extend(inputs)
+        inferred_model.graph.inputs.extend(new_inputs)
         # Even though modified, we know the pass will not change the model if we ran it again.
         # So set modified to False
         return ir.passes.PassResult(inferred_model, modified=False)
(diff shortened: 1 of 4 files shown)


onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        original_inputs_len = len(model.graph.inputs)
+        initializer_names = {v.name for v in initializer_values}
+
+        # Turn the initializers into inputs and clear the initializers
+        # to limit the model size
+        for initializer in initializer_values:
+            # Make sure the initializer has its shape/type set
+            assert initializer.const_value is not None
+            if initializer.shape is None:
+                initializer.shape = initializer.const_value.shape
```
justinchuby, line 52:
```suggestion
                initializer.shape = initializer.const_value.shape  # type: ignore[assignment
```

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        try:
+            proto = ir.serde.serialize_model(model)
+            proto = onnx.shape_inference.infer_shapes(
+                proto,
+                check_type=self.check_type,
+                strict_mode=self.strict_mode,
+                data_prop=self.data_prop,
+            )
+            inferred_model = ir.serde.deserialize_model(proto)
+        except Exception:
```
justinchuby, line 70:
```suggestion
        except Exception:  # pylint: disable=broad-exception-caught
```

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        original_inputs_len = len(model.graph.inputs)
+        initializer_names = {v.name for v in initializer_values}
+
+        # Turn the initializers into inputs and clear the initializers
+        # to limit the model size
+        for initializer in initializer_values:
+            # Make sure the initializer has its shape/type set
+            assert initializer.const_value is not None
+            if initializer.shape is None:
+                initializer.shape = initializer.const_value.shape  # type: ignore[assignment
```
justinchuby, line 52:
```suggestion
                initializer.shape = initializer.const_value.shape  # type: ignore[assignment]
```

onnxscript/ir/passes/common/shape_inference_test.py
```diff
@@ -0,0 +1,48 @@
...
+                nodes=[add_node],
+                opset_imports={"": 20},
+            ),
+            ir_version=10,
+        )
+        self.assertIsNone(add_node.outputs[0].shape)
+        self.assertIsNone(add_node.outputs[0].dtype)
+
+        # Perform shape inference
+        result = shape_inference.ShapeInferencePass()(model)
```
gramalingam, line 39: may be useful to expose a utility function so that this looks as close as possible to calling onnx's shape inference ... that is, so user can write something like below:
```py
result = shape_inference.infer_shapes(model, check_type=True, strict_mode=False, data_prop=True)
```

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,98 @@
...
+            # Assign the tensors back to the initializers
+            if new_input.name in initializer_names:
+                new_input.const_value = tensors[new_input.name]
+                inferred_model.graph.register_initializer(new_input)
+
+        # Remove the inputs that were added
+        new_inputs = inferred_model.graph.inputs[:original_inputs_len]
+        inferred_model.graph.inputs.clear()
+        inferred_model.graph.inputs.extend(new_inputs)
+        # Even though modified, we know the pass will not change the model if we ran it again.
```
gramalingam, line 96: Seems like we need some thoughts/documentation on what `modified` means ... looks like it could have several different meanings (potentially distinct). Another dimension here is the `in_place` ... so, if in_place is false, the incoming model is guaranteed not to be modified? so, this `modified` field (when in_place=False) could refer to (a) is outgoing model same as incoming model, or (b) will running this pass again on outgoing model produce a different result.

Seems to me (a) is better? Also: we can't guarantee (b) without making some assumptions about ONNX's shape-inference implementation details.

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        self.data_prop = data_prop
+
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        # Store the original initializer values so they can be restored
+        initializer_values = tuple(model.graph.initializers.values())
+        tensors = {v.name: v.const_value for v in initializer_values}
+        original_inputs_len = len(model.graph.inputs)
+        initializer_names = {v.name for v in initializer_values}
+
+        # Turn the initializers into inputs and clear the initializers
```
gramalingam, line 46: Two points to consider: (a) ONNX does allow large-tensors to be stored as attributes in nodes, and (b) What about small tensor-constants (which can be useful for shape-inference to produce better results)?

So, it would be useful if small tensors are NOT removed from the serialized model that is input to shape-inference (for (b))

For attributes: removing them here is tricky, so may need to handle it some other way (as external tensors?)

justinchuby: (b) sounds good. I will add a threshold like we do in ir save
(a) is doable too but depending on how shape inference treats them. I propose we handle them when we see a need?

onnxscript/ir/passes/common/shape_inference.py
justinchuby: It's intended to be used as (b) (i.e. when to early stop if we run iterate over passes a few times) but since there is no way we know (b) we can only use (a) when there is not enough information I think?

onnxscript/ir/passes/common/shape_inference_test.py
justinchuby: Done

onnxscript/ir/passes/common/shape_inference.py
justinchuby: Done (b)

onnxscript/ir/passes/common/shape_inference_test.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import annotations
+
+import unittest
+
+from onnxscript import ir
+from onnxscript.ir.passes.common import shape_inference
+
+
+class TestShapeInference(unittest.TestCase):
+    def test_shape_inference(self):
```
shubhambhokare1, line 12: maybe we could create a test case for model with initializers?

justinchuby: Done, thanks!

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -0,0 +1,132 @@
...
+
+import onnx
+
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+# Temporarily remove initializers larger than this size to keep model size down
+# for the onnx.shape_inference call because it needs to serialize the model
+_BIG_TENSOR_SIZE_LIMIT = 1000  # 1KB
```
gramalingam, line 22: nit: should this be a parameter, with a default value, of a ShapeInferencePass? likely won't matter in near term, but somebody might want to experiment with it some time.

gramalingam (approved):
LGTM, thanks!

onnxscript/ir/passes/common/shape_inference.py
justinchuby: Thanks- I suppose we can expose it as needed when there is a need to keep the public apis limited for now.

onnxscript/ir/passes/common/shape_inference.py
justinchuby: Done. Properly set modified field

PR #2119, opened 2025-03-22
[Pass] Graph extractor pass

justinchuby:
* Added a new pass for extracting subgraphs from a graph (`ExtractGraphPass`) in `onnxscript/ir/passes/common/graph_extration.py`. This pass allows for the extraction of a sub-model defined by specific input and output tensor names.

PR #2120, opened 2025-03-22
[IR] Improve pass infra

justinchuby:
1. Run invariant functions `requires` and `ensures` by default at Pass `__call__` to match pytorch's pass behavior. This means the invariants cannot be too expensive because they are always checked.
2. Make PassManager a `Pass` so that it can be composed.
3. Add `changes_input` attribute to indicate if the input is changed. Turn two class attributes into properties for them to be dynamic. Combining the two attributes we can tell if a pass is destructive. For now the properties are unused but they will become useful when we want to have a better guard on pass usage etc.
4. Create `Sequential`, `InPlacePass`, `FunctionalPass` to help users create passes.

### DIFF
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -225,13 +225,14 @@ class PassManager(Sequential):
     Attributes:
         passes: The passes to run.
         steps: The number of times to run the passes.
+        early_stop: Whether to stop running the passes if the graph stops changing.
     """
 
     def __init__(
         self,
         passes: Sequence[PassBase],
         steps: int = 1,
-        early_stop=True,
+        early_stop: bool = True,
     ):
         # TODO(justinchuby): Implement constraints
         super().__init__(*passes)
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -176,14 +176,16 @@ class Sequential(PassBase):
         if not passes:
             raise ValueError("Sequential must take at least one pass")
         self.passes = passes
+        self._in_place = all(pass_.in_place for pass_ in passes)
+        self._changes_input = self.passes[0].changes_input or self.passes[0].in_place
 
     @property
     def in_place(self) -> bool:
-        return all(pass_.in_place for pass_ in self.passes)
+        return self._in_place
 
     @property
     def changes_input(self) -> bool:
-        return self.passes[0].changes_input or self.passes[0].in_place
+        return self._changes_input
 
     def call(self, model: ir.Model) -> PassResult:
         modified = False
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -70,7 +70,11 @@ class PassBase(abc.ABC):
 
     @property
     def in_place(self) -> bool:
-        """Whether the pass modifies the model in place."""
+        """Whether the pass modifies the model in place.
+
+        If True, the pass will return the same model object that was passed in.
+        If False, the pass will return a new model object.
+        """
         return True
 
     @property
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -23,7 +23,6 @@ __all__ = [
     "Sequential",
     "InPlacePass",
     "OutOfPlacePass",
-    "DestructivePass",
     "PassManager",
     "PassResult",
     # Errors
@@ -168,18 +167,6 @@ class OutOfPlacePass(PassBase):
     @property
     def changes_input(self) -> bool:
         return False
-
-
-class DestructivePass(PassBase):
-    """A pass that modifies the input model and returns a new model."""
-
-    @property
-    def in_place(self) -> bool:
-        return False
-
-    @property
-    def changes_input(self) -> bool:
-        return True
 
 
 class Sequential(PassBase):
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -20,6 +20,10 @@ from typing import Sequence
 
 __all__ = [
     "PassBase",
+    "Sequential",
+    "InPlacePass",
+    "OutOfPlacePass",
+    "DestructivePass",
     "PassManager",
     "PassResult",
     # Errors
@@ -69,8 +73,9 @@ class PassBase(abc.ABC):
     """Base class for all passes."""
 
     @property
+    @abc.abstractmethod
     def in_place(self) -> bool:
-        """Whether the pass modifies the model in place.
+        """Whether the pass modifies the model in place and returns it.
 
         If True, the pass will return the same model object that was passed in.
         If False, the pass will return a new model object.
@@ -78,9 +83,18 @@ class PassBase(abc.ABC):
         return True
 
     @property
+    @abc.abstractmethod
+    def changes_input(self) -> bool:
+        """Whether the pass modifies input model."""
+        return True
+
+    @property
     def destructive(self) -> bool:
-        """Whether the pass will destroy the input model when ``in_place=False``."""
-        return False
+        """Whether the pass will destroy the input model when ``in_place=False``.
+
+        A pass is destructive if it is not in place and it modifies the input model.
+        """
+        return not self.in_place and self.changes_input
 
     def __call__(self, model: ir.Model) -> PassResult:
         # Check preconditions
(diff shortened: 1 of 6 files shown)


justinchuby:
@gramalingam

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +146,112 @@ def ensures(self, model: ir.Model) -> None:
...
+    @property
+    def in_place(self) -> bool:
+        return False
+
+    @property
+    def changes_input(self) -> bool:
+        return False
+
+
+class DestructivePass(PassBase):
```
gramalingam, line 173: Do we really need this? Is there any good example where this is useful? Seems to me it is better to encourage users to use just one of above two ... (which may mean we don't need these two boolean properties) ... I can see the value of "analysis" passes that compute some information but don't modify the model, but not sure whether those are within the scope of "passes" as designed ...

justinchuby: https://github.com/microsoft/onnxscript/pull/2119 this is an example where the pass is DestructivePass. Maybe possible to implement it as a copy as well? I can remove the class to discourage the use though.

justinchuby: Done

onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -22,12 +22,9 @@
 _BIG_TENSOR_SIZE_LIMIT = 1000  # 1KB
 
 
-class ShapeInferencePass(ir.passes.PassBase):
+class ShapeInferencePass(ir.passes.OutOfPlacePass):
```
gramalingam, line 25: Just wondering aloud ... we are forced to make this OutOfPlace because of the existing API (to ONNX's shape inference) ... but that does come with some disadvantages ... eg., if we ever store extra information in the IR (that doesn't get serialized and is lost) ... even if some user/pass uses a dictionary mapping values/nodes in IR to some extra information ... should we, in the long run, prefer in-place? We could potentially do it (at least in the long run) ...

justinchuby: I think being out of place is ok. The issues arise when we serialize. In the long run if we don't need to serialize then the info are more easily preserved, if we copy the objects properly.

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +145,100 @@ def ensures(self, model: ir.Model) -> None:
...
+        self.early_stop = early_stop
 
-    def __call__(self, model: ir.Model) -> PassResult:
+    def call(self, model: ir.Model) -> PassResult:
         """Run the set of passes `steps` number of times or until the graph stops changing."""
         overall_modified = False
         for step in range(self.steps):
-            step_result = self._run_one_step(model, step)
+            try:
+                step_result = super().__call__(model)
```
shubhambhokare1, line 234: would this be super.call()?

justinchuby: `__call__` is intended because it includes invariant checks.

onnxscript/ir/passes/_pass_infra.py
gramalingam: > #2119 this is an example where the pass is DestructivePass. Maybe possible to implement it as a copy as well? I can remove the class to discourage the use though.
> 
> (but the boolean properties are still useful. For example, we can turn a destructive pass to an out of place pass by chaining a "Copy" pass with the destructive pass automatically by looking at these two properties)

I see. I think I would classify #2219 as an example of an in-place pass that returns some extra information (the extracted sub-model) ... and a version that returns only the extracted sub-model without modifying the input-model could be an out-of-place pass (though, in general, we might need the ability to return values other than a model for more complex analysis passes).

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +145,100 @@ def ensures(self, model: ir.Model) -> None:
...
+            except Exception as e:
+                raise PassError(f"An error occurred at step {step}") from e
             model = step_result.model
             modified = step_result.modified
             overall_modified = overall_modified or modified
             # If the graph no longer changes, then we can stop running these passes
-            if not modified:
+            if not modified and self.early_stop:
```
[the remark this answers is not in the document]
justinchuby: Good catch thanks. Let me fix that

onnxscript/ir/passes/_pass_infra.py
justinchuby: Although: the two properties are still useful I suppose: When we compose passes with into a sequence, whether a pass is out of place or destructive requires some careful logic: https://github.com/microsoft/onnxscript/pull/2120/files#diff-70c7e5b3422f4daaf1611d4f76578c96e4c5894cced3d51718efa0290219f7f5R180-R186

onnxscript/ir/passes/_pass_infra.py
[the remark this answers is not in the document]
justinchuby: I just checked again: I think the behavior is good: After calling `super().__call__(model)`, all passes would have been run once because `super()` is `Sequential`

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +145,102 @@ def ensures(self, model: ir.Model) -> None:
...
+
+class Sequential(PassBase):
+    """Run a sequence of passes in order."""
+
+    def __init__(self, *passes: PassBase):
+        if not passes:
+            raise ValueError("Sequential must take at least one pass")
+        self.passes = passes
+        self._in_place = all(pass_.in_place for pass_ in passes)
+        self._changes_input = self.passes[0].changes_input or self.passes[0].in_place
```
titaiwangms, line 180: Changes_input looks exactly the same as in_place currently.

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +145,102 @@ def ensures(self, model: ir.Model) -> None:
...
+    @property
+    def in_place(self) -> bool:
+        return True
+
+    @property
+    def changes_input(self) -> bool:
+        return True
+
+
+class OutOfPlacePass(PassBase):
```
gramalingam, line 160: nit: I wonder if `FunctionalPass` would be a better name ... something that conveys the does-not-modify-input nature of the pass

justinchuby: Sure!

justinchuby: Done

onnxscript/ir/passes/_pass_infra.py
justinchuby: Do you mean in a PassBase? They will look different when a user changes them or when passes are composed in Sequential (here)

titaiwangms: But I don't see a setter though. And why is changes_inputs is decided by the first pass?

justinchuby: The properties are declared at init time. Users can subclass PassBase and define them.

The reason changes_inputs by the first pass is that if the first pass is either in-place, or if it is not designed to be in-place but somehow changes the input (destructive), this pass sequence will change inputs.

The relationship of the two properties is:

in_place x changes_inputs => in place
not in_place x changes_inputs => destructive
in_place x not changes_inputs => not possible, invalid
not in_place x not changes_inputs => functional

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +145,102 @@ def ensures(self, model: ir.Model) -> None:
...
+
+class Sequential(PassBase):
+    """Run a sequence of passes in order."""
+
+    def __init__(self, *passes: PassBase):
+        if not passes:
+            raise ValueError("Sequential must take at least one pass")
+        self.passes = passes
+        self._in_place = all(pass_.in_place for pass_ in passes)
+        self._changes_input = self.passes[0].changes_input or self.passes[0].in_place
```
justinchuby, line 180:
```suggestion
        # The reason changes_inputs is decided by the first pass is that if the first pass is either in-place,
        # or if it is not designed to be in-place but somehow changes the input (destructive),
        # this pass sequence will change inputs.
        self._changes_input = self.passes[0].changes_input or self.passes[0].in_place
```

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -97,76 +158,105 @@ def ensures(self, model: ir.Model) -> None:
...
+    """Run a sequence of passes in order."""
+
+    def __init__(self, *passes: PassBase):
+        if not passes:
+            raise ValueError("Sequential must take at least one pass")
+        self.passes = passes
+        self._in_place = all(pass_.in_place for pass_ in passes)
+        # The reason changes_inputs is decided by the first pass is that if the first pass is either in-place,
+        # or if it is not designed to be in-place but somehow changes the input (destructive),
+        # this pass sequence will change inputs.
```
titaiwangms, line 195: I mean would second or other passes that changes inputs after the first pass?

justinchuby: If the first pass is functional, the second pass will take the new model that the first pass returns, which means the second pass has no chance to affect the input model.

titaiwangms: But if the first pass is side-effect only pass. Shouldn't we check the following passes?

justinchuby: Good point. In that case we just assume it changes the model for now. I think that's ok?

onnxscript/ir/passes/_pass_infra.py
```diff
@@ -68,14 +71,72 @@ class PassResult:
 class PassBase(abc.ABC):
     """Base class for all passes.
 
-    Class attributes:
-        in_place: Whether the pass modifies the model in place.
+
```
xadupre, line 74: Is it possible to add a paragraph in the documentation of the exporter (torch.onnx.export) to mention the list of passes applied to the fx graph?

justinchuby:
I will merge since this is the foundation for some other PRs. Please feel free to comment and I will address in a follow up

onnxscript/ir/passes/_pass_infra.py
justinchuby: Actually you were right. Calling `__call__` creates an infinite loop. We don't need to `__call__` becase there is not invariants and we need to call the proper method in Sequential.

PR #2124, opened 2025-03-24
[rewriter] Enable llama rule sets

titaiwangms:
Enable llama_rule_sets. We might need to come up with a better name.

Follow-up issue: https://github.com/microsoft/onnxscript/issues/2128

### DIFF
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -291,7 +291,7 @@ def llama_p0_rule_set() -> orp.RewriteRuleSet:
     """
     return orp.RewriteRuleSet(
         [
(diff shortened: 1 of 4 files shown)


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -296,7 +296,7 @@ def llama_p0_rule_set() -> orp.RewriteRuleSet:
             no_op.add_0_rule,
             no_op.add_0_rule,
             no_op.div_by_1_rule,
```
justinchuby, line 298: There are `no_op.rules.rules` enabled in the default rules. Are these repetitive?

shubhambhokare1 (approved):

justinchuby (approved):

PR #2126, opened 2025-03-24
[IR] Convenience constructor for Node

justinchuby:
Create a convenience constructor for `Node`. Refactor the constructors to a separate module.

## Motivation

Currently users when interacting with the IR needs to use the raw `ir.Node` constructor for creating nodes. This constructor is designed for performance and not ease-of-use. For users I created a new `ir.node` that exposes a more natural calling style that supports plain python values as attributes and an optional `domain` argument.


onnxscript/ir/_convenience/_constructors_test.py
```diff
@@ -1,20 +1,20 @@
...
+from onnxscript.ir._convenience import _constructors
 
 
 class ConvenienceTest(unittest.TestCase):
     def test_tensor_accepts_torch_tensor(self):
         import torch as some_random_name  # pylint: disable=import-outside-toplevel
 
         torch_tensor = some_random_name.tensor([1, 2, 3])
-        tensor = _convenience.tensor(torch_tensor)
+        tensor = _constructors.tensor(torch_tensor)
```
shubhambhokare1, line 17: nit: Maybe a test showing usage of _constructors,node

justinchuby: Added doctests

justinchuby:
> Would be better to include the motivation of this PR, and a demo of constructing a node and a tensor.

Done by updating the PR description and added doctests

shubhambhokare1 (approved):
LGTM, thanks

PR #2132, opened 2025-03-25
Add Gelu Tanh fusion rule

gramalingam:
Add Gelu Tanh fusion rule

PR #2134, opened 2025-03-25
[rewriter | torchlib] respect ops order in torchscript graph

titaiwangms:
This helps us to match the optimization pattern in https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py

ref: https://github.com/microsoft/onnxscript/pull/2132#discussion_r2013039689

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -487,8 +487,8 @@ def _aten_gelu_approximate_none(self: TReal) -> TReal:
     inner = op.Div(self, 1.4142135623730951)
     erf = op.Erf(inner)
     inner = op.Add(erf, 1)
-    inner = op.Mul(self, inner)
-    result = op.Mul(0.5, inner)
+    inner = op.Mul(0.5, inner)
+    result = op.Mul(self, inner)
     return result
 
 
@@ -505,8 +505,8 @@ def _aten_gelu_approximate_tanh(self: TReal) -> TReal:
     inner = op.Mul(op.Sqrt(two_over_pi), inner)
     inner = op.Tanh(inner)
     inner = op.Add(inner, 1)
-    inner = op.Mul(self, inner)
-    result = op.Mul(0.5, inner)
+    inner = op.Mul(0.5, inner)
+    result = op.Mul(self, inner)
     return result
 
 
--- a/onnxscript/rewriter/ort_fusions/gelu.py
+++ b/onnxscript/rewriter/ort_fusions/gelu.py
@@ -20,8 +20,8 @@ class GeluTanhFusion(pattern.RewriteRuleClassBase):
         t4 = op.Mul(_sqrt_two_over_pi, t3)
         t5 = op.Tanh(t4)
         t6 = op.Add(t5, 1)
-        t7 = op.Mul(x, t6)
-        result = op.Mul(0.5, t7)
+        t7 = op.Mul(0.5, t6)
+        result = op.Mul(x, t7)
         return result
 
     def rewrite(self, op, x):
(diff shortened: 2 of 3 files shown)


shubhambhokare1 (commented):
I see that based on https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py

The previous order matches the fuse_4 pattern and the new order matches the fuse_1 pattern, just a general question whether it might be a good idea to support both these patterns, maybe via using the check function/control parameter?

titaiwangms:
> I see that based on https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py
> 
> The previous order matches the fuse_4 pattern and the new order matches the fuse_1 pattern, just a general question whether it might be a good idea to support both these patterns, maybe via using the check function/control parameter?

Hmm you are right. I missed the 4th pattern. I wonder how it mismatched then..

shubhambhokare1:
> > I see that based on https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py
> > The previous order matches the fuse_4 pattern and the new order matches the fuse_1 pattern, just a general question whether it might be a good idea to support both these patterns, maybe via using the check function/control parameter?


Actually sorry I meant the fuse_2 pattern.

Fuse_1
` +---------------------------+
    |                                        |
    |                                        v
  [root] -> Pow --> Mul -> Add  --> Mul --> Tanh --> Add --> Mul
    |       (Y=3)    (B=0.0447...)       (B=0.7978...)              (B=1)          ^
    |                                                                                                   |
    +------> Mul(B=0.5)----------------------------------------------+`
    
Fuse_2

Fuse Gelu with tanh into one node:
    +---------------------------+
    |                                          |
    |                                          v
  [root] -> Pow -> Mul-> Add -> Mul -> Tanh -> Add -> Mul(B=0.5)->Mul-->
    |           (Y=3)   (B=0.0447...)    (B=0.7978...)         (B=1)                           ^
    |                                                                                                                |
    +---------------------------------------------------------------------------+

shubhambhokare1:
> > I see that based on https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py
> > The previous order matches the fuse_4 pattern and the new order matches the fuse_1 pattern, just a general question whether it might be a good idea to support both these patterns, maybe via using the check function/control parameter?
> 
> ~Hmm you are right. I missed the 4th pattern. I wonder how it mismatched then..~
> 
> No the previous pattern does not match anything. It has Pow but fuse_4 does not.

Actually, sorry I meant the fuse_2 pattern, seems like only the last two Mul are switched

https://github.com/microsoft/onnxruntime/blob/9dbfee91ca9c2ba2074d19805bb6dedccedbcfe3/onnxruntime/python/tools/transformers/fusion_fastgelu.py#L31
https://github.com/microsoft/onnxruntime/blob/9dbfee91ca9c2ba2074d19805bb6dedccedbcfe3/onnxruntime/python/tools/transformers/fusion_fastgelu.py#L139

shubhambhokare1:
> > I see that based on https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_fastgelu.py
> > The previous order matches the fuse_4 pattern and the new order matches the fuse_1 pattern, just a general question whether it might be a good idea to support both these patterns, maybe via using the check function/control parameter?
> 
> ~Hmm you are right. I missed the 4th pattern. I wonder how it mismatched then..~
> 
> No the previous pattern does not match anything. It has Pow but fuse_4 does not.

You're right, Although I wonder about
https://github.com/microsoft/onnxruntime/blob/9dbfee91ca9c2ba2074d19805bb6dedccedbcfe3/onnxruntime/python/tools/transformers/fusion_fastgelu.py#L31
https://github.com/microsoft/onnxruntime/blob/9dbfee91ca9c2ba2074d19805bb6dedccedbcfe3/onnxruntime/python/tools/transformers/fusion_fastgelu.py#L139

and if we would need cover the second pattern as well

titaiwangms:
Screenshot 2025-03-25 165915

The second pattern is still different from our op implementation. I guess we just need to follow torchscript ops order.

justinchuby (approved):
Thanks!

shubhambhokare1 (approved):

gramalingam (approved):

PR #2136, opened 2025-03-26
A couple of ort fusion fixes

gramalingam:
* Enable the use of SDPA fusions, along with undoing it when it does not lead to some subsequent final fusion (such as MHA or GQA).
* Fix the use of constants in extracted functions from fusion.
* Fix the use of Gelu instead of FastGelu in the new fusion introduced earlier today.


onnxscript/rewriter/ort_fusions/_test_utils.py
```diff
@@ -33,7 +33,7 @@ def ort_run(model_name: str, model, inputs):
     return session.run(None, inputs)
 
 
-def assert_allclose(outputs, expected_outputs, rtol=1e-2, atol=1e-2):
+def assert_allclose(outputs, expected_outputs, rtol=1e-4, atol=1e-4):
     for i, (baseline_output, optimized_output) in enumerate(zip(expected_outputs, outputs)):
         try:
             np.testing.assert_equal(baseline_output.shape, optimized_output.shape)
```
justinchuby, line 39:
```suggestion
```

Possible to set `strict=True` instead to compare shape and dtype https://numpy.org/doc/2.2/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose

onnxscript/rewriter/pattern.py
```diff
@@ -1511,6 +1512,7 @@ def rule(cls, *args, **kwargs):
         instance = cls(*args, **kwargs)
         setup = instance.setup if hasattr(instance, "setup") else None
         cleanup = instance.cleanup if hasattr(instance, "cleanup") else None
+        as_function = instance.as_function if hasattr(instance, "as_function") else False
```
justinchuby, line 1515: Is it possible to ensure all instances have the fields defined?

onnxscript/rewriter/pattern.py
```diff
@@ -1542,23 +1545,34 @@ def _copy_for_function(
...
+        )  # dummy parameter for a None input
         value_map[input] = new_value
         function_inputs.append(new_value)
 
     def copy_value(value: ir.Value | None) -> ir.Value | None:
         if value is None:
             return None
         if value not in value_map:
+            const_value = value.const_value
+            if isinstance(const_value, (ir.Tensor, ir.TensorProtoTensor)):
```
justinchuby, line 1569:
```suggestion
            if const_value is not None:
```

justinchuby: The const value can be of different types too, but if it is not None it should implement TensorProtocol

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -4,7 +4,7 @@
 
 import onnxscript.ir as ir
 from onnxscript.ir.passes.common import shape_inference
-from onnxscript.optimizer import optimize, remove_unused_nodes
+from onnxscript.optimizer import optimize
 from onnxscript.rewriter import rewrite
 from onnxscript.rewriter.ort_fusions import (
```
justinchuby, line 9: Would it be possible to import modules only? This is only to help prevent import cycles down the line

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -11,6 +11,7 @@
 class SDPA(pattern.RewriteRuleClassBase):
     def __init__(self, name: str, *, use_mask: bool, pre_scale: bool, use_mul: bool):
         super().__init__(name=name)
+        self.as_function = True
```
justinchuby, line 14: I wonder if this can be controlled by the pattern engine instead of the rule itself.

onnxscript/rewriter/pattern.py
gramalingam: So, I assume any valid value will be acceptable as parameter to `ir.AttrTensor` down below? Or, is a conversion required in any case?

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: How? I assume you mean the engine should figure it out automatically, but that's hard. It could choose to extract the replaced set of nodes as a function whenever it is legal, but that could cause the model size to increase unnecessarily in some cases (at least temporarily). I guess it may be doable if we don't mind this performance overhead. We *need* it only for speculative fusions that introduce a temporary intermediate op (like SDPA above) that is expected to be subsequently removed by another fusion (like MHA or GQA) ... and if it is not, we would like to undo the original speculative fusion. Anyway: this is an interesting extension to consider later on.

onnxscript/rewriter/pattern.py
gramalingam: Done

onnxscript/rewriter/ort_fusions/_test_utils.py
gramalingam: Hmmm ... that is new in numpy version 2.0. So, I am going to undo this change.

onnxscript/rewriter/pattern.py
justinchuby: Yes. Any TensorProtocol can be used for `ir.AttrTensor`

onnxscript/rewriter/ort_fusions/_test_utils.py
justinchuby: I see ok

justinchuby: But this is test so I guess that's ok?

onnxscript/rewriter/ort_fusions/sdpa.py
justinchuby: SG, thanks! Just a thought to explore

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -58,8 +57,26 @@ def fuse_xformers(model: ir.Model) -> None:
     # Finally: inline any intermediate fusion functions introduced that were not
     # consumed by other fusions, and eliminate any remaining unused nodes.
     optimize(model)
+    return model
+
+
+def optimize_for_ort(model: ir.Model, config_name: str | None = None) -> ir.Model:
```
justinchuby, line 63: Should we call it target? Or maybe config_name is more general?

gramalingam: Current usage (in ort/olive) is EP (which is ORT-equivalent to target, I guess). But I think we could envision other configs (at least for debugging, testing, perhaps more generally too). Hence, I made it config_name ...

onnxscript/rewriter/ort_fusions/_core.py
gramalingam: This file is supposed to depend on most of the remaining files (but other files should not depend on this). So, I am not sure replacing 10 odd lines by 20 is worth it (because we'd like to expose imported fuse_X methods in __init__ also ... at the least, it seems worth doing it) where "from X import y" becomes "import X"; "y = X.y" ...

PR #2138, opened 2025-03-26
Make the return type of rewrite check functions a MatchResult object

shubhambhokare1:
- Check function returns a MatchResult object instead of bool
- This allows propagating the failure reason to the tracer to help in debugging
- #TODO Improve fail message (Reviewers feel free to suggest improvements to current failure messages)

### DIFF
--- a/onnxscript/rewriter/generic_pattern_test.py
+++ b/onnxscript/rewriter/generic_pattern_test.py
@@ -546,8 +546,8 @@ class GenericPatternTest(unittest.TestCase):
             # replace by return [perm0[p] for p in perm1] ?
             return new_perm
 
-        def transpose_transpose_check(op, **_) -> bool:
-            return True
+        def transpose_transpose_check(op, **_) -> pattern.MatchResult:
+            return pattern.MatchResult()
 
         def transpose_transpose_apply_pattern(op, X, XT: ir.Value, Y, **_):
             perm0 = XT.producer().attributes.get("perm")
--- a/onnxscript/rewriter/generic_pattern_test.py
+++ b/onnxscript/rewriter/generic_pattern_test.py
@@ -546,8 +546,8 @@ class GenericPatternTest(unittest.TestCase):
             # replace by return [perm0[p] for p in perm1] ?
             return new_perm
 
-        def transpose_transpose_check(op, **_) -> pattern.MatchResult:
-            return pattern.MatchResult()
+        def transpose_transpose_check(op, **_) -> bool:
+            return True
 
         def transpose_transpose_apply_pattern(op, X, XT: ir.Value, Y, **_):
             perm0 = XT.producer().attributes.get("perm")
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -207,7 +207,7 @@ class TransposeIdentity(orp.RewriteRuleAsClass):
     def check(cls, context, x: ir.Value, perm: ir.Attr) -> orp.MatchResult:
         check_result = orp.MatchResult()
         if isinstance(perm, ir.RefAttr):
-            return check_result.fail("Permutation is not a reference attr.")
+            return check_result.fail("Permutation is a reference attribute.")
         if perm.type == ir.AttributeType.INTS:
             if perm.value == list(range(len(perm.value))):
                 return check_result
@@ -231,7 +231,7 @@ class TransposeTranspose(orp.RewriteRuleAsClass):
     def check(cls, context, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr) -> orp.MatchResult:
         check_result = orp.MatchResult()
         if isinstance(perm1, ir.RefAttr) or isinstance(perm2, ir.RefAttr):
-            return check_result.fail("Permutation values are not reference attributes.")
+            return check_result.fail("Permutation is a reference attribute.")
         return check_result
 
     @classmethod
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -26,9 +26,12 @@ class SqueezeReshape(orp.RewriteRuleClassBase):
     def rewrite(self, op, x: ir.Value):
         return op.Identity(x)
 
-    def check(self, context, x) -> bool:
+    def check(self, context, x) -> orp.MatchResult:
         del context  # Unused
-        return ir_utils.has_rank(x, 1)
+        check_result = orp.MatchResult()
+        if not ir_utils.has_rank(x, 1):
+            return check_result.fail("Input is not 1D")
+        return check_result
 
 
 class CastIdentity(orp.RewriteRuleAsClass):
@@ -43,8 +46,11 @@ class CastIdentity(orp.RewriteRuleAsClass):
         return op.Identity(x)
 
     @classmethod
-    def check(cls, context, x, to) -> bool:
-        return x.dtype == to.value
+    def check(cls, context, x, to) -> orp.MatchResult:
+        check_result = orp.MatchResult()
+        if x.dtype != to.value:
+            return check_result.fail("Input and output types are not the same")
+        return check_result
 
 
 class CastCast(orp.RewriteRuleAsClass):
@@ -62,11 +68,13 @@ class CastCast(orp.RewriteRuleAsClass):
         return op.Cast(op.Cast(x, to=to_ignored), to=to)
 
     @classmethod
-    def check(cls, context, x: ir.Value, to: ir.Attr, to_ignored: ir.Attr) -> bool:
-        return (
-            to.value in cls._allowed_tensor_types
-            and to_ignored.value in cls._allowed_tensor_types
-        )
+    def check(cls, context, x: ir.Value, to: ir.Attr, to_ignored: ir.Attr) -> orp.MatchResult:
+        check_result = orp.MatchResult()
+        if to.value not in cls._allowed_tensor_types:
+            return check_result.fail(f"Output type {to.value} is not allowed")
+        if to_ignored.value not in cls._allowed_tensor_types:
+            return check_result.fail(f"Ignored type {to_ignored.value} is not allowed")
+        return check_result
 
     @classmethod
     def rewrite(cls, op, x: ir.Value, to: ir.Attr, to_ignored: ir.Attr):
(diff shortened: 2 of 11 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -1448,8 +1452,8 @@ def rewrite(cls, op, *_) -> Any:
         raise NotImplementedError("Method 'rewrite' must be overwritten.")
 
     @classmethod
-    def check(cls, context, *_, **__) -> bool:
-        return True
+    def check(cls, context, *_, **__) -> MatchResult:
```
gramalingam, line 1455: nit: I think allowing check to return `bool | MatchResult` would help with some backward compatibility (especially if there are external users using this). There is nothing wrong with `True` in successful cases. The calling logic within the pattern matcher can handle all the conversion.

onnxscript/rewriter/ort_fusions/rotary_embedding.py
```diff
@@ -66,22 +71,23 @@ def pattern(self, op, x, end1, start2):
...
         if not isinstance(end1_value, int) or not isinstance(start2_value, int):
-            return False
+            return check_result.fail("end1 and start2 are not integers.")
         if end1_value != start2_value:
-            return False
+            return check_result.fail("end1 and start2 are not equal.")
         rotary_embedding_attributes = x_part_1_rope.producer().attributes
         if "rotary_embedding_dim" in rotary_embedding_attributes:
-            return False
+            return check_result.fail("rotary_embedding_dim is not a singleton value.")
```
gramalingam, line 84: change to something like "rotary_embedding_dim attribute already specified"

onnxscript/rewriter/pattern.py
```diff
@@ -1371,7 +1371,11 @@ def try_rewrite(
                 if var.name is not None:
                     if var.name not in match.bindings:
                         match.bindings[var.name] = None
-            if not self._condition_function(context, **match.bindings):
+            check_match_result = self._condition_function(context, **match.bindings)
+            if not check_match_result:
+                # If check function was provided, but it failed, return the reason for failure to the tracer.
+                if not isinstance(check_match_result, bool):
```
gramalingam, line 1377: nit: changing the condition to `isinstance(check_match_result, MatchResult)` may be more robust

onnxscript/rewriter/ort_fusions/rotary_embedding.py
```diff
@@ -66,22 +71,23 @@ def pattern(self, op, x, end1, start2):
...
         return op.Concat(x_part_1_rope, x_part_2, axis=-1)
 
-    def check(self, op, x, end1, start2, x_part_1_rope, **_):
+    def check(self, op, x, end1, start2, x_part_1_rope, **_) -> pattern.MatchResult:  # type: ignore[name-defined]
+        check_result = pattern.MatchResult()
         end1_value = _ir_utils.get_singleton_value(end1)
         start2_value = _ir_utils.get_singleton_value(start2)
         if not isinstance(end1_value, int) or not isinstance(start2_value, int):
-            return False
+            return check_result.fail("end1 and start2 are not integers.")
```
gramalingam, line 79: Suggestion: Extend the MatchResult api (eg., the `fail` method or a new one) to optionally specify the failure-causing node. We could potentially also allow failure-causing values ... or even allow a list of node-or-values. Could be helpful.

The message above can be made a bit more explicit ... may be "The end1 value of first slice and start2 of second slice".

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -38,12 +39,13 @@ def pattern(cls, op, x, y, cst):
         return op.Div(op.FusedMatMul(x, y, _domain="com.microsoft"), cst)
 
     @classmethod
-    def check(cls, context, x, y, cst) -> bool:
+    def check(cls, context, x, y, cst) -> orp.MatchResult:
+        check_result = orp.MatchResult()
         if cst.const_value is None:
-            return False
+            return check_result.fail("Divisor is a None value.")
```
gramalingam, line 45: It is not a None value. Suggest "Divisor is not a constant value".

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -15,13 +15,14 @@ def pattern(cls, op, x, y, cst):
         return op.Div(op.MatMul(x, y), cst)
 
     @classmethod
-    def check(cls, context, x, y, cst) -> bool:
+    def check(cls, context, x, y, cst) -> orp.MatchResult:
+        check_result = orp.MatchResult()
         if cst.const_value is None:
-            return False
+            return check_result.fail("Divisor is a None value.")
```
gramalingam, line 21: "is not a constant value"

onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -96,24 +96,32 @@ def pattern(
...
+                return check_result
         if (
             _ir_utils.has_rank(position_ids, 2) and _ir_utils.is_singleton_value(extra_dims, 1)
         ) or (
             _ir_utils.has_rank(position_ids, 1) and _ir_utils.is_1d_value(extra_dims, [0, 1])
         ):
             pass
         else:
-            return False
+            return check_result.fail("position_ids are not 1D or 2D tensors.")
```
gramalingam, line 116: "is not a 1D or 2D tensor"

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -210,10 +228,11 @@ def pattern(cls, op, x, perm1, perm2):
         return op.Transpose(op.Transpose(x, perm=perm1), perm=perm2)
 
     @classmethod
-    def check(cls, context, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr) -> bool:
+    def check(cls, context, x: ir.Value, perm1: ir.Attr, perm2: ir.Attr) -> orp.MatchResult:
+        check_result = orp.MatchResult()
         if isinstance(perm1, ir.RefAttr) or isinstance(perm2, ir.RefAttr):
-            return False
-        return True
+            return check_result.fail("Permutation values are not reference attributes.")
```
gramalingam, line 234: change to "is a reference attribute"

onnxscript/rewriter/ort_fusions/rotary_embedding.py
shubhambhokare1: The MatchResult API currently supports passing a `failure_node` to `fail`, so I just extended it to support a node, value or a collection of nodes and values. I added example usage in sdpa.py. If this looks good, I'll extend it other such `fail` calls stemming from check functions.

onnxscript/rewriter/pattern.py
```diff
@@ -1374,8 +1381,14 @@ def try_rewrite(
             check_match_result = self._condition_function(context, **match.bindings)
             if not check_match_result:
                 # If check function was provided, but it failed, return the reason for failure to the tracer.
-                if not isinstance(check_match_result, bool):
-                    match.fail(check_match_result.reason)
+                if isinstance(check_match_result, MatchResult):
+                    if check_match_result._failure_nodes_and_values:
```
gramalingam, line 1385: nit: this if-then-else condition is not required? I think the then branch should work correctly in both cases, right?

onnxscript/rewriter/pattern.py
```diff
@@ -1812,10 +1825,20 @@ def print(self):
...
                 print("Graph matching failed.")
-            failure_node = self.match_result._failure_node
-            if failure_node:
-                print("Failure at or around node:")
-                failure_node.display()
+            failure_nodes_and_values = self.match_result._failure_nodes_and_values
+            print("Failure at or around nodes/values:")
+
+            if failure_nodes_and_values:
+                if isinstance(failure_nodes_and_values, list):
```
gramalingam, line 1832: Don't we expect a list in all cases? However, I think we can special-case when the list has a single element (just to produce a slightly better wording of error message).

onnxscript/rewriter/pattern.py
```diff
@@ -1812,10 +1825,20 @@ def print(self):
...
+                    for failure_cause in failure_nodes_and_values:
+                        if isinstance(failure_cause, ir.Node):
+                            failure_cause.display()
+                        elif isinstance(failure_cause, ir.Value):
+                            print(failure_cause)
+                elif isinstance(failure_nodes_and_values, ir.Node):
+                    failure_nodes_and_values.display()
+                elif isinstance(failure_nodes_and_values, ir.Value):
+                    print(failure_nodes_and_values)
```
gramalingam, line 1841: Even a value has a display method, so you can call it, I think. (But there may be scope for improving the display of values ... but that is an orthogonal issue, can deal with it separately.)

justinchuby: The display methods were designed for terminal interactive use. Would print be enough? If not we can improve

shubhambhokare1: Using the display call for now, we can improve if needed.

onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -546,8 +546,8 @@ def transpose_transpose_mapping(perm0, perm1):
             # replace by return [perm0[p] for p in perm1] ?
             return new_perm
 
-        def transpose_transpose_check(op, **_) -> bool:
-            return True
+        def transpose_transpose_check(op, **_) -> pattern.MatchResult:
+            return pattern.MatchResult()
```
gramalingam, line 550: I believe that returning True here should work, correct? Just a minor nit: I think "True" is more readable than "pattern.MatchResult()" and more efficient (no need to construct a new object with sub-objects). If we really want a MatchResult, we could also get the same advantages using a readonly object like `pattern.SUCCESS`

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -26,9 +26,12 @@ def pattern(self, op, x):
...
         return op.Identity(x)
 
-    def check(self, context, x) -> bool:
+    def check(self, context, x) -> orp.MatchResult:
         del context  # Unused
-        return ir_utils.has_rank(x, 1)
+        check_result = orp.MatchResult()
+        if not ir_utils.has_rank(x, 1):
+            return check_result.fail("Input is not 1D")
+        return check_result
```
gramalingam, line 34: Another nit thought: it might be slightly better if this looked like
```py
   if ...
      return orp.failure("reason")
   return orp.SUCCESS
```
Just a stylistic suggestion to keep in mind. I realize this pattern has already been used in many places down below and may not be worth changing at this point.

gramalingam (approved):

PR #2142, opened 2025-03-27
[DRAFT] Add basic test for GQA (fusion)

gramalingam:
Adds a test to verify equivalence of an expanded graph and GQA operator (which will serve as the basis for a fusion rule).


shubhambhokare1 (commented):
Given it is a just a check to see whether the GQA operates in the same way as we think and not really checking the rewriting, maybe we can create a folder called something like op_equivalence_tests and place this there?

gramalingam:
> nit: Given it is a just a check to see whether the GQA operates in the same way as we think and not really checking the rewriting, maybe we can create a folder called something like op_equivalence_tests and place this there?

Before discussing where to put this, let me explain why I created this. I have a GQA fusion implementation, which successfully replaces a subgraph with GQA ... but it produces a different output (with and without the fusion). It is very complicated pattern, and it was not clear what exactly was wrong.  An advantage of a test-case like this one is that I can execute through the "expanded subgraph" and the 'replacement subgraph' using a debugger, stopping at intermediate points, to verify values look the way I think they should look (for example: does the attention mask look like a causal mask, etc.). (All I have to do is call the script-function with the inputs, instead of using its model-proto for an ORT session.)

The proof of correctness of a fusion rule comes from a proof of equivalence of code-fragments as captured in this test-case.

gramalingam:
> > nit: Given it is a just a check to see whether the GQA operates in the same way as we think and not really checking the rewriting, maybe we can create a folder called something like op_equivalence_tests and place this there?
> 
> Before discussing where to put this, let me explain why I created this. I have a GQA fusion implementation, which successfully replaces a subgraph with GQA ... but it produces a different output (with and without the fusion). It is very complicated pattern, and it was not clear what exactly was wrong. An advantage of a test-case like this one is that I can execute through the "expanded subgraph" and the 'replacement subgraph' using a debugger, stopping at intermediate points, to verify values look the way I think they should look (for example: does the attention mask look like a causal mask, etc.). (All I have to do is call the script-function with the inputs, instead of using its model-proto for an ORT session.)
> 
> The proof of correctness of a fusion rule comes from a proof of equivalence of code-fragments as captured in this test-case.

To continue the discussion: once a working fusion rule is in place, this test can be easily turned into a fusion rule (by applying the fusion to the expanded model defined in the test case).

gramalingam:
Closing this. Will create an updated PR with GQA fusion added.

PR #2148, opened 2025-03-28
Add fusion rules for com.microsoft.Attention

shubhambhokare1:
#TODO 
- Find a model and create a test case to test this rewrite rule 
- Add rotaryembedding to pattern incorporating do_rotary


onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,188 @@
...
+        return False
+    for actual, expected in zip(val.shape, shape):
+        if expected not in bindings:
+            bindings[expected] = actual  # type: ignore[assignment]
+        elif actual != bindings[expected]:
+            return False
+    return True
+
+
+class Attention(pattern.RewriteRuleClassBase):
```
gramalingam, line 27: nit: suggest a name like "AttentionFusion".

Another potential issue (unfortunately) is that there is com.microsoft.Attention and onnx Attention (and we may end up needing a fusion for both). But I guess we have a separate folder for ORT contrib op fusion, so I guess that should be enough to distinguish the two.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,188 @@
...
+        # Ensure Dh = Dh_q + Dh_k + Dh_v
+        Dh = bindings.get("Dh")
+        Dh_q = bindings.get("Dh_q")
+        Dh_k = bindings.get("Dh_k")
+        Dh_v = bindings.get("Dh_v")
+
+        if Dh is None or Dh_q is None or Dh_k is None or Dh_v is None:
+            return False  # Missing bindings, cannot verify
+
+        if Dh != Dh_q + Dh_k + Dh_v:  # type: ignore[operator]
```
gramalingam, line 134: The dimensions can be symbolic (in worst case) ... which is why there is operator warning. May be worth changing previous check from `Dh is None` to `not instance(Dh, int)`.

onnxscript/rewriter/ort_fusions/attention_basic_test.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (c) Microsoft Corporation.
```
gramalingam, line 1: To clarify PR #2142  further: this type of test is more for debugging (the mathematical rewrite rule), and I was suggesting it in case the relative behavior of the two ops was unclear. 

For the rewrite rule implementation, you should have a test case where the rewrite rule is applied. Eg., you can apply it to the "expanded model" defined below, and verify that the rewritten model has the expected attention op, and that it behaves equivalent to the original model.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,283 @@
...
+                _outputs=1,
+            )
+
+
+attention_with_input_bias_rule = AttentionFusion.rule(
+    "attention_input_bias",
+    has_input_bias=True,
+    has_past=False,
+)
+attention_with_no_input_bias_rule = AttentionFusion.rule(
```
gramalingam, line 251: Just a minor suggestion for the future (this looks fine for now): I found that when the number of combinations increases, it is easier to use names listing only positive flags: that is, names like "attention_with_bias" and "attention" (which implicitly has no bias).  Avoids very long names if you have 3 flags, for example. In this case, for example, "attention", "attention_with_bias", "attention_with_past", "attention_with_bias_and_past".

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,283 @@
...
+        Dh_k = bindings.get("Dh_k")
+        Dh_v = bindings.get("Dh_v")
+
+        if (
+            not isinstance(Dh, int)
+            or not isinstance(Dh_q, int)
+            or not isinstance(Dh_k, int)
+            or not isinstance(Dh_v, int)
+        ):
+            return False  # Missing bindings, cannot verify
```
gramalingam, line 189: nit: can return a suitable fail error message

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,283 @@
...
+        # Add bias if present
+        if self._has_input_bias:
+            projected = op.Add(projected, qkv_bias)
+
+        # Slice packed Matmul QKV into Q, K, and V
+        # Q, K, and V are of shape (B, S, D)
+        query_BSD = op.Slice(
+            projected,
+            _allow_other_inputs=True,
+            _allow_other_attributes=True,
```
gramalingam, line 57: We can drop the _allow_other_attributes, Slice doesn't have any attributes. Further, True is the default value for this flag anyway. (I realize this probably comes from a copy of pre-existing code, which was meant to be cleaned up, but wasn't).

onnxscript/rewriter/ort_fusions/attention.py
shubhambhokare1: Modified

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -53,6 +54,7 @@ def fuse_xformers(model: ir.Model) -> ir.Model:
     fuse_cos_sin_cache(model)
     fuse_sdpa(model)
     fuse_mha(model)
+    fuse_attention(model)
```
gramalingam, line 57: This is fine for now. At some point, we should consider a potential optimization: typically we expect a MHA or Attention or GQA etc. ... so if one of them applies, the others may not be necessary to apply.

shubhambhokare1: That makes sense, although won't the Attention op expect the MHA fusion to be done beforehand?

gramalingam: Right ... so, it will need to be a bit more complicated. If MHA doesn't succeed, Attention need not be tried, but GQA can be tried, etc.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,287 @@
...
+                past,
+                # attention_bias,
+                # past_sequence_length
+                num_heads=num_heads,
+                qkv_hidden_sizes=qkv_hidden_sizes,
+                # scale=scale,
+                _domain="com.microsoft",
+                _outputs=2,
+            )
+            # Return present output first as it captures the complete rewrite pattern graph
```
gramalingam, line 231: This comment should be in the return statement of the pattern method above. Here, I suggest something like "Use same output ordering as in pattern"

onnxscript/rewriter/ort_fusions/attention_test.py
```diff
@@ -0,0 +1,156 @@
...
+        input_types = (
+            FLOAT["B", "S", D],
+            FLOAT[D, Dh_qkv],
+            FLOAT[Dh_qkv],
+        )
+        output_types = (FLOAT["B", "S", self.v_hidden_size],)
+
+        if with_past:
+            input_types += (FLOAT[2, "B", self.num_heads, "S", self.headsize],)
+            output_types += (FLOAT[2, "B", self.num_heads, "NS", self.headsize],)
```
gramalingam, line 116: Is "NS" supposed to be total_sequence_length? It would be nice to settle on some common notation we can use across the different fusions. What do you think about using "T"? A single letter will make it easier to use abbreviations like BHTd to indicate a shape. (I also tried using abbreviations like "Dh" and "Hkv", but if a single letter works, that would be even better. Not sure if the community has any well-accepted conventions for this.)

onnxscript/rewriter/ort_fusions/attention_test.py
```diff
@@ -0,0 +1,156 @@
...
+        self.input_hidden_size = self.headsize * self.num_heads
+        self.q_hidden_size = 160
+        self.k_hidden_size = 160
+        self.v_hidden_size = 160
+
+    def random_inputs(self, with_past=False):
+        """Generate random inputs for the model."""
+        B = self.batchsize
+        S = self.seqlen
+        M = self.max_seqlen
```
gramalingam, line 39: I guess "max_seqlen" is a bit different from "total_seqlen"? Does it impact the fusion or the test below which of the two we use?

gramalingam: I mean, do we need to separately test the two cases where the buffer is (a) equal-to and (b) greater-than what is needed for this run? Or, does it not matter?

onnxscript/rewriter/ort_fusions/attention_test.py
gramalingam: Elsewhere, we also need abbreviations for "past seq len" ("P" or "Sp" or "Skvp" ?).

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -0,0 +1,287 @@
...
+from typing import Sequence, Union
+
+import onnxscript.ir as ir
+from onnxscript.rewriter import pattern
+
+Dim = Union[int, ir.SymbolicDim]
+
+
+# TODO: Maybe add this check to utilities
+def _check_shape(bindings: dict[str, Dim], val: ir.Value, shape: Sequence[str]) -> bool:
```
gramalingam, line 14: Can we move this into a shared utility? And add appropriate failure messages?

onnxscript/rewriter/ort_fusions/attention_test.py
```diff
@@ -0,0 +1,156 @@
...
+        self.headsize = 16
+        self.num_heads = 10
+        self.input_hidden_size = self.headsize * self.num_heads
+        self.q_hidden_size = 160
+        self.k_hidden_size = 160
+        self.v_hidden_size = 160
+
+    def random_inputs(self, with_past=False):
+        """Generate random inputs for the model."""
+        B = self.batchsize
```
justinchuby, line 37: Avoid upper case variable names to follow python naming conventions.

onnxscript/rewriter/ort_fusions/attention_test.py
```diff
@@ -0,0 +1,156 @@
...
+                key_BSDh,
+                value_BSDh,
+                num_heads=self.num_heads,
+            )
+            return mha
+
+        @script()
+        def model_with_mha_past(input, weight, bias, past):
+            QKV_no_bias = op.MatMul(input, weight)
+            QKV = op.Add(QKV_no_bias, bias)
```
justinchuby, line 79: Avoid upper case variable names to follow python naming conventions.

onnxscript/rewriter/ort_fusions/attention_test.py
gramalingam: So, I think it is an acceptable exception to the general rule.

onnxscript/rewriter/ort_fusions/_core.py
shubhambhokare1: Will add this as a follow up PR once this is merged.

onnxscript/rewriter/ort_fusions/attention_test.py
shubhambhokare1: Leaving variable names for shapes as they are, as I agree (B, S, ...) for example are commonly used for shape notations and might be easier to recognize in this context.

onnxscript/rewriter/ort_fusions/attention_test.py
shubhambhokare1: T for total_sequence_length makes sense to me.
We should probably come with a standard for these naming conventions, I think we should document this somehow.

But here is what I had in mind,
`
B: batch size

S: sequence length (if query/key/value sequence lengths are the same)
Sq: query sequence length,
Skv: key/value sequence length, when Sq and Skv are different

Following B, S, N, H conventions
H: head size (if query/key/value head sizes are the same)
Hqk: query/key head sizes
Hv: value head size, when Hqk and Hv are different

N: num of heads (if query/key/value num heads are the same)
Nq: query num of heads
Nkv: key/value num of heads, when Nq and Nqk are different

Dq : query hidden size
Dk: key hidden size
Dv: value hidden size

# For input, weight, and bias type inputs
H: head size or embedding dimension per head
D: input embedding dimension (hidden size) = N * H
D = Dq + Dk + Dv

T: total sequence length (after concatenation of past and current key/value)
Sp/Spkv: past sequence length
Sm: Maximum sequence

`

onnxscript/rewriter/ort_fusions/attention_test.py
shubhambhokare1: That is correct, max_seq_len is different from total_seq_len. However max_seq_len is only used when past and present share the same buffer, the op's parameter `past_present_share_buffer` is set to 1 (it is zero by default) and `past_sequence_length` is provided as an input. I guess this PR doesn't cover this pattern so we dont need to test this here. I changed past shape sequence length from total_seq_len to past_seq_len to reflect this.

gramalingam (approved):

PR #2149, opened 2025-03-29
Turn inliner into a pass and use it in rewriter & optimizer

justinchuby:
Use passes in optimizer and rewriter.

1. By opting into using the pass infra early, we get the benefit of getting the additional features in pass infra w/o having to pay higher refactoring cost in the future. We will be able to add more sophisticated debug utilities/snapshot capabilities etc. to the passes.
2. Since we are offering the pass infra to users, we can start validating it internally by using it here. If order altering becomes a valid use case we can expect users may need that and we can create appropriate facilities to support the usage.

### DIFF
--- a/onnxscript/optimizer/_optimizer.py
+++ b/onnxscript/optimizer/_optimizer.py
@@ -72,4 +72,5 @@ def optimize_ir(
         onnxscript.ir.passes.common.unused_removal.RemoveUnusedNodesPass(),
     )
     assert optimizer_pass.in_place
-    optimizer_pass(model)
+    result = optimizer_pass(model)
+    assert result.model is model
--- a/onnxscript/optimizer/_optimizer.py
+++ b/onnxscript/optimizer/_optimizer.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 
+import onnxscript.ir.passes.common.unused_removal
 import onnxscript.optimizer
 from onnxscript import ir, rewriter
 from onnxscript.optimizer import _constant_folding, _inliner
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/__init__.py
```diff
@@ -17,37 +17,51 @@
 from onnxscript.ir.passes.common import unused_removal
 from onnxscript.rewriter import pattern
 
-RewriteRuleSet = pattern.RewriteRuleSet
```
justinchuby, line 20: Keep

onnxscript/optimizer/_optimizer.py
```diff
@@ -50,15 +51,25 @@ def optimize_ir(
...
-    for _ in range(num_iterations):
-        _constant_folding.fold_constants(
-            model,
-            onnx_shape_inference=onnx_shape_inference,
-            input_size_limit=input_size_limit,
-            output_size_limit=output_size_limit,
-        )
-        rewriter.rewrite(model, pattern_rewrite_rules=_DEFAULT_REWRITE_RULES)
-    onnxscript.optimizer.remove_unused_nodes(model)
+    optimizer_pass = ir.passes.Sequential(
```
gramalingam, line 54: While I am okay with this, I would like to ask: what are the advantages of switching to using a PassManager as below?

A pass-manager, as below, is a programming-abstraction. Here, it is an abstraction of sequential-composition of several passes. But python already allows sequential-composition, and programmers are very familiar with it. So, a new abstraction must offer some compelling advantage.

The flipside is that premature design of a programming abstraction (just like premature optimization) can end up hampering us ... if we don't anticipate/know what all users are likely to want to do. (As an example, if we decide to switch between Pass1 and Pass2 in alternative iterations in the above loop: easy to encode in above loop, but we will be handicapped down below.) Do we know enough to design this abstraction, and do we see enough value?

justinchuby: I think there are two motivations:

1. By opting into using the pass infra early, we get the benefit of getting the additional features in pass infra w/o having to pay higher refactoring cost in the future. We will be able to add more sophisticated debug utilities/snapshot capabilities etc. to the passes.
2. Since we are offering the pass infra to users, we can start validating it internally by using it here. If order altering becomes a valid use case we can expect users may need that and we can create appropriate facilities to support the usage.

onnxscript/optimizer/_optimizer.py
```diff
@@ -50,15 +51,25 @@ def optimize_ir(
...
+                onnxscript.ir.passes.common.unused_removal.RemoveUnusedFunctionsPass(),
+                onnxscript.ir.passes.common.unused_removal.RemoveUnusedOpsetsPass(),
+            ],
+            steps=num_iterations,
+            early_stop=stop_if_no_change,
+        ),
+        onnxscript.ir.passes.common.unused_removal.RemoveUnusedNodesPass(),
+    )
+    assert optimizer_pass.in_place
+    optimizer_pass(model)
```
justinchuby, line 75:
```suggestion
    result = optimizer_pass(model)
    assert result.model is model
```

gramalingam (approved):

onnxscript/rewriter/__init__.py
```diff
@@ -17,37 +17,50 @@
...
-    if pattern_rewrite_rules:
-        if not isinstance(pattern_rewrite_rules, RewriteRuleSet):
-            # Create a pattern rule-set using provided rules
-            pattern_rewrite_rules = pattern.RewriteRuleSet(pattern_rewrite_rules)
-        count = pattern_rewrite_rules.apply_to_model(model_ir)
-        if count:
-            print(f"Applied {count} of general pattern rewrite rules.")
-    unused_remover = ir.passes.PassManager(
+
+    rewrite_pass = ir.passes.PassManager(
```
gramalingam, line 55: I think this is causing a regression when no rewrite-rules are specified. I think we should retain an `if pattern_rewrite_rules` condition and avoid the RewritePass if no rules are specified.

PR #2150, opened 2025-03-31
Optimize away zero-length concat operands

gramalingam:
We optimize `Concat (x1, x2, x3)` if one or more the concat operands has zero length along the concatenated axis-dimension.

This pattern shows up, for example, in Phi models. See [this line](https://github.com/huggingface/transformers/blob/786d9c5ed920a099573ea7b6dbf265f1aeb32fc0/src/transformers/models/phi3/modeling_phi3.py#L152) in the implementation of partial-rotary-embedding:
```py
q_embed = torch.cat([(q_rot * cos) + (rotate_half(q_rot) * sin), q_pass], dim=-1)
```
In the special case of total-rotary-embedding, the second operand `q_pass` of the concat is empty. This also interferes with the pattern-matching for GQA in the generated graph. Optimizing the redundant Concat away will help with GQA fusion as well.

Handle the edge case when all operands have zero size.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -575,7 +575,8 @@ def concat(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
         if (shape := operand.shape) is None:
             return False
         try:
-            dim_size = shape[axis]
+            # We have already checked that axis is an int value (!= None)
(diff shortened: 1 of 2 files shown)


justinchuby (approved):

justinchuby:
Some tests are failing

PR #2159, opened 2025-04-02
Allow fuse_xformers to return a count of different fusions applied

shubhambhokare1:


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -44,20 +46,33 @@ def _pre_optimize(model: ir.Model) -> ir.Model:
...
-    fuse_rms_normalization(model)
-    fuse_normalization(model)
-    fuse_rotary_embedding(model)
-    fuse_partial_rotary_embedding(model)
-    fuse_cos_sin_cache(model)
-    fuse_sdpa(model)
-    fuse_mha(model)
-    fuse_gelu(model)
+    fusion_count["rms_normalization"] = fuse_rms_normalization(model)
+    fusion_count["rms_and_skip_normalization"] = fuse_normalization(model)
```
gramalingam, line 65: Why "rms_and_skip"? It should be either "skip_" or "skip_rms_" ? This reminds me: we should cleanup fuse_normalization so it doesn't apply rms-normalization, which is redundant in this setting.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -44,20 +46,33 @@ def _pre_optimize(model: ir.Model) -> ir.Model:
...
+    fusion_count["mha"] = fuse_mha(model)
+    fusion_count["gelu"] = fuse_gelu(model)
     # Finally: inline any intermediate fusion functions introduced that were not
     # consumed by other fusions, and eliminate any remaining unused nodes.
     optimize(model)
-    return model
+    return model, fusion_count
 
 
 def optimize_for_ort(model: ir.Model, config_name: str | None = None) -> ir.Model:
```
gramalingam, line 78: We should extend above to this `optimize_for_ort` function as well, which is what external users are expected to call anyway.

onnxscript/rewriter/ort_fusions/gelu.py
```diff
@@ -33,5 +33,10 @@ def rewrite(self, op, x):
 gelu_rules = pattern.RewriteRuleSet([_rule])
 
 
-def fuse_gelu(model: ir.Model) -> None:
-    gelu_rules.apply_to_model(model)
+def fuse_gelu(model: ir.Model, debug: bool = False) -> int:
+    count = gelu_rules.apply_to_model(model)
```
gramalingam, line 37: Reminds me: may be useful to lift this pattern (which is repeated in every fuse_* method) into a utility.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -44,20 +46,33 @@ def _pre_optimize(model: ir.Model) -> ir.Model:
     return model
 
 
-def fuse_xformers(model: ir.Model) -> ir.Model:
+def fuse_xformers(model: ir.Model) -> Tuple[ir.Model, Dict[str, int]]:
```
justinchuby, line 49:
```suggestion
def fuse_xformers(model: ir.Model) -> tuple[ir.Model, dict[str, int]]:
```

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -2,6 +2,8 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
+from typing import Dict, Tuple
```
justinchuby, line 5:
```suggestion
```

onnxscript/rewriter/ort_fusions/_core.py
shubhambhokare1: Cleaned up, and also add SkipLayerNorm fusion, I have them as separate fusion calls for now so we can isolate the counts for the different types of skip fusions.

onnxscript/rewriter/ort_fusions/gelu.py
shubhambhokare1: Add utility to _fusion_utils.py. Let me if you think there is a better place for this.

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -32,15 +32,55 @@ def _skip_normalization(op, input, skip, gamma, epsilon, stash_type):
...
-normalization_rules = rms_normalization_rules + skip_normalization_rules
-normalization_ruleset = pattern.RewriteRuleSet(normalization_rules)
+skip_rms_normalization_rules = [_skip_rms_rule]
+skip_rms_normalization_ruleset = pattern.RewriteRuleSet(skip_rms_normalization_rules)
 
 
-def fuse_normalization(model):
-    count = normalization_ruleset.apply_to_model(model)
-    print(f"Normalization count: {count}")
+def _skip_layernorm_pattern(op, input, skip, gamma, beta, epsilon, stash_type):
```
gramalingam, line 43: nit: rename the "layernorm" part to "layer_norm" to be consistent with other usage

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -2,11 +2,11 @@
...
 from __future__ import annotations
 
-from onnxscript.rewriter import pattern
-from onnxscript.rewriter.ort_fusions.rms_normalization import rms_normalization_rules
+import onnxscript.ir as ir
+from onnxscript.rewriter import _fusion_utils, pattern
 
 
-def _skip_norm_pattern(op, input, skip, gamma, epsilon, stash_type):
+def _skip_rmsnorm_pattern(op, input, skip, gamma, epsilon, stash_type):
```
gramalingam, line 9: nit: rename "rmsnorm" part to "rms_norm" to be consistent with other usage

onnxscript/rewriter/_fusion_utils.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+from __future__ import annotations
+
+import onnxscript.ir as ir
+from onnxscript.rewriter import pattern
+
+
+def apply_fusion_rules(
```
gramalingam, line 9: An alternative to consider: changing this as below has some advantages:
```py
def apply_fusion_rules(rules: pattern.RewriteRule | pattern.RewriteRuleSet):
   def apply_to (model: ir.Model, debug: bool = False) -> int:
       ...
   return apply_to
```
would let us define
```py
   fuse_gelu = apply_fusion_rules (gelu_rules)
```
The main advantage is that you can extend the interface by adding other options beyond debug in one place, without having to go and update every fusion file to add the same option.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -74,9 +93,11 @@ def optimize_for_ort(model: ir.Model, config_name: str | None = None) -> ir.Mode
...
+        A tuple containing:
+        - The optimized `ir.Model` after applying transformer-specific fusions.
+        - A dictionary with a count of each of the fusions applied.
     """
 
-    fuse_xformers(model)
+    model, fusion_count = fuse_xformers(model)
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
-    return model
+    return model, fusion_count
```
gramalingam, line 103: This is fine for now. (Later it might be worth extending fusion_count to return stats from ORT_PATTERN_REWRITE_RULES too, but that requires updating the pattern-matcher to return this info, so better to do it separately.

onnxscript/rewriter/_fusion_utils.py
shubhambhokare1: That is a good point, changed it. nit question about what the call would look like, 
For example,
`def fuse_gelu(model: ir.Model, debug: bool = False) -> int:
    fuse_gelu = _fusion_utils.apply_fusion_rules(gelu_rules)
    return fuse_gelu(model, debug=debug)
`
Would it be better name the callable function as fuse_gelu_fn or something along those lines?

onnxscript/rewriter/ort_fusions/_core.py
shubhambhokare1: Made a note of this, will do this as part of a separate PR.

onnxscript/rewriter/ort_fusions/gelu.py
```diff
@@ -34,4 +34,5 @@ def rewrite(self, op, x):
 
 
 def fuse_gelu(model: ir.Model, debug: bool = False) -> int:
-    return _fusion_utils.apply_fusion_rules(gelu_rules, model, debug=debug)
+    fuse_gelu = _fusion_utils.apply_fusion_rules(gelu_rules)
```
gramalingam, line 37: A nit clarification: what I meant was that these 3 lines can be replaced by `fuse_gelu = _fusion_utils.apply_fusion_rules(gelu_rules)` no need for lines 36 or 38. (You will still need some extra lines in cases where there is an extra `remove_unused_nodes` ... which will eventually disppear, because it is not strictly necessary, but that is a separate cleanup, don't want to overload this PR with other unrelated changes.)

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -32,15 +31,53 @@ def _skip_normalization(op, input, skip, gamma, epsilon, stash_type):
     return normalized, skip_sum
 
 
-_rule = pattern.RewriteRule(
-    _skip_norm_pattern, _skip_normalization, matcher=pattern.SimplePatternMatcher
+_skip_rms_rule = pattern.RewriteRule(
+    _skip_rms_norm_pattern, _skip_rms_normalization, matcher=pattern.SimplePatternMatcher
```
gramalingam, line 35: Curious: do we need to specify the pattern-matcher here?

gramalingam (approved):

onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -170,11 +170,8 @@ def rewrite(
 
 
 def fuse_cos_sin_cache(model: ir.Model, debug: bool = False) -> int:
-    count = cos_sin_cache_rules.apply_to_model(model)
-    if count == 0 and debug:
-        tracer = pattern.MatchingTracer()
-        cos_sin_cache_rules.apply_to_model(model, tracer=tracer)
-        tracer.report()
+    fuse_cos_sin_cache = _fusion_utils.apply_fusion_rules(cos_sin_cache_rules)
+    count = fuse_cos_sin_cache(model, debug=debug)
```
gramalingam, line 174: The remove_unused_nodes part below is no longer necessary, because it has been made automatic (based on the rule's settings ... so, it will be called as the last step in apply_fusion_rules.) So, that should simply this part. Feel free to do this as part of this or as a separate PR, as convenient.

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -150,8 +150,8 @@ def rewrite(
 gqa_rules = pattern.RewriteRuleSet([_rule1])
 
 
-def fuse_gqa(model: ir.Model) -> int:
-    count = gqa_rules.apply_to_model(model)
-    print(f"GQA count: {count}")
+def fuse_gqa(model: ir.Model, debug: bool = False) -> int:
+    fuse_gqa = _fusion_utils.apply_fusion_rules(gqa_rules)
+    count = fuse_gqa(model, debug=debug)
```
gramalingam, line 155: As mentioned above, the remove_unused_nodes below is no longer needed.

onnxscript/rewriter/ort_fusions/skip_normalization.py
shubhambhokare1: Not needed, removed.

gramalingam (approved):

Issue #2157, opened 2025-04-01
Pass to lift subgraph initializers to main graph


justinchuby:

titaiwangms:
I thought in subgraph we don't support initializers? cc @gramalingam

PR #2160, opened 2025-04-02
[Pass] Support lift constants to initializers pass

### LINKED ISSUE #2156: Create a constant to initializer pass
https://gist.github.com/justinchuby/cf1699d05baeac281fb3e82f9d0fc473

titaiwangms:
Fix #2156


onnxscript/ir/passes/common/lift_constants_to_initializers.py
```diff
@@ -0,0 +1,50 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        """Convert constant nodes in main graph to initializers."""
+        count = 0
+        for node in model.graph:
+            if node.op_type != "Constant":
```
gramalingam, line 23: Must also check that the domain is either "" or "onnx.ai"

onnxscript/ir/passes/common/lift_constants_to_initializers.py
```diff
@@ -0,0 +1,50 @@
...
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+
+class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        """Convert constant nodes in main graph to initializers."""
+        count = 0
+        for node in model.graph:
```
gramalingam, line 22: This won't process nodes within subgraphs (in an If node, for example). This also introduces a design question: should we move the constant nodes from an If node subgraph into the main graph's initializer? The tricky part is that it could introduce name conflicts ... if we have different tensors in the then-branch and else-branch which have same name. I think the IR naming may handle it correctly, but worth checking.

onnxscript/ir/passes/common/lift_constants_to_initializers.py
```diff
@@ -0,0 +1,50 @@
...
+            if node.op_type != "Constant":
+                continue
+            if "value" not in node.attributes:
```
gramalingam, line 25: This doesn't handle all the variations supported by the Constant op. I think you may be able to use [this utility function](https://github.com/microsoft/onnxscript/blob/2962a09e5b4dd5e0e5afc2e975f66c8bac2452dd/onnxscript/optimizer/_constant_folding.py#L65), but it is in the optimizer module, may need to move it to a common usable place.

onnxscript/ir/_core.py
```diff
@@ -1999,6 +2003,12 @@ def __iter__(self) -> Iterator[Node]:
     def __reversed__(self) -> Iterator[Node]:
         return reversed(self._nodes)
 
+    def _set_input_and_initializer_value_names_into_name_authority(self):
```
justinchuby, line 2006: Feel free to add a todo for me to trigger this when inputs and initializers are modified

onnxscript/ir/passes/common/lift_constants_to_initializers.py
justinchuby, line 1: the module can be `constant_manipulation` or something? We can put some constant related passes here

onnxscript/ir/passes/common/lift_constants_to_initializers.py
justinchuby: I think we should avoid subgraphs, for now. The constant nodes in subgraphs tend to not be model weights

onnxscript/ir/passes/common/lift_constants_to_initializers.py
```diff
@@ -0,0 +1,52 @@
...
+            assert initializer_name is not None
+            initializer = ir.Value(
+                name=initializer_name,
+                shape=tensor.shape,  # type: ignore[arg-type]
+                type=ir.TensorType(tensor.dtype),
+                const_value=tensor,
+            )
+            # TODO(titaiwang): Is it possible that the initializer name has
+            # been taken?
+            model.graph.initializers[initializer_name] = initializer
```
justinchuby, line 42:
```suggestion
            model.graph.register_initializer(initializer)
```

it's now possible to do this

onnxscript/ir/passes/common/lift_constants_to_initializers_test.py
```diff
@@ -0,0 +1,55 @@
...
+        # Check that the initializer is not in the graph yet
+        self.assertEqual(len(model.graph.initializers), 0)
+        # And 1 constant node
+        self.assertEqual(len([node for node in model.graph if node.op_type == "Constant"]), 1)
+
+        # Perform lift constants to initializers
+        result = lift_constants_to_initializers.LiftConstantsToInitializersPass()(model)
+        self.assertTrue(result.modified)
+        # Check that the constant node is lifted to an initializer
+        self.assertEqual(len(result.model.graph.initializers), 1)
```
justinchuby, line 51: Also check the actual tensor value?

onnxscript/ir/_core.py
```diff
@@ -1924,6 +1924,10 @@ def __init__(
         # Be sure the initialize the name authority before extending the nodes
         # because it is used to name the nodes and their outputs
         self._name_authority = _name_authority.NameAuthority()
+        # NOTE: input and initializer value names could be duplicated
+        # to auto-generated value names from name authority and crash ort
+        # https://github.com/microsoft/onnxruntime/blob/bc7b07dbb41a2f441dbed1a91855563ba0dd8a31/onnxruntime/core/graph/graph.cc#L1536
```
justinchuby, line 1929:
```suggestion
```
I assume the following line fixes this problem? We could update the comment to explain what is accomplished by the following line instead.

onnxscript/ir/passes/common/lift_constants_to_initializers.py
gramalingam: The second part is right ... but why do you say we should avoid subgraphs for now? Eg., one option would be to simply replace the for loop to iterate over all nodes (I believe there is an iterator for that, right?). Do you mean the naming part could be challenging?

justinchuby: Actually I suppose it’s ok doing it either way. Initially I thought the naming part can be challenging, but since we already have some knowledge of the used names in the main graph, pulling values from subgraphs and giving them new names should be safe

onnxscript/ir/_core.py
titaiwangms: Added

onnxscript/ir/passes/common/lift_constants_to_initializers.py
titaiwangms: Done

onnxscript/ir/passes/common/lift_constants_to_initializers_test.py
titaiwangms: Done

onnxscript/ir/passes/common/lift_constants_to_initializers.py
justinchuby: In that case you may want to make the subgraph constants to subgraph initializers first, and use the pass #2157 to lift them from subgraphs to the main graph. 

To find the owning graph of the node you can do node.graph.

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+            np.array(attr_value.as_floats(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_string":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_string(), dtype=np.object_), name=initializer_name
+        )
+    elif attr_name == "value_strings":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_strings(), dtype=np.object_), name=initializer_name
+        )
```
justinchuby, line 98:
```suggestion
        tensor = ir.tensor(attr_value.as_strings(), name=initializer_name)
```

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+        tensor = ir.Tensor(
+            np.array(attr_value.as_float(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_floats":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_floats(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_string":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_string(), dtype=np.object_), name=initializer_name
```
justinchuby, line 93: as below

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+            np.array(attr_value.as_ints(), dtype=np.int64), name=initializer_name
+        )
+    elif attr_name == "value_float":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_float(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_floats":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_floats(), dtype=np.float32), name=initializer_name
+        )
```
justinchuby, line 90:
```suggestion
        tensor = ir.tensor(attr_value.as_floats(), dtype=ir.DataType.FLOAT, name=initializer_name)
```

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+
+class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        """Convert constant nodes in main graph to initializers."""
+        count = 0
+        for node in model.graph:
```
justinchuby, line 24:
```suggestion
        for node in Recrusive...(model.graph):
```

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+            # Register an initializer with the tensor value
+            initializer = ir.Value(
+                name=initializer_name,
+                shape=tensor.shape,  # type: ignore[arg-type]
+                type=ir.TensorType(tensor.dtype),
+                const_value=tensor,
+            )
+            # TODO(titaiwang): Is it possible that the initializer name has
+            # been taken?
+            model.graph.register_initializer(initializer)
```
justinchuby, line 56:
```suggestion
            node.graph.register_initializer(initializer)
```

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+            # Replace the constant node with the initilizer
+            ir.convenience.replace_all_uses_with(node.outputs[0], initializer)
+            model.graph.remove(node, safe=True)
```
justinchuby, line 59:
```suggestion
            node.graph.remove(node, safe=True)
```

onnxscript/ir/passes/common/constant_manipulation.py
gramalingam: Spelling of recursive seems off in above comment

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,101 @@
...
+            np.array(attr_value.as_float(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_floats":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_floats(), dtype=np.float32), name=initializer_name
+        )
+    elif attr_name == "value_string":
+        tensor = ir.Tensor(
+            np.array(attr_value.as_string(), dtype=np.object_), name=initializer_name
+        )
```
justinchuby, line 94:
```suggestion
        tensor = ir.StringTensor(attr_value.as_string()), shape=ir.Shape([]), name=initializer_name)
```

onnxscript/ir/passes/common/constant_manipulation.py
justinchuby, line 98:
```suggestion
        strings = attr_value.as_strings()
        tensor = ir.StringTensor(strings, shape=ir.Shape([len(strings)]), name=initializer_name)
```

onnxscript/ir/passes/common/lift_constants_to_initializers.py
titaiwangms: Done

onnxscript/ir/passes/common/lift_constants_to_initializers.py
titaiwangms: The PR supports lifting subgraph constants to initializers as well now. We will have another pass to lift subgraph initializers to main graph if it's necessary.

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,142 @@
...
+    def test_pass_with_lifting_constants_to_initializers_within_subgraph(self):
+        input_value = ir.Value(
+            name="input", type=ir.TensorType(ir.DataType.FLOAT), shape=ir.Shape((2, 3))
+        )
+
+        then_constant_tensor = ir.tensor(np.random.rand(2, 3).astype(np.float32))
+        attribute = ir.convenience.convert_attributes({"value": then_constant_tensor})
+        then_const_node = ir.Node(
+            "", "Constant", inputs=[], attributes=attribute, num_outputs=1
+        )
```
justinchuby, line 78: you can use `ir.node` to simplify this construction. We can leverage it here as reference for future readers.

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,142 @@
...
+                self.assertEqual(
+                    node.graph.initializers["val_0"].const_value,
+                    then_constant_tensor,
+                )
+            if node.op_type == "Mul":
+                self.assertEqual(len(node.graph.initializers), 1)
+                self.assertEqual(
+                    node.graph.initializers["val_0"].const_value,
+                    else_constant_tensor,
+                )
```
justinchuby, line 142: Possible to directly look at initializers of `then_graph` and `else_graph`?

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -29,10 +31,11 @@ def test_pass_with_lifting_constants_to_initializers(self, ir_dtype, numpy_dtype
         ]
 
         constant_tensor = ir.tensor(np.random.rand(2, 3).astype(numpy_dtype))
```
justinchuby, line 33: numpy_dtype can be obtained with `ir_dtype.numpy()`

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -128,15 +125,69 @@ def test_pass_with_lifting_constants_to_initializers_within_subgraph(self):
...
-                )
-            if node.op_type == "Mul":
-                self.assertEqual(len(node.graph.initializers), 1)
-                self.assertEqual(
-                    node.graph.initializers["val_0"].const_value,
-                    else_constant_tensor,
-                )
+        self.assertEqual(len(else_graph.initializers), 1)
+        self.assertEqual(len(then_graph.initializers), 1)
+        self.assertEqual(
```
justinchuby, line 130: `assertIs` may be more precise

justinchuby (approved):

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -0,0 +1,95 @@
...
+                continue
+            # Register an initializer with the tensor value
+            initializer = ir.Value(
+                name=initializer_name,
+                shape=tensor.shape,  # type: ignore[arg-type]
+                type=ir.TensorType(tensor.dtype),
+                const_value=tensor,
+            )
+            assert node.graph is not None
+            assert isinstance(node.graph, ir.Graph)
```
titaiwangms, line 55: @justinchuby does this make sense? I think there should not be any ir.Function node coming out from recursive iterator?

justinchuby: Does make sense. Thanks!

justinchuby: Technically the graph in a function is private and not used directly. It is currently an implementation detail

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,191 @@
...
+        assert len(model.graph.initializers) == 0
+        # And 1 constant node
+        assert len([node for node in model.graph if node.op_type == "Constant"]) == 1
+
+        # Perform lift constants to initializers
+        result = constant_manipulation.LiftConstantsToInitializersPass()(model)
+        assert result.modified
+        # Check that the constant node is lifted to an initializer
+        assert len(result.model.graph.initializers) == 1
+        self.assertTrue(
```
justinchuby, line 186: nit: np.testing.assert_array_equal

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,191 @@
...
+        # Check that the initializer is not in the graph yet
+        assert len(model.graph.initializers) == 0
+        # And 1 constant node
+        assert len([node for node in model.graph if node.op_type == "Constant"]) == 1
+
+        # Perform lift constants to initializers
+        result = constant_manipulation.LiftConstantsToInitializersPass()(model)
+        assert result.modified
+        # Check that the constant node is lifted to an initializer
+        assert len(result.model.graph.initializers) == 1
```
justinchuby, line 185: Prefer `self.assertEqual`

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,191 @@
...
+        )
+
+        # Check that the initializer is not in the graph yet
+        assert len(model.graph.initializers) == 0
+        # And 1 constant node
+        assert len([node for node in model.graph if node.op_type == "Constant"]) == 1
+
+        # Perform lift constants to initializers
+        result = constant_manipulation.LiftConstantsToInitializersPass()(model)
+        assert result.modified
```
justinchuby, line 183: `self.assertTrue` (maybe I should change other places)

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,191 @@
...
+                nodes=[identity_node_input, const_node, identity_node_constant],
+                opset_imports={"": 20},
+            ),
+            ir_version=10,
+        )
+
+        # Check that the initializer is not in the graph yet
+        assert len(model.graph.initializers) == 0
+        # And 1 constant node
+        assert len([node for node in model.graph if node.op_type == "Constant"]) == 1
```
justinchuby, line 179: nit: as below

onnxscript/ir/passes/common/constant_manipulation_test.py
```diff
@@ -0,0 +1,189 @@
...
+        # And 1 constant node
+        self.assertEqual(len([node for node in model.graph if node.op_type == "Constant"]), 1)
+
+        # Perform lift constants to initializers
+        result = constant_manipulation.LiftConstantsToInitializersPass()(model)
+        self.assertTrue(result.modified)
+        # Check that the constant node is lifted to an initializer
+        self.assertEqual(len(result.model.graph.initializers), 1)
+        np.testing.assert_array_equal(
+            result.model.graph.initializers["val_1"].const_value.raw,
```
justinchuby, line 187:
```suggestion
            result.model.graph.initializers["val_1"].const_value.numpy(),
```

maybe this?

justinchuby (approved):

justinchuby (approved):

onnxscript/ir/passes/common/constant_manipulation.py
justinchuby: sorry I was wrong. It is pointed to a graph when we call `function.append`, but it is not when we call ir.Node(graph=function). I need to figure out how to reconcile this. Suggestions appreciated. https://github.com/microsoft/onnxscript/issues/2181

justinchuby: https://github.com/microsoft/onnxscript/pull/2183

PR #2164, opened 2025-04-07
Handle empty rewrite rules in rewrite function

justinchuby:
In https://github.com/microsoft/onnxscript/pull/2149 the logic for skipping rewrite when no rules are provided was removed. This PR adds the logic back and hardens input checks. Now if no rules are provided to `rewrite()`, it will only run cleanup passes.

### DIFF
--- a/onnxscript/rewriter/__init__.py
+++ b/onnxscript/rewriter/__init__.py
@@ -38,8 +38,8 @@ _DEFAULT_REWRITE_RULES: tuple[pattern.RewriteRule, ...] = (
 class RewritePass(ir.passes.InPlacePass):
     def __init__(
         self,
-        /,
         rules: Sequence[pattern.RewriteRule] | pattern.RewriteRuleSet,
+        /,
     ) -> None:
         super().__init__()
         if isinstance(rules, Sequence):
--- a/onnxscript/rewriter/__init__.py
+++ b/onnxscript/rewriter/__init__.py
@@ -14,9 +14,25 @@ import onnx
 
 from onnxscript import ir
 from onnxscript.ir.passes.common import unused_removal
-from onnxscript.rewriter import pattern
+from onnxscript.rewriter import (
+    broadcast_to_matmul,
+    cast_constant_of_shape,
+    collapse_slices,
+    gemm_to_matmul_add,
+    llama_rule_sets,
+    no_op,
+    pattern,
+)
 
 _ModelProtoOrIr = TypeVar("_ModelProtoOrIr", onnx.ModelProto, ir.Model)
+_DEFAULT_REWRITE_RULES: tuple[pattern.RewriteRule, ...] = (
+    *no_op.rules.rules,  # TODO: merge this rule into constant folding?
+    *broadcast_to_matmul.rules.rules,
+    gemm_to_matmul_add.rule,  # type: ignore[has-type]
+    *cast_constant_of_shape.rules.rules,
+    *collapse_slices.rules.rules,
+    *llama_rule_sets.llama_p0_rule_set().rules,
+)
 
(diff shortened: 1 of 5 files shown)


onnxscript/rewriter/__init__.py
```diff
@@ -42,19 +41,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
     else:
         model_ir = model
         proto = False
 
+    rewrite_or_empty = [RewritePass(pattern_rewrite_rules)] if pattern_rewrite_rules else ()
+
     rewrite_pass = ir.passes.PassManager(
         (
-            RewritePass(pattern_rewrite_rules),
+            *rewrite_or_empty,
```
gramalingam, line 70: The input to the PassManager constructor allows nested lists? Sounds a bit unusual ...

onnxscript/rewriter/__init__.py
```diff
@@ -42,19 +41,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+    pattern_rewrite_rules: Union[Sequence[pattern.RewriteRule], pattern.RewriteRuleSet] = (),
+) -> _ModelProtoOrIr:
+    """Rewrite the model using the provided pattern rewrite rules.
+
+    Unused nodes, functions, and opsets will be removed after the rewrite.
+
+    Args:
+        model: The model to be rewritten. Can be an ONNX ModelProto or an ir.Model.
+        pattern_rewrite_rules: A sequence of pattern rewrite rules or a RewriteRuleSet.
+            If not provided or empty, only clean up passes will be applied.
```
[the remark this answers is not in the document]
justinchuby: Sounds good! Will change to default set

onnxscript/rewriter/__init__.py
justinchuby: This should be a flat list? I’m trying to unpack here. In any case will change to default rewrite set

gramalingam: Sorry, I missed the * ... you will still need to have the special handling in case user explicitly passes an empty list (though it will hopefully not happen much)

onnxscript/rewriter/__init__.py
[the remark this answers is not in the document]
justinchuby: Done

onnxscript/rewriter/__init__.py
justinchuby: Done

onnxscript/rewriter/__init__.py
```diff
@@ -5,53 +5,85 @@
...
+        The rewritten model, either as an ONNX ModelProto or an ir.Model.
+    """
     if isinstance(model, onnx.ModelProto):
         model_ir = ir.serde.deserialize_model(model)
         proto = True
     else:
         model_ir = model
         proto = False
 
+    if not pattern_rewrite_rules:
```
gramalingam, line 84: (Discussing a real edge case here): Should we distinguish between None and an empty list here? If the user does not specify an explicit rewrite-rule set (i.e., None), using the default rules makes sense. If the user explicitly specifies an empty list, ideally it should be treated as a no-op (OR be treated as an error, which is slightly less preferable)?

I suggest
```py
   if pattern_rewrite_rules is None:
      pattern_rewrite_rules = _DEFAULT_REWRITE_RULES
   else if pattern_rewrite_rules == []:
      return model
```

And suggest moving this to the beginning of the function (avoid unnecessary deserialization in the return-case).

justinchuby: Sure! Done.

gramalingam (approved):

PR #2167, opened 2025-04-08
Avoid using fixed value for max_pos_ids in cos_sin_cache fusion

shubhambhokare1:
To apply the cos sin fusion pattern-rewrite, we need to know the maximum position id.
- If model/config has this information, use it calculate max_pos_id
- If not, calculate max_pos_id using position ids using ONNX ops
- Removes dependence of pre-setting the max_pos_id for each rewrite rule

### DIFF
--- a/onnxscript/rewriter/ort_fusions/cos_sin_cache.py
+++ b/onnxscript/rewriter/ort_fusions/cos_sin_cache.py
@@ -29,15 +29,11 @@ from onnxscript.rewriter import _fusion_utils, _ir_utils, pattern
 #
 # This produces cos/sin values in a form that can be used by ORT's custom ops.
 
-# TODO: To apply the pattern-rewrite, we need to know the maximum position id.
-# Need to find a way to get this information from the model or its config.
-
 
 class CosSinCacheFusion(pattern.RewriteRuleClassBase):
     def __init__(
         self,
         name: str,
-        max_pos_id: int,
         *,
         cast: bool = False,
         reshape: bool = False,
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -47,13 +43,66 @@ def __init__(
         # matched nodes as part of the rewrite-step. We apply a separate final
         # pass to remove unused nodes.
         super().__init__(name, remove_nodes=False)
-        self._max_pos_id = max_pos_id
+        # TODO: Determine what should be the default max_pos_id value
+        self._max_pos_id = 2048  # Set a default max_pos_id value
+        self._max_pos_id_from_model = False
```
gramalingam, line 48: I think it would be simpler to combine the above two, using a None value for _max_pos_id (to indicate dynamic calculation of max-pos-id).

onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -47,13 +43,66 @@ def __init__(
...
         # map from inv_freq to (cos, sin) values for transformed graph
         self._inv_freq_cos_sin_cache: dict[ir.Value, tuple[ir.Value, ir.Value]] = {}
         self._reshape = reshape
         self._cast = cast
         self._const_freqs = const_freqs
 
+    def _set_max_position_ids_from_model(self, model: ir.Model):
```
gramalingam, line 55: I think it would be better to decouple the computation of max_position_id from model from setting it here. This could be a plain setter for the _max_pos_id field. Do the computation outside ... in future, we might want to move it up elsewhere.

onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -47,13 +43,66 @@ def __init__(
...
+            max_pos_id = self._max_pos_id
+        elif position_ids.const_value is not None:
+            # Calculate max_pos_id from the position_ids tensor
+            max_pos_id = int(np.max(position_ids.const_value.numpy()))
+            self._max_pos_id = max_pos_id
+        else:
+            # Dynamically compute max_pos_id from position_ids using ONNX ops
+            inv_freq = op.Reshape(inv_freq, op.Constant(value_ints=[1, -1]))
+            max_pos_id = op.ReduceMax(position_ids)
+            max_pos_id = op.Squeeze(max_pos_id, op.Constant(value_ints=[0]))
```
gramalingam, line 84: Use the keepdims attribute of ReduceMax? No need to squeeze?

onnxscript/rewriter/ort_fusions/cos_sin_cache.py
```diff
@@ -47,13 +43,66 @@ def __init__(
...
+            max_pos_id = op.Add(max_pos_id, op.Constant(value_int=1))
```
gramalingam, line 85: What's the type of position_ids? Is it guaranteed to be int64?

gramalingam: Why is there a +1 here, but not in the case down below where the value is known?

shubhambhokare1: Added a + 1 to the case down below

gramalingam (approved):

PR #2171, opened 2025-04-08
[IR] Create sort() on Model

justinchuby:


onnxscript/ir/_core.py
```diff
@@ -2553,6 +2553,21 @@ def __repr__(self) -> str:
     graph={textwrap.indent(repr(self.graph), " " * 4).strip()}
 )"""
 
+    def sort(self) -> None:
```
shubhambhokare1, line 2556: where would this be called?

shubhambhokare1, line 2556: Could you add a unit test for this?

justinchuby:
Replaced by the topological sort pass

PR #2172, opened 2025-04-09
Modify constant-folder to return computed symbolic value map

gramalingam:
Modify constant-folder to return computed symbolic value map, which may be useful to the caller. (Eg., fusion optimizations can make use of this information.)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -154,6 +154,10 @@ def __init__(self):
         self._sym_value_map: dict[ir.Value, Any] = {}
         self._initializer_inputs: list[set[ir.Value]] = []
 
+    @property
+    def symbolic_value_map(self) -> dict[ir.Value, Any]:
```
justinchuby, line 158: Is it possible to make the value type of the dictionary more precise?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1104,23 +1116,31 @@ def fold_constants(
...
     Applies constant folding optimization to the model.
-    Returns true iff the model was modified.
+
+    Args:
+        model (ir.Model): The ONNX model to optimize.
+        external_data_folder (str, optional): Path to the folder containing external data
+            for the model. Defaults to an empty string.
+        onnx_shape_inference (bool, optional): Whether to enable ONNX shape inference during
+            constant folding. Defaults to False.
+        input_size_limit (int, optional): The maximum size (in bytes) of input tensors
```
justinchuby, line 1129:
```suggestion
        model : The ONNX model to optimize.
        external_data_folder: Path to the folder containing external data
            for the model. Defaults to an empty string.
        onnx_shape_inference: Whether to enable ONNX shape inference during
            constant folding. Defaults to False.
        input_size_limit: The maximum size (in bytes) of input tensors
```

No need to add types in docstring as we have type annotations already

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1104,23 +1116,31 @@ def fold_constants(
...
+            that can be considered for constant folding. Defaults to
+            `DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT`.
+        output_size_limit (int, optional): The maximum size (in bytes) of output tensors
```
justinchuby, line 1132:
```suggestion
        output_size_limit: The maximum size (in bytes) of output tensors
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1094,7 +1098,15 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
-        return ir.passes.PassResult(model, self.modified)
+        return FoldConstantsResult(model, self.modified, self._state.symbolic_value_map)
+
+
+@dataclasses.dataclass
+class FoldConstantsResult(ir.passes.PassResult):
+    symbolic_value_map: dict[ir.Value, Any]
+
+    def __bool__(self) -> bool:
+        return self.modified
```
justinchuby, line 1109: Is this method created for compatibility reasons for fold_constants? If so I would add a comment to document this.

gramalingam: Done

onnxscript/optimizer/_constant_folding.py
gramalingam: Done

justinchuby (approved):

onnxscript/optimizer/__init__.py
```diff
@@ -35,7 +35,9 @@ def optimize(model: ir.Model, *args, **kwargs) -> ir.Model:
         return legacy_optimizer.optimize(model, *args, **kwargs)
 
 
-def fold_constants(model: ir.Model | onnx.ModelProto, *args, **kwargs) -> bool:
+def fold_constants(
+    model: ir.Model | onnx.ModelProto, *args, **kwargs
+) -> constant_folding.FoldConstantsResult | bool:
```
justinchuby, line 40: Should we remove legacy_constant_folding at some point?

shubhambhokare1 (approved):

onnxscript/optimizer/__init__.py
gramalingam: Yes, I think so too.

PR #2174, opened 2025-04-09
Add fusion to rule to fuse (query, key, value) to a packed QKV for GQA

shubhambhokare1:


onnxscript/rewriter/ort_fusions/fuse_qkv.py
```diff
@@ -0,0 +1,159 @@
...
+            return check_result.fail(
+                "Could not determine the hidden sizes of query, key, and value.",
+            )
+
+        if Dq + (2 * Dkv) != D:  # type: ignore[operator]
+            return check_result.fail(
+                f"Hidden size of query, key and value do not add up to hidden size: {D} != {Dq} + (2 * {Dkv})",
+            )
+
+        return True
```
gramalingam, line 113: The above checks are not sufficient to guarantee correctness, I think. For example: what if the model stores Q, K, and V in a different order than what the GQA op expects? So, you will need to verify that the slice starts/ends match what is needed to ensure this.

onnxscript/rewriter/ort_fusions/fuse_qkv.py
```diff
@@ -0,0 +1,159 @@
+# Copyright (c) Microsoft Corporation.
```
gramalingam, line 1: Nit: suggest expanding the file name .. may something like "fuse_packed_qkv_gqa"

onnxscript/rewriter/ort_fusions/fuse_qkv.py
shubhambhokare1: Added slice checks to check start, end values

onnxscript/rewriter/ort_fusions/fuse_packed_qkv_qga.py
```diff
@@ -71,6 +85,29 @@ def check(
...
+        head_size = hidden_size // (q_num_heads + (2 * kv_num_heads))
+        q_hidden_size = head_size * q_num_heads
+        kv_hidden_size = head_size * kv_num_heads
+        if not (
+            _ir_utils.is_singleton_value(start1, 0)
+            and _ir_utils.is_singleton_value(end1, q_hidden_size)
+            and _ir_utils.is_singleton_value(start2, q_hidden_size)
+            and _ir_utils.is_singleton_value(end2, (q_hidden_size + kv_hidden_size))
+            and _ir_utils.is_singleton_value(start3, (q_hidden_size + kv_hidden_size))
+            and _ir_utils.is_singleton_value(end3, hidden_size)
```
gramalingam, line 104: In other examples, I have seen the use of MAXINT to slice up to the end. Eg. try what happens if we export `x[from:]`. So, it is better to check if end3 >= hidden_size. (I think I have an example in another rule doing this.)

gramalingam: I mean it is ok if end3 >= hidden_size. So, fail if end3 < hidden_size.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -14,7 +14,7 @@
 )
 from onnxscript.rewriter.ort_fusions.attention import fuse_attention
 from onnxscript.rewriter.ort_fusions.cos_sin_cache import fuse_cos_sin_cache
-from onnxscript.rewriter.ort_fusions.fuse_qkv import fuse_qkv_gqa
+from onnxscript.rewriter.ort_fusions.fuse_packed_qkv_qga import fuse_qkv_gqa
```
gramalingam, line 17: nit: typo: "qga" => "gqa"

onnxscript/rewriter/ort_fusions/fuse_packed_qkv_qga.py
shubhambhokare1: Updated

gramalingam:
Can you add a unit test? Thanks!

shubhambhokare1:
> Can you add a unit test? Thanks!

Added 👍

onnxscript/rewriter/ort_fusions/fuse_packed_qkv_gqa_test.py
```diff
@@ -0,0 +1,147 @@
...
+        def gqa(packed_qkv, past_key, past_value, cos, sin):
+            # Generate seqlens_k and total_seqlen inputs for GQA:
+            # In this test case, all batch elements have same sequence length.
+            S = op.Shape(packed_qkv, start=1, end=2)
+            past_seq_length = op.Shape(past_key, start=2, end=3)
+            total_seq_length = op.Add(past_seq_length, S)
+            total_seqlen_int32 = op.Cast(total_seq_length, to=6)
+            total_seqlen_int32_minus_1 = op.Sub(total_seqlen_int32, 1)
+            batchsize = op.Shape(packed_qkv, start=0, end=1)
+            seqlens_k = op.Tile(total_seqlen_int32_minus_1, batchsize)
```
gramalingam, line 87: I am curious why we need the above code fragment (for a minimal test case). Would it be possible to simply make seqlens_k and total_seqlen_int32 just be inputs of this model (just like cos, sin, etc.)? It's a nit, but seems better

shubhambhokare1: Done

gramalingam (approved):

PR #2175, opened 2025-04-09
Introduce pattern.any_value

gramalingam:
Introduce pattern.any_value as a convenience when writing patterns. It is more precise than using `_allow_other_inputs=True` (which will allow any number of inputs).


onnxscript/rewriter/pattern.py
```diff
@@ -634,6 +634,20 @@ def _is_pattern_variable(x: Any) -> bool:
...
+
+    def __init__(self) -> None:
+        super().__init__(None)
+
+    def clone(self, node_map: dict[NodePattern, NodePattern]) -> AnyValue:
+        # A single instance of AnyValue suffices.
+        return self
+
+
+any_value = AnyValue()
```
gramalingam, line 648: Any suggestions for alternative names? Another possibility is `ignore`. Or, would altering the casing to `ANY_VALUE` be better?

justinchuby: I like `ANY_VALUE` or more simply `ANY`

shubhambhokare1: Leaning towards 'ANY_VALUE' as well

gramalingam:
> Could this be extended for _allow_other_attributes as well?

Please clarify: what are you proposing? Thanks!

onnxscript/rewriter/pattern.py
gramalingam: done

justinchuby (approved):

PR #2179, opened 2025-04-10
[torchlib] Precompute the constant for gelu to avoid precision loss

justinchuby:
I think this improves accuracy for gelu under float16.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -483,10 +483,10 @@ def _aten_gelu_approximate_none(self: TReal) -> TReal:
     """gelu(Tensor self, *, str approximate='none') -> Tensor"""
 
     # GELU(x) = 0.5 * x * [1 + ERF(x/sqrt(2)]
-    inner = op.Div(self, 1.4142135623730951)
+    inner = op.Div(self, ir.tensor(1.4142135623730951, dtype=self.dtype))
     erf = op.Erf(inner)
-    inner = op.Add(erf, 1)
-    inner = op.Mul(0.5, inner)
+    inner = op.Add(erf, ir.tensor(1, dtype=self.dtype))
+    inner = op.Mul(ir.tensor(0.5, dtype=self.dtype), inner)
     result = op.Mul(self, inner)
     return result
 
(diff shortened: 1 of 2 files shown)


justinchuby:
We have tests right? I can update the pattern

justinchuby:
OK this should be fine I think. The patterns only concern approximate="none", not "tanh". Can you confirm?

gramalingam:
Do you know what is the source of the precision improvement? Trying to understand this ... is it because some computation happens at higher precision in python vs. using float32 in onnx? Which one? Would help in other cases if we understood this.

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -479,33 +479,31 @@ def aten_gelu(self: TReal, approximate: str = "none") -> TReal:
...
-    cubed = op.Pow(self, 3)
-    inner = op.Mul(0.044715, cubed)
+    cubed = op.Pow(self, ir.tensor(3, dtype=self.dtype))
+    inner = op.Mul(ir.tensor(0.044715, dtype=self.dtype), cubed)
     inner = op.Add(self, inner)
-    # Prefer explicit graph construction over precomputed constants for clarity.
-    two_over_pi = op.CastLike(op.Div(2.0, _MATH_PI), self)
-    inner = op.Mul(op.Sqrt(two_over_pi), inner)
+    # math.sqrt(2.0/math.pi) = 0.7978845608028654
+    sqrt_two_over_pi = ir.tensor(0.7978845608028654, dtype=self.dtype)
```
justinchuby, line 502: @gramalingam I think it's from here

justinchuby:
The computation happens at higher precision than if we did it in onnx at float16

justinchuby:
ok thanks. I looked at the patterns. This should still work because the computation is previously done by the constant folder before being pattern matched on. Here we just changed the implementation so the pre-optimized model is accurate.

titaiwangms (approved):

gramalingam:
> The computation happens at higher precision than if we did it in onnx at float16

I see. Because there is a CastLike before the Sqrt, and it should have been after the Sqrt. Got it.

But, agree, I prefer that the computation happen ahead of time ... WRT outdated comment about preferring explicit graph computation over precomputed constants, there is a way of having the cake and eating it too here ... we express the pre-computation explicitly in python, but outside the script-function, clarifying both the computation that happens and *when* it happens, like below:
```py
sqrt_two_over_pi = math.sqrt(2.0/math.pi)
def _aten_gelu(...):
   ... sqrt_two_over_pi ...
```

justinchuby:
> > The computation happens at higher precision than if we did it in onnx at float16
> 
> I see. Because there is a CastLike before the Sqrt, and it should have been after the Sqrt. Got it.
> 
> But, agree, I prefer that the computation happen ahead of time ... WRT outdated comment about preferring explicit graph computation over precomputed constants, there is a way of having the cake and eating it too here ... we express the pre-computation explicitly in python, but outside the script-function, clarifying both the computation that happens and _when_ it happens, like below:
> 
> ```python
> sqrt_two_over_pi = math.sqrt(2.0/math.pi)
> def _aten_gelu(...):
>    ... sqrt_two_over_pi ...
> ```

That's a great idea. Thanks!

PR #2180, opened 2025-04-10
Remove legacy optimizer

justinchuby:
- Remove legacy optimizer and support proto inputs with the IR based optimizer.
- Add a new `inline=True` option in `optimize()` to control whether function inlining is done when optimizing
- Implement identity folding for graph outputs
- Migrate constant folding tests to run on IR models

Fix https://github.com/microsoft/onnxscript/issues/2185


onnxscript/optimizer/_function_folding_test.py
```diff
@@ -9,6 +9,7 @@
 
 
 class FunctionFoldingTest(unittest.TestCase):
+    @unittest.expectedFailure  # FIXME
```
justinchuby, line 12: @gramalingam It looks like we are missing a few features

onnxscript/optimizer/__init__.py
```diff
@@ -2,46 +2,109 @@
...
 
+def optimize(
+    model: _ModelProtoOrIr,
+    num_iterations: int = 2,
+    *,
+    onnx_shape_inference: bool = True,
+    stop_if_no_change: bool = True,
+    input_size_limit: int = constant_folding.DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT,
+    output_size_limit: int = constant_folding.DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT,
+    inline: bool = True,
```
justinchuby, line 42: Added a new option

onnxscript/optimizer/__init__.py
```diff
@@ -2,46 +2,109 @@
...
+) -> _ModelProtoOrIr:
+    """Optimizes a model.
 
-def optimize(model: ir.Model, *args, **kwargs) -> ir.Model:
```
xadupre, line 29: The API changes. If nobody uses optimize, it is ok, if somebody is, then his code will break. Not really a good option.

justinchuby: Which change is concerning for you? Is it model or args? If args and kwargs, changes to the options are minor and shouldn't affect existing usage. If model the behavior remains the same (only the type annotations are updated)

onnxscript/optimizer/_function_folding_test.py
```diff
@@ -60,13 +60,11 @@ def test_sequence_concat(self):
...
         optimized = optimizer.optimize(
-            model,
-            onnx_shape_inference=False,
-            num_iterations=1,
+            model, onnx_shape_inference=False, num_iterations=1, inline=False
         )
         function_node = optimized.functions[0].node
-        self.assertEqual(len(function_node), 3)
         self.assertEqual(function_node[2].op_type, "Concat")
+        self.assertEqual(len(function_node), 3)
```
gramalingam, line 67: Why reorder? The length assertion guards the indexing operation, which may be invalid if length < 3.

justinchuby: Makes sense. Will revert.

onnxscript/optimizer/_function_folding_test.py
```diff
@@ -60,13 +60,11 @@ def test_sequence_concat(self):
         """
         )
         optimized = optimizer.optimize(
-            model,
-            onnx_shape_inference=False,
-            num_iterations=1,
+            model, onnx_shape_inference=False, num_iterations=1, inline=False
```
gramalingam, line 63: Line [58] above should be changed to `z = ConcatFromSequence <axis=0> (t3)` ... the required attribute axis was missing in test-case (ignored and optimized by old optimizer, not by new).

onnxscript/optimizer/_function_folding_test.py
```diff
@@ -114,12 +110,11 @@ def test_single_user_function_is_modified_inplace_after_folding(self):
...
         )
         self.assertEqual(optimized.functions[0].name, "fun1")
 
+    @unittest.expectedFailure  # FIXME
```
gramalingam, line 117: This is as designed. This requires creating a copy of the called function to optimize it (without inlining). Our current strategy is to inline and optimize, so this type of optimization is not planned.

onnxscript/optimizer/_function_folding_test.py
```diff
@@ -138,14 +133,13 @@ def test_multi_users_function_is_not_modified_inplace_after_folding(self):
...
-            model,
-            onnx_shape_inference=False,
-            num_iterations=1,
+            model, onnx_shape_inference=False, num_iterations=1, inline=False
         )
         self.assertEqual(len(optimized.functions), 2)
         self.assertNotEqual(optimized.functions[0].name, "fun1")
         self.assertNotEqual(optimized.functions[1].name, "fun1")
 
+    @unittest.expectedFailure  # FIXME
```
gramalingam, line 142: Error in test case model. Imported opset in main graph not same as in function (21 vs 18) ... was being ignored by previous inliner, caught by current inliner.

onnxscript/optimizer/_function_folding_test.py
gramalingam: I checked. This is one of the limitations of the current Identity-elimination ... when the main graph output is computed as `graph-output = Identity(some-value)` ... here we can't replace graph-output by some-value trivially. We should
* use the ir.Value some-value
* rename it to be `graph-output`, and
* update graph-outputs to include this renamed some-value
* And we can do this only if some-value is not a graph-input

For the third step above: I see that `graph.outputs` returns a `list` ... so, I assume it is intended to be mutable? Can we simply assign `graph.outputs[j] = some-value` ?

onnxscript/optimizer/_function_folding_test.py
gramalingam: I merged in the changes I made while debugging these ... hope that's fine.

justinchuby: Thanks!

onnxscript/optimizer/_function_folding_test.py
justinchuby: Removed test case

onnxscript/optimizer/_function_folding_test.py
justinchuby: > Can we simply assign graph.outputs[j] = some-value ?

Yes - that works!

justinchuby: I the logic you described in `visit_graph`. Please take a look.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1087,6 +1087,23 @@ def visit_graph(self, graph: ir.Graph) -> None:
         for node in graph:
             self.visit_node(node, graph)
 
+        # Replace outputs if output nodes can be folded. This is typically outputs from
```
justinchuby, line 1090:
```suggestion
        # Replace outputs if output nodes can be folded. This are typically outputs from
```

onnxscript/optimizer/_optimizer.py
```diff
@@ -54,7 +55,11 @@ def optimize_ir(
         ),
         onnxscript.ir.passes.common.unused_removal.RemoveUnusedNodesPass(),
         onnxscript.ir.passes.common.constant_manipulation.LiftConstantsToInitializersPass(),
-    )
+    ]
+    if inline:
+        # Inline all functions first before optimizing
+        passes.insert(0, _inliner.InlinePass())
```
justinchuby, line 61:
```suggestion
        passes = [inliner.InlinePass(), *passes]
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1087,6 +1087,23 @@ def visit_graph(self, graph: ir.Graph) -> None:
...
+            if output is None:
+                continue
+            sym_value = self._state.get_sym_value(output)
+            if not isinstance(sym_value, ir.Value):
+                # An output must be a Value
+                continue
+            if sym_value in graph.inputs:
+                # ONNX does not allow a graph output to be a graph input
+                continue
+            # Rename sym_value to match the output name
```
gramalingam, line 1102: I just realized that there is another edge case: we can't do this for two different outputs: eg.
```
   x = Add(...)
   graph_output1 = Identity(x)
   graph_output2 = Identity(x)
```
We can rename x to either graph_output1 or graph_output2, not both.

justinchuby: Thanks - updated

justinchuby: Added a test

onnxscript/optimizer/_constant_folding_test.py
```diff
@@ -610,22 +552,38 @@ def test_large_transpose(self):
...
         self.assertEqual(ops, ["Transpose", "MatMul"])
 
         # Input size limit will allow folding of Transpose op
         # Since there is no increase in model-size, output-size is not a concern.
-        optimized = self._fold(irmodel, input_size_limit=4 * 512 * 256)
+        optimized = self._fold(model, input_size_limit=4 * 512 * 256)
         ops = [node.op_type for node in optimized.graph]
         self.assertEqual(ops, ["Constant", "MatMul"])
 
+    def test_multi_graph_identity_output_preserves_output_name(self):
```
justinchuby, line 571: tested here

onnxscript/optimizer/_constant_folding.py
justinchuby: Looks like the subgraphs are still not handled correctly: https://github.com/microsoft/onnxscript/actions/runs/14414536354/job/40428982831?pr=2180#step:6:809

justinchuby: Fixed

onnxscript/optimizer/__init__.py
justinchuby: Let me know if you have further concerns, thanks!

titaiwangms (approved):
Approve to unblock @justinchuby. @gramalingam feel free to follow up on this merged PR when you have time.

PR #2182, opened 2025-04-10
[Pass] Remove metadata_props and doc_string from the model

### LINKED ISSUE #2163: [Pass] Clear metadata
https://github.com/microsoft/onnxruntime/blob/a28da4b90a17c9729c9448c5702c302e3bd2566b/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L172-L179

titaiwangms:
Fix #2163


onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
                 modified |= self._process_graph_like(function, used_domains=set())
 
         return ir.passes.PassResult(model, modified=modified)
+
+
+class RemoveUnusedMetadataAndDocStringPass(ir.passes.InPlacePass):
```
justinchuby, line 191: I don't think there is a notion of "unused" for metadata. For the other passes we always know if something is being "used" in the graph. Suggest renaming

```suggestion
class RemoveMetadataAndDocStringPass(ir.passes.InPlacePass):
```

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        modified = False
+        # 1. Clean up all of the nodes metadata properties
+        for node in ir.traversal.RecursiveGraphIterator(model.graph):
+            if len(node.metadata_props) > 0:
```
justinchuby, line 196:
```suggestion
            if node.metadata_props:
```

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+                modified = True
+                logger.debug("Removed metadata from %s nodes", node.name)
+            node.metadata_props.clear()
+
+            # TODO: Use model.graphs to get the graph instead of node.graph
+            # 2. Clean up the main graph metadata properties
+            # and doc_string
+            assert node.graph is not None
+            if len(node.graph.metadata_props) > 0 or node.graph.doc_string is not None:
```
justinchuby, line 205:
```suggestion
            if node.graph.metadata_props or node.graph.doc_string:
```

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+                modified = True
+                logger.debug("Removed metadata from %s graph", node.graph.name)
+            node.graph.metadata_props.clear()
+            node.graph.doc_string = None
+
+        # 3. Clean up all of the functions metadata properties
+        for function in model.functions.values():
```
justinchuby, line 212: loop though nodes and subgraphs in a function too?

onnxscript/ir/passes/common/unused_removal.py
justinchuby, line 205: We can record the visited graphs so we don't check this again for every node

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,33 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
                 modified |= self._process_graph_like(function, used_domains=set())
 
         return ir.passes.PassResult(model, modified=modified)
+
+
+class RemoveUnusedMetadataAndDocStringPass(ir.passes.InPlacePass):
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
```
justinchuby, line 192: I think there is also model metadata and docstring

gramalingam: As well as other objects, if IIRC (tensors?). I wonder, though, if removing model metadata is a bit extreme. Of course, it is up to the user who calls this ... but perhaps there should be options for users to control this ... which types of objects to remove metadata from.

onnxscript/ir/passes/common/unused_removal.py
titaiwangms: I guess I can use ir.traversal.RecursiveGraphIterator(function)?

justinchuby: Yes. Should be the exact same usage.

onnxscript/ir/passes/common/unused_removal.py
justinchuby: Is there a better file that can host this pass?

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -186,3 +186,44 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+
+        checked_graphs = set()
+        # 1. Clean up all of the nodes metadata properties
+        for node in ir.traversal.RecursiveGraphIterator(model.graph):
+            if node.metadata_props:
+                modified = True
+                logger.debug("Removed metadata from %s nodes", node.name)
+            node.metadata_props.clear()
+
+            # TODO: Use model.graphs to get the graph instead of node.graph
```
justinchuby, line 209: I think the current is ok. We loop over all nodes anyways

onnxscript/ir/passes/common/unused_removal.py
titaiwangms: For reference, ORT manages to remove graphs and nodes only for model shipping.

titaiwangms: https://github.com/microsoft/onnxruntime/blob/d7c688e15c1dc40f57140bff08c78e01a88b19fc/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L176-L179

onnxscript/ir/passes/common/unused_removal.py
titaiwangms: Created a new file

onnxscript/ir/_core.py
```diff
@@ -2605,15 +2605,13 @@ def __init__(
...
         graph: Graph,
         attributes: Sequence[Attr],
-        metadata_props: dict[str, str] | None = None,
     ) -> None:
         self._domain = domain
         self._name = name
         self._overload = overload
         self._graph = graph
         self._attributes = OrderedDict((attr.name, attr) for attr in attributes)
         self._metadata: _metadata.MetadataStore | None = None
```
justinchuby, line 2614: This can be removed too

onnxscript/ir/_core.py
```diff
@@ -2605,15 +2605,13 @@ def __init__(
         # and not from an outer scope
         graph: Graph,
         attributes: Sequence[Attr],
```
justinchuby, line 2607: Could you remove the `__slots__` too?

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
```diff
@@ -0,0 +1,58 @@
...
+class ClearMetadataAndDocStringPass(ir.passes.InPlacePass):
+    def __init__(self) -> None:
+        super().__init__()
+        self.modified = False
+
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        # 0. TODO: Should we clean model metadata and docstring?
+
+        # 1. Clean up the graph and the belonged nodes metadata properties
+        self._clear_graph_or_function_metadata_and_docstring(model.graph)
```
justinchuby, line 27: I would make `_clear_graph_or_function_metadata_and_docstring` return a boolean (modified). This way we don't need to manage a state in the class. Note that currently I think you would need to reset `modified` otherwise.

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
```diff
@@ -0,0 +1,58 @@
...
+
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+
+class ClearMetadataAndDocStringPass(ir.passes.InPlacePass):
+    def __init__(self) -> None:
+        super().__init__()
+        self.modified = False
```
justinchuby, line 21:
```suggestion
```

onnxscript/ir/serde.py
```diff
@@ -711,7 +712,6 @@ def deserialize_function(proto: onnx.FunctionProto) -> _core.Function:
         overload=getattr(proto, "overload", ""),
         graph=graph,
         attributes=typing.cast(List[_core.Attr], attributes),
```
justinchuby, line 714: We should maintain the api

justinchuby: Ref: https://github.com/microsoft/onnxscript/commit/dff5381a113783012daa63bef56a24119b7669ae

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
```diff
@@ -0,0 +1,58 @@
...
+]
+
+import logging
+
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+
+class ClearMetadataAndDocStringPass(ir.passes.InPlacePass):
```
shubhambhokare1, line 18: Would this be called in _optimizer.py (by default)?

onnxscript/ir/passes/common/unused_removal.py
shubhambhokare1: Adding onto this, #2167 for example, could provide a case where metadata_props could be used to propagate model properties (max_seq_len in this case). If we remove all model metadata by default, is there an alternative method to propagate model specific parameters or properties?

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
justinchuby: I don't think it should. This pass should only be used when we ship models. By default the optimizer produces a debugable (but optimized) model for users, which contains all the traces and metadata.

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
titaiwangms: Done

onnxscript/ir/_core.py
titaiwangms: Done

onnxscript/ir/passes/common/unused_removal.py
titaiwangms: This pass happens in the end of optimization call. Could you confirm if this works in the case?

onnxscript/ir/serde.py
titaiwangms: Could you check if the modification meets your expectation?

onnxscript/ir/passes/common/unused_removal.py
titaiwangms: Sorry, it's not used by default. It's an option for users.

onnxscript/ir/_core.py
```diff
@@ -2605,15 +2603,12 @@ def __init__(
...
-        metadata_props: dict[str, str] | None = None,
```
justinchuby, line 2608: I would preserve this line. In the ref commit there is a line of logic that calls dict.update

onnxscript/ir/serde.py
justinchuby: I think we should revert changes in this file, since it is the call site, which should rename unchanged from the changes we make in the implementation

justinchuby (approved):
Discussed offline. Changes to the metadata_props in ir.Function lgtm. Thanks!

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
```diff
@@ -0,0 +1,56 @@
...
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        # 0. TODO: Should we clean model metadata and docstring?
+
+        # 1. Clean up the graph and the belonged nodes metadata properties
+        modified = self._clear_graph_or_function_metadata_and_docstring(model.graph)
+
+        # 3. Clean up all of the functions metadata properties
+        for function in model.functions.values():
+            modified = self._clear_graph_or_function_metadata_and_docstring(function)
```
justinchuby, line 27:
```suggestion
            modified = self._clear_graph_or_function_metadata_and_docstring(function) or modified
```

onnxscript/ir/passes/common/clear_metadata_and_docstring.py
```diff
@@ -0,0 +1,56 @@
...
+        return ir.passes.PassResult(model, modified=modified)
+
+    def _clear_graph_or_function_metadata_and_docstring(
+        self,
+        graph_or_function: ir.Graph | ir.Function,
+    ) -> bool:
+        """Clear metadata and docstring from the graph or function."""
+        checked_graphs_or_functions = set()
```
justinchuby, line 35: mypy may ask you to annotate the type of this set

justinchuby (approved):

Issue #2181, opened 2025-04-10
[IR] Reconcile the graph attribute in ir.Node


justinchuby:
It is pointed to a graph when we call `function.append()`, but it is not when we call ir.Node(graph=function). Need to figure out how to reconcile this.

1. Can a node.graph point to a Function? (no)
2. The graph in Function is private, should it be exposed to users? (will expose when user calls node.graph)
3. What should users pass in to `ir.Node(graph=)` when the node is meant for a function being constructed? (it can be a function. `function.append` will do the right thing as assign the underlying graph to the node)

PR #2183, opened 2025-04-11
[IR] Reconcile graph in Node

justinchuby:
Always assign a `Graph` object to the node's graph.

Fix https://github.com/microsoft/onnxscript/issues/2181

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1145,22 +1145,24 @@ class Node(_protocols.NodeProtocol, _display.PrettyPrintable):
         Args:
             domain: The domain of the operator. For onnx operators, this is an empty string.
             op_type: The name of the operator.
-            inputs: The input values. When an input is None, it is an empty input.
+            inputs: The input values. When an input is ``None``, it is an empty input.
             attributes: The attributes. RefAttr can be used only when the node is defined in a Function.
             overload: The overload name when the node is invoking a function.
             num_outputs: The number of outputs of the node. If not specified, the number is 1.
-            outputs: The output values. If None, the outputs are created during initialization.
-            version: The version of the operator. If None, the version is unspecified and will follow that of the graph.
-            graph: The graph that the node belongs to. If None, the node is not added to any graph.
-                A `Node` must belong to zero or one graph.
-            name: The name of the node. If None, the node is anonymous.
+            outputs: The output values. If ``None``, the outputs are created during initialization.
+            version: The version of the operator. If ``None``, the version is unspecified and will follow that of the graph.
+            graph: The graph that the node belongs to. If ``None``, the node is not added to any graph.
+                A `Node` must belong to zero or one graph. If a :class:`Function`, the underlying graph
+                of the function is assigned to the node.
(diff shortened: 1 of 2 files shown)


onnxscript/ir/_core.py
```diff
@@ -1187,18 +1189,18 @@ def __init__(
         self._version: int | None = version
         self._metadata: _metadata.MetadataStore | None = None
         self._metadata_props: dict[str, str] | None = metadata_props
-        self._graph: Graph | Function | None = graph
+        # _graph is set by graph.append
+        self._graph: Graph | None = None
+        # Add the node to the graph if graph is specified
+        if graph is not None:
+            graph.append(self)
```
xadupre, line 1196: What happens if you append a node into a graph != node.graph? I tend to reuse nodes when I tweak graphs.

gramalingam, line 1196: Doesn't this alter the behavior (wrt what was happening previously)? Trying to understand what happens during optimization/rewriting (now vs. previously): I thought we had the ability to process a node N, and replace it with say new nodes M1, M2 ... and that they get inserted in the place where N was. Trying to reconcile how the above `append` interacts with that?

justinchuby: It will raise. You must set `node.graph=None` before adding it into a new graph to make the intention clear

justinchuby: I think this should be the same behavior as before (only moving the lines from https://github.com/microsoft/onnxscript/pull/2183/files#diff-33a5ea7182ac2a3ba4127e9fea3cd2db2dbe54810994ca210727d28fccf2b5cfL1198-L1201)

When we insert by calling graph.insert*, node.graph will be set in the insert methods; when we create the replacement nodes `graph` is typically None so they do not get appended.

onnxscript/ir/_core.py
```diff
@@ -1187,18 +1189,18 @@ def __init__(
         self._version: int | None = version
         self._metadata: _metadata.MetadataStore | None = None
         self._metadata_props: dict[str, str] | None = metadata_props
-        self._graph: Graph | Function | None = graph
+        # _graph is set by graph.append
+        self._graph: Graph | None = None
```
shubhambhokare1, line 1193: Where does self._graph get set to be =graph, does the setter for _graph get called somewhere else?

onnxscript/ir/_core.py
```diff
@@ -1187,18 +1189,18 @@ def __init__(
...
         self.doc_string = doc_string
 
         # Add the node as a use of the inputs
         for i, input_value in enumerate(self._inputs):
             if input_value is not None:
                 input_value._add_usage(self, i)  # pylint: disable=protected-access
 
-        # Add the node to the graph if graph is specified
-        if self._graph is not None:
-            self._graph.append(self)
```
titaiwangms, line 1200: I don't understand how moving these up make any difference?

titaiwangms (approved):

onnxscript/ir/_core.py
justinchuby: I moved it closer to were graph is created so it is more readable.

PR #2186, opened 2025-04-11
[torchlib] Fix calls to Unsqueeze to provide correct 1d axes

justinchuby:
Discovered in https://github.com/onnx/onnx/issues/6886#issuecomment-2797339394, the `axes` input in calls to unsqueeze are sometimes 0d. This is incorrect according to the ONNX spec. The PR fixes the instances I could find.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -2992,7 +2992,7 @@ def _aten_embedding_bag_onnx(
     # Get weight out according to indices_1d,
     new_weight = op.Gather(weight, indices_1d)
     # This happends after first step of Gather. Because Shape(indices)==Shape(per_sample_weights)
-    new_weight = op.Mul(new_weight, op.Unsqueeze(per_sample_weights, axes=1))
+    new_weight = op.Mul(new_weight, op.Unsqueeze(per_sample_weights, axes=[1]))
     weight_dim_1 = op.Reshape(op.Shape(weight, start=1), neg_1)
     indices_size = op.Shape(indices_1d)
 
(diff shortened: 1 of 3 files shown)


xadupre (approved):

justinchuby:
image

It is now correct

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -2992,7 +2992,7 @@ def _aten_embedding_bag_onnx(
     # Get weight out according to indices_1d,
     new_weight = op.Gather(weight, indices_1d)
     # This happends after first step of Gather. Because Shape(indices)==Shape(per_sample_weights)
```
justinchuby, line 2994:
```suggestion
    # This happens after first step of Gather. Because Shape(indices)==Shape(per_sample_weights)
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3132,7 +3132,7 @@ def _aten_embedding_bag_1d_padding_idx_onnx(
     # e.g. indices=[3,1,4,5,3] means get weight[[3,1,4,5,3]]
     indices_weight = op.Gather(weight, indices)
     # This happends after first step of Gather. Because Shape(indices)==Shape(per_sample_weights)
```
justinchuby, line 3134:
```suggestion
    # This happens after first step of Gather. Because Shape(indices)==Shape(per_sample_weights)
```

gramalingam (approved):
I remember noticing something like this while writing some fusion patterns. May need to double-check fusion patterns in case they are impacted by this. (Just noting it down here before I forget!)

PR #2190, opened 2025-04-12
[pass] Avoid lifting tensors that are too small to initializers

justinchuby:
Tensors with too few elements are usually not weights and are plenty. Lifting them will make the initializer list very noisy. I added a parameter `size_limit` to control this and defaulted it to 16.

### DIFF
--- a/onnxscript/ir/passes/common/constant_manipulation.py
+++ b/onnxscript/ir/passes/common/constant_manipulation.py
@@ -24,7 +24,7 @@ class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
         lift_all_constants: Whether to lift all Constant nodes, including those that does not contain a tensor attribute (e.g. with value_ints etc.)
             Default to False, where only Constants with the ``value`` attribute are lifted.
         size_limit: The minimum size of the tensor to be lifted. If the tensor contains
-            less than this number of elements, it will not be lifted.
+            number of elements less than size_limit, it will not be lifted.
     """
 
     def __init__(self, lift_all_constants: bool = False, size_limit: int = 16):
--- a/onnxscript/ir/passes/common/constant_manipulation.py
+++ b/onnxscript/ir/passes/common/constant_manipulation.py
@@ -90,34 +90,39 @@ class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
             )
             return None
 
+        tensor: ir.TensorProtocol
         if attr_name == "value":
             tensor = attr_value.as_tensor()
-            if tensor.size < self.size_limit:
-                logger.debug(
-                    "Tensor from node '%s' has less than %s elements",
-                    node.name,
-                    self.size_limit,
-                )
-                return None
-            return tensor
-        if attr_name == "value_int":
-            return ir.tensor(
+        elif attr_name == "value_int":
+            tensor = ir.tensor(
                 attr_value.as_int(), dtype=ir.DataType.INT64, name=initializer_name
             )
-        if attr_name == "value_ints":
-            return ir.tensor(
+        elif attr_name == "value_ints":
+            tensor = ir.tensor(
                 attr_value.as_ints(), dtype=ir.DataType.INT64, name=initializer_name
             )
-        if attr_name == "value_float":
-            return ir.tensor(
+        elif attr_name == "value_float":
+            tensor = ir.tensor(
                 attr_value.as_float(), dtype=ir.DataType.FLOAT, name=initializer_name
             )
-        if attr_name == "value_floats":
-            return ir.tensor(
+        elif attr_name == "value_floats":
+            tensor = ir.tensor(
                 attr_value.as_floats(), dtype=ir.DataType.FLOAT, name=initializer_name
             )
-        if attr_name in ("value_string", "value_strings"):
-            return ir.StringTensor(
+        elif attr_name in ("value_string", "value_strings"):
+            tensor = ir.StringTensor(
                 np.array(attr_value.value, dtype=np.bytes_), name=initializer_name
             )
-        return None
+        else:
+            raise ValueError(
+                f"Unsupported constant node '{node.name}' attribute '{attr_name}'"
+            )
+
+        if tensor.size < self.size_limit:
+            logger.debug(
+                "Tensor from node '%s' has less than %s elements",
+                node.name,
+                self.size_limit,
+            )
+            return None
+        return tensor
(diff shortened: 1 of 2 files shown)


onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -23,11 +23,14 @@ class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
     Attributes:
         lift_all_constants: Whether to lift all Constant nodes, including those that does not contain a tensor attribute (e.g. with value_ints etc.)
             Default to False, where only Constants with the ``value`` attribute are lifted.
+        size_limit: The minimum size of the tensor to be lifted. If the tensor contains
+            less than this number of elements, it will not be lifted.
     """
 
-    def __init__(self, lift_all_constants: bool = False):
+    def __init__(self, lift_all_constants: bool = False, size_limit: int = 16):
         super().__init__()
```
shubhambhokare1, line 31: How do we determine what an appropriate default size_limit value will be?

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -23,11 +23,14 @@ class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
     Attributes:
         lift_all_constants: Whether to lift all Constant nodes, including those that does not contain a tensor attribute (e.g. with value_ints etc.)
             Default to False, where only Constants with the ``value`` attribute are lifted.
+        size_limit: The minimum size of the tensor to be lifted. If the tensor contains
+            less than this number of elements, it will not be lifted.
```
shubhambhokare1, line 27: nit: confusing wording, maybe
```suggestion
            number of elements less than size_limit, it will not be lifted.
```

onnxscript/ir/passes/common/constant_manipulation.py
justinchuby: Just empirically. We don't want constants that represents shapes, axes etc. to be lifted

onnxscript/ir/passes/common/constant_manipulation.py
```diff
@@ -23,11 +23,14 @@ class LiftConstantsToInitializersPass(ir.passes.InPlacePass):
     Attributes:
         lift_all_constants: Whether to lift all Constant nodes, including those that does not contain a tensor attribute (e.g. with value_ints etc.)
             Default to False, where only Constants with the ``value`` attribute are lifted.
+        size_limit: The minimum size of the tensor to be lifted. If the tensor contains
+            number of elements less than size_limit, it will not be lifted.
```
justinchuby, line 27:
```suggestion
            number of elements less than size_limit, it will not be lifted. Default is 16.
```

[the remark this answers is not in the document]
justinchuby: Done

shubhambhokare1 (approved):

titaiwangms:
Understand the motivation, but this will create a different behavior to shipped models from GenAI. They lifted all constants with "value". I am not sure whether it's easier to implement for them or they have specific request/process, so they are lifting all tensors.

justinchuby:
> Understand the motivation, but this will create a different behavior to shipped models from GenAI. They lifted all constants with "value". I am not sure whether it's easier to implement for them or they have specific request/process, so they are lifting all tensors.

I think it is possible to users to tweak the value of the size limit to control this behavior. As part of the general optimization passes I think it makes more sense to maintain the proposed default behavior from this PR. If there are special requirements (we need to understand them better than looking at the current proto processing behavior) we can do (1) change the size limit in ort optimization passes (2) update genai so that it is less constrained

justinchuby:
@kvaishnavi I assume the logic is for pattern matching. Could you share the assumptions the patterns make? What are the patterns? Or is it for different purposes?

kunal-vaishnavi:
> @kunal-vaishnavi I assume the logic is for pattern matching. Could you share the assumptions the patterns make? What are the patterns? Or is it for different purposes?

The ORT transformer optimizer [applies](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/fusion_attention.py#L708-L716) its graph fusions on initializers instead of `Constant` nodes. In order for it to run successfully, all `Constant` nodes have to be [converted](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139) to initializers.

justinchuby:
> > @kunal-vaishnavi I assume the logic is for pattern matching. Could you share the assumptions the patterns make? What are the patterns? Or is it for different purposes?
> 
> The ORT transformer optimizer [applies](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/fusion_attention.py#L708-L716) its graph fusions on initializers instead of `Constant` nodes. In order for it to run successfully, all `Constant` nodes have to be [converted](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139) to initializers.

Is this the only fusion? I can see the example is on model weights/biases so this should already be accounted for.

kunal-vaishnavi:
> > > @kunal-vaishnavi I assume the logic is for pattern matching. Could you share the assumptions the patterns make? What are the patterns? Or is it for different purposes?
> > 
> > 
> > The ORT transformer optimizer [applies](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/fusion_attention.py#L708-L716) its graph fusions on initializers instead of `Constant` nodes. In order for it to run successfully, all `Constant` nodes have to be [converted](https://github.com/microsoft/onnxruntime/blob/9a993c3720fee8164e7f9469e50dcf16e64bf43b/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139) to initializers.
> 
> Is this the only fusion? I can see the example is on model weights/biases so this should already be accounted for.

This is just one example. Almost all fusions in the ORT transformer optimizer expect initializers instead of `Constant` nodes.

justinchuby:
Thanks. To be on the safe side we can set size_limit to 0 when running the `optimize_for_ort` api. @titaiwangms

titaiwangms:
> Thanks. To be on the safe side we can set size_limit to 0 when running the `optimize_for_ort` api. @titaiwangms

We can just apply this pass again with ClearMetadata pass after the export. It's not necessarily to have this change by default if we believe size_limit=16 benefits more in general cases.

PR #2193, opened 2025-04-13
Update ort-nightly version in test

justinchuby:


titaiwangms (approved):

shubhambhokare1 (commented):
Could you check what is causing the failing test in ci, (i believe it could be tolerance issues)

justinchuby:
> Could you check what is causing the failing test in ci, (i believe it could be tolerance issues)

I am not exactly sure what cause the issue. I think it is ephemeral due to randomness.

PR #2196, opened 2025-04-14
Updated docs for rewrite-patterns

### LINKED ISSUE #2169: [docs] RewriteRuleSet's order dependency should be documented
Greetings!
Using rewriter, I expected `RewriteRuleSet` is ,namely a set, rules not ordered and applying while any is applicable. But, as I see, order is matters. Setting `commute` flag didn't helped. Is this behaviour is supposed?

_edit:_------------------------

Yes, it is fixed as simple as

      while(
          pattern.RewriteRuleSet(
              [rule_relu, rule_leak, rule_tanh, rule_bnorm ]
          ).apply_to_model(ir_model)
      ):
          pass

LuniaKunal:
Description
Added a note to the rewrite patterns tutorial clarifying that the order of rules in `pattern_rewrite_rules` matters. Some rules depend on others being applied first, so incorrect order may lead to unexpected results.

@justinchuby let me know, does this helps Or something else is needed?

Fixes #2169

### DIFF
--- a/docs/tutorial/rewriter/rewrite_patterns.md
+++ b/docs/tutorial/rewriter/rewrite_patterns.md
@@ -152,6 +152,20 @@ In order to apply this method to the example above, first create the two separat
 :pyobject: erf_gelu_pattern_2
 ```
 
+:::{note}
+:name: rule-application-order-matters
+
+### Order of Rewrite Rules Matters
+
+When you pass multiple rules in `pattern_rewrite_rules`, the **order in which they appear is important**. 
+
+This is because some rules may depend on patterns created or modified by earlier rules. For example, if `rule1` can only match after `rule2` has made a specific change in the model, then `rule2` must come **before** `rule1` in the list.
+
+If you're not seeing expected results, try adjusting the order or applying the rule set in a loop until no more changes occur.
+
+:::
+
+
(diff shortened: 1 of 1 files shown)


justinchuby (approved):
Thank you!

justinchuby:
Could you remove the trailing whitespaces?

docs/tutorial/rewriter/rewrite_patterns.md
```diff
@@ -152,6 +152,20 @@ In order to apply this method to the example above, first create the two separat
...
 ```
 
+:::{note}
+:name: rule-application-order-matters
+
+### Order of Rewrite Rules Matters
+
+When you pass multiple rules in `pattern_rewrite_rules`, the **order in which they appear is important**. 
+
+This is because some rules may depend on patterns created or modified by earlier rules. For example, if `rule1` can only match after `rule2` has made a specific change in the model, then `rule2` must come **before** `rule1` in the list.
```
justinchuby, line 162:
```suggestion
This is because some rules may depend on patterns created or modified by earlier rules. For example, if `rule2` can only match after `rule1` has made a specific change in the model, then `rule1` must come **before** `rule2` in the list.
```

nit: slightly less cognitive load

shubhambhokare1 (approved):

PR #2199, opened 2025-04-14
[pass] Implement checker pass and refactor shape inference

justinchuby:
- Refactor shape_inference pass to extract logic for handling large models for onnx c-api.
- Implement an onnx checker pass leveraging the refactored logic.

### DIFF
--- a/onnxscript/ir/passes/common/_c_api_utils.py
+++ b/onnxscript/ir/passes/common/_c_api_utils.py
@@ -58,8 +58,10 @@ def call_onnx_api(func: Callable[[onnx.ModelProto], _R], model: ir.Model) -> _R:
             assert initializer.name is not None
             model.graph.initializers.pop(initializer.name)
 
+    proto = ir.serde.serialize_model(model)
+
     try:
-        proto = ir.serde.serialize_model(model)
+        # Call the ONNX C API function
         result = func(proto)
     finally:
         # Restore the original initializer values so the model is unchanged
--- a/onnxscript/ir/passes/common/_c_api_utils.py
+++ b/onnxscript/ir/passes/common/_c_api_utils.py
@@ -20,28 +20,26 @@ _BIG_TENSOR_SIZE_LIMIT = 1000  # 1KB
 
 
 def call_onnx_api(
-    func: Callable[[onnx.ModelProto], onnx.ModelProto],
-    model: ir.Model,
-    merge_func: Callable[[ir.Model, onnx.ModelProto], tuple[ir.Model, bool]],
-) -> tuple[ir.Model, bool]:
+    func: Callable[[onnx.ModelProto], onnx.ModelProto], model: ir.Model
+) -> onnx.ModelProto:
     """Call an ONNX C API function by temporarily removing initializers.
+
     This is necessary because the ONNX C API does not support large models
-    with initializers that have large tensor values.
+    with initializers that have large tensor values. The input model is left
+    unchanged no matter the call succeeds or not.
 
     Args:
         func: Partially applied function that takes a model proto and returns a model proto.
         model: The IR model to pass to the API function.
-        merge_func: Function that merges IR model with information from the model proto.
 
     Returns:
-        A tuple containing the modified model and a boolean indicating whether the model was modified.
+        The resulting ModelProto that contains the result of the API call.
     """
 
     # Store the original initializer values so they can be restored
     initializer_values = tuple(model.graph.initializers.values())
     tensors = {v.name: v.const_value for v in initializer_values}
     original_inputs_len = len(model.graph.inputs)
-    initializer_names = {v.name for v in initializer_values}
 
     # Turn the initializers into inputs and clear the initializers
     # to limit the model size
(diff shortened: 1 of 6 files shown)


onnxscript/ir/passes/common/shape_inference.py
```diff
@@ -9,20 +9,49 @@
...
+    Args:
+        model: The original IR model.
+        inferred_proto: The inferred ONNX model.
+
+    Returns:
+        A tuple containing the modified model and a boolean indicating whether the model was modified.
+    """
+    inferred_model = ir.serde.deserialize_model(inferred_proto)
+    modified = False
+    for original_graph, inferred_graph in (model.graphs(), inferred_model.graphs()):
```
justinchuby, line 35:
```suggestion
    for original_graph, inferred_graph in zip(model.graphs(), inferred_model.graphs()):
```

onnxscript/ir/passes/common/_c_api_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    import onnx
+
+
+logger = logging.getLogger(__name__)
+# Temporarily remove initializers larger than this size to keep model size down
+# for the onnx.shape_inference call because it needs to serialize the model
+_BIG_TENSOR_SIZE_LIMIT = 1000  # 1KB
+
+
+def call_onnx_api(
```
titaiwangms, line 22: What are other usages? It looks like it's very specifically designed for shape type inference pass regarding the initializer processings?

justinchuby: It is used in the checker as well. It is designed to remove the initializers temporarily to that calls to onnx capis are not constrained by the 2gb limit.

onnxscript/ir/passes/common/shape_inference_test.py
```diff
@@ -7,10 +7,13 @@
 import numpy as np
 
 from onnxscript import ir
-from onnxscript.ir.passes.common import shape_inference
+from onnxscript.ir.passes.common import _c_api_utils, shape_inference
 
 
 class TestShapeInferencePass(unittest.TestCase):
+    def test_pass_is_in_place(self):
```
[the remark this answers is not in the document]
justinchuby: Could you explain?

onnxscript/ir/passes/common/debugging.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Passes for debugging purposes."""
```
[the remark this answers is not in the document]
justinchuby: If we run checker directly, models larger than 2gb will not be processed unless we externalize the data. Doing it this way (in this pass) avoids externalizing the data or hitting the 2gb limit.

onnxscript/ir/passes/common/debugging.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
justinchuby: SG. Do you have suggestions?

onnxscript/ir/passes/common/shape_inference_test.py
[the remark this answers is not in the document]
titaiwangms: Sorry, thought it was a tool

titaiwangms (approved):

onnxscript/ir/passes/common/_c_api_utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns:
+        A tuple containing the modified model and a boolean indicating whether the model was modified.
+    """
+
+    # Store the original initializer values so they can be restored
+    initializer_values = tuple(model.graph.initializers.values())
+    tensors = {v.name: v.const_value for v in initializer_values}
+    original_inputs_len = len(model.graph.inputs)
+    initializer_names = {v.name for v in initializer_values}
```
gramalingam, line 44: nit: Looks like the check is line 73 is not really being used effectively? Wouldn't it be better to ensure that we create only a list of the temporarily removed initializers in the loop below?

onnxscript/ir/passes/common/_c_api_utils.py
```diff
@@ -0,0 +1,83 @@
...
+                initializer.const_value = tensors[initializer.name]
+                model.graph.register_initializer(initializer)
+
+        # Restore the original inputs
+        inputs = model.graph.inputs[:original_inputs_len]
+        model.graph.inputs.clear()
+        model.graph.inputs.extend(inputs)
+
+    # Merge the result with the original model
+    return merge_func(model, result_proto)
```
gramalingam, line 83: Given that this is the last step, I wonder if it is worth the bother to introduce the merge_func parameter and explaining it ... may be easier for caller to do it with the returned value?

onnxscript/ir/passes/common/_c_api_utils.py
gramalingam: Another possibility is to provide two primitives, one for the "before" part (removing large initializers from model) and one for the "after" part (adding the large initializers back).

onnxscript/ir/passes/common/debugging.py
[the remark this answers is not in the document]
gramalingam: Perhaps `onnx_checker_pass.py` ? Preferable to stick to name used by the underlying onnx utility ... I guess at some point we could have other alternative (pure-python) implementations ... (at least for shape-inference, but probably not for the checker). So, adding the `onnx_` prefix.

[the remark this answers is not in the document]
justinchuby: Initially I was hoping to group debug passes like checker, model print, model dump etc. in the same file, thus "debugging". But splitting them up works for me too

onnxscript/ir/passes/common/_c_api_utils.py
justinchuby: I updated the function to use a dictionary and removed the check. Does it look like what you had in mind?

onnxscript/ir/passes/common/_c_api_utils.py
justinchuby: I removed merge_func and made caller handle the result.

onnxscript/ir/passes/common/_c_api_utils.py
justinchuby: I went with the recommendation below. I wanted to ensure the model is unchanged if the call fails, which is handled by the try finally block inside the call_api function.

onnxscript/ir/passes/common/debugging.py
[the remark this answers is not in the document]
justinchuby: Changed to onnx_checker

titaiwangms (approved):

PR #2200, opened 2025-04-14
[IR] Implement `model.graphs()`

justinchuby:
Implement model.graphs() as a way to retrieve the main graph and all subgraphs of it in the model.

Given

(1) how useful the method is
(2) I couldn't find an appropriate name for it in `traversal.py`
(3) Users familiar with onnxruntime optimization tools expect this method. In PyTorch a similar `modules()` method exists.

I created this method as a core method instead of an iterator in `traversal.py`.

Depends on https://github.com/microsoft/onnxscript/pull/2183

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -50,6 +50,7 @@ from onnxscript.ir import (
     _name_authority,
     _protocols,
     _type_casting,
+    traversal,
 )
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


onnxscript/ir/_core.py
```diff
@@ -2561,6 +2561,25 @@ def __repr__(self) -> str:
...
+        `onnxscript.ir.traversal.RecursiveGraphIterator` for more advanced
+        traversals on nodes.
+        """
+        # NOTE(justinchuby): Given
+        # (1) how useful the method is
+        # (2) I couldn't find an appropriate name for it in `traversal.py`
+        # (3) Users familiar with onnxruntime optimization tools expect this method
+        # I created this method as a core method instead of an iterator in
+        # `traversal.py`.
+        seen_graphs: set[Graph] = set()
```
shubhambhokare1, line 2577:
```suggestion
        seen_graphs: set[Union[Function, Graph]] = set()
```

shubhambhokare1 (approved):

PR #2205, opened 2025-04-16
Move the gemm_to_matmul_add rule to ort_fusions rules

justinchuby:
Instead, remove the reshape nodes and keep Gemm nodes.


justinchuby:
NVM. There are rank issues.

PR #2206, opened 2025-04-16
[passes] Move inliner to common passes

justinchuby:
Expose inliner to common passes for general usage.

Fix https://github.com/microsoft/onnxscript/issues/2194


onnxscript/optimizer/__init__.py
```diff
@@ -35,6 +35,12 @@ def optimize(model: ir.Model, *args, **kwargs) -> ir.Model:
         return legacy_optimizer.optimize(model, *args, **kwargs)
 
 
+def inline(model: ir.Model) -> None:
+    """Inline all function calls (recursively) in the model."""
+    if model.functions:
+        onnxscript.ir.passes.common.inliner.InlinePass()(model)
```
titaiwangms, line 41: I wonder if we should do this to all the passes that are not applied by default? What is the pass API we have in mind?

titaiwangms (approved):

onnxscript/optimizer/__init__.py
justinchuby: Do you mean creating a function for them? I think only for convenience. I imagine the pass api is how we currently do in `optimize()`, with a pass manager that's driving them. A little bit like how you can compose nn.Module inside nn.Module, we can do the same with passes too.

titaiwangms: OK.

I found another InlinePass under_inliner.py. Should we clean that up?

justinchuby: Let me check, thanks

justinchuby: I think this is the only one after this PR? Could you confirm?

PR #2209, opened 2025-04-16
Test fft normalization

justinchuby:
The numbers seems closer-ish for fft_c2r, but don't really match.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/fft.py
+++ b/onnxscript/function_libs/torch_lib/ops/fft.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 
 from typing import Optional, Sequence
 
-from onnxscript import INT64
+from onnxscript import INT64, ir
 from onnxscript.function_libs.torch_lib.registration import torch_op
 from onnxscript.function_libs.torch_lib.tensor_typing import TFloat
 from onnxscript.onnx_opset import opset18 as op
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/fft.py
```diff
@@ -118,12 +118,18 @@ def aten__fft_c2r(
...
+            transformed,
+            dft_length=last_dim_size,
+            axis=dimension,
+            inverse=True,
+            onesided=False,
+        )
+        * scale
     )
     transformed = _fftn_onnx_normalization(
         transformed,
```
bmehta001, line 135: Can you replace line 137 (op.Shape...) with `op.CastLike(last_dim_size, self)` and then remove `scale`? Would that yield the same/better results?

bmehta001 (changes requested):
Perhaps simplify the normalization by passing in the dimension size you want directly to the normalization function.  I think there is an underlying issue with my c2r implementation, though, because it should be able to support multiple axes. I can try to see if something like what is done in https://github.com/onnx/onnx/issues/6016 will help

justinchuby:
> Perhaps simplify the normalization by passing in the dimension size you want directly to the normalization function. I think there is an underlying issue with my c2r implementation, though, because it should be able to support multiple axes. I can try to see if something like what is done in [onnx/onnx#6016](https://github.com/onnx/onnx/issues/6016) will help

👍 looks like they recreated the conjugate part. It's like doubling the work that was meant to be saved but I guess that's the best option right now?

PR #2210, opened 2025-04-17
Allow sdpa fusion to accept custom scale factor

shubhambhokare1:


onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,55 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
+        self, op, query, key_transposed, value, mask, query_scale, key_scale, qk_scale, **_
+    ):
+        sdpa_args = [query, key_transposed, value]
         if self._use_mask:
-            return op.SDPA(query, key_transposed, value, mask, _domain="ai.onnxruntime.fusion")
-        else:
-            return op.SDPA(query, key_transposed, value, _domain="ai.onnxruntime.fusion")
+            sdpa_args.append(mask)
+        if self._custom_scale:
+            scale = _ir_utils.get_singleton_value(query_scale if self._pre_scale else qk_scale)
```
gramalingam, line 107: Should it be `query_scale * query_scale` in the _pre_scale case?

gramalingam: I think it might be helpful to keep the logic for this in one place: eg., if we want it in `check` method, we can let `self._custom_scale` be an optional float, and set its value there. We can just directly use it here.

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,55 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
+            else:
+                self._scale = expected_scaling_factor
+            # Ensure the scaling factor for key is the same as for query
+            if _ir_utils.get_singleton_value(key_scale) is None:
                 return check_result.fail(
-                    "Key scale is not a scalar or does not match the expected scaling factor.",
+                    "Key scale is not a scalar.",
                     key_scale,
                 )
+            if not _ir_utils.is_singleton_value(key_scale, sqrt_scaling_factor, rtol=1e-3):
```
gramalingam, line 80: Isn't this test redundant? I think we want the check down below to ensure that query_scale and key_scale are same. And we are checking (above) whether query_scale equals sqrt(...) or not?

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,55 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
+                if _ir_utils.get_singleton_value(query_scale) != _ir_utils.get_singleton_value(
```
gramalingam, line 81: nit: this is a float-value comparison. I think some comparison with atol/rtol would be better (but we need some thought on what those values should be, even in other places where we use 1e-3 ... that's a generic pre-existing TODO.)

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,55 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
-            if not _ir_utils.is_singleton_value(query_scale, sqrt_scaling_factor, rtol=1e-3):
+            # Calculate the scaling factor for query
+            if _ir_utils.get_singleton_value(query_scale) is None:
                 return check_result.fail(
-                    "Query scale is not a scalar or does not match the expected scaling factor.",
+                    "Query scale is not a scalar.",
                     query_scale,
                 )
-            if not _ir_utils.is_singleton_value(key_scale, sqrt_scaling_factor, rtol=1e-3):
+            if not _ir_utils.is_singleton_value(query_scale, sqrt_scaling_factor, rtol=1e-3):
```
gramalingam, line 69: minor nit: Once we start using the value of query_scale, it is better not to use an utility like is_singleton_value. It is simpler and more efficient to do:
```py
   if (query_scale_value := _ir_utils.get_singleton_value(query_scale)) is None:
      fail ...
   # Here, use query_scale_value directly
```

onnxscript/rewriter/ort_fusions/sdpa_test.py
```diff
@@ -161,6 +270,15 @@ class TestSDPAFusion(unittest.TestCase):
...
             ("post_mul", _masked_post_mul_sdpa_script),
+            ("custom_scale_post_mul", _custom_scale_post_mul_sdpa_script),
+            ("custom_scale_post_div", _custom_scale_post_div_sdpa_script),
+            ("custom_scale_pre_mul", _custom_scale_pre_mul_sdpa_script),
+            ("custom_scale_pre_div", _custom_scale_pre_div_sdpa_script),
+            ("custom_scale_post_mul_masked", _custom_scale_post_mul_sdpa_script),
+            ("custom_scale_post_div_masked", _custom_scale_post_div_sdpa_script),
+            ("custom_scale_pre_mul_masked", _custom_scale_pre_mul_sdpa_script),
+            ("custom_scale_pre_div_masked", _custom_scale_pre_div_sdpa_script),
+            (_custom_multi_scale_pre_mul_sdpa_script, _custom_multi_scale_pre_mul_sdpa_script),
```
gramalingam, line 281: Is there a typo in above line? Should be a string for first parameter, i guess?

onnxscript/rewriter/ort_fusions/sdpa.py
shubhambhokare1: Used a float-value comparison with rtol for now. Could fix/propose a mechanism to allow caller to set rtol/atol in a future PR as an extension.

onnxscript/rewriter/ort_fusions/sdpa_test.py
```diff
@@ -161,6 +270,15 @@ class TestSDPAFusion(unittest.TestCase):
             ("pre_mul", _masked_pre_mul_sdpa_script),
             ("post_div", _masked_post_div_sdpa_script),
             ("post_mul", _masked_post_mul_sdpa_script),
+            ("custom_scale_post_mul", _custom_scale_post_mul_sdpa_script),
```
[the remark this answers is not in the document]
shubhambhokare1: Added checks to ensure scale-attribute is present and the value is set correctly. If you can think of any other checks that could improve the test robustness, feel free to suggest.

As for the second point, it would be great to have such a test, as you mentioned however this can be only done when have an SDPA implementation that we could use.

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,51 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
-            if not _ir_utils.is_singleton_value(query_scale, sqrt_scaling_factor, rtol=1e-3):
+            # Calculate the scaling factor for query
+            if (query_scale_value := _ir_utils.get_singleton_value(query_scale)) is None:
                 return check_result.fail(
-                    "Query scale is not a scalar or does not match the expected scaling factor.",
+                    "Query scale is not a scalar.",
                     query_scale,
                 )
-            if not _ir_utils.is_singleton_value(key_scale, sqrt_scaling_factor, rtol=1e-3):
+            if not math.isclose(query_scale_value, sqrt_scaling_factor, rel_tol=1e-3):
```
gramalingam, line 69: nit: Can we move this if-then-else down below *after* the check in line 79? Hard for a reader to understand what's happening unless they understand the check down below that query and key scale factors are same.

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,51 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
+                self._scale = query_scale_value * query_scale_value
+            else:
+                self._scale = expected_scaling_factor
```
gramalingam, line 72: Should we be setting this to `None`, if we want to indicate it is the default-value? If we are not doing that, not sure we even need this if-then-else, we can just use like 70 always?

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -57,34 +58,51 @@ def check(self, op, query, key_transposed, value, mask, query_scale, key_scale,
...
+            if (qk_scale_value := _ir_utils.get_singleton_value(qk_scale)) is None:
                 return check_result.fail(
-                    "QK scale is not a scalar or does not match the expected scaling factor.",
+                    "QK scale is not a scalar.",
                     qk_scale,
                 )
+            if not math.isclose(qk_scale_value, expected_scaling_factor, rel_tol=1e-3):
+                self._scale = qk_scale_value
+            else:
+                self._scale = expected_scaling_factor
```
gramalingam, line 95: As discussed above. We should either check if default-value works and use No scaling factor ... or else, no need to for if-then-else, always use line 93.

onnxscript/rewriter/ort_fusions/sdpa.py
shubhambhokare1: Setting it to `None` makes more sense. Pass scale(=None) to SDPA for these cases, so the SDPA implementation knows that default scale value is to be used). Obviously, we might need to adjust the semantics once we have the attention/SDPA to test against, but for now this should be good.

gramalingam (approved):

PR #2214, opened 2025-04-19
[pass] Create version converter pass

justinchuby:
Use both the onnxscript version converter and optionally fall back to the onnx version converter if the target version is unsupported.

Created `version_supported` helper function for users to check if a target version is supported by the onnxscript version converter.

Use the converter in pytorch apis.


onnxscript/version_converter/_version_converter.py
```diff
@@ -38,6 +39,14 @@ class Replacement:
 AdapterFunction = Callable[[ir.Node, orp.RewriterContext], ReturnValue]
 
 
+def version_supported(current_version: int, target_version: int) -> bool:
+    """Check if the target version is supported by the current version."""
+    return (
+        SUPPORTED_MIN_ONNX_OPSET
+        <= current_version
+        <= target_version
+        <= SUPPORTED_MAX_ONNX_OPSET
```
justinchuby, line 48:
```suggestion
        <= SUPPORTED_MAX_ONNX_OPSET
    )
```

onnxscript/version_converter/__init__.py
```diff
@@ -3,19 +3,157 @@
...
 
-    # In functions, we can have attribute-parameters, which means we don't know the value of the attribute.
-    # Hence, we inline all the functions.
-    onnxscript.optimizer.inline(model)
-    _version_converter.convert_version(model, target_version)
+    Args:
+        model: The model to convert.
+        target_version: The target ONNX opset version.
+        fallback: Whether to fallback to the onnx version converter if the
+            target version is not supported.
```
shubhambhokare1, line 157: nit: specify default behavior

onnxscript/version_converter/__init__.py
```diff
@@ -3,19 +3,157 @@
...
```
justinchuby, line 157:
```suggestion
            target version is not supported. Default is False.
```

shubhambhokare1 (approved):

PR #2221, opened 2025-04-23
Add fusion rules (Whisper optimizations)

shubhambhokare1:
Add fusion rules to support the optimization of Whisper models.

Fusions added:

- Basic Fusions:
    * additional pattern for erfgelu [moved to #2222]
- SkipLayerNorm:
    * #2259 
    * Fusion patterns where skip_sum is also an output
    * Bias + SkipLayerNorm -> SkipLayerNorm (with bias) [moved to #2222]
- BiasGelu Fusion [moved to #2222]
- SDPA:
    * Support for pattern where only q is pre-scaled
- MHA:
    * Patterns with/without past/present keys/values
    * Patterns with non-rotary embeddings
    * Patterns with/without mask
    * Patterns with cross-attention (only for past key/value patterns)
 - MHA Bias Fusion:
    * Bias was offloaded to Attention fusion previously, this fusion fixes that
 - Attention:
    * Patterns where Q, K and V do not come from slicing

TODO:
- [x]  Fix SDPA singular prescale case, due to lost shape information
- [x] - Enable check conditions when #2210 is merged
- [x] - Improve/Rewrite whisper model test case to be similar to that of smollm (for eg)
- [x] - Fix failing test cases to account for new patterns
- [x] - Add isolated test cases for new fusions like BiasGelu, SkipLayerNorm etc


onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -8,24 +8,67 @@
...
+                if self._pre_scale_q:
+                    query = op.Div(query, qk_scale)
+                else:
+                    query = op.Div(query, query_scale)
+                    key_transposed = op.Div(key_transposed, key_scale)
+
+        # There might be patterns where the reshape and transpose are done
+        # after the pre-scaling. If the inputs are 3D, we need to reshape them to 4D
+        # and apply the approriate transposes to query.
+        if self._has_3d_inputs and self._pre_scale_q:
```
gramalingam, line 66: Do these 2 options interact? Doesn't 3D/4D variation affect all (query/key/value)?

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -38,7 +38,7 @@ def _skip_rms_normalization(op, input, skip, gamma, epsilon, stash_type):
 
 
 def _skip_layer_norm_pattern(op, input, skip, gamma, beta, epsilon, stash_type):
-    skip_sum = op.Add(input, skip)
+    skip_sum = op.Add(skip, input)
```
gramalingam, line 41: I thought Skip Normalization related fusions were moved into a separate PR? Are these fixes to that PR?

onnxscript/rewriter/ort_fusions/fuse_mha_bias.py
```diff
@@ -0,0 +1,203 @@
...
+        v_bias,
+        mask,
+        past_key,
+        past_value,
+        num_heads,
+        **_,
+    ):
+        if self._q_no_bias:
+            q_bias = op.Constant(
+                value=ir.tensor(numpy.zeros((self.Dh_q,), dtype=numpy.float32))
```
gramalingam, line 153: Do we care only about 32-bit float query/key/value?

onnxscript/rewriter/ort_fusions/fuse_mha_bias.py
```diff
@@ -0,0 +1,203 @@
...
+        )
+
+
+parameter_combinations = [
+    {
+        "q_no_bias": q_no_bias,
+        "k_no_bias": k_no_bias,
+        "v_no_bias": v_no_bias,
+    }
+    for q_no_bias in [False, True]
```
gramalingam, line 184: The (True, True, True) case seems unnecessary? Won't it unnecessary introduce a zero-valued bias and add it? Seems inefficient.

onnxscript/rewriter/ort_fusions/_whisper_tiny.py
```diff
@@ -0,0 +1,112 @@
...
+
+def make_decoder_model():
+    pass
+
+
+class TestMultiHeadAttention(unittest.TestCase):
+    def test_whisper_tiny(self):
+        # Generate encoder model
+        whisper_encoder_model = onnx.load(
+            "/workspace/testing/whisper-opt/whisper-tiny-4.48/whisper-tiny_encoder.onnx"
```
gramalingam, line 30: Where are these files stored? I don't see them in the PR. Would it be possible/worth to convert the model to onnxscript and store it in source-format like the Smollm models?

shubhambhokare1: Yup, that's the current plan (see task 3 of the task list)

shubhambhokare1: Added models now.

onnxscript/rewriter/ort_fusions/skip_normalization.py
shubhambhokare1: Have reverted the change you highlighted above. However, the fusion pattern for the skip normalization fusion with bias was modified in this PR as the pattern previously was too general and was causing replacements to happen if when they shouldn't have.

onnxscript/rewriter/ort_fusions/fuse_mha_bias.py
shubhambhokare1: Modified to have dtype of the input to the mha, the contrib op only supports float and float16, so that will be the constraint on the bias as well

onnxscript/rewriter/ort_fusions/fuse_mha_bias.py
shubhambhokare1: Set bias=None for this case.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -75,15 +76,18 @@ def fuse_xformers(model: ir.Model) -> tuple[ir.Model, dict[str, int]]:
     fusion_count["partial_rotary_embedding"] = fuse_partial_rotary_embedding(model)
     fusion_count["cos_sin_cache"] = fuse_cos_sin_cache(model)
     fusion_count["sdpa"] = fuse_sdpa(model)
+    model = _pre_optimize(model)
```
gramalingam, line 79: Can you add a comment to the source why this was necessary? I suspect this may be something we need to cleanup later.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -29,32 +33,37 @@ def pattern(
...
-        value_BSD = op.Slice(
-            projected,
-            _allow_other_inputs=True,
-            _outputs=["value_mm_sliced"],
-        )
+            # Slice packed Matmul QKV into Q, K, and V
+            # Q, K, and V are of shape (B, S, D)
+            query_BSD = op.Slice(
+                projected,
+                _allow_other_inputs=True,
```
gramalingam, line 54: A generic comment: whenever we use this (that is, not check the other input parameters of the Slice op), we may need some suitable checks in the check method to ensure correctness. Typically that would mean checking that the axes, starts, ends, steps are the right value. If we expect "axes" to be a specific value, it may be easier to specify it directly in the pattern. The recently introduced `pattern.ANY_VALUE` will allow you to skip values checked in the checker method, while specifying expected values like `[3]` elsewhere.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -99,11 +108,11 @@ def pattern(
                 query_BSD,
                 key_BSD,
                 value_BSD,
-                # bias
-                # key_padding_mask
-                # attention_bias,
-                # past_key
-                # past_value
+                qkv_bias,
+                None,
```
gramalingam, line 112: nit: adding `# key_padding_mask` or whatever this parameter corresponds to will help the reader.

onnxscript/rewriter/ort_fusions/fuse_mha_bias.py
gramalingam: But does the rule do anything in the (True, True, True) case? Based on my reading, it does nothing. So, why not exclude this combination from the rules in the first place? You just need to add a `if not (q_no_bias and k_no_bias and v_no_bias)` down below?

onnxscript/rewriter/ort_fusions/mha.py
```diff
@@ -69,7 +90,11 @@ def pattern(
             _outputs=["key_BSHDh"],
         )
         # Transpose from (B, S, H, D/H) to (B, H, S, D/H)
-        key_BHSDh = op.Transpose(key_BSHDh, perm=[0, 2, 1, 3])
+        # TODO: Fix condition
+        if not self._is_cross_attention and self._has_past_present:
+            key_BHSDh = op.Transpose(key_BSHDh, perm=[0, 2, 1, 3])
+        else:
+            key_BHSDh = op.Transpose(key_BSHDh, perm=[0, 2, 3, 1])
```
gramalingam, line 97: Minor forward-looking suggestion: such variations can be handled in the check-condition, by allowing perm to be a pattern variable and using `perm=perm` in the pattern. Mentioning this since the restricted disjunction will not allow two such variants with same op.

gramalingam: But this is fine for current PR ... this just FYI

onnxscript/rewriter/ort_fusions/attention.py
shubhambhokare1: Modified

onnxscript/rewriter/ort_fusions/sdpa.py
shubhambhokare1: If all query/key/values are just 3D or just 4D, SDPA should be able to handle them. But this a very specific scenario in whisper, where the Mul for the query happens before the Reshape/Transpose, so we are looking at 3D query but a 4D key/value.

`pre_scale_q` handles the case where only q is scaled however it doesn't care whether query is 3D or 4D, as there could be cases where just 4D query is pre-scaled as well.

Now, `has_3d_inputs` which in retrospect might be confusing as a name, only checks in cases where query is pre-scaled, whether that query was 3d/4d. I think changing the parameter name to `has_3d_query` might be a better idea for now.

onnxscript/rewriter/ort_fusions/attention.py
```diff
@@ -29,32 +33,43 @@ def pattern(
...
-        value_BSD = op.Slice(
-            projected,
-            _allow_other_inputs=True,
-            _outputs=["value_mm_sliced"],
-        )
+            # Slice packed Matmul QKV into Q, K, and V
+            # Q, K, and V are of shape (B, S, D)
+            query_BSD = op.Slice(
+                projected,
+                pattern.ANY_VALUE,  # starts
```
gramalingam, line 54: Are there checks down below to ensure that the starts/ends have the right values? I couldn't easily locate them. But it is a big rule, so I may have missed it.

shubhambhokare1: Added

gramalingam (approved):

PR #2223, opened 2025-04-23
[IR] Handle invalid output deserialization

justinchuby:
Handle deserializing a graph if an output that is not produced by any nodes.

This is discovered when working on https://github.com/microsoft/onnxruntime-genai/pull/1416


titaiwangms (approved):
Is it a bug or it's a special need from gen AI?

justinchuby:
> Is it a bug or it's a special need from gen AI?

It is for the ability to work with (slightly) invalid graphs rather than crash. So not a bug or a special need but an improvement.

PR #2225, opened 2025-04-24
[IR] Support specifying output value in Tape

justinchuby:
When a user wants to specify names for output values, they can initialize the value first then supply them to tape op() call.

Also renamed symbolic_multi_output to symblic_multi_out to match https://pytorch.org/docs/main/onnx_ops.html#torch.onnx.ops.symbolic_multi_out


onnxscript/ir/_tape.py
```diff
@@ -85,17 +85,23 @@ def op(
         name: str | None = None,
         doc_string: str | None = None,
         metadata_props: dict[str, str] | None = None,
+        output: ir.Value | None = None,
     ) -> ir.Value:
         if attributes is None:
             attrs: Sequence[ir.Attr | ir.RefAttr] = ()
         else:
             attrs = _convenience.convert_attributes(attributes)
+        output_kwargs: dict[str, Any]
```
xadupre, line 94: op and op_multi_output are very similar. Maybe we could merge them into a single method.

gramalingam (approved):
Are there any examples/documentation illustrating the creation of output value and then using it to create an op?

PR #2230, opened 2025-04-25
Update type annotations for passes

justinchuby:
More clear on the documentation site.

### DIFF
--- a/onnxscript/ir/passes/_pass_infra.py
+++ b/onnxscript/ir/passes/_pass_infra.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import dataclasses
 import logging
-from typing import Sequence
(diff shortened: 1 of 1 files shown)


onnxscript/ir/passes/_pass_infra.py
```diff
@@ -180,23 +180,31 @@ class InPlacePass(PassBase):
     """A pass that modifies the input model in place and returns it."""
 
     @property
-    def in_place(self) -> bool:
+    @final
+    def in_place(self) -> Literal[True]:
```
titaiwangms, line 184: Maybe just turn these to be class attributes, so does in PassBase? We don't need them in runtime anyway.

titaiwangms (approved):

PR #2236, opened 2025-04-28
Update optimize_for_ort call to allow debug and shape_inference modes

shubhambhokare1:
- debug=True, can be called for all the ort-fusion rules
- apply_shape_inference=True, can be called, if we want to apply shape_inference after each fusion rule is applied

### DIFF
--- a/onnxscript/rewriter/_fusion_utils.py
+++ b/onnxscript/rewriter/_fusion_utils.py
@@ -38,7 +38,7 @@ def apply_fusion_rules(rules: pattern.RewriteRule | pattern.RewriteRuleSet) -> C
     ) -> int:
         count = rules.apply_to_model(model)
         if apply_shape_inference:
-            model = shape_inference.infer_shapes(model)
+            shape_inference.infer_shapes(model)
         if count == 0 and debug:
             tracer = pattern.MatchingTracer()
             rules.apply_to_model(model, tracer=tracer)
--- a/onnxscript/rewriter/_fusion_utils.py
+++ b/onnxscript/rewriter/_fusion_utils.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import Callable, Sequence, Union
 
 import onnxscript.ir as ir
+from onnxscript.ir.passes.common import shape_inference
 from onnxscript.rewriter import pattern
 
 Dim = Union[int, ir.SymbolicDim]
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -67,35 +71,42 @@ def fuse_xformers(model: ir.Model) -> tuple[ir.Model, dict[str, int]]:
...
     optimize(model)
     return model, fusion_count
 
 
 def optimize_for_ort(
-    model: ir.Model, config_name: str | None = None
+    model: ir.Model,
+    config_name: str | None = None,
+    debug: bool = False,
+    apply_shape_inference: bool = False,
```
gramalingam, line 109: Is it necessary to expose this to an external user? Or, can the implementation  do the right thing?

gramalingam: I mean the "apply_shape_inference" flag.

shubhambhokare1: I think keeping the flag internal would be better, although we might need to determine which fusions would be need/benefit from having the shape inference call after fusion. For example, SDPA -> shape_inf -> MHA would be a good candidate for this, given SDPA could create some nodes in the rewrite for the only_q_pre_scale pattern

onnxscript/rewriter/_fusion_utils.py
```diff
@@ -26,11 +27,18 @@ def _check_shape(bindings: dict[str, Dim], val: ir.Value, shape: Sequence[str])
...
+    apply_shape_inference: If True, apply shape inference after fusions.
     """
 
-    def apply_to(model: ir.Model, debug: bool = False) -> int:
+    def apply_to(
+        model: ir.Model, debug: bool = False, apply_shape_inference: bool = False
+    ) -> int:
         count = rules.apply_to_model(model)
+        if apply_shape_inference:
+            model = shape_inference.infer_shapes(model)
```
gramalingam, line 41: This is not going to work, since shape-inference does not update model in-place. The returned model is ignored here. If your tests are working, maybe you don't need this feature?

shubhambhokare1: Yep, I thought it was in-place given https://github.com/microsoft/onnxscript/blob/b63ba43ca2eb64d3ee7186117ececd810ac7adaf/onnxscript/ir/passes/common/shape_inference.py#L53

gramalingam: oh, interesting! Didn't realize that it merges it back to original model. In that case, I suggest dropping the assignment to model, and writing it as `shape_inference.infer_shapes(model)`.

gramalingam: This implies that `fuse_xformers` itself becomes an in-place fusion. Which is good for a performance perspective. But we should probably update the signature (back) to clarify this or at least document it.

gramalingam (approved):

gramalingam (approved):

PR #2238, opened 2025-04-28
[IR] Allow to copy an unfrozen version of the Shape

justinchuby:
When a shape is frozen, the dims of the shape cannot be modified. Users can call

```
new_shape = shape.copy()
new_shape[0] = 1
```

to assign to the new shape.

Added examples and the `frozen` property.


shubhambhokare1 (approved):

onnxscript/ir/_core.py
```diff
@@ -1090,10 +1140,6 @@ def __init__(
             )
         self._frozen: bool = frozen
 
-    def copy(self):
```
justinchuby, line 1093: moved

titaiwangms (approved):

PR #2242, opened 2025-04-29
Add support for a non-backtracking version of pattern disjunction

gramalingam:
Several fusions need to support multiple variants of a pattern (such as the optional presence of an Add or some such op). This PR adds support for a non-backtracking version of pattern disjunction. We can now use an "Or" between variants such as "Add(...)" and "MatMul(...)", for example.

Supporting unrestricted Or patterns is more complicated, since failure of one alternative will require backtracking, which will require unbinding any bindings added during the unsuccessful partial search. (We can consider that later, if it seems useful.)


onnxscript/rewriter/pattern_test.py
```diff
@@ -688,6 +688,39 @@ def test_model(x: FLOAT[1024], y: FLOAT[1024]) -> FLOAT[1024]:
...
+        def test_model1(x: FLOAT[16, 32], y: FLOAT[32, 16]) -> FLOAT[16, 16]:
+            return op.Relu(op.MatMul(x, y))
+
+        model_proto = test_model1.to_model_proto()
+        model = ir.serde.deserialize_model(model_proto)
+        rule.apply_to_model(model)
+        self.assertEqual([x.op_type for x in model.graph], ["WithoutBias"])
+
+        @script()
+        def test_model2(x: FLOAT[16, 32], y: FLOAT[32, 16], bias: FLOAT[16]) -> FLOAT[16, 16]:
```
shubhambhokare1, line 716: Does the output of the OrValue pattern always supposed to have the same shape/type, regardless of what path is chosen? (Is this part of what you mean by unrestricted or patterns)

onnxscript/rewriter/pattern_test.py
```diff
@@ -688,6 +688,39 @@ def test_model(x: FLOAT[1024], y: FLOAT[1024]) -> FLOAT[1024]:
...
         self.assertEqual([x.op_type for x in model.graph], ["Constant", "Identity"])
 
+    def test_or_pattern(self):
+        def source_pattern(op, x, y, bias):
+            t1 = op.MatMul(x, y)
+            t2 = op.Add(t1, bias)
+            t1_or_t2 = pattern.OrValue([t1, t2], tag_var="has_bias", tag_values=[False, True])
+            return op.Relu(t1_or_t2)
+
+        def replacement(op, x, y, bias, has_bias):
```
shubhambhokare1, line 698: Might be nice to have an example showing how the check function interacts with patterns with OrValue's

shubhambhokare1 (commented):
Thanks! This will be super useful in avoiding bloating the patterns with parameters. Just mentioning a couple clarifying questions.

onnxscript/rewriter/pattern.py
```diff
@@ -467,6 +467,9 @@ def __str__(self) -> str:
         return self._name if self._name is not None else "anonymous:" + str(id(self))
 
 
+OpIdentifier = Tuple[str, str, str]
```
justinchuby, line 470: You may use ir.OperatorIdentifier

onnxscript/rewriter/pattern_test.py
```diff
@@ -688,6 +688,39 @@ def test_model(x: FLOAT[1024], y: FLOAT[1024]) -> FLOAT[1024]:
         self.assertEqual(len(model.graph), 2)
         self.assertEqual([x.op_type for x in model.graph], ["Constant", "Identity"])
 
+    def test_or_pattern(self):
+        def source_pattern(op, x, y, bias):
+            t1 = op.MatMul(x, y)
+            t2 = op.Add(t1, bias)
+            t1_or_t2 = pattern.OrValue([t1, t2], tag_var="has_bias", tag_values=[False, True])
```
justinchuby, line 695: Would the name `EitherValue` or `Either` be more descriptive? Never mind me if "Or" is more standard in PL

onnxscript/rewriter/pattern_test.py
```diff
@@ -688,6 +688,39 @@ def test_model(x: FLOAT[1024], y: FLOAT[1024]) -> FLOAT[1024]:
...
+                return op.WithoutBias(x, y)
+
+        rule = pattern.RewriteRule(source_pattern, replacement)
+
+        @script()
+        def test_model1(x: FLOAT[16, 32], y: FLOAT[32, 16]) -> FLOAT[16, 16]:
+            return op.Relu(op.MatMul(x, y))
+
+        model_proto = test_model1.to_model_proto()
+        model = ir.serde.deserialize_model(model_proto)
```
justinchuby, line 711: We can create a util function in this test class to avoid having to call serde everytime.

onnxscript/rewriter/pattern_test.py
gramalingam: No, to Shubham's question. Shape/type restrictions are al the user's responsibility. The restrictions are spelled out in the Constructor. We can do an "Or" of `Add(...)` and `Mul(...)` ... because we can distinguish between `Add` and `Mul` and choose the correct alternative. But we cannot do an OR of `Add(x, y)` and `Add(y, x)`, for example: because looking at the `Add` we cannot determine which of the two alternatives to continue the match with. So, simple summary is: the alternatives must all correspond to distinct ops.

onnxscript/rewriter/pattern_test.py
gramalingam: Yes, onnx functions should have a method to return the "ir model" ... in fact, the current version is logically a double-conversion (since script-function => IR => proto should be the logical route, eventually, to create proto).

onnxscript/rewriter/pattern_test.py
gramalingam: I think relaxing the constraints may be useful. I will try that in a separate PR. It will need refactoring the state to support backtracking (to "undo" changes done in a failed alternative).

onnxscript/rewriter/pattern.py
gramalingam: Done

justinchuby (approved):
Could you add an entry to https://github.com/microsoft/onnxscript/blob/main/docs/api/rewriter_pattern.md ? Thanks

shubhambhokare1 (approved):
Thanks, will create a follow up PR converting some of the patterns in ort-fusion to utilize OrPatterns

shubhambhokare1 (approved):

Issue #2235, opened 2025-04-28
If we not removed initialized input, should we keep initializer also?


leshabirukov:
Me still haunted by https://github.com/microsoft/onnxscript/issues/2211  
We allow hanging inputs, because it is part of interface, but we are removing it's initializer if exist. May be we should keep it, because it is affects interface either, - now user can ignore it before DCE pass, but can not ignore after.

justinchuby:
I am not following this. Could you share a concrete example?

leshabirukov:
In the code below first run is successful, while second run gives

ValueError: Required inputs (['two']) are missing from input feed (['x']).

after remove_unused_nodes

```
import onnx
from onnxscript.optimizer import remove_unused_nodes
import onnxruntime as ort
import numpy as np


def util_random( shape ):
    return np.round(np.random.random( shape )*10).astype(np.float32)

def test_unused_initialized_inputs_are_kept_by_default():
    model = onnx.parser.parse_model(
        """
        <ir_version: 10, opset_import: [ "" : 17]>
        agraph (float[2] x, float[2] two) => (float[2] z)
        <float[2] two = {2.0,2.0}> {
            four = Add(two, two)
            z = Mul(x, x)
        }
    """
    )
    inputTensor = util_random([2], )
    print(inputTensor)
    session_1 = ort.InferenceSession(model.SerializeToString())
    session_1.run(None, {"x": inputTensor})

    remove_unused_nodes(model)

    session_2 = ort.InferenceSession(model.SerializeToString())
    session_2.run(None, {"x": inputTensor})


test_unused_initialized_inputs_are_kept_by_default()

```

justinchuby:
So I guess we need to keep the initializer values as well, right?

leshabirukov:
Should it be ` warnings.warn` or `logger.warning`?

justinchuby:
`logger.warning` is preferable

leshabirukov:
Done.  
Now I'm annoyed by this marginal functionality all over `RemoveUnusedNodesPass::call`. Move it to a separate function?

justinchuby:
Sounds reasonable. Maybe create a pass that is "remove unused inputs"?

leshabirukov:
Err, I think rather about just intrnal function in `RemoveUnusedNodesPass`, anyway, warning belongs there.

leshabirukov:
`count += 1`
Should we count initialized input as 1 or 2?

justinchuby:
[...]

PR #2245, opened 2025-04-29
[pass] Fix DCE to keep initializers that are inputs

leshabirukov:
Fix https://github.com/microsoft/onnxscript/issues/2235


onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -108,14 +108,14 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
+        graph_inputs = model.graph.inputs
         if self.remove_initialized_inputs:
-            graph_inputs = model.graph.inputs
             for i, inp in reversed(list(enumerate(graph_inputs))):
                 if inp.name in initializers and not (inp in graph_outputs or inp.uses()):
                     del graph_inputs[i]
                     count += 1
         for init in list(initializers.values()):
-            if not (init in graph_outputs or init.uses()):
+            if not (init in graph_outputs or init in graph_inputs or init.uses()):
```
justinchuby, line 118:
```suggestion
            if not (init.uses() or init in graph_outputs or init in graph_inputs):
```

We can reorder the expression to make it more efficient

justinchuby (approved):
Thanks!

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -104,18 +104,32 @@ def __init__(self, remove_initialized_inputs: bool = False):
...
         graph_outputs = frozenset(model.graph.outputs)
         initializers = model.graph.initializers
-        if self.remove_initialized_inputs:
-            graph_inputs = model.graph.inputs
-            for i, inp in reversed(list(enumerate(graph_inputs))):
-                if inp.name in initializers and not (inp in graph_outputs or inp.uses()):
-                    del graph_inputs[i]
-                    count += 1
+        graph_inputs = model.graph.inputs
+        self._remove_unused_initialized_inputs(model)
```
justinchuby, line 130: check `self.remove_initialized_inputs` before callig the function?

leshabirukov: No, if `not remove_initialized_inputs` it still warns about funny things in graph.

justinchuby: Got it, then two suggestions:

1. rename `_remove_unused_initialized_inputs` -> `_maybe_remove_unused_initialized_inputs` because it doesn't always remove, depending on the option
2. Lift it into a module function (instead of a class method), because the only state used in the function is the `self.remove_initialized_inputs` option, which can be passed in as function input for clarity.

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -104,18 +104,32 @@ def __init__(self, remove_initialized_inputs: bool = False):
...
 
+    def _remove_unused_initialized_inputs(self, model: ir.Model) -> None:
+        graph_outputs = model.graph.outputs
+        initializers = model.graph.initializers
+        graph_inputs = model.graph.inputs
+        unused_init_inputs = []
+        for i, inp in reversed(list(enumerate(graph_inputs))):
+            if inp.name in initializers and not (inp.uses() or inp in graph_outputs):
+                if self.remove_initialized_inputs:
+                    del graph_inputs[i]
```
gramalingam, line 115: Why would a user want to remove a graph-input? It is part of its signature. Even if the input is never used. It could potentially fail in a system when the user specifies an input-value, and the system rejects it as an invalid input (eg., could happen with onnxruntime). Against this disadvantage, there is nothing much gained by removing the model input.

The initializer value can, however, be removed, if there is no use of that value within the model. May be that will save some memory for large initializers.

justinchuby: I think the scenario is that when an initializer happens to be declared an input (as it can be in some converters), if only the initializer is removed but not the input, the same calling will fail. A user may decide to not care about those inputs since for example in ort inputs are provided by keyword.

onnxscript/ir/passes/common/unused_removal.py
leshabirukov: 1. Done
2. The question is, do we allow call it outside the `RemoveUnusedNodesPass`, if user introduce fake inputs for probing (my understanding of net from initial issue), he probably want to clean off whose at some point, but it may be bad practice, that not to be encoraged.

justinchuby: We don't need to decide if we allow calling it outside. By prepending the name with `_`, we should that it is intended to be private. The suggestion is to move `_maybe_remove_unused_initialized_inputs` out from the class.

`onnxruntime::remove_initializer_from_input` does something different. They removes the inputs, but keep the initializers; so the model is still functionally the same. This is something we can implemement as a pass as well.

leshabirukov: Well, fine for me

onnxscript/ir/passes/common/unused_removal.py
```diff
@@ -92,6 +92,25 @@ def _remove_unused_nodes_in_graph_like(function_or_graph: ir.Function | ir.Graph
...
+        if inp.name in initializers and not (inp.uses() or inp in graph_outputs):
+            if do_remove:
+                del graph_inputs[i]
+            else:
+                unused_init_inputs.append(inp.name)
+    if unused_init_inputs:
+        logger.warning(
+            "RemoveUnusedNodesPass: Found unused initialized inputs %s,"
+            " consider turning `remove_initialized_inputs` on",
+            unused_init_inputs,
```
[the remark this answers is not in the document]
justinchuby: Never mind

justinchuby (approved):

justinchuby (changes requested):
@leshabirukov Apologize but after discussion with @gramalingam , we think that instead of creating an option, it is better to separately handle the initialized inputs. I am working on this in https://github.com/microsoft/onnxscript/pull/2253. If you may, could you retain the line `if not (init.uses() or init in graph_outputs or init in graph_inputs):`, and remove the rest of the changes in the PR? I will update this other PR to provide the needed functionality

justinchuby:
@leshabirukov I have created https://github.com/microsoft/onnxscript/pull/2257. Unfortunately we will have to revert some of the changes you made in favor of https://github.com/microsoft/onnxscript/pull/2253. Will the proposed passes work for you?

leshabirukov:
> @leshabirukov I have created #2257. Unfortunately we will have to revert some of the changes you made in favor of #2253. Will the proposed passes work for you?

No problem, once I was aware of dangling initialized input issue, I new what to do with that. But think about people hit the same issue, I insist, there should be the warning at least. I propose to preserve this function, (in another file if it should):

```
def _maybe_remove_unused_initialized_inputs(model: ir.Model, do_remove: bool) -> None:
    graph_outputs = model.graph.outputs
    initializers = model.graph.initializers
    graph_inputs = model.graph.inputs
    unused_init_inputs = []
    for i, inp in reversed(list(enumerate(graph_inputs))):
        if inp.name in initializers and not (inp.uses() or inp in graph_outputs):
            if do_remove:
                del graph_inputs[i]
            else:
                unused_init_inputs.append(inp.name)
    if unused_init_inputs:
        logger.warning(
            "RemoveUnusedNodesPass: Found unused initialized inputs %s,"
            " consider turning `remove_initialized_inputs` on",
            unused_init_inputs,
```

And call it from DCE with `do_remove=False`

justinchuby:
Thanks. I think that’s reasonable

leshabirukov:
ok, then two questions:
1. api: How user (me) would invoke `_maybe_remove_unused_initialized_inputs` and in which file it should be.
2. technical: I'm closing this PR, right?

leshabirukov:
ok, don't forget about warning for DCE

Issue #2073, opened 2025-02-22
[IR] Display small values for readability


justinchuby:
Display constant values with <10 elements.

PR #2248, opened 2025-04-30
[IR] Display constant tensors for Value

justinchuby:
Display constant values and simplify the value repr string when fields are empty.

```py
>>> from onnxscript import ir
>>> v = ir.Value(name="v1", const_value=ir.tensor(1))
>>> v
Value(name='v1', const_value=Tensor<INT64,[]>(array(1), name=None))
>>> v = ir.Value(name="v1", const_value=ir.tensor([[1]]))
>>> v
Value(name='v1', const_value=Tensor<INT64,[1,1]>(array([[1]]), name=None))
>>> print(v)
%"v1"<?,?>{Tensor<INT64,[1,1]>(array([[1]]), name=None)}
```

Fix https://github.com/microsoft/onnxscript/issues/2073

### DIFF
--- a/onnxscript/ir/_convenience/__init__.py
+++ b/onnxscript/ir/_convenience/__init__.py
@@ -212,7 +212,7 @@ def convert_attributes(
         ...     "type_protos": [ir.TensorType(ir.DataType.FLOAT), ir.TensorType(ir.DataType.FLOAT)],
         ... }
         >>> convert_attributes(attrs)
-        [Attr('int', INT, 1), Attr('float', FLOAT, 1.0), Attr('str', STRING, 'hello'), Attr('ints', INTS, [1, 2, 3]), Attr('floats', FLOATS, [1.0, 2.0, 3.0]), Attr('strings', STRINGS, ['hello', 'world']), Attr('tensor', TENSOR, Tensor<DOUBLE,[3]>(array([1., 2., 3.]), name=None)), TensorProtoTensor<FLOAT,[3]>(array([1., 2., 3.], dtype=float32), name='proto')), Attr('graph', INTS, Graph(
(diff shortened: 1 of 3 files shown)


gramalingam:
Does this show up in the displayed form of a Node? Eg., if we have a node `Y = Add(X, c)` where `c` is a constant 1, for example? Would the displayed representation inside a node for an (input) value be too dense (with type/shape etc.)? If so, can we have a more compact representation for an input-value representation inside a node? May be just the name "c" and the value "1"?

justinchuby:
In a node, it won't show. In a graph it will show in the initializers section

```
graph(
    name=anonymous_graph:124652572372048,
    inputs=(

    ),
    outputs=(

    ),
    initializers=(
        %"init"<?,?>{Tensor<INT64,[5]>(array([1, 1, 1, 1, 1]), name=None)}
    ),
) {
    0 |  # node_Add_0
         %"val_0"<?,?> ⬅️ ::Add(%"init", %"init")
    return 
}
```

Is that ok? Or maybe we want to see it directly in the node?

justinchuby:
I updated for it to show

```
%"anonymous:132887616237136"<?,?> ⬅️ ::Add(%"init"{Tensor<INT64,[5]>(array([1, 1, 1, 1, 1]), name=None)}, %"init"{Tensor<INT64,[5]>(array([1, 1, 1, 1, 1]), name=None)})
```

titaiwangms (approved):
What do we have when it's larger than 10 elements?

justinchuby:
It will show `{...}`

gramalingam:
> I updated for it to show
> 
> ```
> %"anonymous:132887616237136"<?,?> ⬅️ ::Add(%"init"{Tensor<INT64,[5]>(array([1, 1, 1, 1, 1]), name=None)}, %"init"{Tensor<INT64,[5]>(array([1, 1, 1, 1, 1]), name=None)})
> ```

Great! (My original request, at least, was for the ability to see it in the node.) What would be even better (IMO) is if this can be made "less busy" ... in typical usage, just the values will suffice (you want to see "0" or "1" or "0.36712") ... given that the full information is available elsewhere, if a user really wants to know, a shorter version should work better here.

I do think there is a difference between seeing `%"Y" <- ::Add (%"X", %"init"{1})` and `%"Y" <- ::Add (%"X", %"init"{Tensor<INT64,[1]>(array([1]), name=None)})` ... in favor of the first.

PR #2249, opened 2025-04-30
[IR] Introduce short name for dtypes

justinchuby:
Introduce short name for dtypes as a more compact way of describing the data types in strings. Users can already access the enums by name with e.g. `ir.DataType["DOUBLE"]`.


gramalingam (approved):
Where do we use the short-names (when generating some output, for eg)? Where can we use the short-names (when parsing which input)? Just wondering

justinchuby:
I am still debating. Initially I was thinking model builder constant value names but I realized the long names may be more preferable there

PR #2254, opened 2025-04-30
Add initializer constructors

justinchuby:
---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/onnxscript/pull/2254?shareId=2cb9008b-1127-42ed-b87e-318cbb835b2e).


justinchuby:
Update: not a good idea until we record graph ownership for values. When a ir.initializer() is passed into tape it needs to know which graph to add to or if it is owned by a graph already.

Issue #2158, opened 2025-04-02
[rewrite] Transpose initializer -> initializer (transposed)


titaiwangms:
Essentially, we are upstreaming https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_constant_fold.py
https://github.com/microsoft/onnxruntime/blob/838b97e73289cd11caf969f9f5c01ce153d6069f/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L182
If initializer is not consumed by other inputs, we can transpose the initializer in advance.

titaiwangms:
Related https://github.com/microsoft/onnxscript/pull/2025/files

PR #2255, opened 2025-04-30
[rewriter] Transpose initializer rule

justinchuby:
Create a rule to absorb a Transpose node into the initializer (fold constant). This rule is in place to handle initializers of any size. So users do not need to change the size limit of the optimizer when constant folding.

TODO
- tests

Fix https://github.com/microsoft/onnxscript/issues/2158


justinchuby:
I think this is less efficient than an IR pass. Because it will match all transpose nodes, but we only need to look at users of the initializers. Thoughts? @titaiwangms @shubhambhokare1 @gramalingam

justinchuby:
But this is op-specific. I don't think it fits in the common passes. So I prefer that it is a rewrite rule.

onnxscript/rewriter/transpose_initializer.py
```diff
@@ -0,0 +1,59 @@
...
+        return op.initializer(ir.tensor(transposed, name=new_name))
+
+    def check(self, context, initializer: ir.Value) -> orp.MatchResult:
+        del context  # Unused
+        check_result = orp.MatchResult()
+        if initializer.const_value is None:
+            return check_result.fail("Value is not an initializer, const_value is None")
+        if initializer.producer() is not None:
+            return check_result.fail("Value is not an initializer, producer is not None")
+        if initializer.uses() != 1:
```
shubhambhokare1, line 53:
```suggestion
        if len(initializer.uses()) != 1:
```

justinchuby: Good catch thanks

titaiwangms:
In the original issue, there are also transpose -> transpose and transpose -> MatMul/Gemm pattern matching. Should we have those in this rule as well? If we are going to go through Transpose nods anyway.

justinchuby:
I think we should have different rules since they will match different nodes

justinchuby:
@shubhambhokare1 I see it's already in https://github.com/microsoft/onnxscript/blob/605e06ec11c75049398178b227f439a309b6ae96/onnxscript/rewriter/llama_rule_sets.py#L318, was it not applied?

shubhambhokare1:
> @shubhambhokare1 I see it's already in
> 
> https://github.com/microsoft/onnxscript/blob/605e06ec11c75049398178b227f439a309b6ae96/onnxscript/rewriter/llama_rule_sets.py#L318
> 
> , was it not applied?

Ah I see, however the one in https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_constant_fold.py seems to be doing, 
inp -> Transpose -> Transpose -> out
to 
inp -> out if transpose perm values are equal

Whereas the current rewrite rule only replace two transposes with an equivalent single transpose

justinchuby:
> > @shubhambhokare1 I see it's already in
> > https://github.com/microsoft/onnxscript/blob/605e06ec11c75049398178b227f439a309b6ae96/onnxscript/rewriter/llama_rule_sets.py#L318
> > 
> > , was it not applied?
> 
> Ah I see, however the one in [microsoft/onnxruntime@`main`/onnxruntime/python/tools/transformers/fusion_constant_fold.py](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_constant_fold.py) seems to be doing, inp -> Transpose -> Transpose -> out to inp -> out if transpose perm values are equal
> 
> Whereas the current rewrite rule only replace two transposes with an equivalent single transpose

Sounds good. I will update

justinchuby:
@shubhambhokare1 does this one https://github.com/microsoft/onnxscript/blob/ac87a1cf0c0be4684c6f5d9d8ae705026cb7b675/onnxscript/rewriter/llama_rule_sets.py#L216 or this one https://github.com/microsoft/onnxscript/blob/ac87a1cf0c0be4684c6f5d9d8ae705026cb7b675/onnxscript/rewriter/llama_rule_sets.py#L257 do it?

onnxscript/rewriter/transpose_initializer.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self):
+        super().__init__("TransposeInitializer", remove_nodes=True)
+
+    def pattern(self, op, initializer):
+        return op.Transpose(initializer, _allow_other_attributes=True)
+
+    def rewrite(self, op, initializer: ir.Value) -> ir.Value:
+        original_transpose = initializer.consumers()[0]
+        perm_attr = original_transpose.attributes.get("perm")
+        assert isinstance(perm_attr, ir.Attr)
```
xadupre, line 30: What else could it be?

onnxscript/rewriter/transpose_initializer.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if perm_attr is not None:
+            perm = perm_attr.as_ints()
+        else:
+            perm = None
```
xadupre, line 35: Can we eliminate that case in def check?

justinchuby: When perm is None the transpose can still be evaluated. So I don't think it can be eliminated in check()?

justinchuby:
Replaced by https://github.com/microsoft/onnxscript/pull/2355

PR #2253, opened 2025-04-30
Add add/remove initializers passes

justinchuby:
Implement `RemoveInitializersFromInputsPass` and `AddInitializersToInputsPass` in `onnxscript/ir/passes/common/constant_manipulation.py`.

* **RemoveInitializersFromInputsPass**
  - Add `RemoveInitializersFromInputsPass` class to find and remove graph inputs with `const_value`.
  - Implement `call` method to remove inputs with `const_value` from `graph.inputs`.
  - Register `RemoveInitializersFromInputsPass` in the `__all__` list.

* **AddInitializersToInputsPass**
  - Add `AddInitializersToInputsPass` class to find and add initializers to the graph inputs.
  - Implement `call` method to add all initializers to the graph inputs if not already present.
  - Register `AddInitializersToInputsPass` in the `__all__` list.

* **Tests**
  - Add test cases for `RemoveInitializersFromInputsPass` in `onnxscript/ir/passes/common/constant_manipulation_test.py` to verify removal of inputs with `const_value`.
  - Add test cases for `AddInitializersToInputsPass` in `onnxscript/ir/passes/common/constant_manipulation_test.py` to verify addition of initializers to the graph inputs.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/microsoft/onnxscript/pull/2253?shareId=321de9a6-ee5d-4a2a-a84d-27ef7a4c6d6f).

PR #2257, opened 2025-04-30
[pass] Update DCE passes

justinchuby:
- Remove the `remove_initialized_inputs` option in dce because the contract of the pass it that it does not modify model signature. Fixed bugs where initializers are removed. Instead, users can use https://github.com/microsoft/onnxscript/pull/2253 to remove the initialized inputs first.
- Additionally updated RemoveUnusedOpsetsPass to always retain the default opset.

### DIFF
--- a/onnxscript/ir/passes/common/unused_removal_test.py
+++ b/onnxscript/ir/passes/common/unused_removal_test.py
@@ -13,15 +13,13 @@ from onnxscript import ir
 class RemoveUnusedTest(unittest.TestCase):
     using_ir: bool
 
-    def remove_unused_nodes(
-        self, model: onnx.ModelProto, remove_initialized_inputs: bool = False
-    ):
+    def remove_unused_nodes(self, model: onnx.ModelProto):
         if self.using_ir:
             model_ir = ir.serde.deserialize_model(model)
-            onnxscript.optimizer.remove_unused_nodes(model_ir, remove_initialized_inputs)
+            onnxscript.optimizer.remove_unused_nodes(model_ir)
             model = ir.serde.serialize_model(model_ir)
             return model
-        onnxscript.optimizer.remove_unused_nodes(model, remove_initialized_inputs)
+        onnxscript.optimizer.remove_unused_nodes(model)
         return model
 
     def test_remove_unused_nodes(self):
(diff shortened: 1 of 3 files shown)


shubhambhokare1 (commented):
Might need to update tests such as
https://github.com/microsoft/onnxscript/blob/b63ba43ca2eb64d3ee7186117ececd810ac7adaf/onnxscript/ir/passes/common/unused_removal_test.py#L105-L107
and/or the function called in these tests
https://github.com/microsoft/onnxscript/blob/b63ba43ca2eb64d3ee7186117ececd810ac7adaf/onnxscript/optimizer/__init__.py#L115-L127
to reflect the removal of the attribute

justinchuby:
@shubhambhokare1 thanks! Done.

onnxscript/ir/passes/common/unused_removal_test.py
```diff
@@ -13,15 +13,13 @@
...
             model_ir = ir.serde.deserialize_model(model)
-            onnxscript.optimizer.remove_unused_nodes(model_ir, remove_initialized_inputs)
+            onnxscript.optimizer.remove_unused_nodes(model_ir)
             model = ir.serde.serialize_model(model_ir)
             return model
-        onnxscript.optimizer.remove_unused_nodes(model, remove_initialized_inputs)
+        onnxscript.optimizer.remove_unused_nodes(model)
         return model
 
     def test_remove_unused_nodes(self):
```
xadupre, line 25: There should a test with a model including subgraphs.

titaiwangms (commented):
From the discussion with Kunal on phi4_mm/builder.py, It seems from the user side, they don't really benefit from delicately/cautiously split passes. The end of the day, they just call the higher level API. So that means these delicately/cautiously split passes only serve as debugging usage for ourselves. From debugging point of view, would it be too trivial to separate unused initializers from DCE?

justinchuby:
I don't think the passes are designed for debugging purposes only. There can be users who want a high level api which we can provide, but we will also support usages that are more fine grained & users can design their own transformation pipelines. Each pass should be modularized: it should have a clear contract and a self contained behavior.

titaiwangms (approved):

onnxscript/ir/passes/common/unused_removal_test.py
justinchuby: Thanks - will update

justinchuby: https://github.com/microsoft/onnxscript/issues/2264

PR #2259, opened 2025-05-01
Rewrite Skip fusions with check functions

shubhambhokare1:
- Rewrite SkipLayerNorm fusions and SkipRMSNorm fusions to match format of other ort-fusion patterns.
- Added check functions for ensuring shapes are as expected.
- Moving these fusions out of PR #2221 

Fusion support patterns with:
- `Add(input, skip) -> Norm`
- `Add(input, skip) -> Add (result, bias) -> Norm`
- `Add(input, bias) -> Add (result, skip) -> Norm`

NOTE:
These fusions should support:
- Planned whisper-related optimizations
- Benchmark failures stemming from wrong bias shapes for SkipLayerNorm fusions

### DIFF
--- a/onnxscript/rewriter/ort_fusions/skip_normalization.py
+++ b/onnxscript/rewriter/ort_fusions/skip_normalization.py
@@ -13,14 +13,17 @@ Dim = Union[int, ir.SymbolicDim]
 
 
 class SkipRmsNormFusion(pattern.RewriteRuleClassBase):
-    def __init__(self, name: str, has_bias: bool = False):
+    def __init__(self, name: str, has_bias: bool = False, bias_pre_add: bool = False):
         """Fusion rule for SkipRMSNormalization."""
         super().__init__(name=name)
         self._has_bias = has_bias
+        self._bias_pre_add = bias_pre_add
 
     def pattern(self, op, input, skip, gamma, bias, epsilon, stash_type):
+        if self._has_bias and self._bias_pre_add:
+            input = op.Add(input, bias)
         skip_sum = op.Add(input, skip)
-        if self._has_bias:
+        if self._has_bias and not self._bias_pre_add:
             skip_sum = op.Add(skip_sum, bias)
         normalized = op.SimplifiedLayerNormalization(
             skip_sum,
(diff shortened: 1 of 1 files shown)


justinchuby (approved):

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -2,115 +2,195 @@
...
+
+    def check(self, op, input, skip, gamma, bias, epsilon, stash_type) -> pattern.MatchResult:  # type: ignore[name-defined]
+        """Check if the pattern matches conditions for use of SkipSimplifiedLayerNormalization op."""
+        check_result = pattern.MatchResult()
+        bindings: dict[str, Dim] = {}
+
+        def no_match(val: ir.Value, dims: Sequence[str]) -> bool:
+            return not _fusion_utils._check_shape(bindings, val, dims)
+
+        if no_match(input, ["B", "S", "D"]):
```
justinchuby, line 45: For my understanding, how does no match work?

justinchuby: What happens when the shapes do not take these names?

gramalingam: The symbolic dim names like "B" are local to this check function ... it is used to ensure equivalence of different dims that are expected to be equivalent. "B" will be bound to the actual dimension, whether it is 1 or "BatchSize" or whatever.

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -2,115 +2,195 @@
...
+        self._has_bias = has_bias
+        self._bias_pre_add = bias_pre_add
+
+    def pattern(self, op, input, skip, gamma, bias, epsilon, stash_type):
+        if self._has_bias and self._bias_pre_add:
+            input = op.Add(input, bias)
+        skip_sum = op.Add(input, skip)
+        if self._has_bias and not self._bias_pre_add:
+            skip_sum = op.Add(skip_sum, bias)
+        normalized = op.SimplifiedLayerNormalization(
```
gramalingam, line 28: Isn't this from com.microsoft domain? If so, we should specify that

shubhambhokare1: Ported over a similar note to what we have in RmsNormFusion

justinchuby (approved):

PR #2271, opened 2025-05-05
Add New Metadata and Pattern Features

jsmonson:
This PR will add new metadata and pattern features.


onnxscript/utils/graph_view_utils.py
```diff
@@ -0,0 +1,263 @@
+from onnxscript import ir
+
+import ast
+
+
+def is_initializer(value):
+    return input.producer() == None
```
justinchuby, line 7: When a value is a graph input, its producer is also None. I will add a method to make this check more robust

onnxscript/utils/graph_view_utils.py
```diff
@@ -0,0 +1,263 @@
...
+    usage = set()
+    for use in value.uses():
+        user_node = use[0]
+        if user_node in nodes:
+            usage.add("INTERNAL")
+        else:
+            usage.add("EXTERNAL")
+    return usage
+
+def find_subgraph_inputs(nodes):
```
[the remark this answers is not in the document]
justinchuby: I see. It's for constructing the graph view

onnxscript/rewriter/pattern_builder_jsm.py
```diff
@@ -0,0 +1,546 @@
...
+    node._outputs = (*node._outputs, output)
+    node._num_outputs = len(node._outputs)
+
+def prepend_output_to_node(node, output):
+    output._producer = node
+    output._index = 0
+    for outp in node._outputs:
+        outp._index += 1
+    node._outputs = (output, *node._outputs)
+    node._num_outputs = len(node._outputs)
```
justinchuby, line 216: We should add this to node so there is no need to modify internal states

onnxscript/rewriter/pattern_builder_jsm.py
```diff
@@ -0,0 +1,546 @@
...
+                            num_outputs =1)
+            LoopBody.function.append(Ident)
+
+            #Add Output to Function Call Nodes
+            for i,node in enumerate(nodes):
+                output_copy = copy.copy(noutput)
+
+                #preserve single_assignment
+                output_copy.name += f'_{i}'
+                append_output_to_node(node,output_copy)
```
justinchuby, line 333: Is there a way to collect all outputs before constructing the node?

onnxscript/rewriter/pattern_builder_jsm.py
```diff
@@ -0,0 +1,546 @@
...
+def direct_convert_ir_graph_to_pattern(graph):
+
+
+    # Transform IR values to ValuePatterns
+    vmap = {}
+    for input in graph.inputs:
+        vmap[input] = ValuePattern(input.name)
+
+    for init in graph.initializers:
+        vmap[init] = ValuePattern(init.name)
```
gramalingam, line 31: Looks like you want to map constants and initializers to unconstrained variables in the pattern? I wonder if it would make sense to map them to "Constants" in the pattern that require a matching contstant-value in the graph for a successful match? That makes reasonable, at least for simple and small constants. If it should be abstracted, wouldn't it be better for the user themselves to do that explicitly by mapping them to graph inputs?

onnxscript/rewriter/pattern_builder_jsm.py
```diff
@@ -0,0 +1,546 @@
...
+    builder = OpsetPatternBuilder("", record=True)
+
+    with pattern_builder(builder):
+        for node in graph._nodes:
+            ninputs = []
+            for ninput in node.inputs:
+                ninputs.append(vmap[ninput])
+
+            #if len(node.outputs) > 1:
+            vp_outputs = builder.__getattr__(node.op_type)(*ninputs,_domain=node.domain, _outputs=len(node.outputs))
```
gramalingam, line 47: So, the attributes of the node are abstracted away, and not matched?

onnxscript/rewriter/pattern_builder_jsm.py
jsmonson: Yes. Thanks for this suggestion. I think we should do this. This would provide a good level of control for the user.

onnxscript/rewriter/pattern_builder_jsm.py
jsmonson: Good point. We ought to match the attributes. I'll make these changes.

onnxscript/utils/graph_view_utils.py
justinchuby: https://github.com/microsoft/onnxscript/pull/2282

onnxscript/rewriter/pattern_builder_jsm.py
justinchuby: In fact we recommend creating new nodes if you want to update inputs and outputs. This way invariance of the graph is maintained. You may consider accumulating the inputs and outputs into two lists before constructing the node.

onnxscript/utils/test_PytorchHierarchyNode.py
```diff
@@ -0,0 +1,190 @@
+import pytest
+import ast
+
+import onnx
+from onnxscript import script
```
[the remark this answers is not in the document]
jsmonson: Removed unused import

onnxscript/rewriter/pattern_builder_jsm.py
```diff
@@ -0,0 +1,546 @@
...
+
+def prepend_output_to_node(node, output):
+    output._producer = node
+    output._index = 0
+    for outp in node._outputs:
+        outp._index += 1
+    node._outputs = (output, *node._outputs)
+    node._num_outputs = len(node._outputs)
+
+def prepend_input_to_node(node, input):
```
justinchuby, line 218: Modifying private fields is not recommended. I would consider initializing a new node

tests/loop_rolling/test_loop_rolling.py
```diff
@@ -0,0 +1,198 @@
...
+    sess_options.log_severity_level = 0
+    sess_options.graph_optimization_level = onnxrt.GraphOptimizationLevel.ORT_ENABLE_BASIC
+
+    return onnxrt.InferenceSession(filename, sess_options=sess_options).run([output_name], input_dict)
+
+input_dict, outputs = ort_make_rand_io(args.filename)
+golden_results      = ort_run_graph(args.filename, input_dict, outputs[0].name)
+
+
+LoopBody = LoopBodyTemplate(args.patternfilename)
```
justinchuby, line 70:
```suggestion
loop_body = LoopBodyTemplate(args.patternfilename)
```

note: always use snake_case for variable names

onnxscript/utils/graph_view_utils.py
```diff
@@ -0,0 +1,263 @@
...
+            usage = classify_usage(noutput, nodes)
+            if has_external_usage(usage):
+                if has_internal_usage(usage):
+                    used_output.add(noutput)
+                else:
+                    output.add(noutput)
+    return [output, used_output]
+
+
+def bGraphView(name, nodes):
```
justinchuby, line 71: nit: snake case for function names. Is this `build_graph_view`?

onnxscript/utils/graph_view_utils.py
justinchuby, line 41: This function is useful. I wonder if we should put it in ir.convenience https://github.com/onnx/ir-py/blob/main/src/onnx_ir/_convenience/__init__.py

justinchuby: As well as the one below for outputs and bGraphView

PR #2272, opened 2025-05-05
Guard ort-fusions against onnxruntime version

shubhambhokare1:
- Guard ort fusion against ort version
- Optimize fusion calls for attention fusions

### DIFF
--- a/onnxscript/rewriter/_fusion_utils.py
+++ b/onnxscript/rewriter/_fusion_utils.py
@@ -4,9 +4,12 @@ from __future__ import annotations
 
 from typing import Callable, Sequence, Union
 
+import packaging.version
+
 import onnxscript.ir as ir
 from onnxscript.ir.passes.common import shape_inference
 from onnxscript.rewriter import pattern
+from onnxscript.rewriter.ort_fusions._test_utils import ORT_VERSION
 
 Dim = Union[int, ir.SymbolicDim]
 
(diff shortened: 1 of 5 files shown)


onnxscript/rewriter/_fusion_utils.py
```diff
@@ -31,18 +34,25 @@ def apply_fusion_rules(rules: pattern.RewriteRule | pattern.RewriteRuleSet) -> C
...
     ) -> int:
-        count = rules.apply_to_model(model)
-        if apply_shape_inference:
-            shape_inference.infer_shapes(model)
-        if count == 0 and debug:
-            tracer = pattern.MatchingTracer()
-            rules.apply_to_model(model, tracer=tracer)
-            tracer.report()
+        count = 0
+        if packaging.version.Version(ort_version) <= ORT_VERSION:
```
gramalingam, line 48: May be worth rethinking this. Let us not confuse the installed version of ORT where the optimizer is running with the intended target version. (Eg., I could optimize the model, and send the optimized model to some other user.) ONNX opset-versioning is good, in this regard, since it captures the intended opset version explicitly in the model. Unfortunately, we have no way of capturing the intended ORT version in the model. So, only the caller of the main fusion method can specify what the intended target version is.

This is the reason I felt handling this may be a lower priority. But if we are going to handle it, I think we should do as below:
* The caller of `fuse_xformers` should specify the intended target ort version.
* The fusion rules can specify their ort version requirement (which you do down below).

Does this make sense?

shubhambhokare1: This makes sense. Coverting this PR to a draft version until I come up with a suitable design (also seems like this is low priority for now)

shubhambhokare1:
Closing for now.

PR #2273, opened 2025-05-05
Add support for backtracking in pattern matcher

gramalingam:
Extends the recently introduced "Or" patterns fully (without any restrictions). The general case is handled via backtracking. The Or pattern constructor function will automatically determine whether the optimized (deterministic) matching implementation can be used or if the backtracking-based implementation should be used.

### DIFF
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -354,8 +354,12 @@ class MatchResult:
         return self._reason
 
     @property
-    def nodes(self) -> MutableSequence[ir.Node]:
-        return self._matched_nodes
+    def nodes(self) -> Sequence[ir.Node]:
+        return tuple(self._matched_nodes)
+
+    def add_node(self, node: ir.Node) -> None:
+        """Adds a node to the list of matched nodes."""
+        self._matched_nodes.append(node)
 
     def bind(self, var: str, value: Any) -> bool:
         """Binds a pattern variable name to a value from the matched IR.
@@ -1117,7 +1121,7 @@ class SimplePatternMatcher(PatternMatcher):
         if self._verbose:
             print(f"Matched: {node.op_type}")
 
-        match.nodes.append(node)
+        match.add_node(node)
         self._matched[pattern_node] = node
 
         # TODO: Revisit this to handle optional trailing inputs better.
--- a/onnxscript/rewriter/pattern.py
+++ b/onnxscript/rewriter/pattern.py
@@ -299,7 +299,7 @@ def _to_value_pattern(
 
 
 class MatchResult:
-    """Represents the result of a match operation.
+    """The state object used by the pattern-matching algorithm.
 
     A match can either succeed or fail.
     If it succeeds, it returns a list of nodes that matched the pattern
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/pattern.py
```diff
@@ -755,47 +875,66 @@ def __init__(
...
-            if id is not None and id in self._op_to_pattern:
-                return self._op_to_pattern[id]
-        return self._default_pattern
+            if id is None or id in mapping:
+                return None
+            mapping[id] = (tag_values[i], alternative)
+        return _OpIdDispatchOr(mapping, name, tag_var)
+
+    optimized_pattern = make_op_id_or_pattern()
+    return optimized_pattern or _BacktrackingOr(
```
shubhambhokare1, line 935: Given _BacktrackingOr is sort of a fallback mechanism here, could this cause any speedup issues where we have a lot of OrValues (that fallback to backtracking) used in the pattern?

onnxscript/rewriter/pattern.py
```diff
@@ -1216,10 +1349,23 @@ def _match_value(self, pattern_value: ValuePattern, value: ir.Value | None) -> b
             if value is None:
                 return self.fail("Mismatch: Constant pattern does not match None.")
             return self._match_constant(pattern_value, value)
-        if isinstance(pattern_value, OrValue):
+        if isinstance(pattern_value, _BacktrackingOr):
```
shubhambhokare1, line 1352: Curious as to what happens in this case,

t1 = someOp(x)
t2 = OrValue(choice1, choice2, choice3)
t3 = OrValue(choice1, choice4)
 let's say choice 1 is a successful option for both t2, t3

Would the matching for t3 fail for choice1 as its already bound to t2 and then would check for choice4 instead, or is it the other way around where choice1 is bound to t3 and then choice2 and choice3 are checked for t2

onnxscript/rewriter/pattern.py
gramalingam: The alternative would be to convert them into multiple rules, which would take even more time. That is what we need to compare against. If it is not fast enough, well, then we need to design an even more efficient algorithm.

onnxscript/rewriter/pattern.py
gramalingam: You have two different questions? Neither of your two alternatives would happen. First, nothing prevents two "pattern value variables" from being bound to the same value. For example, if you match the pattern `Add(x, y)` against a reall subgraph node `Add(v, v)`, then the pattern variables x and y would both be bound to the value v. There is nothing problematic here. The same thing applies in your scenario ... if you associate pattern-variables to t2 and t3 (you would have to specify them, in order to get the values bound to them), then can both match against the same value.

However, there is an interaction. When we are trying for a match for t3, if choice1 gets matched to some value, then when we try to match t2, choice1 would need to be matched to the same value. This is because a single pattern-variable, choice1, can only be bound to a single value even if it is repeated multiple times.

shubhambhokare1: Understood, thanks!

shubhambhokare1 (approved):
lgtm, thanks!

PR #2274, opened 2025-05-05
Extend optimize_for_ort to cover passes

### LINKED ISSUE #2261: Higher level API for post-processing/optimization
We are having more and more graph-level passes for users to combine their own needs. But it could be inconvenient for the users who only want a standard combination (clear metadata, lift constants, and fold transpose): https://github.com/kunal-vaishnavi/phi4-mm/pull/1/files#r2069106112.

The `optimize_model` api currently used in onnxruntime.

https://github.com/microsoft/onnxruntime/blob/9194cadbb02cb89bd0706f654f9abc06b39ce8bf/onnxruntime/python/tools/transformers/optimizer.py#L275

```python
def optimize_model(
    input: str | ModelProto,
    model_type: str = "bert",
    num_heads: int = 0,
    hidden_size: int = 0,
    optimization_options: FusionOptions | None = None,
    opt_level: int | None = None,
    use_gpu: bool = False,
    only_onnxruntime: bool = False,
    verbose: bool = False,
    *,
    provider: str | None = None,
) -> OnnxModel:
```

titaiwangms:
Fix #2261

### DIFF
--- a/onnxscript/rewriter/ort_fusions/_core.py
+++ b/onnxscript/rewriter/ort_fusions/_core.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import onnxscript.ir as ir
-from onnxscript.ir.passes.common import shape_inference
+import onnxscript.ir.passes.common as common_passes
 from onnxscript.optimizer import optimize
 from onnxscript.rewriter import rewrite
 from onnxscript.rewriter.ort_fusions import (
@@ -47,7 +47,7 @@ def _pre_optimize(model: ir.Model) -> ir.Model:
     # TODO: Do we need this dependence on ONNX's partial-data-propagation? There are some
     # extra shape-propagation and partial-data-propagation rules in ONNX that are not yet
     # incorporated in our optimizer.
-    shape_inference.infer_shapes(model)
+    common_passes.ShapeInferencePass()(model)
     optimize(model)
     return model
 
@@ -130,4 +130,17 @@ def optimize_for_ort(
     )
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
+
+    # TODO(exporter team): Fold transpose into initializers
+    # Apply the ORT optimization passes.
+    # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L172
+    common_passes.ClearMetadataAndDocStringPass()(model)
+    # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139
+    common_passes.LiftConstantsToInitializersPass(lift_all_constants=False, size_limit=1)(
+        model
+    )
+    common_passes.RemoveInitializersFromInputsPass()(model)
+    common_passes.ShapeInferencePass()(model)
+    common_passes.CheckerPass()(model)
+
     return model, fusion_count


gramalingam:
Please also consider whether this method should be optimize in-place or not. I think we can make it in-place now that shape-inference itself is in-place.

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -130,4 +130,17 @@ def optimize_for_ort(
     )
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
+
+    # TODO(exporter team): Fold transpose into initializers
+    # Apply the ORT optimization passes.
+    # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L172
+    common_passes.ClearMetadataAndDocStringPass()(model)
```
justinchuby, line 137: You may put all the passes into a pass manager like we do in optimize()

justinchuby:
> Please also consider whether this method should be optimize in-place or not. I think we can make it in-place now that shape-inference itself is in-place.

I think making it out-of-place is safer, in case we have passes in the future that need to be functional?

onnxscript/rewriter/ort_fusions/_core.py
titaiwangms: Done

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -130,4 +130,18 @@ def optimize_for_ort(
...
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
-    return model, fusion_count
+
+    passes = [
+        # TODO(exporter team): Fold transpose into initializers
+        # Apply the ORT optimization passes.
+        # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L172
+        common_passes.ClearMetadataAndDocStringPass(),
+        # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139
+        common_passes.LiftConstantsToInitializersPass(lift_all_constants=False, size_limit=1),
```
titaiwangms, line 140: We have another pass called `LiftSubgraphInitializersToMainGraphPass`. Do we know if it's needed in genAI? @kunal-vaishnavi

kunal-vaishnavi: If the pass logic is in `DynamoOnnxHelper`, then it is used for ONNX Runtime GenAI.

justinchuby: We don't really produce graphs with subgraph initializers. I think we are ok either way

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -135,4 +135,18 @@ def optimize_for_ort(
     )
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
-    return model, fusion_count
+
+    passes = [
```
justinchuby, line 139:
```suggestion
    passes = ir.passes.Sequential(
```

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -135,4 +135,18 @@ def optimize_for_ort(
...
+        # Apply the ORT optimization passes.
+        # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L172
+        common_passes.ClearMetadataAndDocStringPass(),
+        # https://github.com/microsoft/onnxruntime/blob/74dcf7e296639095dfa55d31336998b6f719ed76/onnxruntime/python/tools/transformers/dynamo_onnx_helper.py#L139
+        common_passes.LiftConstantsToInitializersPass(lift_all_constants=False, size_limit=1),
+        common_passes.RemoveInitializersFromInputsPass(),
+        common_passes.ShapeInferencePass(),
+        common_passes.CheckerPass(),
+    ]
+    optimize_for_ort_passes = ir.passes.Sequential(*passes)
```
justinchuby, line 150:
```suggestion
    )
```

justinchuby (approved):

titaiwangms:
I will set up Whisper and test it before merge this PR https://github.com/microsoft/onnxruntime/pull/24382

PR #2277, opened 2025-05-07
Use OrPatterns to support SkipLayerNormalization rewrite vaiations

shubhambhokare1:


onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -22,7 +22,15 @@ def __init__(self, name: str, has_bias: bool = False, bias_pre_add: bool = False
     def pattern(self, op, input, skip, gamma, bias, epsilon, stash_type):
         if self._has_bias and self._bias_pre_add:
             input = op.Add(input, bias)
-        skip_sum = op.Add(input, skip)
+
+        # Support different combinations of addition of input and skip
+        skip_sum_pattern_1 = op.Add(skip, input)
```
gramalingam, line 27: Do you see both these patterns arising in examples? Or, is it just to try it out?

shubhambhokare1: The Whisper model consists of this pattern. I've seen this in a couple other models in the benchmark as well.

gramalingam: Just to clarify my question: I mean, have you seen *both* variations of doing the Add (both `Add(skip,input)` and `Add(input,skip)`)?

gramalingam (approved):
After #2284, you should be able to do this without naming the OrValue explicitly. In general, supporting arbitrary OrValues as return values can be complicated, but we should be able to handle this one.

Issue #1440, opened 2024-04-23
[IR] Point graph input and initializers back to the Graph


justinchuby:
Point graph input and initializers back to the Graph. Use a tracked list for graph.input (and maybe output) to update the pointer.

PR #2282, opened 2025-05-07
[IR] Record owning graph for input/output/initializers

justinchuby:
Fix https://github.com/microsoft/onnxscript/issues/1440 by pointing graph input, output and initializers back to the Graph using a tracked list.

Users can now check if a value is a graph input/output/initializer, and find the owning graph of a value with `.graph`.


onnxscript/ir/_core.py
```diff
@@ -1757,6 +1763,9 @@ class Value(_protocols.ValueProtocol, _display.PrettyPrintable):
 
     __slots__ = (
         "_const_value",
+        "_graph_initializer_of",
```
justinchuby, line 1766:
```suggestion
```

onnxscript/ir/_core.py
```diff
@@ -1808,6 +1814,13 @@ def __init__(
         self._uses: dict[Usage, None] = {}
         self.doc_string = doc_string
 
+        # The graph this value belongs to. It is set *only* when the value is added as
+        # a graph input, output or initializer.
+        # The two properties can only be set by the Graph class (GraphIO).
```
gramalingam, line 1819: Is it two or three?

onnxscript/ir/_core.py
```diff
@@ -1846,11 +1859,33 @@ def _constant_tensor_part(self) -> str:
                 return f"{{{self.const_value.__class__.__name__}(...)}}"
         return ""
 
+    @property
+    def graph(self) -> Graph | None:
```
gramalingam, line 1863: A couple of inter-related thoughts:
(a) Would it be a bit simpler and more efficient to store a single "graph" slot, along with 1 or more fields to indicate the type of value it is?
(b) Unclear why graph_output_of would be necessary ... this sort of relates to the question of what are the different kinds of values: we have inputs, produced values (outputs of nodes), initializers that are/are-not graph-inputs. Graph outputs would have to be one of the above kinds: though onnx doesn't allow inputs to be outputs ... i guess outputs would have to be either produced values or initializers-that-are-not-inputs.

gramalingam: I guess the main thing is to be clear what possibilities/combinations are legal/possible.

onnxscript/ir/_core.py
```diff
@@ -1808,6 +1815,13 @@ def __init__(
         self._uses: dict[Usage, None] = {}
         self.doc_string = doc_string
 
+        # The graph this value belongs to. It is set *only* when the value is added as
+        # a graph input, output or initializer.
+        # The two properties can only be set by the Graph class (GraphIO).
```
justinchuby, line 1820:
```suggestion
        # The three properties can only be set by the Graph class (_GraphIO and GraphInitializers).
```

onnxscript/ir/_graph_containers.py
```diff
@@ -0,0 +1,213 @@
...
+import logging
+from typing import TYPE_CHECKING, Iterable, SupportsIndex
+
+import onnxscript
+
+if TYPE_CHECKING:
+    from onnxscript.ir import _core
+
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
justinchuby: Done

onnxscript/ir/_core.py
justinchuby: Updated to three. Thanks!

onnxscript/ir/_core.py
justinchuby: A value can be any combination of the three if we parse from a protobuf format, some cases even though invalid may still appear in an onnx file. So I was hoping to cover these case and not embed too much logic here. It seems like among the combinations we can have the following valid ones:

| input | output | initializer | valid? |
|-------|--------|-------------|--------|
| N     | N      | N           | ✅      |
| N     | N      | Y           | ✅      |
| N     | Y      | N           | ✅      |
| N     | Y      | Y           | ✅      |
| Y     | N      | N           | ✅      |
| Y     | N      | Y           | ✅      |
| Y     | Y      | N           | ❌      |
| Y     | Y      | Y           | ❌      |

justinchuby: So maybe we can instead store
1. a field that indicates input or output
2. a field that indicates initializer
3. the graph

onnxscript/ir/_graph_containers.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def __init__(self, graph: _core.Graph, initlist=None):
+        self._graph = graph
+        # Use a ref counter to track the number of references to each value
+        # in the input/output list. This is used to determine when to unset the graph
+        # reference in the value.
+        # Even though a duplicated value is invalid in inputs and not recommended in outputs,
+        # it is still possible to have duplicated inputs/outputs in an ONNX graph so we
+        # need to properly handle this case and maintain the graph reference properly.
+        self._ref_counter = collections.Counter()
```
justinchuby, line 34:
```suggestion
        self._ref_counter: collections.Counter[_core.Value] = collections.Counter()
```

onnxscript/ir/_graph_containers.py
```diff
@@ -21,18 +20,23 @@
...
     def __init__(self, graph: _core.Graph, initlist=None):
-        super().__init__(initlist)
         self._graph = graph
+        # Use a ref counter to track the number of references to each value
+        # in the input/output list. This is used to determine when to unset the graph
+        # reference in the value.
+        # Even though a duplicated value is invalid in inputs and not recommended in outputs,
+        # it is still possible to have duplicated inputs/outputs in an ONNX graph so we
+        # need to properly handle this case and maintain the graph reference properly.
+        self._ref_counter: collections.Counter[_core.Value] = collections.Counter()
```
gramalingam, line 34: nit: It might be more efficient to simply recheck if a removed input is still in the input list (given than removing an input is an unlikely operation ... I doubt if there is much functionality today that actually does this). On the other hand, it may not matter much (adding inputs does happen, but probably not that much).

PR #2284, opened 2025-05-07
Returning choice values in patterns

gramalingam:
A fix for the use of Choice values (i.e., `OrValue([choice1, choice2])` as a return-value in the pattern (that is, as a root of the pattern). This is a bit complicated to support with the current implementation, which is oriented towards iterating over the nodes in the graph, and matching them against pattern-nodes. This PR provides a limited extension (which handles the case in PR 2277): when the returned choice-values are already covered by the other output values, they can be supported by the existing matcher.

Also, on a related note, fix how value-bindings are handled in the pattern-matcher to make it easier to return these values as the outputs of the pattern.


justinchuby:
For my understanding, could you explain more on what "returned choice-values are already covered by the other output values" means?

justinchuby (approved):

onnxscript/rewriter/pattern.py
justinchuby, line 1: Would be helpful to document the limitation somewhere in the documentation

gramalingam:
> For my understanding, could you explain more on what "returned choice-values are already covered by the other output values" means?

"value1 is covered by value2" is just shorthand here for "value1 is transitively used to compute value2". In the pattern-matching situation, a successful backward match for "value2" would also have found a match for "value1". Consider a pattern that includes
```
   value1 = ...
   value2 = op.Relu(value1)
  return value2, value1
```
Here, if we have successfully matched a subgraph for value2, we no longer need to do any more search to match value1, it is guaranteed to have already been found.

docs/sg_execution_times.rst
justinchuby, line 1: nit: Before to exclude this file

docs/tutorial/rewriter/attributes.md
```diff
@@ -0,0 +1,22 @@
+## Specifying attributes in the pattern
```
justinchuby, line 1:
```suggestion
# Specifying attributes in the pattern
```

docs/tutorial/rewriter/commute.md
```diff
@@ -0,0 +1,71 @@
+(heading-target-commute)=
+## Utilizing `commute` parameter for pattern-matching
```
justinchuby, line 2:
```suggestion
# Utilizing `commute` parameter for pattern-matching
```

docs/tutorial/rewriter/simple_example.md
```diff
@@ -0,0 +1,71 @@
+(heading-target-simple)=
+## A Simple Example
```
justinchuby, line 2:
```suggestion
# A Simple Example
```

etc.

docs/sg_execution_times.rst
gramalingam: I was surprised to see this file, wasn't sure of its origin/intent. Are you suggesting excluding it?

justinchuby: It's generated by sphinx build. I think we can add it to gitignore

docs/tutorial/rewriter/attributes.md
gramalingam: Sounds good. Looked at the generated html, looks as below: I think the existing heading-depth makes sense if we anticipate having multiple tutorials under "Rewriter Tutorials", but I don't see that happening anytime soon. We should rename this to be "Rewriter Tutorial" and eliminate the top-level "Pattern-based Rewrite Using Rules".
image

gramalingam: Did it. But, on the flip side, when you go into the tutorial sections, there is no top-level title, as below: so, it is not so clear-cut which is better: 
image

justinchuby: I see. Are the pages intended to be sections of the same page?

docs/ir/tensors.md
```diff
@@ -188,7 +188,7 @@ To fully support arrays from other frameworks, it is usually a good idea to crea
 
 ```{eval-rst}
 .. exec_code::
-
+    from __future__ import annotations
```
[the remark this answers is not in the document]
gramalingam: Yes, I needed it locally. Else, I need to recreate a new conda environment. Is there any harm in leaving this? Is there a strong dependence on 3.10 or higher now?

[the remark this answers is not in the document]
justinchuby: I guess it's ok. onnxscript is still 3.9&up

docs/tutorial/rewriter/rewrite_patterns.md
```diff
@@ -1,231 +1,26 @@
-# Pattern-based Rewrite Using Rules
-
-## Introduction
+# Introduction
 
 The ONNX Rewriter tool provides the user with the functionality to replace certain patterns in an ONNX graph with another pattern based on rewrite rules provided by the user.
 
-## Usage
+# Usage
```
justinchuby, line 5: I think there should only be one level 1 heading in each file?

docs/tutorial/rewriter/attributes.md
justinchuby, line 1: I didn't realize the file is included as part of the main file. (though it was an independent page) If so having the top most heading to be level 2 makes sense.

gramalingam: Will revisit this later.

justinchuby:
@gramalingam looks like the docs workflow failed: https://github.com/microsoft/onnxscript/actions/runs/14933269595/job/41954640986

PR #2288, opened 2025-05-10
Unify rule implementations with classes

justinchuby:
- Replace RewriteRuleAsClass with RewriteRuleClassBase to unify rule implementations.
- Also: Use as ints() on attributes in rewrite rules

### DIFF
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -190,7 +190,8 @@ class TransposeIdentity(orp.RewriteRuleClassBase):
         if isinstance(perm, ir.RefAttr):
             return check_result.fail("Permutation is a reference attribute.")
         if perm.type == ir.AttributeType.INTS:
-            if perm.as_ints() == list(range(len(perm.as_ints()))):
+            perm_ints = perm.as_ints()
+            if perm_ints == list(range(len(perm_ints))):
                 return check_result
         return check_result.fail("Permutation is not identity.")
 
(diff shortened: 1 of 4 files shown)


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -32,26 +32,23 @@ def check(self, context, x) -> orp.MatchResult:
...
-    def rewrite(cls, op, x: ir.Value, to: ir.Attr):
+    def rewrite(self, op, x: ir.Value, to: ir.Attr):
         return op.Identity(x)
 
-    @classmethod
-    def check(cls, context, x, to) -> orp.MatchResult:
+    def check(self, context, x, to) -> orp.MatchResult:
         check_result = orp.MatchResult()
-        if x.dtype != to.value:
+        if x.dtype != to.as_int():
```
gramalingam, line 46: Trying to understand this: does `as_int` throw an exception if it is not an int attribute? If so, the existing version seems preferable?

One design question worth thinking about is: what the optimization rules should do when dealing with invalid nodes: eg., if we have a model with a Cast node where the "to" attribute is a string or float. The comparison with `to.value` will choose to not optimize the node, but not fail, while the changed version might fail.

gramalingam (approved):
LGTM, just a minor question about the use of `.value` vs `.asint`

onnxscript/rewriter/llama_rule_sets.py
justinchuby: >  does as_int throw an exception if it is not an int attribute
I think so.

Good point thanks. I was assuming the graph to be valid (can pass checker). I suppose we can assume graphs to be always valid in rewrites, other wise moving responsibilities of ensuring graph validity to the rules may be burdensome?

justinchuby: Will add the explicit check to as_int() as discussed

PR #2295, opened 2025-05-12
Update opset imports in version_converter

shubhambhokare1:
- Update version converter logic to only up-convert if all nodes in the graph can be successfully upconverted to that opset version
- Assign opset import to be the highest opset version to which all the nodes were able to be successfully converted to.

### DIFF
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -220,6 +220,38 @@ class _VersionConverter:
     def __init__(self, target_version: int):
         self.target_version = target_version
 
+    def _maybe_set_opset_version(
+        self, model_or_function: ir.Model | ir.Function, domain: str, version: int | None
+    ):
+        """Set the opset version for the domain."""
+        current_version = model_or_function.opset_imports.get(domain)
+        if version is None or current_version is None:
+            return
+        if domain == "":
+            model_or_function.opset_imports[domain] = max(version, current_version)
+            return
+        elif domain == "ai.onnx":
+            model_or_function.opset_imports[domain] = max(version, current_version)
+            return
+        else:
+            return
+
+    def _update_opset_imports(self, model: ir.Model) -> None:
+        """Collect all opsets used and add opset imports to the model and functions."""
+        for node in ir.traversal.RecursiveGraphIterator(model.graph):
+            domain = node.domain
+            self._maybe_set_opset_version(model, domain, node.version)
+
+        for function in model.functions.values():
+            for node in ir.traversal.RecursiveGraphIterator(function):
+                domain = node.domain
+                self._maybe_set_opset_version(function, domain, node.version)
+            for domain, version in function.opset_imports.items():
+                # Add all opsets used in the function to the model, because ONNX Runtime
+                # does not handle adding the opset imports to the model after inlining during inference.
+                # This should happen after all opsets are collected for the function from its nodes.
+                self._maybe_set_opset_version(model, domain, version)
+
     def _upgrade_version(self, node: ir.Node, opset_version: int, up_conversion: bool) -> None:
         if up_conversion is True:
             node.version = opset_version + 1
@@ -320,6 +352,8 @@ class _VersionConverter:
                 return None
         self.model_version = model_version
         self.visit_graph(model.graph)
+        # Finally, update the opset imports for the model
+        self._update_opset_imports(model)
         return None
 
 
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -277,7 +277,7 @@ class _VersionConverter:
             self.replace_node(node, replacement, root)
         return None
 
-    def visit_graph(self, graph: ir.Graph) -> None:
+    def visit_graph(self, graph: ir.Graph) -> ir.Graph | None:
         if self.target_version > SUPPORTED_MAX_ONNX_OPSET:
             logger.warning(
                 "Conversion to target opset: %s not currently supported.",
@@ -296,15 +296,14 @@ class _VersionConverter:
                 if node.version is None:
                     node.version = self.model_version
                 if self.target_version < node.version:
-                    up_conversion = False
+                    # up_conversion = False
                     # TODO(shubhambhokare1): Remove once down-conversion adapters are supoorted
                     logger.warning(
                         "Target opset: %s less than %s, downstream version conversion not currently handled.",
                         self.target_version,
                         node.version,
                     )
-                    graph = pre_conversion_graph
-                    return None
+                    return pre_conversion_graph
                 try:
                     self.visit_node(node, graph, self.model_version, up_conversion)
                     self._upgrade_version(node, self.model_version, up_conversion)
@@ -314,8 +313,7 @@ class _VersionConverter:
                         node.op_type,
                         e,
                     )
-                    graph = pre_conversion_graph
-                    return None
+                    return pre_conversion_graph
             self.model_version += 1
             del pre_conversion_graph
         return None
@@ -328,7 +326,9 @@ class _VersionConverter:
             if model_version is None:
                 return None
         self.model_version = model_version
-        self.visit_graph(model.graph)
+        graph = self.visit_graph(model.graph)
+        if graph is not None:
+            model.graph = graph
 
         # Finally, update the opset imports for the model
         if self.model_version != self.target_version:
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -4,6 +4,7 @@
 
 from __future__ import annotations
 
+import copy
 import dataclasses
 import functools
 import logging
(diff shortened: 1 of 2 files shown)


onnxscript/version_converter/_version_converter.py
```diff
@@ -220,6 +220,38 @@ class _VersionConverter:
...
+        """Set the opset version for the domain."""
+        current_version = model_or_function.opset_imports.get(domain)
+        if version is None or current_version is None:
+            return
+        if domain == "":
+            model_or_function.opset_imports[domain] = max(version, current_version)
+            return
+        elif domain == "ai.onnx":
+            model_or_function.opset_imports[domain] = max(version, current_version)
+            return
```
justinchuby, line 235:
```suggestion
        if domain == "":
            model_or_function.opset_imports[domain] = max(version, current_version)
            return
        elif domain == "ai.onnx":
            model_or_function.opset_imports[domain] = max(version, current_version)
            return
```

since `""` and `"ai.onnx"` is the same, they should share the same opset import. In practice it may be good to assert that we don't have any `"ai.onnx"` nodes if handling both domains over complicates the logic

onnxscript/version_converter/_version_converter.py
```diff
@@ -320,6 +352,8 @@ def visit_model(self, model: ir.Model) -> None:
                 return None
         self.model_version = model_version
         self.visit_graph(model.graph)
+        # Finally, update the opset imports for the model
+        self._update_opset_imports(model)
```
justinchuby, line 356: With this logic, if a user request opset 22, but all nodes were updated to opset 21 max, is it still possible for the model to be set at 22?

onnxscript/version_converter/_version_converter.py
```diff
@@ -220,6 +220,38 @@ class _VersionConverter:
...
+        else:
+            return
+
+    def _update_opset_imports(self, model: ir.Model) -> None:
+        """Collect all opsets used and add opset imports to the model and functions."""
+        for node in ir.traversal.RecursiveGraphIterator(model.graph):
+            domain = node.domain
+            self._maybe_set_opset_version(model, domain, node.version)
+
+        for function in model.functions.values():
```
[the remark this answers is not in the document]
shubhambhokare1: True, I see we are inlining before this pass is called.

onnxscript/version_converter/_version_converter.py
```diff
@@ -220,6 +220,38 @@ class _VersionConverter:
     def __init__(self, target_version: int):
         self.target_version = target_version
 
+    def _maybe_set_opset_version(
```
gramalingam, line 223: A couple of questions:
* I assume that the version-converter will not change anything for opsets other than the standard onnx opset. Is this correct? If so, wouldn't it be better if it updates only the standard onnx opset version in imports?
* I don't understand the use of "max" down below. 
* I guess it might have to do with a failure to update some nodes to the requested opset version? It would be useful if we have some documentation about what happens in the various special cases, like failure to version-convert etc.

onnxscript/version_converter/_version_converter.py
shubhambhokare1: Currently, all nodes are upconverted to target_version. If no adapter is registered for a particular node, only the node.version is updated to be =target_version.

Now, the real issue would arise, in situations in which up_conversion for a node is not possible, let's take the example for GroupNorm with no bias, if no bias was provided the up conversion from 20->21 would not possible, meaning we would have a model, where GroupNorm is version 20, but all other ops are 21.

I think we need a rethink of the version conversion logic, instead of iterating (each node, then up-converting verA->verB->verC) we would have to switch the loop order to be (verA->verB, iterate each node, verB->verC, iterate each node, and so on), maintaining a copy of the original graph pre-conversion for that opset during each up conversion from opset -> opset +1, if any conversion fails, we fall back to the graph from before that up conversion, so all the nodes are of a particular opset. Once this process is done, we assign the model_imports to be whatever the last successful up conversion was. 

Marking the PR as draft as I flesh out/implement this new logic.

onnxscript/version_converter/_version_converter.py
shubhambhokare1: For the first point, that is correct. I'm just returning, if node's domain is not the standard onnx opset.

For the second question, yes, the "max" is call to account for the failure of up-converts of some nodes. So for example, if GroupNorm fails to up-convert from 20->21, we wouldn't want opset imports to be "":21,

However I think this is a flaw with the current logic, we ideally don't want to have nodes with mixed opset versions, so the loop order might need to change (mentioned in the reply to Justin above)

onnxscript/version_converter/_version_converter.py
shubhambhokare1: Updated logic.

onnxscript/version_converter/_version_converter.py
```diff
@@ -276,39 +277,45 @@ def visit_node(
...
+
+        # TODO(shubhambhokare1): Support down-conversion
+        while self.model_version < self.target_version:
+            pre_conversion_graph = copy.copy(graph)
+            # Up-convert each node in the graph from opset_version -> opset_version + 1
+            # or down-convert from opset_version -> opset_version - 1
+            # Return non-converted graph if any node fails to convert.
+            for node in graph:
+                up_conversion = True
+                if node.version is None:
```
gramalingam, line 296: We should check if it is an onnx domain op first: else skip, and don't do anything (even updating node.version).

onnxscript/version_converter/_version_converter.py
```diff
@@ -276,39 +277,45 @@ def visit_node(
...
-                    "Target opset: %s less than %s, downstream version conversion not currently handled.",
-                    self.target_version,
-                    self.model_version,
-                )
-                return None
-            for opset_version in range(node.version, self.target_version):
+
+        # TODO(shubhambhokare1): Support down-conversion
+        while self.model_version < self.target_version:
+            pre_conversion_graph = copy.copy(graph)
```
gramalingam, line 290: @justinchuby : how expensive is copy? Specifically, what does it do for large tensors?

justinchuby: We can't copy graphs yet. This won't work (sorry)

shubhambhokare1: @justinchuby What would be the correct way to store pre-conversion state graphs?

justinchuby: Maybe implement a proper graph.duplicate()? Before that maybe aborting on failures is acceptable?

onnxscript/version_converter/_version_converter.py
gramalingam: Hmmm ... I guess #2283 does NOT do it for opset-imports, that remains a TODO, is that right?

onnxscript/version_converter/_version_converter.py
gramalingam: Thanks for the update. I think it will also help to verify/document the global-strategy. I think some of the earlier complications stem back to how onnxscript functions are registered for torchlib ops. Suppose I register an onnxscript function for torch.gelu targetting onnx opset 20. Suppose the user requests onnx opset 21. If we use the abovementioned onnxscript function, will it generate the decomposition with calls to opset 20 or 21? I assume that they will be 20? And will need to be version-converted to 21?

So, in general: the assumption is that the model we start with (before version-conversion) could contain calls to several different onnx opsets (say both 20 and 21), with the assumption that they will be typically <= the target version.

What is the self.model_version's initial value in your loop? How is it determined?

onnxscript/version_converter/_version_converter.py
justinchuby: Yeah we need to create a new data structure for opset imports to do this, which is a todo.

onnxscript/version_converter/_version_converter.py
shubhambhokare1: #2298

gramalingam:
My understanding is this:
* The (python) IR allows different nodes in the graph to call different opset versions (of the standard domain)
* The models generated by torchlib may actually have such nodes (with different opset versions). (We may be able to change this, but that will require some changes in torchlib. And would have some disadvantages.)

Hence:
* The version-converter should support input models where different nodes may use different opset versions.
* It will be unable to guarantee that it will be able to produce a model where all nodes use same opset-version.
* I suggest settling for a weaker guarantee: we migrate each node to the opset closest to the target-opset that we can get to, which however might not be the target-opset or even be the same as for other nodes. We return a boolean or status indicating where we were unsuccessful. (Ideally, we should map the error back to a torchlib-registered function, which likely needs a new version for failing-opset-version). The user can also try conversion with other opset versions if they want.
* This will not require creating a copy of the graph.

The goal of producing a model with same opset-version through-out will require extra work anyway.

@justinchuby : does this sound reasonable?

justinchuby:
That sounds reasonable to me!

titaiwangms:
Replaced by #2318

PR #2304, opened 2025-05-13
Support common subexpression elimination pass (CSE)

### LINKED ISSUE #2105: CSE pass
Tracker issue

titaiwangms:
Fix #2105 

For the logic, this PR follows https://github.com/pytorch/pytorch/blob/main/torch/fx/passes/dialect/common/cse_pass.py.

Essentially, this PR traverses the original graph and examines whether the values or the nodes are duplicated. If it's not, the values or the nodes is saved in mappings, and added to the new graph. If it is duplicated, the value or the node is replaced with the mapped/saved value or node. (FunctionalPass)

TODO: Subgraph is not covered yet. The current pass should fail on control flow nodes.

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -2734,7 +2734,7 @@ class Model(_protocols.ModelProtocol, _display.PrettyPrintable):
         model_version: int | None = None,
         doc_string: str | None = None,
         functions: Sequence[Function] = (),
-        meta_data_props: dict[str, str] | None = None,
+        metadata_props: dict[str, str] | None = None,
     ) -> None:
         self.graph: Graph = graph
         self.ir_version = ir_version
@@ -2745,7 +2745,7 @@ class Model(_protocols.ModelProtocol, _display.PrettyPrintable):
         self.doc_string = doc_string
         self._functions = {func.identifier(): func for func in functions}
         self._metadata: _metadata.MetadataStore | None = None
-        self._metadata_props: dict[str, str] | None = meta_data_props
+        self._metadata_props: dict[str, str] | None = metadata_props
 
     @property
     def functions(self) -> dict[_protocols.OperatorIdentifier, Function]:
--- a/onnxscript/ir/passes/common/common_subexpression_elimination.py
+++ b/onnxscript/ir/passes/common/common_subexpression_elimination.py
@@ -15,12 +15,9 @@ from onnxscript import ir
 logger = logging.getLogger(__name__)
 
 
-class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
+class CommonSubexpressionEliminationPass(ir.passes.FunctionalPass):
     """Eliminate common subexpression in ONNX graphs."""
 
-    # TODO(titaiwang): Logger
-    # TODO(titaiwang): Add more tests
-    # TODO(titaiwang): FunctionalPass?
     def call(self, model: ir.Model) -> ir.passes.PassResult:
         """Return the same ir.Model but with CSE applied to the graph."""
         modified = False
@@ -49,10 +46,12 @@ class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
             new_input = _copy_value(input)
             new_graph.inputs.append(new_input)
             old_value_to_new_value[input] = new_input
+            logger.debug("Adding input %s", new_input.name)
         for initializer in old_graph.initializers.values():
             new_initializer = _copy_value(initializer)
             new_graph.register_initializer(new_initializer)
             old_value_to_new_value[initializer] = new_initializer
+            logger.debug("Adding initializer %s", new_initializer.name)
 
         # 4. Create nodes in the new graph from the old graph.
         for old_node in old_graph:
@@ -66,21 +65,27 @@ class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
                     old_node_inputs.append(old_input)
             # 4.1. Construct the (node, inputs, attributes) hash to
             #      check if the node is a common subexpression.
+            # Attr is not hashable
+            attributes = {}
+            for k, v in old_node.attributes.items():
+                assert isinstance(v, ir.Attr)
+                attributes[k] = v.value
             hash_value = hash(
                 (
                     old_node.op_identifier(),
                     tuple(old_node_inputs),
-                    tuple(old_node.attributes.items()),
+                    tuple(attributes),
                 )
             )
-            # TODO(exporter team): Subgraphs are not supported yet.
-            # TODO(exporter team): Skip control flow nodes?
+            # TODO(titaiwang): Subgraphs are not supported yet.
+            # TODO(titaiwang): Skip control flow nodes?
             # 4.2. Check if the node is a common subexpression.
             if hash_value in old_node_hash_to_new_node:
                 # 4.2.1. If it is, this node is already in the new graph, so
                 #         we don't need to create a new node.
                 modified = True
                 new_node = old_node_hash_to_new_node[hash_value]
+                logger.debug("Reusing node %s", new_node.name)
             else:
                 # 4.2.2. If it is not, create a new node and add it to the graph.
                 new_node = _copy_node(old_node, old_value_to_new_value)
@@ -92,8 +97,9 @@ class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
         for output in old_graph.outputs:
             new_output = old_value_to_new_value[output]
             new_graph.outputs.append(new_output)
+            logger.debug("Adding output %s", new_output.name)
         # 6. Replace the old graph with the new graph.
-        model.graph = new_graph
+        model = _copy_model(original_model=model, new_graph=new_graph)
         return ir.passes.PassResult(
             model,
             modified=modified,
@@ -135,3 +141,23 @@ def _copy_node(
         version=original_node.version,
     )
     return new_node
+
+
+def _copy_model(
+    original_model: ir.Model,
+    new_graph: ir.Graph,
+) -> ir.Model:
+    """Copy an IR model but with the new graph."""
+    functions = tuple(original_model.functions.values())
+    new_model = ir.Model(
+        graph=new_graph,
+        ir_version=original_model.ir_version,
+        producer_name=original_model.producer_name,
+        producer_version=original_model.producer_version,
+        domain=original_model.domain,
+        model_version=original_model.model_version,
+        doc_string=original_model.doc_string,
+        functions=functions,
+        metadata_props=original_model.metadata_props,
+    )
+    return new_model
--- a/onnxscript/ir/passes/common/common_subexpression_elimination.py
+++ b/onnxscript/ir/passes/common/common_subexpression_elimination.py
@@ -1,0 +1,137 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Eliminate common subexpression in ONNX graphs."""
+
+from __future__ import annotations
+
+__all__ = [
+    "CommonSubexpressionEliminationPass",
+]
+
+import logging
+
+from onnxscript import ir
+
+logger = logging.getLogger(__name__)
+
+
+class CommonSubexpressionEliminationPass(ir.passes.InPlacePass):
+    """Eliminate common subexpression in ONNX graphs."""
+
+    # TODO(titaiwang): Logger
+    # TODO(titaiwang): Add more tests
+    # TODO(titaiwang): FunctionalPass?
+    def call(self, model: ir.Model) -> ir.passes.PassResult:
+        """Return the same ir.Model but with CSE applied to the graph."""
+        modified = False
+        # 1. Initialize a new graph. It will be used to store the new nodes.
+        #    and replace the old graph.
+        old_graph = model.graph
+        # Values and nodes point to the old graph, so they need to be
+        # created in the new graph.
+        new_graph = ir.Graph(
+            inputs=[],
+            outputs=[],
+            nodes=[],
+            initializers=[],
+            name=old_graph.name,
+            metadata_props=old_graph.metadata_props,
+            opset_imports=old_graph.opset_imports,
+            doc_string=old_graph.doc_string,
+        )
+        # 2. Create a mapping from olds to news.
+        old_node_hash_to_new_node: dict[int, ir.Node] = {}
+        old_value_to_new_value: dict[ir.Value, ir.Value] = {}
+
+        # 3. Create inputs and initializers in the new graph
+        #    from the old graph.
+        for input in old_graph.inputs:
+            new_input = _copy_value(input)
+            new_graph.inputs.append(new_input)
+            old_value_to_new_value[input] = new_input
+        for initializer in old_graph.initializers.values():
+            new_initializer = _copy_value(initializer)
+            new_graph.register_initializer(new_initializer)
+            old_value_to_new_value[initializer] = new_initializer
+
+        # 4. Create nodes in the new graph from the old graph.
+        for old_node in old_graph:
+            # 4.0. Iterate and update node inputs.
+            old_node_inputs: list[ir.Value] = []
+            for old_input in old_node.inputs:
+                assert old_input is not None
+                if old_input in old_value_to_new_value:
+                    old_node_inputs.append(old_value_to_new_value[old_input])
+                else:
+                    old_node_inputs.append(old_input)
+            # 4.1. Construct the (node, inputs, attributes) hash to
+            #      check if the node is a common subexpression.
+            hash_value = hash(
+                (
+                    old_node.op_identifier(),
+                    tuple(old_node_inputs),
+                    tuple(old_node.attributes.items()),
+                )
+            )
+            # TODO(exporter team): Subgraphs are not supported yet.
+            # TODO(exporter team): Skip control flow nodes?
+            # 4.2. Check if the node is a common subexpression.
+            if hash_value in old_node_hash_to_new_node:
+                # 4.2.1. If it is, this node is already in the new graph, so
+                #         we don't need to create a new node.
+                modified = True
+                new_node = old_node_hash_to_new_node[hash_value]
+            else:
+                # 4.2.2. If it is not, create a new node and add it to the graph.
+                new_node = _copy_node(old_node, old_value_to_new_value)
+                new_graph.append(new_node)
+                old_node_hash_to_new_node[hash_value] = new_node
+            # 4.3 Add the node outputs to the mapping.
+            old_value_to_new_value.update(dict(zip(old_node.outputs, new_node.outputs)))
+        # 5. Create outputs in the new graph from the old graph.
+        for output in old_graph.outputs:
+            new_output = old_value_to_new_value[output]
+            new_graph.outputs.append(new_output)
+        # 6. Replace the old graph with the new graph.
+        model.graph = new_graph
+        return ir.passes.PassResult(
+            model,
+            modified=modified,
+        )
+
+
+def _copy_value(original_value: ir.Value) -> ir.Value:
+    """Copy an IR value."""
+    new_input = ir.Value(
+        name=original_value.name,
+        shape=original_value.shape,
+        type=original_value.type,
+        doc_string=original_value.doc_string,
+        const_value=original_value.const_value,
+    )
+    return new_input
+
+
+def _copy_node(
+    original_node: ir.Node, old_value_to_new_value: dict[ir.Value, ir.Value]
+) -> ir.Node:
+    """Copy an IR node."""
+    new_inputs: list[ir.Value] = []
+    for original_input in original_node.inputs:
+        if original_input in old_value_to_new_value:
+            new_inputs.append(old_value_to_new_value[original_input])
+        else:
+            raise ValueError(f"Input {original_input} not found in old_value_to_new_value")
+    new_node = ir.node(
+        domain=original_node.domain,
+        op_type=original_node.op_type,
+        inputs=new_inputs,
+        attributes=original_node.attributes,
+        overload=original_node.overload,
+        num_outputs=len(original_node.outputs),
+        metadata_props=original_node.metadata_props,
+        doc_string=original_node.doc_string,
+        name=original_node.name,
+        version=original_node.version,
+    )
+    return new_node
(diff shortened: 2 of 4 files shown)


onnxscript/ir/_core.py
```diff
@@ -2788,7 +2788,7 @@ def __init__(
         model_version: int | None = None,
         doc_string: str | None = None,
         functions: Sequence[Function] = (),
-        meta_data_props: dict[str, str] | None = None,
```
justinchuby, line 2791: thanks!

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+            new_graph.register_initializer(new_initializer)
+            old_value_to_new_value[initializer] = new_initializer
+            logger.debug("Adding initializer %s", new_initializer.name)
+
+        # 4. Create nodes in the new graph from the old graph.
+        for old_node in old_graph:
+            # 4.0. Iterate and update node inputs.
+            old_node_inputs: list[ir.Value] = []
+            for old_input in old_node.inputs:
+                assert old_input is not None
```
gramalingam, line 61: There should be no problem allowing None ... we can add None to the hashed list of inputs, and it should work, I think.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+                if old_input in old_value_to_new_value:
+                    old_node_inputs.append(old_value_to_new_value[old_input])
+                else:
+                    old_node_inputs.append(old_input)
+            # 4.1. Construct the (node, inputs, attributes) hash to
+            #      check if the node is a common subexpression.
+            # Attr is not hashable
+            attributes = {}
+            for k, v in old_node.attributes.items():
+                assert isinstance(v, ir.Attr)
```
gramalingam, line 71: If any attribute is a Graph or list of Graphs, we should skip the equality check (and handle it as if the check failed ... that is, as if it is a new expression, not equal to any old one).

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+                attributes[k] = v.value
+            hash_value = hash(
```
gramalingam, line 73: I don't think we should hash: we should use the tuple as-is (using equality check).

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+                modified = True
+                new_node = old_node_hash_to_new_node[hash_value]
+                logger.debug("Reusing node %s", new_node.name)
+            else:
+                # 4.2.2. If it is not, create a new node and add it to the graph.
+                new_node = _copy_node(old_node, old_value_to_new_value)
+                new_graph.append(new_node)
+                old_node_hash_to_new_node[hash_value] = new_node
+            # 4.3 Add the node outputs to the mapping.
+            old_value_to_new_value.update(dict(zip(old_node.outputs, new_node.outputs)))
```
gramalingam, line 95: We will need to handle the possibility of optional outputs as well: that is a minor complication. We need to encode the outputs-present as an extra entry in the tuple used above. If we have two nodes which look identical but one has an optional output and the other doesn't, the above code could break. So, better to not treat them as being same. So, `output1 = SomeOp(A, B)` and `outputX, outputY = SomeOp(A, B)` will be treated as being different. Unlikely to happen, but good to handle it.

onnxscript/ir/serde.py
```diff
@@ -511,7 +511,7 @@ def deserialize_model(proto: onnx.ModelProto) -> _core.Model:
         model_version=_get_field(proto, "model_version"),
         doc_string=_get_field(proto, "doc_string"),
         functions=functions,
-        meta_data_props=deserialize_metadata_props(proto.metadata_props),
+        metadata_props=deserialize_metadata_props(proto.metadata_props),
```
gramalingam, line 514: May be good to add a negative test as well: eg., two nodes which have identical inputs, but differ in one attribute value.

xadupre: You should also have a test checking that names in subgraphs are renamed (operator If, Scan, Loop).

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+                    tuple(sorted(attributes.items())),
+                )
+            )
+            # TODO(titaiwang): Subgraphs are not supported yet.
+            # TODO(titaiwang): Skip control flow nodes?
+            # 4.2. Check if the node is a common subexpression.
+            if hash_value in old_node_hash_to_new_node:
+                # 4.2.1. If it is, this node is already in the new graph, so
+                #         we don't need to create a new node.
+                modified = True
```
gramalingam, line 86: To convert this pass to an in-place one: we just need to modify the input graph in this one place: we just need to call `ir.convenience.replace_nodes_and_values` to delete the current node, and replace the output-values of current node with the output-values of reused-node. (If necessary, we can modify the replace_nodes_and_values utility to fit the needs ... eg., to control what information we want to retain from old and new values.)

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,163 @@
...
+                # 4.2.2. If it is not, create a new node and add it to the graph.
+                new_node = _copy_node(old_node, old_value_to_new_value)
+                new_graph.append(new_node)
+                old_node_hash_to_new_node[hash_value] = new_node
+            # 4.3 Add the node outputs to the mapping.
+            old_value_to_new_value.update(dict(zip(old_node.outputs, new_node.outputs)))
+        # 5. Create outputs in the new graph from the old graph.
+        for output in old_graph.outputs:
+            new_output = old_value_to_new_value[output]
+            new_graph.outputs.append(new_output)
```
gramalingam, line 99: We need to ensure that the *names* of outputs (and inputs) of the main model/graph are not changed.

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: Do you have an example op in mind? I am not sure if the following is the optional input case you have in mind:

```python
        def f(x)
            a, b = topk(x, k=1)
            c, _ = topk(x, k=1)
            return a, b, c
```

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: Done

onnxscript/ir/serde.py
titaiwangms: Added test_the_ops_with_the_same_inputs_but_different_attributes_are_not_csed

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: I can't think of an op that could have different number of outputs when the inputs and attributes are the same. Could you think of any?

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: Changed to one by one check with op_identifier, length of outputs (this is for output comparison if it's needed), input value id, and attributes.

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: Deleted

titaiwangms:
Changed to inplace pass. Subgraph is not yet tackled.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -15,149 +15,70 @@
...
-) -> ir.Node:
-    """Copy an IR node."""
-    new_inputs: list[ir.Value] = []
-    for original_input in original_node.inputs:
-        if original_input in old_value_to_new_value:
-            new_inputs.append(old_value_to_new_value[original_input])
+    # node to node identifier, length of outputs, inputs, and attributes
+    existing_node_info_to_the_node: dict[
+        tuple[
+            tuple[str, str, str],  # op_identifier
```
gramalingam, line 40: nit: can use `ir.OperatorIdentifier`

justinchuby:
I think it is ok to leave subgraphs untouched.

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: This should be taken care of by `ir.convenience.replace_nodes_and_values()`

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: I added len(node.outputs) to op_info (as unique identifier) but do not think of a test to guard it. however, it should work.

justinchuby (approved):
lgtm thanks! I would allow Rama to do another review

justinchuby:
Looks like the attribute value hashing part needs to be stronger

onnxscript/ir/passes/common/common_subexpression_elimination.py
justinchuby, line 1: Does the pass remove duplicated initializers as well? Should we create a separate pass or add an option?

titaiwangms: I suggest it's better we keep this in node-level to reduce the debugging complexity.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,107 @@
...
+    for node in graph:
+        # Skip control flow ops like Loop and If.
+        control_flow_op: bool = False
+        # Use equality to check if the node is a common subexpression.
+        attributes = {}
+        for k, v in node.attributes.items():
+            # TODO(exporter team): CSE subgraphs.
+            # NOTE: control flow ops like Loop and If won't be CSEd
+            # because attribute: graph won't match.
+            if isinstance(v, ir.Graph):
```
gramalingam, line 58: Change condition to `v.type in (ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS)` to handle the Graphs case also.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,107 @@
...
+        # Check if the node is a common subexpression.
+        if node_info in existing_node_info_to_the_node:
+            # If it is, this node has an existing node with the same
+            # operator, number of outputs, inputs, and attributes.
+            # We replace the node with the existing node.
+            modified = True
+            existing_node = existing_node_info_to_the_node[node_info]
+            ir.convenience.replace_nodes_and_values(
+                graph,
+                insertion_point=previous_node or node,
```
gramalingam, line 95: Does `insertion_point=node` not work? If it works, that would be simpler. I don't remember similar logic in the optimizer or rewriter though they also invoke the same replace_nodes_and_values.

onnxscript/ir/passes/common/common_subexpression_elimination_test.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Check if the models produce the same output
+        # with the same inputs
+        for idx, original_model_result in enumerate(original_model_results):
+            np.testing.assert_allclose(
+                original_model_result, result_results[idx], rtol=1e-5, atol=1e-5
+            )
+
+    def test_two_branches_with_the_same_operations_is_csed(self):
+        """Test if two branches with the same operations are CSEd.
```
gramalingam, line 69: nit: why do you say "branches"? Somehow it suggests an "If" node branches to me.

onnxscript/ir/passes/common/common_subexpression_elimination.py
gramalingam: Conceptually, they are similar, but the implementation is a bit different. I agree it can be added in a separate PR (no matter how we organize it, within this or outside).

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+        # Check if the node is a common subexpression.
+        if node_info in existing_node_info_to_the_node:
+            # If it is, this node has an existing node with the same
+            # operator, number of outputs, inputs, and attributes.
+            # We replace the node with the existing node.
+            modified = True
+            existing_node = existing_node_info_to_the_node[node_info]
+            ir.convenience.replace_nodes_and_values(
```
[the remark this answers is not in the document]
titaiwangms: This is the reason of https://github.com/microsoft/onnxscript/pull/2304#discussion_r2112436884. `replace_nodes_and_values` insert the new node after the insertion point, which if we use for loop, the node will show up next. That's why I have to find the previous node in this case.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,107 @@
...
+                graph,
+                insertion_point=previous_node or node,
+                old_nodes=[node],
+                new_nodes=[existing_node],
```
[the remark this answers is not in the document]
gramalingam: I think this is what we should do. We are reusing existing_node's outputs ... we don't need to add any new node. So, if we start with nodes [n1, n2, n3, n2duplicate], when we process n2duplicate, we will simply remove it and have [n1, n2, n3]. The main thing is we are going to replace uses of the outputs of n2duplicate with uses of the outputs of n2.

onnxscript/ir/passes/common/common_subexpression_elimination.py
gramalingam: But see the point down below ... we should have `new_nodes=[]` ... so, these two points interact with each other, I guess

onnxscript/ir/passes/common/common_subexpression_elimination.py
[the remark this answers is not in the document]
gramalingam: Right, but there should be no new node inserted.

onnxscript/ir/passes/common/common_subexpression_elimination.py
titaiwangms: Deleted

onnxscript/ir/_convenience/__init__.py
```diff
@@ -373,5 +373,6 @@ def replace_nodes_and_values(
             graph_or_function.outputs[idx] = replacement_mapping[graph_or_function_output]
 
     # insert new nodes after the index node
-    graph_or_function.insert_after(insertion_point, new_nodes)
+    if new_nodes:
```
titaiwangms, line 376: This is allowing empty new_nodes. I can create another function if you think it's out of the scope in this function. @justinchuby

onnxscript/ir/passes/common/common_subexpression_elimination.py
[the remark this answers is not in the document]
justinchuby: Yeah modifying while iterating is safe. Generally: We do need to insert before the current iterator as Ti-Tai mentioned (_if_ there is any nodes to be inserted) if we don't want the new nodes to participate in the iteration.

onnxscript/ir/_convenience/__init__.py
justinchuby: I want to reason about this function a bit more carefully because the renaming behavior currently is also no ideal. Can you copy the logic to the pass and use it privately for now? I intend to freeze this function before the onnx ir release.

gramalingam: That makes sense. We just want to replace uses of Y... by X..., and remove the node computing Y...

titaiwangms: Created another function without adding new node.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -88,16 +89,50 @@ def _eliminate_common_subexpression(graph: ir.Graph, modified: bool) -> bool:
...
+    """
+
+    for old_value, new_value in zip(remove_values, new_values):
+        # Propagate relevant info from old value to new value
+        # TODO(Rama): Perhaps this should be a separate utility function. Also, consider
+        # merging old and new type/shape info.
+        new_value.type = old_value.type
+        new_value.shape = old_value.shape
+        new_value.const_value = old_value.const_value
+        new_value.name = old_value.name
```
justinchuby, line 128: Is this needed? I think we can preserve the original info of the values.

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,138 @@
...
+
+    # Reconnect the users of the deleted values to use the new values
+    ir.convenience.replace_all_uses_with(remove_values, new_values)
+    # Update graph/function outputs if the node generates output
+    replacement_mapping = dict(zip(remove_values, new_values))
+    for idx, graph_or_function_output in enumerate(graph.outputs):
+        if graph_or_function_output in replacement_mapping:
+            graph.outputs[idx] = replacement_mapping[graph_or_function_output]
```
justinchuby, line 136:
```suggestion
    if any(v.is_graph_output() for v in remove_values):
	    replacement_mapping = dict(zip(remove_values, new_values))
	    for idx, graph_or_function_output in enumerate(graph.outputs):
	        if graph_or_function_output in replacement_mapping:
	            graph.outputs[idx] = replacement_mapping[graph_or_function_output]
```

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,138 @@
...
+        # TODO(Rama): Perhaps this should be a separate utility function. Also, consider
+        # merging old and new type/shape info.
+        new_value.type = old_value.type
+        new_value.shape = old_value.shape
+        new_value.const_value = old_value.const_value
+        new_value.name = old_value.name
+
+    # Reconnect the users of the deleted values to use the new values
+    ir.convenience.replace_all_uses_with(remove_values, new_values)
+    # Update graph/function outputs if the node generates output
```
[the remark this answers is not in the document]
titaiwangms: I see. This was what @gramalingam suggested.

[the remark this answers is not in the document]
titaiwangms: Can we just use the same old_value name?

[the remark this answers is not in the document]
justinchuby: I think so.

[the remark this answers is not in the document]
titaiwangms: But if both new and old values are graph outputs, the output names will be changed anyway. Do we want to cover this kind of case?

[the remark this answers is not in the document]
justinchuby: So the old value can have an identity and retain the old names, the new value will not change. Does that work?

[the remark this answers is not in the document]
titaiwangms: OK now when the value1 (new_value) is also graph output, we create an Identity node to take value1 (new_value), and output another value, which uses value2's name (old_value).

But if value1 (new_value) is not graph output, we just update it to use value2's name (old_value).

onnxscript/ir/passes/common/common_subexpression_elimination_test.py
```diff
@@ -0,0 +1,303 @@
...
+class TestCommonSubexpressionEliminationPass(unittest.TestCase):
+    def check_graph(self, model: ir.Model, inputs: list[ir.Value], delta_nodes: list[int]):
+        """Check if the model applied the CSE pass correctly.
+
+        Args:
+            model (ir.Model): The model to check.
+            inputs (list[ir.Value]): The inputs to the model.
+            delta_nodes (list[int]): The expected change in the number of nodes in the model.
+                                     The length of this list should match the number of graphs
+                                     in the model. (to support subgraphs in the future)
```
justinchuby, line 24:
```suggestion
            model: The model to check.
            inputs: The inputs to the model.
            delta_nodes: The expected change in the number of nodes in the model.
                The length of this list should match the number of graphs
                in the model. (to support subgraphs in the future)
```

onnxscript/ir/passes/common/common_subexpression_elimination.py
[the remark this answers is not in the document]
justinchuby: That's a good idea, thanks!

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,146 @@
...
+                    # If the new value is also a graph output, we need to
+                    # create a Identity node to preserve the remove_value.
+                    identity_node = ir.node(
+                        "Identity",
+                        inputs=[new_value],
+                        num_outputs=1,
+                    )
+                    # reuse the name of the graph output
+                    identity_node.outputs[0].name = graph_output.name
+                    graph.outputs[idx] = identity_node.outputs[0]
```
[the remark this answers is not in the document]
titaiwangms: Done

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,146 @@
...
+                    graph.insert_before(
+                        remove_node,
+                        identity_node,
+                    )
+                else:
+                    new_value.name = graph_output.name
```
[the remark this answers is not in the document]
titaiwangms: Added

onnxscript/ir/passes/common/common_subexpression_elimination.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    # Reconnect the users of the deleted values to use the new values
+    ir.convenience.replace_all_uses_with(remove_values, new_values)
+    # Update graph/function outputs if the node generates output
+    if any(remove_value.is_graph_output() for remove_value in remove_values):
+        replacement_mapping = dict(zip(remove_values, new_values))
+        for idx, graph_output in enumerate(graph.outputs):
+            if graph_output in replacement_mapping:
+                new_value = replacement_mapping[graph_output]
+                if new_value.is_graph_output():
```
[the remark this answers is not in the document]
titaiwangms: Is there a kind of case that we would cse graph inputs? In this PR, it's deleting nodes and affecting their outputs.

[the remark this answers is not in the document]
titaiwangms: I see what you mean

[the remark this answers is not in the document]
titaiwangms: I will add that in https://github.com/onnx/ir-py/pull/36

PR #2307, opened 2025-05-14
[IR] introduce slice support on graph

### LINKED ISSUE #2302: [IR] Unsupported slice and ellipsis indexes to iter a graph
# Code to reproduce the error:

```python
import onnx
from onnxscript import ir

model = onnx.parser.parse_model(
"""<ir_version: 10, opset_import: [ "" : 17]>
     agraph (float[N] x) => (float[N] z) {
         y = Sigmoid(x)
         z = Mul(y, y)  
     }""")
ir_model = ir.from_proto(model)
ir_model.graph[0:]
```

# Error logs
Errors when executing `ir_model.graph[0:]`:

```
  File "onnxscript\ir\_core.py", line 2232, in __getitem__
    return self._nodes[index]
           ~~~~~~~~~~~^^^^^^^
  File "onnxscript\ir\_linked_list.py", line 144, in __getitem__
    if index >= self._length or index < -self._length:
       ^^^^^^^^^^^^^^^^^^^^^
TypeError: '>=' not supported between instances of 'slice' and 'int'
```

Johansmm:
Introduce slice support
Close #2302

### DIFF
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -2282,7 +2282,12 @@ class Graph(_protocols.GraphProtocol, Sequence[Node], _display.PrettyPrintable):
     def opset_imports(self) -> dict[str, int]:
         return self._opset_imports
 
-    def __getitem__(self, index: int) -> Node:
+    @typing.overload
+    def __getitem__(self, index: int) -> Node: ...
+    @typing.overload
+    def __getitem__(self, index: slice) -> Sequence[Node]: ...
+
+    def __getitem__(self, index):
         return self._nodes[index]
 
     def __len__(self) -> int:
@@ -2712,7 +2717,12 @@ class GraphView(Sequence[Node], _display.PrettyPrintable):
         self._metadata_props: dict[str, str] | None = metadata_props
         self._nodes: tuple[Node, ...] = tuple(nodes)
 
-    def __getitem__(self, index: int) -> Node:
+    @typing.overload
+    def __getitem__(self, index: int) -> Node: ...
+    @typing.overload
+    def __getitem__(self, index: slice) -> Sequence[Node]: ...
+
+    def __getitem__(self, index):
         return self._nodes[index]
 
     def __len__(self) -> int:
@@ -2961,7 +2971,12 @@ class Function(_protocols.FunctionProtocol, Sequence[Node], _display.PrettyPrint
     def attributes(self) -> OrderedDict[str, Attr]:
         return self._attributes
 
-    def __getitem__(self, index: int) -> Node:
+    @typing.overload
+    def __getitem__(self, index: int) -> Node: ...
+    @typing.overload
+    def __getitem__(self, index: slice) -> Sequence[Node]: ...
+
+    def __getitem__(self, index):
         return self._graph.__getitem__(index)
 
     def __len__(self) -> int:
--- a/onnxscript/ir/_linked_list.py
+++ b/onnxscript/ir/_linked_list.py
@@ -4,7 +4,7 @@
 
 from __future__ import annotations
 
-from typing import Generic, Iterable, Iterator, Sequence, TypeVar
+from typing import Generic, Iterable, Iterator, Sequence, TypeVar, overload
 
 T = TypeVar("T")
 
@@ -136,11 +136,18 @@ class DoublyLinkedSet(Sequence[T], Generic[T]):
         )
         return self._length
 
-    def __getitem__(self, index: int) -> T:
+    @overload
+    def __getitem__(self, index: int) -> T: ...
+    @overload
+    def __getitem__(self, index: slice) -> Sequence[T]: ...
+
+    def __getitem__(self, index):
         """Get the node at the given index.
 
         Complexity is O(n).
         """
+        if isinstance(index, slice):
+            return tuple(self)[index]
         if index >= self._length or index < -self._length:
             raise IndexError(
                 f"Index out of range: {index} not in range [-{self._length}, {self._length})"
--- a/onnxscript/ir/_linked_list_test.py
+++ b/onnxscript/ir/_linked_list_test.py
@@ -373,6 +373,15 @@ class DoublyLinkedSetTest(unittest.TestCase):
         self.assertEqual(len(other_linked_list), 1)
         self.assertEqual([elem.value for elem in other_linked_list], [42])
 
+    @parameterized.parameterized.expand(
+        [(s, t, p) for s in [-2, 0, 2, 3] for t in [2, -1, -2] for p in [-3, -1, 1, 2]]
+    )
+    def test_get_item_slice(self, start, stop, step):
+        elems = [_TestElement(i) for i in range(5)]
+        linked_list = _linked_list.DoublyLinkedSet(elems)
+        self.assertEqual(len(linked_list), 5)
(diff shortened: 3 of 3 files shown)


justinchuby (approved):
Thanks!

justinchuby:
Could you also update the following text to match the implementation?

```
        although accessing nodes at either end of the set is O(1). I.e.
        ``linked_set[0]`` and ``linked_set[-1]`` are O(1).
```
in the class docstring above.

also https://github.com/Johansmm/onnxscript/blob/862de43e60f51ee4833104a254da9d8db80420a6/onnxscript/ir/_core.py#L2319?

There are two type annotations to update: https://github.com/Johansmm/onnxscript/blob/862de43e60f51ee4833104a254da9d8db80420a6/onnxscript/ir/_core.py#L2715 and https://github.com/Johansmm/onnxscript/blob/862de43e60f51ee4833104a254da9d8db80420a6/onnxscript/ir/_core.py#L2285

onnxscript/ir/_linked_list.py
```diff
@@ -136,27 +136,12 @@ def __len__(self) -> int:
         )
         return self._length
 
-    def __getitem__(self, index: int) -> T:
+    def __getitem__(self, index: int | slice) -> T:
```
justinchuby, line 139: Could you create overloads so the type annotations is correct? Two overloads:

`def __getitem__(self, index: int) -> T:`

and `def __getitem__(self, index: slice) -> tuple[T, ...]:`

onnxscript/ir/_linked_list_test.py
```diff
@@ -60,6 +60,9 @@ def test_append_multiple_elements(self):
         self.assertEqual(linked_list[-3], elems[0])
         self.assertEqual(list(linked_list), elems)
         self.assertEqual(list(reversed(linked_list)), list(reversed(elems)))
+        self.assertEqual(list(linked_list[1:2]), elems[1:2])
```
justinchuby, line 63: These tests are for testing different methods. Please create a separate test method `test_get_item_slice` and add the logic there.

justinchuby (changes requested):
Some comments

Johansmm:
Last push force with suggestions

justinchuby:
Last thing I wanted to make sure is that accessing nodes on either ends (graph.node[0], graph.node[-1]) is not much slower (better if it turns out to be faster) than before.

Johansmm:
@justinchuby, the new complexity is O(n) for any case. 
I made the following python script to evaluate this:
```python
import time

from onnxscript.ir import _linked_list


class _TestElement:
    def __init__(self, value):
        self.value = value

    def __repr__(self) -> str:
        return f"_TestElement({self.value})"


N = 100000
linked_list = _linked_list.DoublyLinkedSet()
elems = [_TestElement(i) for i in range(N)]
linked_list.extend(elems)

time_list_1 = []
idx = -1
for _ in range(1000):
    t = time.time()
    x = tuple(linked_list)[idx]
    t = time.time() - t
    time_list_1.append(t)

time_list_2 = []
for _ in range(1000):
    t = time.time()
    for i in reversed(linked_list):
        x = i
        break
    t = time.time() - t
    time_list_2.append(t)


print(sum(time_list_1) / 1000)
print(sum(time_list_2) / 1000)
```

Results changing `N`: 
| N      | time previous          | time now               |
|--------|------------------------|------------------------|
| 100    | 0.0                    | 2.5733470916748046e-05 |
| 1000    | 0.0                    | 0.00013534140586853028 |
| 10000  | 0.0                    | 0.0014723334312438964  |
| 100000 | 1.3112068176269532e-05 | 0.015811803340911864   |

Therefore:
1. do you want me to reverse the code as long as I handle the integers separately?
2. I leave the code as it is, since it is not common to have models with more than 1000 nodes ?

justinchuby:
It is in fact quite common to have models with more than 1000 nodes in the case of LLMs. I would prefer it to run faster. Could you do (1) as you suggested, and revert the comment changes? Thanks a lot! I think it makes sense to handle slices in an if branch.

justinchuby:
Thank you! The changes look very reasonable. Were you able to run the benchmark again with some examples?

Johansmm:
@justinchuby can you explain what benchmark you are talking about? If you are referring to the IR unit tests, I ran them and they passed without any problem.

justinchuby:
I am referring to https://github.com/microsoft/onnxscript/pull/2307#issuecomment-2885241440 this script in your comment

justinchuby (approved):

onnxscript/ir/_core.py
```diff
@@ -2961,7 +2971,12 @@ def outputs(self) -> MutableSequence[Value]:
     def attributes(self) -> OrderedDict[str, Attr]:
         return self._attributes
 
-    def __getitem__(self, index: int) -> Node:
+    @typing.overload
+    def __getitem__(self, index: int) -> Node: ...
+    @typing.overload
+    def __getitem__(self, index: slice) -> tuple[Node, ...]: ...
```
justinchuby, line 2977: nit

```suggestion
    def __getitem__(self, index: slice) -> Sequence[Node]: ...
```

here and elsewhere, thanks!

onnxscript/ir/_linked_list.py
```diff
@@ -136,27 +136,54 @@ def __len__(self) -> int:
...
+        # Return a single element if index is an integer
+        if isinstance(index, int):
+            return item
+
+        # Return a list if index is a slice
+        items = [item]
+        for i in range(1, abs(stop - start)):
+            item = next(iterator)
+            if i % step == 0:
+                items.append(item)
```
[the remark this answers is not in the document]
justinchuby: This is quite expensive right?

onnxscript/ir/_linked_list.py
```diff
@@ -136,27 +136,54 @@ def __len__(self) -> int:
...
                 item = next(iterator)
-        return item
+
+        # Return a single element if index is an integer
+        if isinstance(index, int):
+            return item
+
+        # Return a list if index is a slice
+        items = [item]
+        for i in range(1, abs(stop - start)):
```
justinchuby, line 182: hmm I think this is a little convoluted. Is this much faster than tuple()[slice]?

justinchuby (commented):
Sorry after taking another look I wonder if the added complexity if worth it (line 160-162, 166-167). Would it be a good idea to (1) keep the original logic for int index, and (2) use tuple for slices?

onnxscript/ir/_linked_list.py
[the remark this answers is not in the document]
Johansmm: yes: this is more expensive than my implementation

Johansmm:
@justinchuby, I see that you don't agree with my algorithm, so I have gone back to the initial solution. Note that this one is a bit more expensive. 
Do not hesitate to ask me for anything else if necessary 👍

justinchuby (approved):
Thank you! I think it’s a good trade off for simplicity

justinchuby:
Hi @Johansmm, we are in the process of migrating the IR component to https://github.com/onnx/ir-py so that it becomes an official onnx project. Would you like to help us duplicate your PR to the new repository as well? We are looking to create an initial release within about a week or less.

Johansmm:
@justinchuby done: https://github.com/onnx/ir-py/pull/25

Issue #2301, opened 2025-05-13
Optimizer Suggestions for Common Graph Fusions


AyoubMDL:
I've been using ONNXScript extensively lately and really appreciate how powerful and flexible it is. As part of that experience, I and @Johansmm would like to suggest a few optimization patterns that could improve performance and simplify common model graphs. These are generally applicable across many models:
 
1. Fold `BatchNormalization` into preceding nodes (`Conv`, `ConvTranspose`, `Gemm`) https://github.com/microsoft/onnxscript/pull/2312
2. Fuse `MatMul + Add` into `Gemm` https://github.com/microsoft/onnxscript/pull/2356
3. Fuse `Min + Max` into `Clip` https://github.com/microsoft/onnxscript/pull/2500
4. Eliminate redundant `Reshape` / `Flatten` nodes https://github.com/microsoft/onnxscript/pull/2518
5. Fold `Pad` into `Conv` / `ConvInteger` https://github.com/microsoft/onnxscript/pull/2363
6. Fuse successive `Clip` / `ReLU` nodes https://github.com/microsoft/onnxscript/pull/2410
 
We're happy to contribute if any of these are a good fit. Feel free to add other suggestions or let us know if some of these are out of scope.

justinchuby:
Thanks! All of the above sounds good. PRs welcomed. For (2) we want it to be optional because some runtimes prefer one way over another. You may add the rules in https://github.com/microsoft/onnxscript/tree/main/onnxscript/rewriter and use the existing ones as reference.

AyoubMDL:
PR: https://github.com/microsoft/onnxscript/pull/2553

AyoubMDL:
All tasks completed. Closing this.

PR #2312, opened 2025-05-18
Rewriter: Fold Batchnorm nodes

AyoubMDL:
Fuses `BatchNormalization` nodes into the following nodes (`Conv`, `ConvTranspose`, `Gemm`)
(https://github.com/microsoft/onnxscript/issues/2301)

### DIFF
--- a/onnxscript/rewriter/fuse_batchnorm.py
+++ b/onnxscript/rewriter/fuse_batchnorm.py
@@ -1,0 +1,183 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Fuses BatchNormalization nodes into preceding nodes. Supported fusion patterns:
+- BatchNormalization + Conv         -> Conv
+- BatchNormalization + ConvTranpose -> ConvTranpose
+- BatchNormalization + Gemm         -> Gemm
+
+Approach:
+    Given an inbound operation output: Y = W * X + B
+    And a BatchNormalization outputs: Y_BN = (gamma * (Y - μ) / std) + β, where std = sqrt(var + eps)
+
+    The fusion updates the inbound weights as follows:
+        - W_fused = W * (gamma / std)
+        - B_fused = (B - μ) * (gamma / std) + β
+"""
+
+from abc import ABC, abstractmethod
+
+import numpy as np
+
+from onnxscript import ir
+from onnxscript.rewriter import pattern as orp
+
+
+class FuseBatchNormBase(orp.RewriteRuleClassBase, ABC):
+    """Interface for BatchNormalization nodes fusion."""
+
+    def __init__(
+        self,
+        op_type: str,
+        name: str | None = None,
+        remove_nodes: bool = True,
+        as_function: bool = False,
+    ) -> None:
+        super().__init__(name=name, remove_nodes=remove_nodes, as_function=as_function)
+        self.op_type = op_type
+
+    @abstractmethod
+    def get_filters_axis(self, attributes) -> int:
+        """Return the axis along which BatchNorm scale should be broadcasted."""
+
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/fold_batchnorm.py
```diff
@@ -0,0 +1,102 @@
...
+def replacement_pattern(op_type: str):
+    def _pattern(op, x: ir.Value, inbound_out: ir.Value, batchnorm: ir.Value):
+        batchnorm_node = batchnorm.producer()
+        # Get BatchNorm parameters
+        gamma, beta, input_mean, input_var = [inp.const_value.numpy() for inp
+                                              in batchnorm_node.inputs[1:]]
+
+        # 1e-5 is the default value for epsilon according to
+        # https://onnx.ai/onnx/operators/onnx__BatchNormalization.html#attributes
+        eps = batchnorm_node.attributes.get("epsilon", 1e-5)
```
justinchuby, line 50: eps is of type ir.Attr. You can use eps.as_float() to get the float value.

justinchuby (commented):
Thank you! Could you follow https://github.com/microsoft/onnxscript/blob/2ae13becc151830c84e952dcda2aed5275ca940b/onnxscript/rewriter/llama_rule_sets.py#L35-L48 to implement the rules as an `pattern.RewriteRuleClassBase`? You may create additional subclasses for the different variants for the rules.

You may also follow the readme (https://github.com/microsoft/onnxscript#coding-style) to format the code with lintrunner.

onnxscript/rewriter/fold_batchnorm.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
```
gramalingam, line 2: Can you provide a short description of what this optimization is doing? Thanks!

AyoubMDL:
> 1. Could you follow https://github.com/microsoft/onnxscript/blob/2ae13becc151830c84e952dcda2aed5275ca940b/onnxscript/rewriter/llama_rule_sets.py#L35-L48
>     to implement the rules as an `pattern.RewriteRuleClassBase`? You may create additional subclasses for the different variants for the rules.

@justinchuby Added the suggestion—please let me know if it aligns with what you had in mind.

justinchuby:
The rules look good to me, although I am not sure it should be part of the default rule set? @gramalingam do you have any recommendations? Should we enable multiple levels of optimization, or is this fine?

onnxscript/rewriter/fuse_batchnorm.py
```diff
@@ -0,0 +1,183 @@
...
+        op_type: str,
+        name: str | None = None,
+        remove_nodes: bool = True,
+        as_function: bool = False,
+    ) -> None:
+        super().__init__(name=name, remove_nodes=remove_nodes, as_function=as_function)
+        self.op_type = op_type
+
+    @abstractmethod
+    def get_filters_axis(self, attributes) -> int:
```
justinchuby, line 39:
```suggestion
    def get_filters_axis(self, attributes: Mapping[str, ir.Attr]) -> int:
```

onnxscript/rewriter/fuse_batchnorm.py
```diff
@@ -0,0 +1,183 @@
...
+            self.op_type,
+            inputs=[
+                x,
+                op.initializer(fused_weights, name=inbound_node.inputs[1].name),
+                op.initializer(fused_bias, name=bias_name),
+            ],
+            attributes=inbound_node.attributes,
+        )
+
+    def check(self, context, x, inbound_out, batchnorm_out) -> orp.MatchResult:
```
justinchuby, line 95:
```suggestion
    def check(self, context, x, inbound_out: ir.Value, batchnorm_out: ir.Value) -> orp.MatchResult:
```

justinchuby (approved):
Thanks a lot!

PR #2317, opened 2025-05-20
[rewrite] Specify transpose op when TransposeMatMul checks the pattern

titaiwangms:
Found in Whisper that `list((x if self._pos == 1 else y).uses())[0][0]` could be not op.Transpose. This PR specifies transpose node so that it does not crash with keyerror on 'perm'.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
+++ b/onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
@@ -45,6 +45,7 @@ class FusedMatMulDiv2(orp.RewriteRuleClassBase):
     def rewrite(self, op, x, y, cst):
         value = cst.const_value.numpy()
         c = float(value[0] if value.shape == (1,) else value)
+        assert x.consumers()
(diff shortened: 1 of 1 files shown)


gramalingam:
Hi @titaiwangms : the rules are somewhat broken. I think a more significant rewrite would be better. Can we remove the matmul fusion rules from the default rule set, so that we get something working first?

The rules should be rewritten so that they don't assume anything about how many users() there are. That's much safer and can be done easily.

This may be a good exercise for Bhagirath, so I can ask him to look into it. What do you think?

titaiwangms:
> Hi @titaiwangms : the rules are somewhat broken. I think a more significant rewrite would be better. Can we remove the matmul fusion rules from the default rule set, so that we get something working first?
> 
> The rules should be rewritten so that they don't assume anything about how many users() there are. That's much safer and can be done easily.
> 
> This may be a good exercise for Bhagirath, so I can ask him to look into it. What do you think?

How do you tell it's broken?

titaiwangms:
Replaced by #2331

PR #2322, opened 2025-05-21
Fix fused matmul check/rewrite functions

bmehta001:
- Patterns now declare `_outputs` filters to bind intermediate values
- Rewrites use `fused.producer()` or `transposed.producer()` instead of scanning `.uses()` which may pick up other nodes that use `x` or `y`
- For ir.Value parameters, use a default of None in case the parameter does not exist
- Attribute extraction updated to use as_float() / as_ints() for type safety
- Since rewrite/check functions will have all `ir.Value` variables passed in, but they may not use all variables, use `**_` to read in unused variables
- Updated docstrings from "by" to "with" for clarity and changed `fusedmatmul` to `matmul` where appropriate

All tests pass

### DIFF
--- a/onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
+++ b/onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
@@ -5,10 +5,11 @@ from __future__ import annotations
 from typing import ClassVar
 
 import onnxscript.rewriter.pattern as orp
+from onnxscript import ir
 
 
 class FusedMatMulDiv1(orp.RewriteRuleClassBase):
-    """Replaces ``MatMul + Div`` by FusedMatMul."""
+    """Replaces ``MatMul + Div`` with FusedMatMul."""
 
     def pattern(self, op, x, y, cst):
         return op.Div(op.MatMul(x, y), cst)
@@ -29,12 +30,12 @@ class FusedMatMulDiv1(orp.RewriteRuleClassBase):
 
 
 class FusedMatMulDiv2(orp.RewriteRuleClassBase):
-    """Replaces ``FusedMatMul + Div`` by FusedMatMul."""
+    """Replaces ``FusedMatMul + Div`` with FusedMatMul."""
 
     def pattern(self, op, x, y, cst):
-        return op.Div(op.FusedMatMul(x, y, _domain="com.microsoft"), cst)
+        return op.Div(op.FusedMatMul(x, y, _domain="com.microsoft", _outputs=["fused"]), cst)
 
-    def check(self, context, x, y, cst) -> orp.MatchResult:
+    def check(self, context, x, y, cst, fused: ir.Value) -> orp.MatchResult:
         check_result = orp.MatchResult()
         if cst.const_value is None:
             return check_result.fail("Divisor is not a constant value.")
@@ -42,14 +43,13 @@ class FusedMatMulDiv2(orp.RewriteRuleClassBase):
             return check_result.fail("Divisor is not a scalar value.")
         return check_result
 
-    def rewrite(self, op, x, y, cst):
+    def rewrite(self, op, x, y, cst, fused: ir.Value):
         value = cst.const_value.numpy()
         c = float(value[0] if value.shape == (1,) else value)
-        node = list(x.uses())[0][0]  # noqa: RUF015
+        node: ir.Node = fused.producer()  # type: ignore[assignment]
 
         kwargs = {}
-        alpha = node.attributes.get("alpha", None)
-        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
+        kwargs["alpha"] = node.attributes["alpha"].as_float() / c
         for name in ["transA", "transB", "transBatchA", "transBatchB"]:
             att = node.attributes.get(name)
             if att:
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -82,42 +86,42 @@ def rewrite(self, op, x, y):
 
 
 class TransposeMatMul1(_TransposeMatMulBase):
-    """Replaces ``Transpose + (Fused)MatMul`` by FusedMatMul."""
+    """Replaces ``Transpose + MatMul`` with FusedMatMul."""
 
     def pattern(self, op, x, y):
-        return op.MatMul(op.Transpose(x), y)
+        return op.MatMul(op.Transpose(x), y, _outputs=["matmul"])
```
gramalingam, line 92: I think that the `perm` attribute is from the transpose node. So, suggest using `_outputs=["transposed"]` in `op.Transpose(x, ...)`.

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -60,17 +60,21 @@ def rewrite(self, op, x, y, cst):
...
     _pos: ClassVar = 1
 
-    def check(self, context, x, y) -> orp.MatchResult:
+    def check(self, context, x, y, matmul) -> orp.MatchResult:
         check_result = orp.MatchResult()
-        perm = list((x if self._pos == 1 else y).uses())[0][0].attributes["perm"].value  # noqa: RUF015
+        node = matmul.producer()
+        if "perm" not in node.attributes:
+            return check_result.fail("Transpose node does not have 'perm' attribute.")
+        perm = node.attributes["perm"].value
```
justinchuby, line 68:
```suggestion
        perm = node.attributes["perm"].as_ints()
```

for better typing

bmehta001: Thanks, fixed here + also added `.as_float()` to the line defining `alpha`

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
bmehta001: Thanks, fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -42,10 +42,10 @@ def check(self, context, x, y, cst) -> orp.MatchResult:
             return check_result.fail("Divisor is not a scalar value.")
         return check_result
 
-    def rewrite(self, op, x, y, cst):
+    def rewrite(self, op, x, y, cst, fused):
```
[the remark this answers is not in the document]
bmehta001: Will add type hinting to provide an idea of what the variable is

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +30,137 @@ def rewrite(self, op, x, y, cst):
...
-        alpha = node.attributes.get("alpha", None)
-        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
+        alpha = node.attributes.get("alpha", None)  # type: ignore[union-attr]
+        kwargs["alpha"] = alpha.as_float() / c if alpha else 1.0 / c  # type: ignore[union-attr]
         for name in ["transA", "transB", "transBatchA", "transBatchB"]:
-            att = node.attributes.get(name)
+            att = node.attributes.get(name)  # type: ignore[union-attr]
             if att:
-                kwargs[name] = att.value
+                kwargs[name] = att.value  # type: ignore[union-attr]
```
justinchuby, line 57: I created https://github.com/microsoft/onnxscript/pull/2328 so you don't need to use all the type ignores hopefully.

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +30,137 @@ def rewrite(self, op, x, y, cst):
...
         value = cst.const_value.numpy()
         c = float(value[0] if value.shape == (1,) else value)
-        node = list(x.uses())[0][0]  # noqa: RUF015
+        node = fused.producer()
 
         kwargs = {}
-        alpha = node.attributes.get("alpha", None)
-        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
+        alpha = node.attributes.get("alpha", None)  # type: ignore[union-attr]
+        kwargs["alpha"] = alpha.as_float() / c if alpha else 1.0 / c  # type: ignore[union-attr]
```
justinchuby, line 53: Can alpha ever be None? What happens when it is?

bmehta001: Oh, since `alpha` is an attribute, it should be defined now that we know we are definitely accessing the `fused` output. I will fix this to eliminate the unnecessary check

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
bmehta001: Actually, I could rebase now, and I was able to simplify some annotations (though I still had to add others to indicate that `.producer` would definitely not be None)

justinchuby: for the none check You may use

```
producer = x.producer()
assert producer is not None
```

bmehta001: Fixed

PR #2325, opened 2025-05-22
Optimize causal mask shape

gramalingam:
The generation of the causal mask's shape (produced by the translation of scalar_dot_product_attention) interferes with the subsequent fusion optimizations (because it makes use of the shape of the intermediate matmul value). 

This PR introduces a very specific fusion/rewrite to eliminate this redundant computation of the "sequence length" dimension.


justinchuby (approved):
So constant folding doesn’t get this properly?

gramalingam:
> So constant folding doesn’t get this properly?

Good question (though it is the "optimizer", though we call it constant-folding, since it goes beyond pure constant folding). I think not. It does the necessary analysis for shape-inference. May be worth checking. I thought we might need a more generic optimization pass, but perhaps not.

PR #2329, opened 2025-05-22
[rewriter] Enable as_function in sdpa

titaiwangms:


justinchuby (approved):

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -18,7 +18,10 @@ def __init__(
         use_mul: bool,
         has_3d_query: bool,
     ):
-        super().__init__(name=name)
+        # SDPA helps to fuse a large set of patterns to attention family
+        # However, it could fail and stay with op.SDPA, so we need to
+        # make sure the node can be kept as a local function
+        super().__init__(name=name, as_function=True)
```
gramalingam, line 24: Hold on. This is more complicated than I thought. Because SDPA has extra nodes in the replacement. The CI will fail because of this anyway.

PR #2331, opened 2025-05-22
Fix fused matmul check/rewrite functions

bmehta001:
- Patterns now declare _outputs filters to bind intermediate values
- Rewrites use fused.producer() or transposed.producer() instead of scanning .uses() which may pick up other nodes that use x or y
- For ir.Value parameters, use a default of None in case the parameter does not exist
- Attribute extraction updated to use as_float() / as_ints() for type safety
- Since rewrite/check functions will have all ir.Value variables passed in, but they may not use all variables, use **_ to read in unused variables
- Updated docstrings from "by" to "with" for clarity and changed fusedmatmul to matmul where appropriate
- Add more patterns:
1. If Transpose.perm indices are [1:-1, 0, -1] and transBatchA is 0, we can change transBatchA to 1
2. If Transpose.perm indices are [-2, 0:-2, -1] and transBatchA is 1, we can change transBatchA to 0.
3. If Transpose.perm indices are [1:, 0] and transBatchA is 0, we can change transBatchA to 1 and transA to 1- transA
4. If Transpose.perm indices are [-1, 0:-1] and transBatchA is 1, we can change transBatchA to 0 and transA to 1- transA
5. If Transpose.perm indices are [-1, 1:-1, 0] and transBatchA is 1, we can change transA to 1- transA
6. And also do all of 1-5 for transBatchB
- Add tests to make sure above changes work for `.producer()` and the added conditions related to `transBatch`
All tests pass


onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -60,91 +60,106 @@ def rewrite(self, op, x, y, cst):
...
-        node = list((x if self._pos == 2 else y).uses())[0][0]  # noqa: RUF015
+    def rewrite(self, op, x, y, fused: ir.Value | None = None, **_):
         kwargs = {}
-        for name in ["alpha", "transA", "transB", "transBatchA", "transBatchB"]:
-            att = node.attributes.get(name)
-            if att:
-                kwargs[name] = att.value
+        if fused:
+            node: ir.Node = fused.producer()  # type: ignore[assignment]
+            for name in ["alpha", "transA", "transB", "transBatchA", "transBatchB"]:
```
gramalingam, line 77: A couple of questions:
* Minor suggestion: I think we can just copy the node.attributes dictionary in one line?
* Correctness question: what interaction is there between the transBatchA and transA attribute? Do we need to check for any extra condition in the check method? Eg., perhaps the node should not have a transBatchA attribute already if we plan to introduce a transA attribute?

bmehta001: Fixed copying the attributes dictionary in 1 line - all attributes should always be defined in the node we are copying from, so this should be correct

I will check how transA and transBatchA can interact

bmehta001: @gramalingam, I looked into how transBatchA/transBatchB is being used [here](https://github.com/microsoft/onnxruntime/blob/f8c13937fa6929c2f574e06be45727c33bdde36a/onnxruntime/core/providers/cpu/math/matmul_helper.h#L72-L158). transBatchA gets called before transA, so as you said, if we have a Transpose that would have caused us to flip the value of `transA`, then we should ensure that `transBatchA = 0`.

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +30,312 @@
...
+        check_result = orp.MatchResult()
+        node = transposed.producer()
+        assert node is not None, "Transpose node should not be None"
+        fused_node = fused.producer()
+        assert fused_node is not None, "FusedMatMul node should not be None"
+        perm = node.attributes["perm"].as_ints()
+        # Check that last two dimensions are swapped
+        list_perm = list(range(len(perm)))
+        expected_perm0 = list_perm[1:-1] + [list_perm[0], list_perm[-1]]
+        expected_perm1 = [list_perm[-2]] + list_perm[0:-2] + [list_perm[-1]]
```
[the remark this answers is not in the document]
bmehta001: Fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +30,312 @@
...
+    ) -> orp.MatchResult:
+        check_result = orp.MatchResult()
+        node = transposed.producer()
+        assert node is not None, "Transpose node should not be None"
+        fused_node = fused.producer()
+        assert fused_node is not None, "FusedMatMul node should not be None"
+        perm = node.attributes["perm"].as_ints()
+        # Check that last two dimensions are swapped
+        list_perm = list(range(len(perm)))
+        expected_perm0 = list_perm[1:] + [list_perm[0]]
```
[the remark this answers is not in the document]
bmehta001: Fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
```diff
@@ -29,8 +29,23 @@
...
+            assert len(A.shape) >= 3 and len(B.shape) >= 3, (
+                f"Batch dimensions must be at least 3 for A: {A.shape} and B: {B.shape}"
+            )
+            assert len(A.shape) == len(B.shape), (
+                f"Batch dimensions must match for A: {A.shape} and B: {B.shape}"
+            )
+        if transBatchA:
+            perm = list(range(len(A.shape)))
+            dim = len(perm)
+            perm = perm[1 : dim - 1] + [perm[0], perm[dim - 1]]
```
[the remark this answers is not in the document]
bmehta001: Fixed

justinchuby:
Thanks - also feel free to point out things that are unintuitive or make suggestions as you use the onnx ir and rewriter apis. Would love to hear about them!

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +30,312 @@ def rewrite(self, op, x, y, cst):
...
-        alpha = node.attributes.get("alpha", None)
-        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
-        for name in ["transA", "transB", "transBatchA", "transBatchB"]:
-            att = node.attributes.get(name)
-            if att:
-                kwargs[name] = att.value
+        node = fused.producer()
+        assert node is not None, "FusedMatMul node should not be None"
+        kwargs = {key: val.value for key, val in node.attributes.items()}
+        kwargs["alpha"] = node.attributes["alpha"].as_float() / c
```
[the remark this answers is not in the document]
bmehta001: Now using functions for this

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +43,299 @@ def rewrite(self, op, x, y, cst):
...
+            transBatchProperty = "transBatchB"
+        transBatch = fused_node.attributes.get(transBatchProperty, 0).as_int()  # type: ignore[union-attr]
+        if (expected_perm0 == perm and transBatch == 0) or (
+            expected_perm1 == perm and transBatch == 1
+        ):
+            return check_result
+        return check_result.fail("Permutation values for Transpose are not correct.")
+
+
+class TransposeFusedMatMulWithFlippedBatch2(_TransposeFusedMatMulBaseWithBatch):
```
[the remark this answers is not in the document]
bmehta001: New pattern might fix this

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +43,247 @@ def rewrite(self, op, x, y, cst):
...
         expected_perm[-2], expected_perm[-1] = expected_perm[-1], expected_perm[-2]
         if perm != expected_perm:
             return check_result.fail("Permutation values for Transpose are not correct.")
+        if fused:
+            fused_node = get_node(fused, "FusedMatMul")
+            if fused_node.attributes.get("transBatchA", 0).as_int() == 1 and self._pos == 1:  # type: ignore[union-attr]
+                return check_result.fail(
+                    "FusedMatMul with transBatchA cannot be used with Transpose(A)."
+                )
+            if fused_node.attributes.get("transBatchB", 0).as_int() == 1 and self._pos == 2:  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
justinchuby: Bad idea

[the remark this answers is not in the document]
justinchuby: You do want to make sure transBatchB_attr is not None I think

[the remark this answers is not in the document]
bmehta001: Ok, will use `.value` instead to be safe

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -81,11 +81,11 @@ def check(
             return check_result.fail("Permutation values for Transpose are not correct.")
         if fused:
             fused_node = get_node(fused, "FusedMatMul")
-            if fused_node.attributes.get("transBatchA", 0).as_int() == 1 and self._pos == 1:  # type: ignore[union-attr]
+            if fused_node.attributes.get("transBatchA", 0).value == 1 and self._pos == 1:  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
bmehta001: Oh, I thought earlier you were saying that node.attributes might have "transBatchA" being mapped to `None`

If "transBatchA" does not exist in the attributes list, wouldn't the `0` part of node.attributes.get("transBatchA", 0) ensure that 0 is returned as the default value instead of `None`? Also, if "transBatchA" is not in `node.attributes`, this means it was not set to 1, so I do not want to return `check_result.fail`, so I think the `if` statement I have should be fine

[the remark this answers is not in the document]
bmehta001: Wait, I see the issue - I need to remove the default value altogether, because I don't want `0` to be returned and then call `.value` on it. This would have the same issue as `.asint()`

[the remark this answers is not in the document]
bmehta001: I have a broader question - are `attributes` guaranteed to exist? From my understanding, something like `perm` in `Transpose` is, but in our tests, because we are re-implementing `FusedMatMul`, it is not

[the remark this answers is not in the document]
justinchuby: perm doesn't need to exist in atttributes. Here https://onnx.ai/onnx/operators/onnx__Transpose.html#attributes it say "by default..." that's when it doesn't exsit. I think we should make it clearer in documentation that it is optional.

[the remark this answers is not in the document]
bmehta001: Ok, fixed the "None" issue

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -81,13 +90,10 @@ def check(
             return check_result.fail("Permutation values for Transpose are not correct.")
         if fused:
             fused_node = get_node(fused, "FusedMatMul")
-            if fused_node.attributes.get("transBatchA", 0).value == 1 and self._pos == 1:  # type: ignore[union-attr]
-                return check_result.fail(
-                    "FusedMatMul with transBatchA cannot be used with Transpose(A)."
-                )
-            if fused_node.attributes.get("transBatchB", 0).value == 1 and self._pos == 2:  # type: ignore[union-attr]
+            transBatchProperty = "transBatchA" if self._pos == 1 else "transBatchB"
```
justinchuby, line 93: use snake case for variable names

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -21,6 +21,15 @@ def get_kwargs(node: ir.Node) -> dict[str, float | int]:
     return kwargs
 
 
+def get_int_or_default(node: ir.Node, name: str, default: int = 0) -> int:
```
justinchuby, line 24: can this be private? `_get_int_or_default`

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +52,241 @@ def rewrite(self, op, x, y, cst):
...
-        kwargs = {}
-        alpha = node.attributes.get("alpha", None)
-        kwargs["alpha"] = alpha.value / c if alpha else 1.0 / c
-        for name in ["transA", "transB", "transBatchA", "transBatchB"]:
-            att = node.attributes.get(name)
-            if att:
-                kwargs[name] = att.value
+        fused_node = get_node(fused, "FusedMatMul")
+        kwargs = get_kwargs(fused_node)
+        kwargs["alpha"] = fused_node.attributes["alpha"].as_float() / c
```
gramalingam, line 73: What happens if there is no attribute called alpha? Is there a default value to be used, or should it be treated as an invalid node/model (we should skip optimization in this case: you can return None IIRC).

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +52,241 @@ def rewrite(self, op, x, y, cst):
...
+    def pattern(self, op, x, y):
+        return op.FusedMatMul(
+            x,
+            op.Transpose(y, _outputs=["transposed"]),
+            _domain="com.microsoft",
+            _outputs=["fused"],
+        )
+
+
+class _TransposeFusedMatMulBaseWithBatch(orp.RewriteRuleClassBase):
```
gramalingam, line 150: Would it be possible to add a couple of lines of documentation to describe the attributes (especially transBatch)? Would help understanding the validating the code below, thanks!

bmehta001: This might be clearer to see the rules spelled out:
```
import itertools as it

def transBatch(x):
    return x[1:-1] + [x[0]] + [x[-1]]

def trans(x):
    return x[0:-2] + [x[-1]] + [x[-2]]

def reverse_trans(x):
    return trans(x)

def reverse_transBatch(x):
    return [x[-2]] + x[0:-2] + [x[-1]]

def get_new_fused_matmul_perm(shape, flipTransBatch, flipTrans, originalTransBatch, originalTrans):
    new_shape = shape
    if flipTransBatch:
        newTransBatch = 1 - originalTransBatch
    else:
        newTransBatch = originalTransBatch
    
    if flipTrans:
        newTrans = 1 - originalTrans
    else:
        newTrans = originalTrans
        
    if newTransBatch:
        new_shape = transBatch(new_shape)
    if newTrans:
        new_shape = trans(new_shape)
    # this is what the new shape would look like with the transposing from FusedMatMul applied
    return new_shape
    
def derive_shape_before(shape_after, originalTransBatch, originalTrans):
    new_shape = shape_after # this is the shape after the transposing from FusedMatMul
    if originalTrans:
        new_shape = reverse_trans(new_shape)

    if originalTransBatch:
        new_shape = reverse_transBatch(new_shape)
        
    # this is what the shape would look like before the transposing from FusedMatMul
    return new_shape

def get_transpose_perm(shape, flipTransBatch, flipTrans, originalTransBatch, originalTrans):
    new_shape = get_new_fused_matmul_perm(shape, flipTransBatch, flipTrans, originalTransBatch, originalTrans)
    original_shape = derive_shape_before(new_shape, originalTransBatch, originalTrans)
    # this is what the shape should look like after Transpose and what the 'perm' of the Transpose should be
    return original_shape

x = list(range(6))
for prod in list(it.product(range(2), repeat=4)):
    flipTransBatch, flipTrans, originalTransBatch, originalTrans = prod
    if not flipTransBatch and not flipTrans:
        continue
    print(f"original transBatch: {originalTransBatch}, original trans: {originalTrans}, flip transBatch: {flipTransBatch}, flip trans: {flipTrans} -> {get_transpose_perm(x, flipTransBatch, flipTrans, originalTransBatch, originalTrans)}")
    ```

File: onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py (hunk 10 above)
Comment (bmehta001) [reply onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py]: Fixed

File: onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py (hunk 9 above)
Comment (bmehta001) [reply onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py]: Fixed

File: onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py [hunk 13]
```diff
@@ -358,6 +385,357 @@ def test_should_not_match(self):
             )
             self._check_model(model_proto, rewritten_model, atol=1e-6)
 
+    # Add unit tests to check that fusion rewrite can work even if MatMul is not the first node.
+    @classmethod
+    def _fused_matmul_with_transpose_or_matmul_in_middle(cls):
```
justinchuby, line 390: Also if we are testing multiple models, consider using `parameterized` to create parameterized tests. I think we can avoid the classmethod.

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -81,29 +94,32 @@ def check(
...
-        transposed_node = get_node(transposed, "Transpose")
-        perm = transposed_node.attributes["perm"].as_ints()
-        # Check that last two dimensions are swapped
-        expected_perm = list(range(len(perm)))
-        expected_perm[-2], expected_perm[-1] = expected_perm[-1], expected_perm[-2]
-        if perm != expected_perm:
-            return check_result.fail("Permutation values for Transpose are not correct.")
+        transposed_node = _get_node(transposed, "Transpose")
+        perm = _get_ints_or_default(transposed_node, "perm")
+        # If perm is not defined, the default transpose behavior is to swap
```
gramalingam, line 99: Looking at the documentation, I think the default behavior is to reverse all dimensions: which is the expected permutation only if the rank of the input tensor is 2. You can check the rank using [this](https://github.com/microsoft/onnxscript/blob/2f04ea865a95cbc2494ce3bfb88e3935a1fb7b11/onnxscript/rewriter/_ir_utils.py#L124) ... on an appropriate input value (not sure if it is x, y).

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
gramalingam: Thanks! Interesting.

gramalingam (approved):
LGTM, thanks! Just one minor comment left, whether the rank should be checked if perm attribute is not specified

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -309,14 +312,16 @@ def check(self, context, x, y, transposed: ir.Value, **_) -> orp.MatchResult:
...
         if perm:
             # Check that last two dimensions are swapped
             expected_perm = list(range(len(perm)))
             expected_perm[-2], expected_perm[-1] = expected_perm[-1], expected_perm[-2]
             if perm != expected_perm:
                 return check_result.fail("Permutation values for Transpose are not correct.")
+        elif not (self._pos == 1 and _ir_utils.has_rank(x, 2)) and (self._pos == 2 and _ir_utils.has_rank(y, 2)):
+            # If perm is not defined, the default transpose behavior is to swap
+            #   all dimensions, which is correct for MatMul with rank = 2.
+            return check_result.fail("Permutation values for Transpose are not correct.")
```
justinchuby: Maybe clarify what "not correct" means for users? E.g. Rank is not 2 or something like that

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -2,13 +2,50 @@
...
+    return node
+
+
+def _get_kwargs(node: ir.Node) -> dict[str, float | int]:
+    """Get the kwargs from the node."""
+    kwargs = {key: val.value for key, val in node.attributes.items()}
+    return kwargs
+
+
+def _get_int_or_default(node: ir.Node, name: str, default: int = 0) -> int:
```
justinchuby: With https://github.com/microsoft/onnxscript/pull/2360, you can use `node.attributes.get_int(name, default)`. Feel free to incorporate it here or as a follow up PR later depending on when the PR above is merged

bmehta001: Thanks, fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
bmehta001: Thanks, fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
bmehta001: Thanks, fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
bmehta001: Fixed

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -29,122 +57,289 @@ def rewrite(self, op, x, y, cst):
...
+        self, context, x, y, transposed: ir.Value, fused: ir.Value | None = None, **_
+    ) -> orp.MatchResult:
         check_result = orp.MatchResult()
-        perm = list((x if self._pos == 1 else y).uses())[0][0].attributes["perm"].value  # noqa: RUF015
-        expected_perm = list(range(len(perm)))
-        expected_perm[-2], expected_perm[-1] = expected_perm[-1], expected_perm[-2]
-        if perm != expected_perm:
-            return check_result.fail("Permutation values for Transpose are not correct.")
+        transposed_node = _get_node(transposed, "Transpose")
+        perm = _get_ints_or_default(transposed_node, "perm")
```
justinchuby:
```suggestion
        perm = transposed_node.attributes.get_ints("perm")
```

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -2,13 +2,41 @@
...
+    node: ir.Node, name: str, default: Optional[Sequence[int]] = None
+) -> Sequence[int]:
+    """Get the Sequence[int] value from the node attribute dictionary or return default."""
+    if name in node.attributes:
+        value = node.attributes[name].as_ints()
+    elif default is not None:
+        value = default
+    else:
+        value = []
+    return value
```
justinchuby:
```suggestion
```

Maybe this is not needed now?

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
```diff
@@ -80,283 +290,176 @@ def _check_model(
...
+        model_proto = script_func.to_model_proto(
+            input_types=input_types,
+            output_types=output_types,
+        )
+        ir_model = ir.serde.deserialize_model(model_proto)
+        rule_set = fused_matmul_rule_sets.fused_matmul_rule_sets()
+        rule_set.apply_to_model(ir_model)
+        rewritten_model = ir.serde.serialize_model(ir_model)
+        self.assertEqual(
+            ["Constant", "FusedMatMul"], [n.op_type for n in rewritten_model.graph.node]
```
justinchuby:
```suggestion
            ["Constant", "FusedMatMul"], [n.op_type for n in ir_model.graph]
```

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
```diff
@@ -165,5 +360,11 @@ def fused_matmul_rule_sets() -> orp.RewriteRuleSet:
             TransposeFusedMatMul1.rule(),
             TransposeMatMul2.rule(),
             TransposeFusedMatMul2.rule(),
+            TransposeFusedMatMulWithFlippedBatch1.rule(),  # type: ignore[attr-defined]
```
justinchuby: Curious why the ignores are needed?

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
```diff
@@ -45,7 +64,195 @@ def _run(
...
 
 
-class OrtRuleSetsTest(unittest.TestCase):
+@script()
+def _fused_matmul_div(A: FLOAT[4, 4], B: FLOAT[4, 4]) -> FLOAT[4, 4]:
+    D = op.Constant(
+        value=onnx.helper.make_tensor("scalar_half", onnx.TensorProto.FLOAT, (), [0.6])
+    )
+    ab = ms_op.FusedMatMul(A, B, alpha=0.4, transA=1)
+    out = op.Div(ab, D)
```
justinchuby:
```suggestion
    out = op.Div(ab, 0.6)
```

I think this can be simplified

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
```diff
@@ -45,7 +64,195 @@ def _run(
         return (np.matmul(A, B) * a,)
 
 
-class OrtRuleSetsTest(unittest.TestCase):
+@script()
+def _fused_matmul_div(A: FLOAT[4, 4], B: FLOAT[4, 4]) -> FLOAT[4, 4]:
+    D = op.Constant(
+        value=onnx.helper.make_tensor("scalar_half", onnx.TensorProto.FLOAT, (), [0.6])
```
justinchuby:
```suggestion
```

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
justinchuby: ditto below

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
bmehta001: Will change everywhere

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
bmehta001: I defined several rewrite rules by defining classes dynamically like below to avoid repetition and consolidate the logic into just 1 check function simply by changing the value of the class vars in each line:
```
TransposeFusedMatMulWithFlippedBatchAndTranspose1 = type(
    "TransposeFusedMatMulWithFlippedBatchAndTranspose1",
    (_TransposeFusedMatMulBaseWithBatch,),
    {"_flip_transpose": True, "_flip_transpose_batch": True},
)
TransposeFusedMatMulWithFlippedBatchAndTranspose2 = type(
    "TransposeFusedMatMulWithFlippedBatchAndTranspose2",
    (_TransposeFusedMatMulBaseWithBatch,),
    {"_pos": 2, "_flip_transpose": True, "_flip_transpose_batch": True},
)
```
However, I think this method of dynamically defining rules can cause MyPy to not recognize where the `.rule()` function comes from, so I get this error when linting:
```
 Error (MYPY) attr-defined
    "type" has no attribute "rule"

    To disable, use `  # type: ignore[attr-defined]`
    ```

Comment (bmehta001) [reply onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py]: I can define the rules like this to avoid the ignore statements
```
class TransposeFusedMatMulWithFlippedBatchAndTranspose1(_TransposeFusedMatMulBaseWithBatch):
    _flip_transpose = True
    _flip_transpose_batch = True

````

bmehta001: Ok, changed everywhere

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
bmehta001: This changes the model from `['Constant', 'FusedMatMul']` to `['FusedMatMul', 'Constant', 'CastLike', 'Div']`. Is there a simple way where I can add a constant without radically changing the graph so that I can use the same expected graph for a few different test cases? Ideally, I could simplify the graph to be just `FusedMatMul`.

bmehta001: Is there something like `ir_model = optimize(ir_model)` that I should be calling?

bmehta001: Otherwise, I can separate out the test cases into two separate functions

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets.py
justinchuby: Yeah declaring the classes the normal way seems more readable

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
justinchuby: Maybe assign 0.6 to a variable above, then use the variable? I haven’t tried hopefully it works. If not keeping the current is ok. Or consider

ir.to_proto(ir.tensor(0.6))

in the future we will update onnxscript to accept a tensor object.

bmehta001: Oh wow, just using `C=0.6` works

onnxscript/rewriter/ort_fusions/fused_matmul_rule_sets_test.py
bmehta001: Fixed

PR #2335, opened 2025-05-23
[torchlib] Fix pow.Tensor_Scalar type promotion

justinchuby:
Fix pow.Tensor_Scalar type promotion by accounting different combination of input dtypes. This change ensures the inputs to Pow is always the same type for compatibility with downstream tools.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6742,13 +6742,23 @@ def aten_positive(self: TensorType) -> TensorType:
 @torch_op(("aten::pow.Tensor_Tensor", "_operator::pow"), trace_only=True)
 def aten_pow(self: TReal, exponent: TTensor) -> TReal:
     """pow(Tensor self, Tensor exponent) -> Tensor"""
+    # TODO(justinchuby): Add type promotion
     return op.Pow(self, exponent)
 
 
 @torch_op("aten::pow.Tensor_Scalar", trace_only=True)
 def aten_pow_tensor_scalar(self: TReal, exponent: float) -> TReal:
     """pow(Tensor self, Scalar exponent) -> Tensor"""
-    return op.Pow(self, exponent)
+    if self.dtype.is_floating_point():
+        # Handle cases when e.g. (1) self is float16 or int
+        return op.Pow(self, op.Cast(exponent, to=self.dtype))
+
+    # For integer types, we need to cast self to the exponent type
+    if isinstance(exponent, int):
+        # The scalar exponent can be an int
+        return op.Pow(self, op.Cast(exponent, to=self.dtype))
+    # exponent is float
+    return op.Pow(op.Cast(self, to=FLOAT.dtype), exponent)
 
 
 @torch_op("aten::pow.Scalar", trace_only=True)
--- a/onnxscript/ir/_core.py
+++ b/onnxscript/ir/_core.py
@@ -1266,7 +1266,7 @@ class Usage(NamedTuple):
     idx: int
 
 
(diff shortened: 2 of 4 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6742,13 +6742,23 @@ def aten_positive(self: TensorType) -> TensorType:
...
-    return op.Pow(self, exponent)
+    if self.dtype.is_floating_point():
+        # Handle cases when e.g. (1) self is float16 or int
+        return op.Pow(self, op.Cast(exponent, to=self.dtype))
+
+    # For integer types, we need to cast self to the exponent type
+    if isinstance(exponent, int):
+        # The scalar exponent can be an int
+        return op.Pow(self, op.Cast(exponent, to=self.dtype))
+    # exponent is float
```
justinchuby, line 6760:
```suggestion
    # exponent is float so we cast self to match the exponent type.
    # More precisely if self is float64, we should cast exponent to float64; but
    # this is uncommon and should be fixed when we create a general type promotion
    # mechanism for torchlib
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6755,13 +6755,26 @@ def aten_positive(self: TensorType) -> TensorType:
 @torch_op(("aten::pow.Tensor_Tensor", "_operator::pow"), trace_only=True)
 def aten_pow(self: TReal, exponent: TTensor) -> TReal:
     """pow(Tensor self, Tensor exponent) -> Tensor"""
+    # TODO(justinchuby): Add type promotion
     return op.Pow(self, exponent)
 
 
 @torch_op("aten::pow.Tensor_Scalar", trace_only=True)
 def aten_pow_tensor_scalar(self: TReal, exponent: float) -> TReal:
```
titaiwangms, line 6763: Add int to exponent

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6755,13 +6755,26 @@ def aten_positive(self: TensorType) -> TensorType:
...
     """pow(Tensor self, Scalar exponent) -> Tensor"""
-    return op.Pow(self, exponent)
+    if self.dtype.is_floating_point():
+        # Handle cases when e.g. (1) self is float16 or int
+        return op.Pow(self, ir.tensor(exponent, dtype=self.dtype))
+    # For integer types, we need to cast self to the exponent type
+    if isinstance(exponent, int):
+        # The scalar exponent can be an int
+        return op.Pow(self, ir.tensor(exponent, dtype=self.dtype))
```
titaiwangms, line 6771: The two conditions do the same thing. Maybe merge them.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: The schema parser we have doesn't support this union. Since in python float "includes" int, I did it this way instead.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6755,13 +6755,26 @@ def aten_positive(self: TensorType) -> TensorType:
 @torch_op(("aten::pow.Tensor_Tensor", "_operator::pow"), trace_only=True)
 def aten_pow(self: TReal, exponent: TTensor) -> TReal:
     """pow(Tensor self, Tensor exponent) -> Tensor"""
+    # TODO(justinchuby): Add type promotion
```
titaiwangms, line 6758: Why not do this within the PR?

titaiwangms (approved):
Have we already deprecated type promotion in exporter? I don't remember the status of type promotion.

PR #2336, opened 2025-05-23
Refine shape optimization

gramalingam:
Refine the recently introduced shape optimization: more patterns showed up in the openai whisper model, extracting different slices of the concatenated shape.

The optimization improves MHA fusions (which are other handicapped by the reuse of some intermediate values that prevent fusion).

### DIFF
--- a/onnxscript/rewriter/ort_fusions/shape_optimization.py
+++ b/onnxscript/rewriter/ort_fusions/shape_optimization.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import onnxscript.ir as ir
+import onnxscript.rewriter._ir_utils as _ir_utils
 import onnxscript.rewriter.pattern as pattern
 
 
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/ort_fusions/shape_optimization.py
```diff
@@ -37,11 +38,22 @@ def check(self, context, dim0, dim1, dim2, dim3, final_shape, **_) -> bool:
...
+        self._end_val = _ir_utils.get_singleton_value(end)
+        if self._start_val is None or self._end_val is None:
+            return False
         return True
 
-    def rewrite(self, op, dim1, **_):
-        return dim1
+    def rewrite(self, op, dim0, dim1, dim2, dim3, **_):
+        transposed_dims = [dim0, dim2, dim1, dim3]
+        sliced_result = transposed_dims[self._start_val : self._end_val]
```
justinchuby, line 51: Looks like the lines are uncovered by tests

justinchuby (approved):

onnxscript/rewriter/ort_fusions/shape_optimization.py
gramalingam: Will add unit tests

onnxscript/rewriter/ort_fusions/shape_optimization.py
```diff
@@ -17,15 +18,19 @@ def __init__(self):
...
+        reshaped = op.Reshape(x, shape, allowzero=1)
         transposed = op.Transpose(reshaped, perm=[0, 2, 1, 3])
-        final_shape = op.Shape(transposed, _outputs=["final_shape"], start=0)
-        final_dim = op.Slice(final_shape, [-2], [-1])
+        final_shape = op.Shape(transposed, _outputs=["final_shape"])
+        final_dim = op.Slice(final_shape, start, end)
         return final_dim
 
-    def check(self, context, dim0, dim1, dim2, dim3, final_shape, **_) -> bool:
+    def check(self, context, dim0, dim1, dim2, dim3, final_shape, start, end, **_) -> bool:
```
justinchuby, line 33: possible to annotate the types?

gramalingam: Thought about it. Unfortunately, that would mean this single line turns into probably 12 lines, taking up valuable screen real estate. Eg. , in a laptop, it might mean we don't get to see the checking code below, or the rewrite method below at the same time as the pattern above. All of the relevant parameters above are `ir.Value` ... which is the common/usual case. More interestingly, it is seeing the pattern above that tells us *where* each value comes from (in the original graph pattern that matched). Which is why it more useful to look at the _pattern_ than just the _type_.

PR #2338, opened 2025-05-25
[IR][fix] Implement copy() for graph inputs/outputs

justinchuby:
Implement copy() for graph inputs/outputs because torch.onnx.verification is using it for version torch 2.7.

### DIFF
--- a/onnxscript/ir/_graph_containers.py
+++ b/onnxscript/ir/_graph_containers.py
@@ -12,7 +12,7 @@ __all__ = [
 ]
 
 import collections
-from typing import TYPE_CHECKING, Iterable, SupportsIndex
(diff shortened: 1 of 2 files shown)


onnxscript/ir/_graph_containers.py
```diff
@@ -12,7 +12,7 @@
 ]
 
 import collections
-from typing import TYPE_CHECKING, Iterable, SupportsIndex
+from typing import TYPE_CHECKING, Iterable, Self, SupportsIndex
```
justinchuby, line 15:
```suggestion
from typing import TYPE_CHECKING, Iterable, SupportsIndex
```

xadupre (approved):

PR #2340, opened 2025-05-25
[torchlib] Update linear implementation to support 1d weights

justinchuby:
It is possible when users call `F.linear()` directly in PyTorch.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -832,7 +832,7 @@ def aten_linear(input: TFloat, weight: TFloat, bias: Optional[TFloat] = None) ->
         # In rare cases the weight can be 1d
         weight_transposed = op.Unsqueeze(weight, [1])
     else:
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -825,10 +825,15 @@ def aten_leaky_relu_backward(
...
-    if len(input.shape) == 2:
+    if len(input.shape) == 2 and len(weight.shape) == 2:
         # Use Gemm for the rank 2 input
         return op.Gemm(input, weight, bias, transB=True)
-    weight_transposed = op.Transpose(weight, perm=[1, 0])
+    if len(weight.shape) == 1:
+        # In rare cases the weight can be 1d
+        weight_transposed = op.Unsqueeze(weight, [1])
+    else:
+        assert len(weight.shape) == 2:
```
justinchuby, line 835:
```suggestion
        assert len(weight.shape) == 2
```

PR #2343, opened 2025-05-26
[torchlib] Implement floor_divide for int inputs

justinchuby:
Fix https://github.com/microsoft/onnxscript/issues/2342

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3654,6 +3654,19 @@ def aten_floor_divide(self: TFloat, other: TFloat) -> TFloat:
     return op.Floor(op.Div(self, other))
 
 
+@torch_op("aten::floor_divide", trace_only=True)
+def aten_floor_divide_int(self: TInt, other: TInt) -> TInt:
+    """floor_divide(Tensor self, Tensor other) -> Tensor"""
+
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3654,6 +3654,19 @@ def aten_floor_divide(self: TFloat, other: TFloat) -> TFloat:
...
 
 
+@torch_op("aten::floor_divide", trace_only=True)
+def aten_floor_divide_int(self: TInt, other: TInt) -> TInt:
+    """floor_divide(Tensor self, Tensor other) -> Tensor"""
+
+    # TODO(justinchuby): This can be simplified if we can constrain the
+    # inputs to be positive integers. Consider how we can embed constraints in the model.
+    dtype = self.dtype
+    self = op.Cast(self, to=FLOAT.dtype)
```
xadupre, line 3664: What if dtype is float16?

justinchuby: This is only for int inputs, floating point inputs are handled above: https://github.com/microsoft/onnxscript/pull/2343/files#diff-773a426eaaf0eb2a1ed6039f2cd50613d2db7d61c239453545413499b2a15776R3654

xadupre (approved):

PR #2346, opened 2025-05-27
[torchlib] Set allowzero=True on Reshape where appropriate

justinchuby:
When we reshape from a dynamic shape, the shape can contain zeros. This change accounts for those cases.


justinchuby:
@gramalingam could you reapprove? I fixed the unflatten test case

PR #2352, opened 2025-05-28
SDPA fusion cleanup

gramalingam:
Remove the need for many different rules for SDPA fusion by (a) Using pattern-disjunction, and (b) Simplifying the handling of scaling factors which can occur in several forms (using either multiplication or division, either separately to query and/or key, or to the product of query and key).

### DIFF
--- a/onnxscript/rewriter/ort_fusions/sdpa.py
+++ b/onnxscript/rewriter/ort_fusions/sdpa.py
@@ -45,102 +45,107 @@ class SDPA(pattern.RewriteRuleClassBase):
         # from 3D to 4D and scaling is applied before the reshaping.
         query_reshape,
     ):
-        if self._pre_scale:
-            # Some implementations scale the query and key before computing the dot product
-            if self._use_mul:
-                if self._pre_scale_q:
-                    query = op.Mul(query, qk_scale)
-                else:
-                    query = op.Mul(query, query_scale)
-                    key_transposed = op.Mul(key_transposed, key_scale)
-            else:
-                if self._pre_scale_q:
-                    query = op.Div(query, qk_scale)
-                else:
-                    query = op.Div(query, query_scale)
-                    key_transposed = op.Div(key_transposed, key_scale)
+        # Some implementations scale the query and key before computing the dot product
+        query = pattern.OrValue([
+            op.Mul(query, query_scale),
+            op.Div(query, query_scale),
+            query,
+        ], tag_var="query_scaling", tag_values=["Mul", "Div", "None"])
+        key_transposed = pattern.OrValue([
+            op.Mul(key_transposed, key_scale),
+            op.Div(key_transposed, key_scale),
+            key_transposed,
+        ], tag_var="key_scaling", tag_values=["Mul", "Div", "None"])
 
-        # There might be patterns where the reshape and transpose are done
-        # after the pre-scaling. If the inputs are 3D, we need to reshape them to 4D
-        # and apply the approriate transposes to query.
-        if self._has_3d_query and self._pre_scale_q:
-            # Reshape and transpose 3D input of shape (B, S, D)
-            # to 4D input of shape (B, N, S, H)
-            queryBNSH = op.Reshape(query, query_reshape)
-            query = op.Transpose(queryBNSH, perm=[0, 2, 1, 3])
 
         attn_score = op.MatMul(query, key_transposed)
-        if not self._pre_scale:
-            # Some implementations scale the dot product.
-            if self._use_mul:
-                attn_score = op.Mul(attn_score, qk_scale)
-            else:
-                attn_score = op.Div(attn_score, qk_scale)
-        if self._use_mask:
-            # Some implementations add a mask to the dot product.
-            attn_score = op.Add(attn_score, mask)
+
+        # Some implementations scale the dot product.
+        attn_score = pattern.OrValues ([
+            op.Mul(attn_score, qk_scale),
+            op.Div(attn_score, qk_scale),
+            attn_score,
+        ], tag_var="qk_scaling", tag_values=["Mul", "Div", "None"])
+
+        # Some implementations add a mask to the dot product.
+        masked_attn_score = op.Add(attn_score, mask)
+        attn_score = pattern.OrValue([masked_attn_score, attn_score], tag_var="has_mask", tag_values=[True, False])
+
         attn_weight = op.Softmax(attn_score, axis=-1)
         attn_output = op.MatMul(attn_weight, value)
         return attn_output
 
     def check(
-        self, op, query, key_transposed, value, mask, query_scale, key_scale, qk_scale, **_
+        self, op, query, key_transposed, value, mask, query_scaling, query_scale, key_scaling, key_scale, qk_scale, **_
     ):
         check_result = pattern.MatchResult()
-        # Check that the scaling factors match what SDPA implements:
-
-        # We need to know the hidden size to check the scaling factors.
-        if query is None or query.shape is None or len(query.shape) < 2:
-            return check_result.fail(
-                "Query shape is not known or has less than 2 dimensions.", query
-            )
-        hidden_size = query.shape[-1]
-        if not isinstance(hidden_size, int):
-            return check_result.fail("Hidden size is not an integer.")
-
-        expected_scaling_factor = math.sqrt(hidden_size)
-        if self._use_mul:
-            expected_scaling_factor = 1.0 / expected_scaling_factor
-
-        if self._pre_scale and not self._pre_scale_q:
-            # Check if query_scale and key_scale are scalars == sqrt(expected_scaling_factor)
-            # If they are scalars but != sqrt(expected_scaling_factor), a custom scale is being used.
-            sqrt_scaling_factor = math.sqrt(expected_scaling_factor)
-            # Calculate the scaling factor for query
-            if (query_scale_value := _ir_utils.get_singleton_value(query_scale)) is None:
+        
+        if query_scaling == "None":
+            query_scale_value = 1.0
+        elif query_scaling == "Mul":
+            if (query_scale_value := _ir_utils.get_singleton_value(query_scale, rank=0)) is None:
                 return check_result.fail(
                     "Query scale is not a scalar.",
                     query_scale,
                 )
-            # Ensure the scaling factor for key is the same as for query
-            if (key_scale_value := _ir_utils.get_singleton_value(key_scale)) is None:
+        else:
+            assert query_scaling == "Div", "Unexpected query scaling operation"
+            if (query_scale_value := _ir_utils.get_singleton_value(query_scale, rank=0)) is None:
+                return check_result.fail(
+                    "Query scale is not a scalar.",
+                    query_scale,
+                )
+            query_scale_value = 1.0 / query_scale_value
+
+        if key_scaling == "None":
+            key_scale_value = 1.0
+        elif key_scaling == "Mul":
+            if (key_scale_value := _ir_utils.get_singleton_value(key_scale, rank=0)) is None:
                 return check_result.fail(
                     "Key scale is not a scalar.",
                     key_scale,
                 )
-            if not math.isclose(query_scale_value, key_scale_value, rel_tol=1e-3):
-                return check_result.fail(
-                    "Query and key scales are not equal.",
-                    query_scale,
-                )
-            if not math.isclose(query_scale_value, sqrt_scaling_factor, rel_tol=1e-3):
-                self._scale = query_scale_value * query_scale_value
-            else:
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/_fusion_utils.py
```diff
@@ -24,6 +25,24 @@ def _check_shape(bindings: dict[str, Dim], val: ir.Value, shape: Sequence[str])
...
+    if val.shape.rank() != len(shape):
+        raise MatchFailureError(
+            f"The rank of {val} ({val.shape.rank()} does not match the expected rank {len(shape)}.",
+            val,
+        )
+    for i, (actual, expected) in enumerate(zip(val.shape, shape)):
+        if expected not in bindings:
+            bindings[expected] = actual  # type: ignore[assignment]
+        elif actual != bindings[expected]:
+            raise MatchFailureError(
```
justinchuby, line 40: Is the error mainly used for propagating the match failure information?

onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -174,7 +174,14 @@ def try_rewrite(
                 if var.name is not None:
                     if var.name not in match.bindings:
                         match.bind(var.name, None)
-            check_match_result = self._condition_function(context, **match.bindings)
+            try:
+                check_match_result = self._condition_function(context, **match.bindings)
+            except _basics.MatchFailureError as e:
```
justinchuby, line 179: I imagine the error is always caught here and not exposed to users, is that accurate?

onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -174,7 +174,14 @@ def try_rewrite(
...
+                check_match_result = _basics.MatchResult()
```
justinchuby, line 180: Looks like this part was uncovered

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -41,173 +25,125 @@ def pattern(
...
+            scaling_type = match_bindings.get(tag_name, "None")
+            if scaling_type == "None":
+                return 1.0
             else:
-                # Pass no scaling factor to SDPA, SDPA will use the default scaling factor
-                self._scale = None
+                scale = match_bindings.get(scale_name)
+                value = _ir_utils.get_singleton_value(scale)
+                if value is None:
+                    raise MatchFailureError(f"{scale_name} is not a scalar.", scale)
```
justinchuby, line 104: Curious what the benefits are with raising the error over check_result.fail?

onnxscript/rewriter/_basics.py
```diff
@@ -16,6 +16,36 @@
     import onnxscript.rewriter._rewrite_rule as _rewrite_rule
 
 
+class MatchFailureInfo:
```
titaiwangms, line 19: Last time, @justinchuby and I were debugging transpose rewrite rule, and then we found that we don't know how to use this to debug. Is there a recommended way we can leverage MatchFailureInfor or MatchResult to debug? Is it just print out?

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -3,33 +3,17 @@
 from __future__ import annotations
 
 import math
+from typing import Union
 
+from onnxscript import ir
```
titaiwangms:
```suggestion
import onnx_ir as ir
```

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -41,173 +25,125 @@ def pattern(
...
+
+        # Some implementations scale the dot product.
+        attn_score = pattern.OrValue(
+            [
+                op.Mul(attn_score, qk_scale),
+                op.Div(attn_score, qk_scale),
+                attn_score,
+            ],
+            tag_var="qk_scaling",
+            tag_values=["Mul", "Div", "None"],
```
titaiwangms, line 60: After reading the description, I still don't understand the importance of this in the model. Is it just the naming of the values?

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -41,173 +25,125 @@ def pattern(
...
+
+        bindings: dict[str, Dim] = {}
+
+        # Check that query/key/value have the expected shapes:
+        # They all should have same batch-size (B) and number of heads (H). Conceptually, it is
+        # different for Q and K/V, but the certain op implementations require them to be the same,
+        # which is usually achieved via tiling/expanding K/V num-heads to match Q num-heads.
+        # Query and Key should have same head-size (Dh) while value can have different head-size (Dv).
+        # Key and Value should have same sequence length (Skv), while Query can have different sequence length (S).
+        _fusion_utils.check_shape(bindings, query, ["B", "H", "S", "Dh"])
```
titaiwangms, line 92: When do these shape names kick in to the graph? I am lost..

onnxscript/rewriter/_fusion_utils.py
gramalingam: Yes

onnxscript/rewriter/_rewrite_rule.py
gramalingam: Yes

onnxscript/rewriter/_rewrite_rule.py
gramalingam: True. We don't have negative-test-cases yet. Will add them (but getting fusions to work on the positive examples where they are expected to is more important, and we are far from getting coverage in that regard).

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Compositionality. If we return an error-value here, then the caller has to check for that and return that. An exception avoids the need for that. Same thing with utility functions we can write that are called from various fusion rules.

justinchuby: Thanks - maybe add this explanation to the exception class comments so future readers understand the rationale?

onnxscript/rewriter/_basics.py
gramalingam: Here are a couple of options I use for debugging, depending on the scenario:

(a) With simple examples (like toy unit-test-cases): I just set breakpoints (usually [here](https://github.com/microsoft/onnxscript/blob/2f04ea865a95cbc2494ce3bfb88e3935a1fb7b11/onnxscript/rewriter/_matcher.py#L71)) and examine why the failure happened.

(b) With large (real-world models), the above can be tedious, since there may be many "correct" failures before we hit the issue we care about. For this, I use something like [this](https://github.com/microsoft/onnxscript/blob/2f04ea865a95cbc2494ce3bfb88e3935a1fb7b11/onnxscript/rewriter/_fusion_utils.py#L45) to find the place where there is a maximum match. 

(c) If we know the node/value where the failure issue happens (as from (b)), we can use that to set a conditional breakpoint to stop when matching reaches that node/value for more information, if needed

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: What we are trying to do is match against three different patterns seen in models: models where qk is NOT scaled, models where qk is scaled using a Mul operator and models where qk is scaled using a Div operator.

The tag_var and tag_values are just used to convey information about which pattern matched (to the check/rewrite methods). So, `qk_scaling` will be assigned one of these three values in `tag_values` indicating which of the three patterns matched.

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Names like "B" are just used as the key in the dictionary `bindings`. These checks serve two purpose: first, whenever "B" is repeated, as it is in these 3 lines, we will (that is, `check_shape` will) check that those dimensions have the same value. Second, it can be used as in line 118 below to get the values of specific dimensions like head_size from the shapes of these values.

The way to think of this is: this is just "matching" the shape of the `query` value against the pattern `["B", "H", "S", "Dh"]` ... "B" is a variable in the shape-pattern that will get bound to the actual dimension (coming from the shape of `query`). Here are some links explaining the concept of structural-pattern-matching in programming-languages (which is the key idea we want to support): https://peps.python.org/pep-0636/ and https://dev.java/learn/pattern-matching/ for example.

justinchuby: Maybe helpful to add this explanation to the code, so that the rule can be used an example for later?

onnxscript/rewriter/ort_fusions/sdpa.py
titaiwangms: I am curious that how do we use OrValue to cover the cases that the inputs are mutually dependent in terms of the patterns? For instance, B will be in its second pattern only if A is in its first pattern.

titaiwangms: We could just catch all of them, but I wonder if there would be false positive on the cases.

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -41,173 +25,125 @@ def pattern(
...
-                queryBNSH = op.Reshape(query_mul, query_reshape)
-                query = op.Transpose(queryBNSH, perm=[0, 2, 1, 3])
-            else:
-                query = query_mul
-
         sdpa_args = [query, key_transposed, value]
-        if self._use_mask:
+        if mask is not None:
             sdpa_args.append(mask)
         return op.SDPA(*sdpa_args, scale=self._scale, _domain="ai.onnxruntime.fusion")
```
titaiwangms, line 144: If self._scale is None, who decides the default, as op.SDPA is not a real op.

onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -41,173 +25,125 @@ def pattern(
...
+                    return 1.0 / value
 
-        # check ranks/shapes
+        query_scale_value = get_scale_value("query_scaling", "query_scale")
+        key_scale_value = get_scale_value("key_scaling", "key_scale")
+        qk_scale_value = get_scale_value("qk_scaling", "qk_scale")
+
+        self._scale = query_scale_value * key_scale_value * qk_scale_value
+
+        # If the scaling factor is the default one, we can skip passing it to SDPA.
```
titaiwangms, line 117: Maybe merge this line to line 125

onnxscript/rewriter/ort_fusions/sdpa.py
titaiwangms: I was confused because I didn't expect that we can check the shape relatively. I think the existing comments should be good enough.

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Two options (if I understood your question correctly): I think you are talking about two different `OrValue` patterns that are correlated? If we use the same tag_var in both patterns, they will be required to be bound to the same value in both ... so, by using same tag_var and tag_values, we can enforce the required correlation.

Otherwise, the user can explicitly check for it in their check method: eg.,
```py
   if qk_scaling == "mul" and some_other == "None": # fail the match
```

onnxscript/rewriter/_rewrite_rule.py
gramalingam: Added a test case.

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: I guess it is a matter of documenting SDPA. It is modeled closely on ONNX opset 23 Attention op. The default scale is common to most attention op variants, which is 1 / sqrt(headsize) as above.

gramalingam: I am working on a follow-on PR that will implement SDPA using MHA. Hopefully that should make things more concrete too. (Will let us run all these test-cases with ORT after fusion.)

titaiwangms (approved):
LGTM

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Done

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Will leave it as-is for now. In longer run, I am thinking of integrating this into the pattern-matching core: so, users would be able to write `query = pattern.Var("query", shape=["B", "H", "S", "Dh"])` or something along those lines, where we will have to distinguish: a pattern-variable for a dimension from actual dimensions (which may themselves be SymbolicDims or int constants).

PR #2355, opened 2025-05-29
Always fold the `Transpose` node in the constant folder

justinchuby:
- Create an `always_fold_ops` option to allow users to specify which ops should always be folded
- Refactored the FoldConstantsPass to hide internal attributes
- Update logic to check for graph initialized inputs and removed the need for tracking in the object states


onnxscript/optimizer/_constant_folding.py
```diff
@@ -999,31 +999,45 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+        if any(x.is_graph_input() for x in node.inputs if x is not None):
+            # Do not fold any graph inputs to preserve graph signature
             return None
 
-        if any(self._state.is_initializer_input(x) for x in node.inputs):  # type: ignore[arg-type]
+        # Ensure all node inputs are constants
+        if any(x.const_value is None for x in node.inputs if x is not None):
+            if logger.isEnabledFor(logging.DEBUG):
+                logger.debug(
+                    "Skipping constant folding for node %s because it has None constant inputs",
```
gramalingam, line 1017: nit: Change "None constant inputs" to "non-constant inputs"?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -40,21 +35,21 @@ def is_control_flow_op(node: ir.Node) -> bool:
...
+    graph_types = {ir.AttributeType.GRAPH, ir.AttributeType.GRAPHS}
+    return any(attr.type in graph_types for attr in node.attributes.values())
 
 
-def is_constant_op(node: ir.Node) -> bool:
-    return node.op_type in {"Constant", "ConstantOfShape"} and utils.is_onnx_domain(
-        node.domain
-    )
+def _is_non_deterministic_op(node: ir.Node) -> bool:
+    return node.op_type in _NON_DETERMINISTIC_OPS and utils.is_onnx_domain(node.domain)
```
gramalingam, line 47: Minor: Now that we normalize domains, the utils.is_onnx_domain can be simplified or replaced by comparison with "".

onnxscript/optimizer/_constant_folding.py
```diff
@@ -999,31 +999,45 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+                    "Skipping constant folding for node %s because it has None constant inputs",
+                    node,
+                    [x.name for x in node.inputs if x is not None],
+                )
             return None
 
-        if any(input.nbytes > self._input_size_limit for input in input_values):  # type: ignore[union-attr]
+        input_tensors = [x.const_value if x is not None else None for x in node.inputs]
+
+        if node.op_type not in self.always_fold_ops and any(
```
gramalingam, line 1025: Minor nit: technically, we should check the domain is "" also?

gramalingam: May be worth checking that the node.input[0] has only this one use? Otherwise, we will do the computation, but not save it (since it would increase model size).

onnxscript/optimizer/_constant_folding.py
justinchuby: Done

onnxscript/optimizer/_constant_folding.py
justinchuby: I am ok with leaving the util to make it more readable. Does that work for you?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1155,6 +1166,7 @@ def fold_constants(
     onnx_shape_inference: bool = False,
     input_size_limit: int = DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT,
     output_size_limit: int = DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT,
+    always_fold_ops: Collection[str] = frozenset(["Transpose"]),
```
gramalingam, line 1169: This also needs to be updated to include the domain?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -999,31 +1005,45 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+                    "Skipping constant folding for node %s because it has non-constant inputs",
+                    node,
+                    [x.name for x in node.inputs if x is not None],
+                )
             return None
 
-        if any(input.nbytes > self._input_size_limit for input in input_values):  # type: ignore[union-attr]
+        input_tensors = [x.const_value if x is not None else None for x in node.inputs]
+
+        if (node.domain, node.op_type) not in self.always_fold_ops and any(
```
gramalingam, line 1031: Should always-fold-ups be restricted to values with a single-use? Eg, If we have a large initializer X that is used in two places, once Tranposed and once directly. Folding will double the memory-used for this X.

onnxscript/optimizer/_constant_folding.py
justinchuby: Ops from the default domain can be op_type only. Adding docs

onnxscript/optimizer/_constant_folding.py
justinchuby: Done

onnxscript/optimizer/_constant_folding.py
```diff
@@ -999,31 +1007,58 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+                    node,
+                    [x.name for x in node.inputs if x is not None],
                 )
             return None
 
-        # Filter out bfloat16 cases?
+        input_tensors = [x.const_value if x is not None else None for x in node.inputs]
+
+        if any(
+            tensor.nbytes > self.input_size_limit
```
titaiwangms, line 1034: It does not look like `always_fold_ops` are still folded when tensor size is small.

justinchuby: When tensor size is small everything is folded (usual path). Or did I miss something?

justinchuby:
@gramalingam

onnxscript/optimizer/_constant_folding.py
justinchuby: Feel free to let me know if a follow up PR is needed. Thanks

PR #2356, opened 2025-06-01
[Rewriter]: Add ∘ MatMul -> Gemm

AyoubMDL:
A Rewriter rule that transforms `MatMul(Add)` to `Gemm`.

### DIFF
--- a/tests/eager_test.py
+++ b/tests/eager_test.py
@@ -163,9 +163,9 @@ class TestOnnxSignal(onnx_script_test_case.OnnxScriptTestCase):
                 np.testing.assert_allclose(expected1, expected2)
                 with self.subTest(
                     c_shape=c.shape,
-                    le=list(le),
+                    le=le.tolist(),
                     expected_shape=expected1.shape,
-                    weights=we,
+                    weights=we.tolist(),
                 ):
                     case = onnx_script_test_case.FunctionTestParams(
                         signal_dft.dft_last_axis, [x, le, False], [expected1]
@@ -192,7 +192,7 @@ class TestOnnxSignal(onnx_script_test_case.OnnxScriptTestCase):
             nax = np.array([ax], dtype=np.int64)
             with self.subTest(
                 x_shape=x.shape,
-                le=list(le),
+                le=le.tolist(),
                 ax=ax,
                 expected_shape=expected.shape,
             ):
@@ -230,7 +230,7 @@ class TestOnnxSignal(onnx_script_test_case.OnnxScriptTestCase):
                 np.testing.assert_allclose(expected1, expected2)
                 with self.subTest(
                     c_shape=c.shape,
-                    le=list(le),
+                    le=le.tolist(),
                     ax=ax,
                     expected_shape=expected1.shape,
                 ):
@@ -256,7 +256,7 @@ class TestOnnxSignal(onnx_script_test_case.OnnxScriptTestCase):
                 nax = np.array([ax], dtype=np.int64)
                 with self.subTest(
                     x_shape=x.shape,
-                    le=list(le),
+                    le=le.tolist(),
                     ax=str(ax),
                     expected_shape=expected.shape,
                 ):
@@ -295,7 +295,7 @@ class TestOnnxSignal(onnx_script_test_case.OnnxScriptTestCase):
                 np.testing.assert_allclose(expected1, expected2)
                 with self.subTest(
                     c_shape=c.shape,
-                    le=list(le),
+                    le=le.tolist(),
                     ax=str(ax),
                     expected_shape=expected1.shape,
                 ):


AyoubMDL:
@justinchuby I wrote the rewriter using functions (as in [gemm_to_matmul_add](https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/gemm_to_matmul_add.py)). Tell me if you want me to do it using Classes.

onnxscript/rewriter/matmul_add_to_gemm.py
```diff
@@ -0,0 +1,24 @@
...
+def gemm_pattern(op, input_a, input_b, input_c):
+    return op.Gemm(input_a, input_b, input_c)
+
+
+def check_shapes(*_, input_a, input_b, **__):
+    # Rank of input_a and input_b must be 2
+    return len(input_a.shape) == 2 and len(input_b.shape) == 2
+
+
+rule = pattern.RewriteRule(matmul_add, gemm_pattern, check_shapes)
```
justinchuby: You can allow the rule to permute input orders: https://microsoft.github.io/onnxscript/api/generated/onnxscript.rewriter.pattern.RewriteRule.html#onnxscript.rewriter.pattern.RewriteRule.commute

onnxscript/rewriter/matmul_add_to_gemm.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Does the following transformation:
+- MatMul ∘ Add -> Gemm
```
justinchuby: This is more like Add ∘ MatMul right?

justinchuby:
> @justinchuby I wrote the rewriter using functions (as in [gemm_to_matmul_add](https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/gemm_to_matmul_add.py)). Tell me if you want me to do it using Classes.

Would be nice to use the class based approach just for consistency, thanks!

onnxscript/rewriter/matmul_add_to_gemm.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+from onnxscript.rewriter import pattern
+
+
+def matmul_add(op, input_a, input_b, input_c):
```
justinchuby: Do you have plans to also create rules to absorb transpose nodes into gemm?

inisis: It's not always a good idea to absorb transpose into gemm, especially when there are tied weights between emebdding and lmhead in LLMs, it will create heavy weights.

justinchuby: I see. Do you mean that’s a case when one user of the weight has a transpose, and another doesn’t?

justinchuby: I think it’s possible to detect the number of users of the weight. Sounds like if Transpose isn’t the only consumer of the weight value then we can skip the rule.

justinchuby: Actually: would this just be setting the trans* argument in the Gemm op and removing the transpose node? Maybe we don’t really need to update the weight? Could you point out what I am missing? Thanks!

inisis: I think that you're right, the purpose is to keep only one copy. By setting trans attribute can reduce one more transpose operator.

onnxscript/rewriter/matmul_add_to_gemm.py
AyoubMDL: Yes, I'll change it (thanks !)

onnxscript/rewriter/matmul_add_to_gemm.py
AyoubMDL: I'll add this to this PR

AyoubMDL:
Added the support for transposed inputs
> I think it’s possible to detect the number of users of the weight. Sounds like if Transpose isn’t the only consumer of the weight value then we can skip the rule.

I think I dont need to check the users, the rewriter does check the number of users by default (if i am not mistaken)

> You can allow the rule to permute input orders

With the current implementation, I don't think I need it, right ?

onnxscript/rewriter/matmul_add_to_gemm.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Does the following transformation:
+- Add(MatMul(X, W), B) -> Gemm
+- Add(MatMul(Transpose(X), W), B) -> Gemm
+- Add(MatMul(X, Transpose(W)), B) -> Gemm
+- Add(MatMul(Transpose(X), Transpose(W)), B) -> Gemm
+"""
+
+from abc import ABC
```
justinchuby:
```suggestion
import abc
```

import modules only, except for the typing module

onnxscript/rewriter/matmul_add_to_gemm.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import ClassVar
+
+from onnxscript.rewriter import pattern as orp
+
+
+class _MatMulAddToGemmBase(orp.RewriteRuleClassBase, ABC):
```
justinchuby:
```suggestion
class _MatMulAddToGemmBase(orp.RewriteRuleClassBase, abc.ABC):
```

onnxscript/rewriter/matmul_add_to_gemm_test.py
```diff
@@ -0,0 +1,314 @@
...
+            inputs.append(b)
+        else:
+            b = ir.tensor(self.rng.uniform(-0.5, 0.5, bias_shape).astype("float32"), name="B")
+            b = tape.initializer(b)
+
+        if transA:
+            x_t = tape.op("Transpose", inputs=[x], attributes={"perm": permA})
+
+        if transB:
+            w_t = tape.op("Transpose", inputs=[w], attributes={"perm": permB})
```
justinchuby, line 75:
```suggestion
        if transA:
            x = tape.op("Transpose", inputs=[x], attributes={"perm": permA})

        if transB:
            w = tape.op("Transpose", inputs=[w], attributes={"perm": permB})
```

maybe just do this to make the code checkers happy

AyoubMDL: I've tried this before but I get errors like `ValueError: Value '%"anonymous:132760704101872"<?,?>' is produced by a node and cannot be an input to the graph. Please create new Values for graph inputs`.
I've added a workaround for the checker

justinchuby:
@titaiwangms could you approve? Thanks

titaiwangms (approved):

PR #2358, opened 2025-06-02
Fix test names for pytest

justinchuby:
With the latest version of pytest we get 

> onnxscript/rewriter/ort_fusions/cos_sin_cache_test.py::test_case_1 - Failed: Expected None, but test returned <onnxscript.rewriter.ort_fusions.models._rotary_embedding_models._TestCase1 object at 0x117c920d0>. Did you mean to use `assert` instead of `return`?

This is because the imported functions `test_case_1` and `test_case_2` are not really test cases but were treated as such by pytest. This PR hides them from the test module so they are not triggered.


titaiwangms (approved):
Looks like a bad naming. Maybe we should consider rename it.

justinchuby:
> Looks like a bad naming. Maybe we should consider rename it.

Agreed

PR #2359, opened 2025-06-03
docs: cleanup documentation for function-based rewrites📄

KarelZe:
With https://github.com/microsoft/onnxscript/pull/2039/files support for `function-based` rewriting was dropped. Some tutorials and the the readme were still referencing function-based rewrites.

@justinchuby / @gramalingam Could you please review? Any feedback is appreciated.


justinchuby (approved):
Thanks!

onnxscript/rewriter/onnxruntime/__init__.py
```diff
@@ -26,14 +26,13 @@ def rewrite(
     model_proto: onnx.ModelProto,
     /,
     function_rules=None,
```
justinchuby:
```suggestion
```

could you help remove the param and the docstring as well? Thanks!

KarelZe: @justinchuby It's fixed now. 🎉 Initially, I thought that the param was kept for backward-compatibility. 

NIT, I also noted some broken links to the rewrite tutorials, that should no be fixed.

PR #2361, opened 2025-06-03
Enable CSE in optimizer

titaiwangms:


onnxscript/optimizer/_optimizer.py
```diff
@@ -51,6 +51,7 @@ def optimize_ir(
             early_stop=stop_if_no_change,
         ),
         onnxscript.ir.passes.common.RemoveUnusedNodesPass(),
+        onnxscript.ir.passes.common.CommonSubexpressionEliminationPass(),
```
titaiwangms, line 54: I was confused by `onnxscript.ir.passes.common` for a minute and learned we have alias.

titaiwangms:
@justinchuby PTAL

justinchuby (approved):

PR #2363, opened 2025-06-04
2301 fold pad into conv

Johansmm:
Fuses Pad nodes into the following nodes (Conv, ConvInteger)
(https://github.com/microsoft/onnxscript/issues/2301)


onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+class _NormalizePadFormatBase(orp.RewriteRuleClassBase):
+    """Interface to normalize pad attributes in conv nodes."""
+
+    def compute_pads(
+        self,
+        input_shape: typing.Sequence[int],
+        output_shape: typing.Sequence[int],
+        attributes: dict[str, typing.Sequence[int] | str],
+    ) -> typing.Sequence[int]:
+        raise ChildProcessError("Child have to implement this function")
```
justinchuby:
```suggestion
        raise NotImplementedError
```

I don't think ChildProcessError is designed for this usage?

Johansmm: Ups you are right

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Fuses Pad nodes into preceding nodes. Supported fusion patterns:
+- Pad ∘ Conv          -> Conv
+- Pad ∘ ConvInteger   -> ConvInteger
+
```
justinchuby: Is this Conv ∘ Pad and ConvInteger ∘ Pad?

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+        else:
+            start_value = pads[axes.index(axis)]
+            end_value = pads[axes.index(axis) + len(axes)]
+        pad_len = len(new_pads) // 2
+        new_pads.insert(pad_len + axis, end_value)
+        new_pads.insert(axis, start_value)
+    return new_pads
+
+
+def read_conv_attributes(ir_conv: ir.Node) -> dict[str, typing.Sequence[int] | str]:
```
justinchuby: I just added some methods you can use to read attributes: https://onnx.ai/ir-py/api/generated/onnx_ir._graph_containers.Attributes.html#onnx_ir._graph_containers.Attributes.get_int

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+    # Read attributes
+    attributes = {}
+    if (kernel_shape := ir_conv.attributes.get("kernel_shape", None)) is not None:
+        attributes["kernel_shape"] = kernel_shape.as_ints()
+    else:
+        attributes["kernel_shape"] = ir_conv.inputs[1].shape[2:]
```
justinchuby:
```suggestion
    attributes["kernel_shape"] =  ir_conv.attributes.get_ints("kernel_shape", ir_conv.inputs[1].shape[2:])
```

justinchuby: What if the shape of `ir_conv.inputs[1]` is not known? Just making sure it is checked

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+        pad_pads = fill_pads_with_axes(pad_pads, axes, x_rank)
+
+        # Get only spatial pads
+        new_pads = pad_pads[2:x_rank] + pad_pads[x_rank + 2 :]
+
+        # Replace conv pads = new + old
+        conv_attr: typing.Mapping[str, ir.Attr] = cnode.attributes.copy()
+        if "pads" in conv_attr:
+            new_pads = [x + y for x, y in zip(conv_attr["pads"].as_ints(), new_pads)]
+        conv_attr["pads"] = ir.convenience.convert_attribute("pads", new_pads)
```
justinchuby:
```suggestion
        conv_attr.add(ir.AttrInt64s("pads", new_pads))
```

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+            axes = list(range(x_rank))
+
+        # Fulfill pad_pads in every dimension (filling with zero the other ones)
+        pad_pads = fill_pads_with_axes(pad_pads, axes, x_rank)
+
+        # Get only spatial pads
+        new_pads = pad_pads[2:x_rank] + pad_pads[x_rank + 2 :]
+
+        # Replace conv pads = new + old
+        conv_attr: typing.Mapping[str, ir.Attr] = cnode.attributes.copy()
```
justinchuby:
```suggestion
        conv_attr = cnode.attributes.copy()
```

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def __init__(self, name: str, as_function: bool = False):
+        # Remove nodes is set to False to remove unused nodes after the rewrite.
+        super().__init__(name=name, remove_nodes=False, as_function=as_function)
+
+    def rewrite(
+        self, op: ir.tape.Tape, x: ir.Value, pad: ir.Value, conv: ir.Value
+    ) -> ir.Value:
+        pnode = pad.producer()
+        cnode = conv.producer()
```
justinchuby:
```suggestion
        pad_node = pad.producer()
        conv_node = conv.producer()
```

avoid abbreviations unless well known

onnxscript/rewriter/fuse_pad_into_conv_test.py
```diff
@@ -0,0 +1,332 @@
...
+        ir_model = ir.Model(
+            ir.Graph(
+                inputs=[x, *inputs],
+                outputs=[y],
+                nodes=tape.nodes,
+                initializers=tape.initializers,
+                opset_imports=opset_imports,
+                name="model",
+            ),
+            ir_version=9,
```
justinchuby:
```suggestion
            ir_version=10,
```

nit: prefer 10 but doesn't matter too much

onnxscript/rewriter/fuse_pad_into_conv_test.py
```diff
@@ -0,0 +1,332 @@
...
+        self,
+        op_type: str,
+        input_shape: ir.Shape,
+        weight_shape: typing.Sequence[int],
+        pad_inputs: typing.Sequence[ir.TensorProtocol | ir.Value | None],
+        pad_attributes: typing.Mapping[str, ir.Attr] | None = None,
+        conv_attributes: typing.Mapping[str, ir.Attr] | None = None,
+        opset_imports: typing.Mapping[str, int] = {"": 20},
+    ) -> ir.Model:
+        tape = ir.tape.Tape()
```
justinchuby, line 46: This is a great example of using the tape module, thanks!

onnxscript/rewriter/fuse_pad_into_conv_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def build_model(
+        self,
+        op_type: str,
+        input_shape: ir.Shape,
+        weight_shape: typing.Sequence[int],
+        pad_inputs: typing.Sequence[ir.TensorProtocol | ir.Value | None],
+        pad_attributes: typing.Mapping[str, ir.Attr] | None = None,
+        conv_attributes: typing.Mapping[str, ir.Attr] | None = None,
+        opset_imports: typing.Mapping[str, int] = {"": 20},
```
justinchuby: Using dictionaries as default values is not recommended. Since this field is not changed by callers, I suggest we set it as a constant inside the method

onnxscript/rewriter/fuse_pad_into_conv_test.py
```diff
@@ -0,0 +1,332 @@
...
+        w = ir.tensor(self.rng.uniform(-0.5, 0.5, shape).astype("float32"), name="W")
+        if tape is not None:
+            w = tape.initializer(w)
+        return w
+
+    def build_model(
+        self,
+        op_type: str,
+        input_shape: ir.Shape,
+        weight_shape: typing.Sequence[int],
```
justinchuby:
```suggestion
        weight_shape: Sequence[int],
```

prefer importing the classes from the typing module: https://google.github.io/styleguide/pyguide.html#2241-exemptions

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: Ups

Johansmm:
Push force rebasing on main and fixing conflicts.

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
...
+import typing
+
+import numpy as np
+import onnx_ir as ir
+
+from onnxscript.rewriter import pattern as orp
+
+
+def fill_pads_with_axes(
+    pads: typing.Sequence[int], axes: typing.Sequence[int], rank: int
```
justinchuby: Was this fixed? Sequence should actually be imported from collections.abc

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,286 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Fuses Pad nodes into preceding nodes. Supported fusion patterns:
+- Pad ∘ Conv          -> Conv
+- Pad ∘ ConvInteger   -> ConvInteger
+
+To make some rules possible, we implicitly transform `auto_pad` attribute into its explicit list.
+"""
+
```
justinchuby, line 9: please from future import annotations

onnxscript/rewriter/fuse_pad_into_conv_test.py
justinchuby: Seems like fixes were not pushed yet

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: you are right. After testing I found an issue here. Adding new unit tests.

Johansmm (commented):
@justinchuby I forgot to explain in the previous message that the changes were not ready yet (just fixing rebase with main).
In the last commits I update the code with all the suggestions.
Now this work is ready to be reviewed.

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: I make this but normally this is not required from python3.7+

justinchuby:
Could you update the PR title and description? Thanks

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+    if "pads" in ir_attributes:
+        attributes["pads"] = ir_attributes.get_ints("pads")
+    return attributes
+
+
+class _FusePadConvBase(orp.RewriteRuleClassBase):
+    """Interface for PadConv nodes fusion."""
+
+    def __init__(self, as_function: bool = False):
+        # Remove nodes is set to False to remove unused nodes after the rewrite.
```
titaiwangms: Could you instead put the reason why you want to remove them after the rewrite? It would be more informative.

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+        # Pad constraints: values
+        pads_list = fill_pads_with_axes(pads.const_value.numpy(), axes_list, x_rank)
+        if np.any(pads_list[:2] + pads_list[x_rank : x_rank + 2]):
+            return check_result.fail(f"{pads.name} must be zero in non-spatial dimensions.")
+
+        return check_result
+
+
+class FusePadConv(_FusePadConvBase):
+    """Replaces ``Pad(Conv(x))`` with ``Conv(x)``."""
```
titaiwangms:
```suggestion
    """Replaces ``Conv(Pad(x))`` with ``Conv(x)``."""
```

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # Pad constraints: values
+        pads_list = fill_pads_with_axes(pads.const_value.numpy(), axes_list, x_rank)
+        if np.any(pads_list[:2] + pads_list[x_rank : x_rank + 2]):
+            return check_result.fail(f"{pads.name} must be zero in non-spatial dimensions.")
+
+        return check_result
+
+
+class FusePadConv(_FusePadConvBase):
```
titaiwangms: Is it supposed to be ConvPad?

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+        conv_node = conv.producer()
+        if (
+            apad := conv_node.attributes.get("auto_pad", None)
+        ) and apad.as_string() != "NOTSET":
+            return check_result.fail(f"{conv_node.name} auto_pad must be 'NOTSET'.")
+        return check_result
+
+
+class FusePadConvInteger(FusePadConv):
+    """Replaces ``Pad(ConvInteger(x))`` with ``ConvInteger(x)``."""
```
titaiwangms: Same question here: ConvInteger(Pad(x))?

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+
+        return op.op(
+            conv_node.op_type,
+            inputs=(x, *conv_node.inputs[1:]),
+            attributes=conv_attr,
+            domain=conv_node.domain,
+            name=conv_node.name,
+        )
+
+    def check(self, context, x: ir.Value, pad: ir.Value, conv: ir.Value) -> orp.MatchResult:
```
titaiwangms, line 96: This is relatively complicated. Could you summarize the check logic here. It would be easier to follow and maintain. Something like this would be helpful: https://github.com/microsoft/onnxscript/blob/b8a831ed49d51da405402bcb10c7b1348197c352/onnxscript/rewriter/broadcast_to_matmul.py#L16-L28

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+                return check_result.fail(error_msg.format("output"))
+            attributes = read_conv_attributes(conv_node)
+            if len(attributes["kernel_shape"]) != len(attributes["strides"]):
+                return check_result.fail(
+                    f"strides must have the same length than kernel_shape on {conv_node.name}."
+                )
+        return check_result
+
+
+class NormalizePadFormatConv(_NormalizePadFormatBase):
```
titaiwangms, line 287: I suppose this does not fallback when the rewritten conv still does not match `FusePad` rules? Do we still benefit from converting auto_pad attribute into 'NOTSET' if it's still a no match?

titaiwangms (commented):
Please run "lintrunner -a"

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,296 @@
...
+
+from typing import List, Sequence
+
+import numpy as np
+import onnx_ir as ir
+
+from onnxscript.rewriter import pattern as orp
+
+
+def fill_pads_with_axes(pads: Sequence[int], axes: Sequence[int], rank: int) -> List[int]:
```
titaiwangms, line 20: Could you document this function?

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: You are right, always `NormalizePadFormat` is applied, even if `FusePad` is not applied.

On the other hand, I consider that having a single pad format is coherent, since it makes it easier to be interpreted by different accelerators. ([bug example in `onnxruntime` ](https://github.com/microsoft/onnxruntime/issues/495) due to the `auto_pad` type).

Let me know if you still disagree with this information. What do you think @justinchuby.

Johansmm (commented):
Last commit with @titaiwangms suggestions

onnxscript/rewriter/fuse_pad_into_conv.py
titaiwangms: cc @gramalingam

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,346 @@
...
+    ) -> Sequence[int]:
+        # Compute pads, following auto_pad/pads attributes
+        if attributes["auto_pad"] in ["NOTSET", "VALID"]:
+            assert len(input_shape) > 0
+            return attributes.get("pads", [0] * len(input_shape) * 2)
+
+        bottom_pads, top_pads = [], []
+        kernel_shape, strides = attributes["kernel_shape"], attributes["strides"]
+        assert len(kernel_shape) == len(strides) == len(input_shape) == len(output_shape)
+        for x, y, k, s in zip(input_shape, output_shape, kernel_shape, strides):
```
justinchuby, line 304: What happens when input shape is dynamic?

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,346 @@
...
+        Returns:
+            True if we need to replace the pattern, False otherwise.
+        """
+        del context
+        check_result = orp.MatchResult()
+
+        # Conv constraints: attributes
+        conv_node = conv.producer()
+        auto_pad = conv_node.attributes.get_string("auto_pad", None)
+        if auto_pad in [None, "NOTSET"]:
```
justinchuby:
```suggestion
        if auto_pad in {None, "NOTSET"}:
```

nit. Here and everywhere else

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: This case is rejected for `_NormalizePadFormatBase.check` function, so input_shape is always static.

onnxscript/rewriter/fuse_pad_into_conv.py
Johansmm: I'll do but why ?

Johansmm:
Last push force rebasing on main and addig @justinchuby suggestions

onnxscript/rewriter/fuse_pad_into_conv.py
justinchuby: It's usually more efficient: https://docs.astral.sh/ruff/rules/literal-membership/ ; https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-set-for-membership.html

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,354 @@
...
+    def pattern(self, op: ir.tape.Tape, x: ir.Value) -> ir.Value:
+        return op.Conv(
+            op.Pad(x, _allow_other_inputs=True, _outputs=["pad"]),
+            _allow_other_inputs=True,
+            _outputs=["conv"],
+        )
+
+    def check(self, context, x: ir.Value, pad: ir.Value, conv: ir.Value) -> orp.MatchResult:
+        check_result = super().check(context, x, pad, conv)
+        if check_result.reason:
```
gramalingam: nit: I suggest `if not check_result:` ... a MatchResult can be treated as a boolean, with True meaning success, and False meaning failure.

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,354 @@
...
+            return check_result
+
+        # Conv constraints: attributes
+        conv_node = conv.producer()
+        if (
+            apad := conv_node.attributes.get("auto_pad", None)
+        ) and apad.as_string() != "NOTSET":
```
gramalingam: minor nit: I think this condition can be simplified to `conv_node.attributes.get_string("auto_pad", "NOTSET") != "NOTSET"` .

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def rewrite(
+        self, op: ir.tape.Tape, x: ir.Value, pad: ir.Value, conv: ir.Value
+    ) -> ir.Value:
+        pad_node = pad.producer()
+        conv_node = conv.producer()
+
+        # Retrieve the padding and axes
+        x_rank = len(x.shape)
+        pad_pads = pad_node.inputs[1].const_value.numpy().tolist()
```
gramalingam: The above line needs to handle various special-conditions and error-situations. I suggest using `onnxscript.rewriter._ir_utils.get_numpy_value(pad_node.inputs[1])` which will handle the cases where the input is None or is not a constant. And check that the value is not None before using it.

gramalingam: Oh, I see, the check is being done in the check method done below.

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,354 @@
...
+                return check_result.fail(f"{constant_value.name} must be equal to 0.")
+        if len(pad_node.inputs) > 3 and (axes := pad_node.inputs[3]) is not None:
+            if axes.const_value is None:
+                return check_result.fail(f"{axes.name} is not a constant/initializer.")
+            axes_list = [x if x >= 0 else x_rank + x for x in axes.const_value.numpy()]
+        else:
+            axes_list = list(range(x_rank))
+
+        # Pad constraints: values
+        pads_list = fill_pads_with_axes(pads.const_value.numpy(), axes_list, x_rank)
```
gramalingam: If this is the same as the `pad_pads` used in the `rewrite` method, you can save it as `self._pads_list` and use it in the rewrite method to avoid redoing the entire computation. It involves implicitly passed state, but works fine.

Johansmm (commented):
Last push force rebasing on upstream/main and with @gramalingam suggestions

justinchuby (approved):

onnxscript/rewriter/fuse_pad_into_conv.py
```diff
@@ -0,0 +1,347 @@
...
+        auto_pad = conv_node.attributes.get_string("auto_pad", None)
+        if auto_pad in {None, "NOTSET"}:
+            return check_result.fail(
+                f"{conv_node.name} ({conv_node.op_type}) auto_pad must be different to 'NOTSET'."
+            )
+
+        # Conv constraints: inputs/outputs
+        input_shape = conv_node.inputs[0].shape
+        output_shape = conv_node.outputs[0].shape
+        if len(input_shape) <= 2:
```
justinchuby: What happens when input_shape/output_shape is None? Do we need to handle it?

Johansmm: Yes, you're right. I forget this case. Done

Johansmm:
Last push with last @justinchuby suggestions

titaiwangms (approved):
LGTM! Thank you! (@justinchuby is on vacation, so feel free to merge it)

PR #2364, opened 2025-06-04
fix: pattern match gelu from contrib and onnx ops🐛

### LINKED ISSUE #2362: Domain of `Gelu` in `bias_gelu_rules` set
Thanks for developing onnxscript 💯  

I'm currently comparing the fusion options integrated in onnxruntime and onnxscript. As such, I came across fusion of bias and gelu  ([`BiasGeluFusion`](https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/ort_fusions/bias_gelu.py) in onnxscript and [`FusionBiasGelu`](https://github.com/microsoft/onnxruntime/blob/c7fdb54c3ef95bc295aef1b0e6a2867be5d0331e/onnxruntime/python/tools/transformers/fusion_biasgelu.py#L16) in onnxruntime). 

I noticed, that the domain for `Gelu` in the [rules implementation](https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/ort_fusions/bias_gelu.py#L11) is restricted to the [contributor ops implementation](https://github.com/microsoft/onnxruntime/blob/rel-1.20.0/docs/ContribOperators.md#com.microsoft.Gelu) and does not fuse `Gelu` from onnx ops ([introduced with opset 20](https://onnx.ai/onnx/operators/onnx__Gelu.html#l-onnx-doc-gelu)). Onnxruntime, on the other hand, would fuse such operations into `BiasGelu`.

Is there a reason, why the implementation is constrained to match contributor ops? Would you be open to match against alternative patterns with Gelu from onnx ops? 

@shubhambhokare1 @justinchuby 

**MWE:**

```python
"""Testing bias + Gelu fusion.
"""

import numpy as np
import onnxscript
from onnxscript import opset20 as op
from onnxscript import script
from onnxscript.onnx_types import FLOAT
from onnxscript.rewriter.ort_fusions.bias_gelu import bias_gelu_rules

import onnxruntime as ort

# compare onnx + onnx-contributor implementation
msft_op = onnxscript.values.Opset("com.microsoft", 1)


@script()
def GemmGelu(Input: FLOAT["N", "K"], Weight: FLOAT["K", "M"], Bias: FLOAT["M"]) -> FLOAT["N", "M"]:
    """
    Gemm + Gelu from onnx domain.
    """
    T1 = op.MatMul(Input, Weight)
    T2 = op.Add(T1, Bias)
    return op.Gelu(T2, approximate="none")


@script()
def GemmGeluCustom(Input: FLOAT["N", "K"], Weight: FLOAT["K", "M"], Bias: FLOAT["M"]) -> FLOAT["N", "M"]:
    """
    Gemm with Gelu from Microsoft domain.
    """
    T1 = op.MatMul(Input, Weight)
    T2 = op.Add(T1, Bias)
    return msft_op.Gelu(T2)


onnx_model = GemmGelu.to_model_proto()
onnx_model = onnxscript.rewriter.rewrite(onnx_model, pattern_rewrite_rules=bias_gelu_rules)

print(onnx_model)

# test model
sess = ort.InferenceSession(onnx_model.SerializeToString(), providers=["CPUExecutionProvider"])
inputs = {
    "Input": np.random.randint(10, size=(3, 8)).astype(np.float32),
    "Weight": np.random.randint(10, size=(8, 4)).astype(np.float32),
    "Bias": np.random.randint(10, size=(4)).astype(np.float32),
}
sess.run(None, inputs)
```

KarelZe:
Previously the domain for Gelu in the [rules implementation](https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/ort_fusions/bias_gelu.py#L11) was restricted to the [contributor ops implementation](https://github.com/microsoft/onnxruntime/blob/rel-1.20.0/docs/ContribOperators.md#com.microsoft.Gelu) and does not fuse Gelu from onnx ops ([introduced with opset 20](https://onnx.ai/onnx/operators/onnx__Gelu.html#l-onnx-doc-gelu)). 

This pr introduces pattern matching + tests for both variants.

closes #2362 .

@shubhambhokare1 @justinchuby Could you please review? Any feedback is greatly appreciated.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/bias_gelu.py
+++ b/onnxscript/rewriter/ort_fusions/bias_gelu.py
@@ -6,19 +6,25 @@ from onnxscript.rewriter import _fusion_utils, pattern
 
 
 class BiasGeluFusion(pattern.RewriteRuleClassBase):
+    """Fuses a Bias-Gelu pattern into a single BiasGelu operator.
+
+    Attributes:
+        contrib_op (bool): If True, matches the Gelu operator from the 'com.microsoft' domain.
+                           If False, matches the standard ONNX Gelu operator.
+    """
+
     def __init__(
         self,
-        name,
+        name: str,
         *,
-        has_contrib_op: bool,
+        contrib_op: bool,
     ):
         super().__init__(name)
-        self._has_contrib_op = has_contrib_op
+        self._contrib_op = contrib_op
 
     def pattern(self, op, x, y):
         gelu_add = op.Add(x, y)
-        # see: gh-2362. Match against Gelu from onnx op or contrib ops.
-        if self._has_contrib_op:
+        if self._contrib_op:
             return op.Gelu(gelu_add, _domain="com.microsoft")
         else:
             return op.Gelu(gelu_add)
@@ -29,8 +35,8 @@ class BiasGeluFusion(pattern.RewriteRuleClassBase):
 
 bias_gelu_rules = pattern.RewriteRuleSet(
     [
-        BiasGeluFusion.rule("gelu_onnx_op", has_contrib_op=False),
-        BiasGeluFusion.rule("gelu_contrib_op", has_contrib_op=True),
+        BiasGeluFusion.rule("gelu_onnx_op", contrib_op=False),
+        BiasGeluFusion.rule("gelu_contrib_op", contrib_op=True),
     ]
 )
 
--- a/onnxscript/rewriter/ort_fusions/bias_gelu.py
+++ b/onnxscript/rewriter/ort_fusions/bias_gelu.py
@@ -31,8 +31,8 @@ class BiasGeluFusion(pattern.RewriteRuleClassBase):
 
     def check(self, op, gelu, **_) -> pattern.MatchResult:
         check_result = pattern.MatchResult()
-        approximate = gelu.producer().attributes.get("approximate", None)
-        if approximate is not None and approximate.value == "tanh":
+        approximate = gelu.producer().attributes.get_string("approximate")
+        if approximate is not None and approximate == "tanh":
             return check_result.fail(
                 "Gelu operator with 'approximate' set to 'tanh' is not supported."
             )
--- a/onnxscript/rewriter/ort_fusions/bias_gelu.py
+++ b/onnxscript/rewriter/ort_fusions/bias_gelu.py
@@ -25,12 +25,20 @@ class BiasGeluFusion(pattern.RewriteRuleClassBase):
     def pattern(self, op, x, y):
         gelu_add = op.Add(x, y)
         if self._contrib_op:
-            return op.Gelu(gelu_add, _domain="com.microsoft")
+            return op.Gelu(gelu_add, _domain="com.microsoft", _outputs=["gelu"])
         else:
-            # match behavior of onnxruntime fusion_biasgelu.py
-            return op.Gelu(gelu_add)
+            return op.Gelu(gelu_add, _outputs=["gelu"])
 
-    def rewrite(self, op, x, y):
+    def check(self, op, gelu, **_) -> pattern.MatchResult:
+        check_result = pattern.MatchResult()
+        approximate = gelu.producer().attributes.get("approximate", None)
+        if approximate is not None and approximate.value == "tanh":
+            return check_result.fail(
+                "Gelu operator with 'approximate' set to 'tanh' is not supported."
+            )
+        return check_result
+
+    def rewrite(self, op, x, y, **_):
         return op.BiasGelu(x, y, _domain="com.microsoft")
 
 
--- a/onnxscript/rewriter/ort_fusions/bias_gelu.py
+++ b/onnxscript/rewriter/ort_fusions/bias_gelu.py
@@ -6,17 +6,33 @@ from onnxscript.rewriter import _fusion_utils, pattern
 
 
 class BiasGeluFusion(pattern.RewriteRuleClassBase):
+    def __init__(
+        self,
+        name,
+        *,
+        has_contrib_op: bool,
+    ):
+        super().__init__(name)
+        self._has_contrib_op = has_contrib_op
+
     def pattern(self, op, x, y):
         gelu_add = op.Add(x, y)
-        return op.Gelu(gelu_add, _domain="com.microsoft")
+        # see: gh-2362. Match against Gelu from onnx op or contrib ops.
+        if self._has_contrib_op:
+            return op.Gelu(gelu_add, _domain="com.microsoft")
+        else:
+            return op.Gelu(gelu_add)
 
     def rewrite(self, op, x, y):
         return op.BiasGelu(x, y, _domain="com.microsoft")
 
 
-_rule = BiasGeluFusion.rule()
-
-bias_gelu_rules = pattern.RewriteRuleSet([_rule])
+bias_gelu_rules = pattern.RewriteRuleSet(
+    [
+        BiasGeluFusion.rule("gelu_onnx_op", has_contrib_op=False),
+        BiasGeluFusion.rule("gelu_contrib_op", has_contrib_op=True),
+    ]
+)
 
 
 fuse_bias_gelu = _fusion_utils.apply_fusion_rules(bias_gelu_rules)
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -6,17 +6,33 @@
 
 
 class BiasGeluFusion(pattern.RewriteRuleClassBase):
+    def __init__(
+        self,
+        name,
+        *,
+        has_contrib_op: bool,
```
justinchuby:
```suggestion
        contrib_op: bool,
```

could you add a docstring (google style) to the class to describe this option?

onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -6,17 +6,33 @@
...
+    ):
+        super().__init__(name)
+        self._has_contrib_op = has_contrib_op
+
     def pattern(self, op, x, y):
         gelu_add = op.Add(x, y)
-        return op.Gelu(gelu_add, _domain="com.microsoft")
+        # see: gh-2362. Match against Gelu from onnx op or contrib ops.
```
justinchuby:
```suggestion
```

Would be good to describe by the docstring

onnxscript/rewriter/ort_fusions/bias_gelu_test.py
```diff
@@ -4,25 +4,41 @@
...
+                gelu_add = op.Add(x, y)
+                gelu = msft_op.Gelu(gelu_add)
+                return gelu
+        else:
+
+            @script()
+            def bias_gelu_model(x, y):
+                gelu_add = op.Add(x, y)
+                gelu = op.Gelu(gelu_add)
+                return gelu
```
justinchuby:
```suggestion
                return msft_op.Gelu(gelu_add)
        else:

            @script()
            def bias_gelu_model(x, y):
                gelu_add = op.Add(x, y)
                return op.Gelu(gelu_add)
```

justinchuby (commented):
Thank you! Some minor comments

onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -6,17 +6,33 @@
...
+        if self._has_contrib_op:
+            return op.Gelu(gelu_add, _domain="com.microsoft")
+        else:
+            return op.Gelu(gelu_add)
```
justinchuby: The gelu op has an optional attribute https://onnx.ai/onnx/operators/onnx__Gelu.html#attributes. Do we want the fusion when it is set?

KarelZe: I was somewhat indecisive about the `approximation` attr. Onnxruntime seems to fuse both variants. https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/fusion_biasgelu.py Otherwise it would be safer to match against default + explicit "none" like: 
```python
...
elif approximate == "none":
	return op.Gelu(gelu_add, approximate=approximate)
else:
	return op.Gelu(gelu_add)
``` 
Would lean towards first. @justinchuby Do you have any opinion on this?

justinchuby: I think matching what the onnxruntime tools does is good. Could you add tests for both when `approximate` is set and when it isn't?

onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -6,17 +6,39 @@
 
 
 class BiasGeluFusion(pattern.RewriteRuleClassBase):
+    """Fuses a Bias-Gelu pattern into a single BiasGelu operator.
+
+    Attributes:
+        contrib_op (bool): If True, matches the Gelu operator from the 'com.microsoft' domain.
+                           If False, matches the standard ONNX Gelu operator.
```
justinchuby: nit formatting
```suggestion
        contrib_op (bool): If True, matches the Gelu operator from the 'com.microsoft' domain.
            If False, matches the standard ONNX Gelu operator.
```

justinchuby (approved):
Thanks!

titaiwangms:
> Sorry for putting this on hold! As we also have the same issue with the newly introduced ops like Attention and RoataryEmbeddings, @gramalingam and I had a discussion about introducing a dispatcher-alike mechanism. We have not decided yet how to do that. We will reply to this PR as soon as we have a conclusion.

Sorry, I just realized this PR is only about adding a pattern not rewrite. Based on the PR: https://github.com/microsoft/onnxruntime/pull/19560, it seems that we need to specify attribute=none in the pattern. tanh is another Gelu.

justinchuby:
> > Sorry for putting this on hold! As we also have the same issue with the newly introduced ops like Attention and RoataryEmbeddings, @gramalingam and I had a discussion about introducing a dispatcher-alike mechanism. We have not decided yet how to do that. We will reply to this PR as soon as we have a conclusion.
> 
> Sorry, I just realized this PR is only about adding a pattern not rewrite. Based on the PR: [microsoft/onnxruntime#19560](https://github.com/microsoft/onnxruntime/pull/19560), it seems that we need to specify attribute=none in the pattern. tanh is another Gelu.

I think we should match all attribute cases right? Both approximate specified and unspecified, as well as all approximate options when specified

onnxscript/rewriter/ort_fusions/bias_gelu.py
gramalingam: This is a bit confusing. I took a look at the CPU implementation of [BiasGelu](https://github.com/microsoft/onnxruntime/blob/46caf476d6069837faa4162c36e0ca86324d2d27/onnxruntime/contrib_ops/cpu/bert/bias_gelu.cc#L24) ... even though the underlying template supports both approximate (Tanh) and precise (Erf), the instantiation shows that BiasGelu is meant to be for the precise (Erf) Gelu.

So, I think it may be better to do the fusion only for the precise Gelu?

As for the onnxruntime fusion: I think it does not check the domain? Which might be an old bug? Perhaps it was meant to fuse only the contrib op and not the ONNX standard one?  I don't know, I am just conjecturing.

In one sense, replacing the approximate by more precise may not be a problem ... on the other hand, it could lead to numeric differences which may matter?

titaiwangms: I think it's meant for contrib op Gelu. The code was written (2021) before ONNX Gelu kernel added (2024). I think it's better we replace `approximate` = none and default.

KarelZe: @titaiwangms Thanks for your responses. I will update the pr accordingly. Just out of curiosity, is there a straight-forward way in onnxscript to match against an optional attr incl. default + explicit `'none'` aside from checking both variants explicitly?

justinchuby: I don’t think there is because the matches does not consider the operator’s schema. Maybe @gramalingam can comment more?

titaiwangms: Yeah, I think we need to specify and check both. It's clear anyway.

KarelZe: @titaiwangms / @justinchuby I now exclude `approximate='tanh'` from fusion. Also, added neg. test case.

onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -6,17 +6,48 @@
...
-    def rewrite(self, op, x, y):
+        if self._contrib_op:
+            return op.Gelu(gelu_add, _domain="com.microsoft", _outputs=["gelu"])
+        else:
+            return op.Gelu(gelu_add, _outputs=["gelu"])
+
+    def check(self, op, gelu, **_) -> pattern.MatchResult:
+        check_result = pattern.MatchResult()
+        approximate = gelu.producer().attributes.get("approximate", None)
+        if approximate is not None and approximate.value == "tanh":
```
justinchuby:
```suggestion
        approximate = gelu.producer().attributes.get_string("approximate")
        if approximate is not None and approximate == "tanh":
```
nit: 
https://onnx.ai/ir-py/api/generated/onnx_ir._graph_containers.Attributes.html#onnx_ir._graph_containers.Attributes.get_string

justinchuby (approved):
A final comment and we are ready to merge. Thank you!

onnxscript/rewriter/ort_fusions/bias_gelu.py
KarelZe: Thanks. Didn't know it existed :-D

justinchuby:
@KarelZe our ci pipeline is broken. We will fix soon

justinchuby (approved):

PR #2365, opened 2025-06-04
Update autocast.py to fix attribute creation error

justinchuby:
This change should fix the type of errors like below (reported in https://github.com/pytorch/pytorch/issues/153214#issuecomment-2941500656):

```pytb
  File "/usr/local/lib/python3.12/dist-packages/google/protobuf/internal/python_message.py", line 589, in init
    _ReraiseTypeErrorWithFieldName(message_descriptor.name, field_name)
  File "/usr/local/lib/python3.12/dist-packages/google/protobuf/internal/python_message.py", line 478, in _ReraiseTypeErrorWithFieldName
    raise exc.with_traceback(sys.exc_info()[2])
  File "/usr/local/lib/python3.12/dist-packages/google/protobuf/internal/python_message.py", line 587, in init
    setattr(self, field_name, field_value)
  File "/usr/local/lib/python3.12/dist-packages/google/protobuf/internal/python_message.py", line 730, in field_setter
    raise TypeError(
TypeError: Cannot set onnx.AttributeProto.type to <AttrType.TENSOR: 4>: <AttrType.TENSOR: 4> has type <class 'onnx.onnx_cpp2py_export.defs.OpSchema.AttrType'>, but expected one of: (<class 'int'>,) for field AttributeProto.type
```

### DIFF
--- a/onnxscript/_internal/autocast.py
+++ b/onnxscript/_internal/autocast.py
@@ -45,7 +45,7 @@ def pyvalue_to_onnx_attribute(
     key: str,
     value: Any,
     name_generator: Callable[[], str],
-    attr_type: Optional[onnx.AttributeProto.AttributeType] = None,
+    attr_type: onnx.AttributeProto.AttributeType | None = None,
 ) -> onnx.AttributeProto:
     """Helper function to create an ONNX AttributeProto.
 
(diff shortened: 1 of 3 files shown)


onnxscript/_internal/autocast.py
```diff
@@ -45,7 +45,7 @@ def pyvalue_to_onnx_attribute(
     key: str,
     value: Any,
     name_generator: Callable[[], str],
-    attr_type: Optional[onnx.AttributeProto.AttributeType] = None,
+    attr_type: onnx.AttributeProto.AttributeType | None = None,
```
titaiwangms, line 48: Curious why this was not caught until now? Is it that more than one type (besides `onnx.AttributeProto.AttributeType`) could be passed in through `attr_type`?

justinchuby: Somehow there is also onnx.onnx_cpp2py_export.defs.OpSchema.AttrType. I think starting from a particular protobuf version it became more strict in accepting numbers (a guess)

justinchuby:
Updated to fix the callers instead

titaiwangms:
Are failing checks more than usual?

justinchuby:
> Are failing checks more than usual?

They seem like the usual windows flakiness & unresolved issues with the latest ort. We should fix them though at some point.

PR #2368, opened 2025-06-07
Cast-cast elimination

gramalingam:
Enable the cast-cast simplification to a single cast in a couple of cases where it is valid. This shows up in examples like SmolLM (FP16) and is needed for fusion-pattern to work.

Also: add display of replaced and replacing nodes in fusion in verbose mode.

### DIFF
--- a/onnxscript/rewriter/llama_rule_sets.py
+++ b/onnxscript/rewriter/llama_rule_sets.py
@@ -51,22 +51,30 @@ class CastIdentity(orp.RewriteRuleClassBase):
 class CastCast(orp.RewriteRuleClassBase):
     """Replaces ``Cast(Cast(X, ...), to=to)`` by ``Cast(X, to=to)``."""
 
-    _allowed_tensor_types: ClassVar = {
-        ir.DataType.FLOAT,
-        ir.DataType.FLOAT16,
-        ir.DataType.BFLOAT16,
-        ir.DataType.DOUBLE,
-    }
+    # Simplify "cast type1 => type2 => type3" to "cast type1 => type3".
+    # This rule is not valid for all combinations of types: e.g.,
+    # it is not valid for float32 => float16 => float32 or float32 => int32 => string.
+    # TODO: fill out the list of allowed combinations: the following is just a couple
+    # that shows up in practice where it is valid
+    _allowed_type2_type3: ClassVar = frozenset(
+        {
+            (ir.DataType.FLOAT, ir.DataType.FLOAT16),
+            (ir.DataType.FLOAT, ir.DataType.BFLOAT16),
+        }
+    )
 
     def pattern(self, op, x, to, to_ignored):
         return op.Cast(op.Cast(x, to=to_ignored), to=to)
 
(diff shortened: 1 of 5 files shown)


onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -51,22 +51,30 @@ def check(self, context, x, to) -> orp.MatchResult:
...
 
     def check(self, context, x: ir.Value, to: ir.Attr, to_ignored: ir.Attr) -> orp.MatchResult:
         check_result = orp.MatchResult()
-        if to.as_int() not in self._allowed_tensor_types:
-            return check_result.fail(f"Output type {to.as_int()} is not allowed")
-        if to_ignored.as_int() not in self._allowed_tensor_types:
-            return check_result.fail(f"Ignored type {to_ignored.as_int()} is not allowed")
+        type2 = to_ignored.as_int()
+        type3 = to.as_int()
+        if (type2, type3) not in self._allowed_type2_type3:
```
justinchuby, line 73: Maybe when type2==type3 or when both `.is_floating_point()` and type2.itemsize > type3.itemsize?

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -51,22 +51,30 @@ def check(self, context, x, to) -> orp.MatchResult:
...
     def pattern(self, op, x, to, to_ignored):
         return op.Cast(op.Cast(x, to=to_ignored), to=to)
 
     def check(self, context, x: ir.Value, to: ir.Attr, to_ignored: ir.Attr) -> orp.MatchResult:
         check_result = orp.MatchResult()
-        if to.as_int() not in self._allowed_tensor_types:
-            return check_result.fail(f"Output type {to.as_int()} is not allowed")
-        if to_ignored.as_int() not in self._allowed_tensor_types:
-            return check_result.fail(f"Ignored type {to_ignored.as_int()} is not allowed")
+        type2 = to_ignored.as_int()
```
justinchuby, line 71:
```suggestion
        type2 = it.DataType(to_ignored.as_int())
```

onnxscript/rewriter/llama_rule_sets.py
```diff
@@ -51,22 +51,30 @@ def check(self, context, x, to) -> orp.MatchResult:
...
+        type3 = to.as_int()
```
justinchuby, line 72:
```suggestion
        type3 = ir.DataType(to.as_int())
```

onnxscript/rewriter/llama_rule_sets.py
gramalingam: The first one is already covered by redundant cast elimination. The second one is what I think too (at least for float types; something similar for int types too may hold), but am keeping it simple here. (Do these methods is_floating_point / itemsize exist in ir.DataType?)

justinchuby: Yes the methods exist. I think it’s pretty simple to implement

justinchuby (approved):

PR #2371, opened 2025-06-10
Add aten_maximum_real

xadupre:


justinchuby (changes requested):
Explained above and for further discussions

titaiwangms:
> Looks like max.other is an alias of maximum (https://github.com/pytorch/pytorch/blob/b44306d3681d5b248e6b439d293ea0d5a8903a61/aten/src/ATen/native/native_functions.yaml#L10064-L10069). I think we can do one of two things:
> 
> 1. Unregister max.other and hope that it gets decomposed to maximum, which we have a type promotion rule for.
> 2. Add `max.other` to entries in https://github.com/pytorch/pytorch/blob/b44306d3681d5b248e6b439d293ea0d5a8903a61/torch/onnx/_internal/fx/passes/type_promotion.py#L828-L833

I will take a look and update it here.

titaiwangms:
#2377 should fix what this PR tried to fix.

PR #2374, opened 2025-06-12
Fusion extensions to improve GQA fusion

gramalingam:
Various extensions to improve GQA fusion.

* Move key-transpose into SDPA fusion and clean it up
* Extend cos-sin-cache fusion to handle a new pattern
* Reorder GQA and MHA rules
* Introduce MaskedGQA, since many uses in practice generated GQA with a mask
* MaskedGQA is subsequently simplified to ORT's GQA if the mask can be verified to be causal.


onnxscript/rewriter/ort_fusions/sdpa.py
```diff
@@ -132,20 +158,29 @@ def rewrite(
...
+        sdpa_args = [query, key, value]
         if mask is not None:
             sdpa_args.append(mask)
         # If the scale is None, SDPA will use the default scaling factor, which is 1/sqrt(head_size).
-        return op.SDPA(*sdpa_args, scale=self._scale, _domain="ai.onnxruntime.fusion")
+        return op.SDPA(
+            *sdpa_args,
+            scale=self._scale,
+            key_format=key_format,
+            _domain="ai.onnxruntime.fusion",
```
justinchuby: I wonder if something like `ai.onnxruntime.__fusion__` or `ai.onnxruntime._fusion` will make it clear that this is internal use only

justinchuby (approved):

onnxscript/rewriter/ort_fusions/sdpa.py
gramalingam: Sure. The underscore makes sense. I wonder, though, about the "onnxruntime" part ... these rules can potentially be migrated to do fusion for opset23 standard onnx as well, at which point some of the fusions may not be ORT specific. Not sure if some other name would be better.

gramalingam: Will do it separately to avoid need for another approval

gramalingam:
@justinchuby your approval expired :-(

justinchuby (approved):

PR #2375, opened 2025-06-12
[DRAFT] Merge metadata props when fusing

bmehta001:


onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -531,6 +531,15 @@ def _apply_to_graph_or_function(
                     f = ir.Function(domain, name, overload, graph=graph, attributes=())
                     model.functions[f.identifier()] = f
 
+                # If we are fusing nodes, update the metadata props of the new node(s)
+                if delta.match.nodes and delta.new_nodes:
+                    # Concatenate metadata props from all original nodes
+                    fused_metadata_props = "Fused: " + "\t\n".join(
+                        n.metadata_props for n in delta.match.nodes if getattr(n, "metadata_props", None)
+                    )
```
justinchuby: n will always have metadata_props.

onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -531,6 +531,15 @@ def _apply_to_graph_or_function(
...
+                    # Assign to all new nodes (or just the first, depending on your policy)
+                    delta.new_nodes[0].metadata_props += fused_metadata_props
```
justinchuby: Prefer metadata_props.update()

onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -531,6 +531,15 @@ def _apply_to_graph_or_function(
                     f = ir.Function(domain, name, overload, graph=graph, attributes=())
                     model.functions[f.identifier()] = f
 
+                # If we are fusing nodes, update the metadata props of the new node(s)
+                if delta.match.nodes and delta.new_nodes:
+                    # Concatenate metadata props from all original nodes
+                    fused_metadata_props = "Fused: " + "\t\n".join(
```
justinchuby: I would consider creating special reconciliation logic for known fields

gramalingam: +1. I am also confused a bit here. Isn't metadata_props effectively a dict[str, str]? I am not sure how the proto is represented in onnx-ir, but assuming it is a dictionary. We should be merging the dictionaries, using special merging logic for known entries in the dictionary.

This may require us to extend the definition/documentation of specific metadata fields. For example, we may need a way to represent a set of source-locations, instead of a single source-location, for a source-location metadata.

onnxscript/rewriter/_rewrite_rule.py
```diff
@@ -531,14 +531,22 @@ def _apply_to_graph_or_function(
                     f = ir.Function(domain, name, overload, graph=graph, attributes=())
                     model.functions[f.identifier()] = f
 
-                # If we are fusing nodes, update the metadata props of the new node(s)
+                # If we are fusing nodes, update the docstring of the new node(s)
+                attributes = ["namespace", "pkg.torch.onnx.class_hierarchy", "pkg.torch.onnx.fx_node", "pkg.torch.onnx.name_scopes", "pkg.torch.onnx.stack_trace"]
```
justinchuby, line 535: For namespace, class_hierarchy and name_scopes, we should parse the info and find the common leading scopes.

PR #2377, opened 2025-06-12
[torchlib] Unregister aten::max.other

titaiwangms:
Issue revealed by https://github.com/microsoft/onnxscript/pull/2371, which aten.max.other is lack of matching overload. It's caused by missing type promotion. The reason is that aten::max.other (binary max) is an alias of aten::maimum.default. Thus, iwhen type promotion pass dispatches torch.max through `__torch__dispatch__`, it does not find aten::max.other (However, I am not sure how `make_fx` dispatches torch.max to aten::max.other).

The existence of aten::max.other looks like a legacy code: https://github.com/pytorch/pytorch/pull/42579.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -5382,14 +5382,14 @@ def aten_min_dim(self: TReal, dim: int, keepdim: bool = False) -> Tuple[TReal, T
     return result, indices
 
 
-@torch_op(("aten::minimum", "aten::min.other"))
+@torch_op("aten::minimum")
 def aten_minimum(self: TReal, other: TReal) -> TReal:
     """minimum(Tensor self, Tensor other) -> Tensor"""
 
     return op.Min(self, other)
 
 
-@torch_op(("aten::minimum", "aten::min.other"))
+@torch_op("aten::minimum")
 def aten_minimum_bool(self: BOOL, other: BOOL) -> BOOL:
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5315,14 +5315,14 @@ def aten_max_dim(self: TReal, dim: int, keepdim: bool = False) -> Tuple[TReal, I
...
-@torch_op(("aten::maximum", "aten::max.other"))
+@torch_op("aten::maximum")
 def aten_maximum(self: TReal, other: TReal) -> TReal:
     """maximum(Tensor self, Tensor other) -> Tensor"""
 
     return op.Max(self, other)
 
 
-@torch_op(("aten::maximum", "aten::max.other"))
+@torch_op("aten::maximum")
```
justinchuby, line 5323: Is there also min, and potentially others?

titaiwangms: Added min. I am not sure about the onw about Add raised by @xadupre with phi3.5 yet. We just need to check them one by one to understand the situation. But personally, I think the type promotion pass is robust enough to cover around 95% cases, although I missed the context of pow?

justinchuby: For pow onnx allows the two operands to be different types, so I removed the promotion rules

justinchuby (approved):

PR #2379, opened 2025-06-13
[CI] Fix execnet.gateway_base.DumpError: can't serialize <class 'numpy.int64'>

titaiwangms:


gramalingam (approved):
Hope this works ... it's a bit mysterious

PR #2393, opened 2025-06-15
fix: check for rank of bias in bias-gelu fusion🐛

KarelZe:
Follow-up to #2364.

I  noticed that the current implementation `BiasGeluFusion` from #2364 does not check for the dimensions of the bias term, which can lead to errors, as the bias input for `BiasGelu(...)` is expected to be 1D (see [here]https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftbiasgelu). 

This pr adds:
- additional checks for dim of bias
- additional test cases

Sorry for the inconvenience.

@justinchuby @titaiwangms

### DIFF
--- a/onnxscript/rewriter/ort_fusions/bias_gelu.py
+++ b/onnxscript/rewriter/ort_fusions/bias_gelu.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
-from onnxscript.rewriter import _fusion_utils, pattern
+from onnxscript.rewriter import _fusion_utils, _ir_utils, pattern
 
 
 class BiasGeluFusion(pattern.RewriteRuleClassBase):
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/bias_gelu.py
```diff
@@ -22,24 +22,29 @@ def __init__(
         super().__init__(name)
         self._contrib_op = contrib_op
 
-    def pattern(self, op, x, y):
-        gelu_add = op.Add(x, y)
+    def pattern(self, op, input, bias):
+        gelu_add = op.Add(input, bias)
```
KarelZe, line 26: Open: I guess we should check here best for both combinations (bias term first + bias term last).

Open: onnxruntime also checks for `MatMul` in parent path https://github.com/microsoft/onnxruntime/blob/c7fdb54c3ef95bc295aef1b0e6a2867be5d0331e/onnxruntime/python/tools/transformers/fusion_biasgelu.py#L16

justinchuby: You can use `commute()` to support this: https://github.com/microsoft/onnxscript/blob/b76e1b324bf3ce955cf7823a3b680ed4831a3209/docs/tutorial/rewriter/commute.md?plain=1#L2

KarelZe: @justinchuby Thanks for your quick response. 👍  Seems to do the trick. Will do some more testing.

titaiwangms (approved):

justinchuby (approved):

PR #2396, opened 2025-06-17
fix: handling of default attrs in SimplifiedLayerNormalization + LayerNormalization🐛

### LINKED ISSUE #2378: Handling of default attrs of `LayerNormalization` in `SkipLayerNormFusion`
Thanks for `onnxscript`:100: . 

I'm currently comparing operator fusion in onnxruntime and onnxscript for BART and thereby noticed, that the `SkipLayerNormFusion` does currently not fuse ops, if `stash_type` is at default (`=1`) or `epsilon` is at default (`=1e-5`) for [`LayerNormalization`](https://onnx.ai/onnx/operators/onnx__LayerNormalization.html#). 

The current rule implementation only fuses if both are args explicitly set.

### Minimal example:

```python
import onnxscript
import onnxscript.ir as ir
from onnxscript.rewriter.ort_fusions.skip_normalization import (
    skip_layer_normalization_ruleset,
)

tape = ir.tape.Tape()

x: ir.Value = tape.initializer(ir.tensor([[[1.0, 2.0, 3.0]]], name="x"))
skip: ir.Value = tape.initializer(ir.tensor([[[0.5, 0.5, 0.5]]], name="skip"))

bias: ir.Value = tape.initializer(ir.tensor([0.5, 0.5, 0.5], name="bias"))
scale: ir.Value = tape.initializer(ir.tensor([1.0, 1.0, 1.0], name="scale"))

optionals = {} # {"stash_type": 1, "epsilon": 1e-4} # <-- switch here.

skip_sum = tape.op("Add", [skip, x])
normalized = tape.op(
    "LayerNormalization",
    [skip_sum, scale, bias],
    attributes= optionals | {"axis": -1},
)

onnx_model = ir.Model(
    graph := ir.Graph(
        inputs=[x],
        outputs=[normalized],
        nodes=tape.nodes,
        initializers=tape.initializers,
        opset_imports={"": 20},
    ),
    ir_version=10,
)

model_with_rewrite_applied = onnxscript.rewriter.rewrite(
    onnx_model, pattern_rewrite_rules=skip_layer_normalization_ruleset
)
ir.save(model_with_rewrite_applied, "layer_norm_fusion.onnx")
```

Original pr #2259. 

Would you be open to match against alternative default args for `LayerNormalization`? If yes, I'd be happy to contribute a fix.

@shubhambhokare1 @justinchuby

KarelZe:
`SkipLayerNormFusion` does currently not fuse ops, if stash_type is at default (=1) or epsilon is at default (=1e-5) for [`LayerNormalization`](https://onnx.ai/onnx/operators/onnx__LayerNormalization.html#) and `SimplifiedLayerNormalization`

This pr:
- fixes handling default attrs in `LayerNormalization`, `SimplifiedLayerNormalization`
- adds BART encoder as new test model. I added this model as some of the stash types are at default. The model is versatile and can also be used to test other fusions e.g., `EmbedLayerNormalization`.
- allows for commuted inputs.

Closes #2378.

@shubhambhokare1 @justinchuby Could you please review? Any feedback is greatly appreciated.


onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -141,24 +136,16 @@ def __init__(self, name: str, has_bias: bool = False, bias_pre_add: bool = False
...
             input = op.Add(input, bias)
 
-        # Support different combinations of addition of input and skip
-        skip_sum_pattern_1 = op.Add(skip, input)
-        skip_sum_pattern_2 = op.Add(input, skip)
-        skip_sum = pattern.OrValue([skip_sum_pattern_1, skip_sum_pattern_2], name="skip_sum")
-
+        skip_sum = op.Add(input, skip)
         if self._has_bias and not self._bias_pre_add:
             skip_sum = op.Add(skip_sum, bias)
```
KarelZe, line 159: I chose to enable commute(...), as we didn't check for all variants in this addition and only in the lines above.

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -93,10 +87,11 @@ def rewrite(
         skip,
         gamma,
         bias,
-        epsilon,
-        stash_type,
+        simplified_layer_norm,
         **_,
     ):
+        epsilon = simplified_layer_norm.producer().attributes.get_float("epsilon")
```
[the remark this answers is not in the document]
KarelZe: I guess there is no stash type for fused layer norm ops? https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.SkipLayerNormalization

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -141,24 +140,21 @@ def __init__(self, name: str, has_bias: bool = False, bias_pre_add: bool = False
...
-        skip_sum = pattern.OrValue([skip_sum_pattern_1, skip_sum_pattern_2], name="skip_sum")
-
+        skip_sum = op.Add(input, skip)
         if self._has_bias and not self._bias_pre_add:
             skip_sum = op.Add(skip_sum, bias)
+
         normalized = op.LayerNormalization(
             skip_sum,
             gamma,
             beta,
```
KarelZe, line 164: @gramalingam `beta` is an optional input. I'd lean toward matching both variants (w and w/o bias).

onnxscript/rewriter/ort_fusions/skip_normalization.py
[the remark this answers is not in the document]
gramalingam: I assume you mean also for SimplifiedLayerNorm and SkipSimplifierLayerNorm? Unfortunately, I don't see the doc for the first op. But if it is absent in both ops, it seems safe to ignore it.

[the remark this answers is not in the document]
gramalingam: Let's consider the two cases separately:

(a) For LayerNorm (the rule down below), we are starting with the ONNX op, which supports stash_type. But the SkipLayerNorm in ORT doesn't seem to support stash_type. Is my understanding correct? If so, the rewrite should have a check condition to see if the stash_type has the value supported by the SkipLayerNorm ... if not, we should skip the optimization.

(b) For SimplifiedLayerNorm, if stash_type is not supported by either op, we can ignore it.

However, for (a): we should understand what the default behavior of the ORT ops are: do they use a value of stash_type == FP32 or do they use a stash_type == input-type? The two are different.

onnxscript/rewriter/ort_fusions/skip_normalization.py
gramalingam: I am curious, did you see patterns that this missed? In principle this is ok, but it could increase the fusion time. We also need to update the implementation of commute() to make use of pattern-disjunction, which will be more efficient.

onnxscript/rewriter/ort_fusions/skip_normalization.py
gramalingam: Sorry, do you mean all 4 combinations (w and w/o beta, w/wo bias)? That can be done by dropping beta here, and specifying `_allow_other_inputs=True`. The rewriter should then forward the corresponding inputs to the rewritten node.

onnxscript/rewriter/ort_fusions/skip_normalization.py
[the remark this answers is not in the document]
KarelZe: Thank you. 👍 Yes, neither `SkipLayerNorm` nor `SimplifiedSkipLayerNorm` seems to support an external stash_type. (see [op.LayerNormalization](https://github.com/microsoft/onnxruntime/blob/rel-1.20.0/docs/ContribOperators.md#commicrosoftskiplayernormalization) + [SkipLayerNormalization](https://github.com/microsoft/onnxruntime/blob/rel-1.20.0/docs/ContribOperators.md#com.microsoft.SkipLayerNormalization) ). I'm not very proficient with c/c++, but to my understanding the internal precision for statistics seems to be depend on `strict` mode *and* from the input type `T` e.g., float or bfloat16. For strict mode the computation should be done in fp32. For non-strict mode, the precision depends on the input precision https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cuda/bert/skip_layer_norm.cc @gramalingam Should we then check for `stash_type=None or 1` to be safe?

onnxscript/rewriter/ort_fusions/skip_normalization.py
KarelZe: Sorry meant with and w/o `beta`. I'm wondering, if I drop `beta` here, would it still be forwarded correctly, as beta is the third input to `LayerNormalization` but fourth input to `SkipLayerNormalization`? (https://onnx.ai/onnx/operators/onnx__LayerNormalization.html and https://github.com/microsoft/onnxruntime/blob/rel-1.20.0/docs/ContribOperators.md#commicrosoftskiplayernormalization ) In my quick tests, I got different outputs.

onnxscript/rewriter/ort_fusions/skip_normalization.py
KarelZe: Thanks for asking, I've only looked at a limited number of models so far, which all had the bias term as second input. I can undo this change  to avoid performance regressions 👍

justinchuby (approved):

KarelZe:
@justinchuby Sorry was busy at work. Resolved one issue, for the other I'd need @gramalingam s input. 👍

onnxscript/rewriter/ort_fusions/skip_normalization.py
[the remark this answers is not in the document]
gramalingam: Sorry about the belated response. Yes, I think that for SkipLayerNorm, we should check stash_type is FLOAT (that is, None or 1). Otherwise, even the output-types will not match (for ONNX's LayerNorm and ORT's SkipLayerNorm), so the optimization seems invalid.

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -218,10 +216,10 @@ def rewrite(
         gamma,
         beta,
         bias,
-        epsilon,
-        stash_type,
+        layer_norm,
         **_,
     ):
+        epsilon = layer_norm.producer().attributes.get_float("epsilon")
```
gramalingam: I think you should use `get_float("epsilon", 1e-5)` to use the default value if it is not explicitly specified.

onnxscript/rewriter/ort_fusions/skip_normalization.py
gramalingam: Well, you will have to explicitly construct the value of beta. Something like below:
```py
   layer_norm_inputs = layer_norm.producer().inputs
   beta = layer_norm_inputs[2] if len(layer_norm_inputs) > 2 else None
```
and you can then use beta just like now.

onnxscript/rewriter/ort_fusions/skip_normalization.py
```diff
@@ -93,10 +91,11 @@ def rewrite(
         skip,
         gamma,
         bias,
-        epsilon,
```
gramalingam, line 97: Is epsilon an attribute with a default value in ORT's SimplifiedLayerNorm? If so, it should be specified and used in the `attributes.get_float("epsilon")` line down below. Otherwise, if it is not optional, there doesn't seem to be a reason for this line to change.

justinchuby:
@KarelZe feel free to take a look at the final round of comments. After that I think we are ready to merge. Thanks!

onnxscript/rewriter/ort_fusions/skip_normalization.py
KarelZe: Done.

onnxscript/rewriter/ort_fusions/skip_normalization.py
KarelZe: Yes, defaults to `1e-5`. https://github.com/microsoft/onnxruntime/blob/d9ce6a9d5e29e02fb5dddb8cd064add819d6118d/onnxruntime/core/graph/contrib_ops/contrib_defs.cc#L2981 Updated the code accordingly. BTW for `SkipSimplifiedLayerNorm` epsilon would have no default. https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.SkipSimplifiedLayerNormalization

onnxscript/rewriter/ort_fusions/skip_normalization.py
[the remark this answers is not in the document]
KarelZe: @gramalingam Thanks for your feedback. Added additional checks for `stash_type`. We now only allow `None` or `1`.

KarelZe:
@justinchuby Should be complete now🙌 

Just one more question, would you be open/interested in to adding `EmbedLayerNormalization` fusion to onnxscript? I could maybe contribute some rewrite rules in a separate pr.

justinchuby:
> @justinchuby Should be complete now🙌
> 
> Just one more question, would you be open to /interested in adding `EmbedLayerNormalization` fusion to onnxscript? I could maybe contribute some rewrite rules in a separate pr.

Thanks! Would let @gramalingam comment

gramalingam (approved):

gramalingam:
> @justinchuby Should be complete now🙌
> 
> Just one more question, would you be open to /interested in adding `EmbedLayerNormalization` fusion to onnxscript? I could maybe contribute some rewrite rules in a separate pr.

Sure, that would be great, thanks!

PR #2397, opened 2025-06-18
Updates to the rewriter tutorial

gramalingam:
Updates to the rewriter tutorial

### DIFF
--- a/docs/tutorial/rewriter/examples/erfgelu.py
+++ b/docs/tutorial/rewriter/examples/erfgelu.py
@@ -106,6 +106,7 @@ def apply_rewrite(model):
     )
     return model_with_rewrite_applied
 
+
 ####################################
 # Rewrite Rule as a Class
 # =====================
--- a/pyproject_pylint.toml
+++ b/pyproject_pylint.toml
@@ -2,6 +2,7 @@
 
 [tool.pylint.messages_control]
 disable = [
+  "arguments-differ",               # TODO: abstract methods in Rewriter
   "attribute-defined-outside-init", # TODO: mostly in onnxscript/converter.py
   "cell-var-from-loop",             # Bugbear B023
   "consider-using-from-import",


docs/tutorial/rewriter/commute.md
```diff
@@ -1,5 +1,15 @@
...
 # Utilizing `commute` parameter for pattern-matching
+
+_WARNING_ : Please note that the section below describes a convenience feature for handling commutative operators
+in pattern matching. However, the implementation is a simple, brute-force, technique that generates a collection
+of rewrite-rules from a given rule, taking commutativity of addition and multiplication into account. This can
+lead to an exponential increase in the number of rewrite-rules. So, it should be used with caution. Pattern
+disjunctions (_OR Patterns_) described earlier can be used judiciously to get a somewhat more efficient
+implementation in practice (even though the potential for exponential increase still exists within the
+pattern matching algorithm). Reimplementing commutativity handling using pattern disjunctions is future
+work.
```
justinchuby:
```suggestion

```{warning}
Please note that the section below describes a convenience feature for handling commutative operators
in pattern matching. However, the implementation is a simple, brute-force, technique that generates a collection
of rewrite-rules from a given rule, taking commutativity of addition and multiplication into account. This can
lead to an exponential increase in the number of rewrite-rules. So, it should be used with caution. Pattern
disjunctions (_OR Patterns_) described earlier can be used judiciously to get a somewhat more efficient
implementation in practice (even though the potential for exponential increase still exists within the
pattern matching algorithm). Reimplementing commutativity handling using pattern disjunctions is future
work.
```

```

File: docs/tutorial/rewriter/conditional_rewrite.md [hunk 2]
```diff
@@ -31,7 +31,11 @@ The target pattern in this case has 5 inputs `input_a`, `input_b`, `shape_a`, `s
 Similarly for writing the condition checking function, we require only `input_a`, `input_b` and `shape_c`. Use `**_` to represent all the unused parameters in the condition matching function signature.
 :::
 
-In order to validate whether matmul broadcast is sufficient, we write a condition checking function as follows:
+In order to validate whether matmul broadcast is sufficient, we write a condition checking function as below.
+Note that the relevant inputs passed to the check function are all instances of `ir.Value`. These represent
```
justinchuby:
```suggestion
Note that the relevant inputs passed to the check function are all instances of :class:`onnx_ir.Value`. These represent
```

gramalingam: Will this work, given that IR has moved to a different repo? Not sure if the old documentation still exists, but still ...

gramalingam: I did add a reference to the overall documentation in the lines below.

justinchuby: There is a cross ref link we need to set up. After that it should work

justinchuby (approved):
Looks like there's lint errors

justinchuby:
Maybe we can disable the # pylint: disable=arguments-differ rule universally in https://github.com/microsoft/onnxscript/blob/f62f3bc81bd6af68b49c2f10c9b0ebc0fd48360e/pyproject_pylint.toml#L4

justinchuby:
> > Maybe we can disable the # pylint: disable=arguments-differ rule universally in
> > https://github.com/microsoft/onnxscript/blob/f62f3bc81bd6af68b49c2f10c9b0ebc0fd48360e/pyproject_pylint.toml#L4
> 
> I think that would be good. It shows up as a "warning", so it wasn't clear to me if it _had_ to be fixed. Anyway: there is no good fix, since the alternatives are to drop the method from the abstract class completely or add an override everywhere. So, this must be a new pylint requirement?

I think we disabled some pylint rules in the rewriter folder but not in the examples folder. That's why it's showing up. Anyways any lintrunner issue (no matter level) will need to be fixed or silenced.

PR #2398, opened 2025-06-18
Move gemm_to_matmul_add rule to ort fusion rules

justinchuby:
Stop decomposing gemm to matmul add by default because it is a more compact representation. Move the ort fusion rules so it keeps functioning for ort.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/_core.py
+++ b/onnxscript/rewriter/ort_fusions/_core.py
@@ -7,9 +7,8 @@ import onnxscript.ir.passes.common as common_passes
 import onnxscript.rewriter.ort_fusions.fused_matmul_rule_sets as fused_matmul_rule_sets
 import onnxscript.rewriter.ort_fusions.shape_optimization as shape_optimization
 from onnxscript.optimizer import optimize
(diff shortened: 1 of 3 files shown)


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -130,7 +129,7 @@ def optimize_for_ort(
         - The optimized `ir.Model` after applying transformer-specific fusions.
         - A dictionary with a count of each of the fusions applied.
     """
-
+    rewrite(model, [gemm_to_matmul_add.rule])
```
gramalingam, line 132: I think we can add the rule to the list of rewrite-rules at the top of the file, which already has other rules.

justinchuby: Does any of the xformer fusion rule depend on it? Currently the list of rewrite rules is run after the xformer fusion rules

gramalingam: Good question. Possible. The fusions do involve MatMul.

gramalingam (approved):

PR #2399, opened 2025-06-18
[optimizer] Replace value.nbytes with  value.size

titaiwangms:
To unify the size limitation in terms of the scale.

Passes use tensor size:
https://github.com/onnx/ir-py/blob/a833ab1e178c70046a414b96c1aafbf78a9b4e17/src/onnx_ir/passes/common/constant_manipulation.py#L124

while optimizer uses nbytes, which could potentially confuse users.


justinchuby (approved):
Thanks!

titaiwangms:
I should probably set a compatible default as the previous one (now in different scale...)

onnxscript/optimizer/_constant_folding.py
```diff
@@ -944,21 +944,21 @@ def new_constant(self, node: ir.Node, value) -> ir.Node | None:
         tensor.name = irvalue.name
         irvalue.const_value = tensor
 
-        if value.nbytes > self.output_size_limit:
+        if value.size > self.output_size_limit:
```
gramalingam, line 947: IIRC, there is probably documentation about the output/input size limit parameters that should also be updated to reflect this.

justinchuby: I copied but forgot to paste: https://github.com/titaiwangms/onnx-script/blob/42fafa475d1d4e44b53b3cecc1ab914b07d2c00e/onnxscript/optimizer/_constant_folding.py#L1193-L1198

justinchuby:
@copilot can you help update the documentation?

onnxscript/optimizer/_constant_folding.py
titaiwangms: Done

onnxscript/optimizer/_constant_folding_test.py
```diff
@@ -15,6 +15,7 @@
 
 class FoldConstantsTest(unittest.TestCase):
     def _fold(self, model: ir.Model | str, onnx_shape_inference=False, **kwargs):
+        print("Folding constants with kwargs:", kwargs)
```
gramalingam: Delete this? I assume this was debugging

onnxscript/optimizer/_constant_folding.py
gramalingam: I don't see the changes in documentation here

justinchuby: I think below

justinchuby (approved):

justinchuby (commented):
there is a merge conflict

PR #2402, opened 2025-06-19
Fix fusion ordering for partial rotary embedding

gramalingam:
The partial-rotary-embedding fusion depends on the cos-sin-cache fusion. Fix the fusion ordering. This is necessary for GQA fusion in models like Phi4 (with partial-rotary-embedding).

TODO: Add test-case. The one I have is huge. Need to create a smaller test-case.


justinchuby (approved):
Add a comment inline?

PR #2408, opened 2025-06-20
Support gpa in aten spda

justinchuby:


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2013,7 +2005,55 @@ def _aten_scaled_dot_product_attention_no_mask_onnx(
...
+        # Same for value
+        value_shape = op.Shape(value)
+        value_shape_expanded = op.Concat(
+            op.Slice(value_shape, [0], [-3]),  # batch and other dims
+            op.Mul(key_heads, repeat_factor),  # expanded head dimension
+            op.Slice(value_shape, [-2], [_INT64_MAX]),  # remaining dims
+            axis=0
+        )
+        value_unsqueezed = op.Unsqueeze(value, [-2])
+        value_tiled = op.Tile(value_unsqueezed, op.Concat(
```
titaiwangms, line 2049: op.Tile does not align to PyTorch inplementation.

```python
    if (
        (q_num_heads != k_num_heads)
        and (q_num_heads % k_num_heads == 0)
        and (k_num_heads == v_num_heads)
    ):
        seq_reps = q_num_heads // k_num_heads
        # Interleave-repeat each KV head: [h0, h0, h1, h1, ...]
        K = np.repeat(K, repeats=seq_reps, axis=1)
        V = np.repeat(V, repeats=seq_reps, axis=1)
```

We should be able to reuse `repeat_interleave` here when it's done.

justinchuby: should we use expand for repeat interleave for simplicity over tile?

titaiwangms: Yeah, https://github.com/onnx/onnx/blob/62f2facfc29b0b6a26247614d56e7c294a6206fc/onnx/defs/nn/defs.cc#L3840-L3856

I wonder if we can just adapt whatever function body is in defs.cc to torchlib? Is there any difference?

justinchuby: Probably not. I must have need using the old implementation

PR #2409, opened 2025-06-20
Add phi2/phi4 test cases for mha/gqa fusion

gramalingam:
Onnxscript extensions:
* Extend onnxscript's toModelProto to allow specification of valueinfos in generated model.
* Extend the onnx-proto to onnxscript converter to serialize valueinfos in the model, so that it can be used in the generated script.

Fusion test cases:
* Add Phi2 (1 layer) and Phi4 (2 layer) test cases for MHA and GQA fusion respectively


onnxscript/rewriter/ort_fusions/models/_phi2lm.py
```diff
@@ -0,0 +1,507 @@
...
+
+import numpy
+import onnx_ir as ir
+
+from onnxscript import script
+from onnxscript.onnx_opset import opset18
+from onnxscript.onnx_types import BOOL, FLOAT, INT64
+
+value_infos = {
+    "model_embed_tokens_weight": FLOAT[51200, 2560],
```
justinchuby, line 15: I understand this is a python type, which doesn't work very well with type checkers. Would it be a good idea to provide for example a tuple like this `(FLOAT, [51200, 2560])` (where `FLOAT:int=ir.DataType.FLOAT`)?

gramalingam: I added these files to the exclusion list in .lintrunner.toml, but doesn't seem to be working. What's wrong with that exclusion line? As for these types, we are already using this for specifying input/output types as well. If we use an alternative format (which would be convenient to avoid these warnings/errors), it would be good to support it uniformly. I don't want to do it in a rush for this PR. Now that we have the IR too, it would be good to have one consistent notation that works across both the IR and here, for representing types with shapes. It may actually be useful to have even more functionality for such shaped-types (like the ability to generate random values).

gramalingam: @justinchuby : any ideas why the .lintrunner.tol exclusion is not working?

justinchuby: MYPY was ignored but ruff was not. I created a suggestion.

onnxscript/rewriter/ort_fusions/models/_phi2lm.py
```diff
@@ -0,0 +1,507 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Generated from Phi2LM 1 Layer ONNX model produced by the new (Dynamo) exporter
```
justinchuby, line 4:
```suggestion
# Generated from Phi2LM 1 Layer ONNX model produced by the new (Dynamo) exporter
# ruff: noqa: F821
```

onnxscript/rewriter/ort_fusions/models/_phi4lm.py
```diff
@@ -0,0 +1,746 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+# Generated from Phi4LM 2 Layer ONNX model produced by the new (Dynamo) exporter
```
justinchuby, line 4:
```suggestion
# Generated from Phi4LM 2 Layer ONNX model produced by the new (Dynamo) exporter
# ruff: noqa: F821
```

onnxscript/rewriter/ort_fusions/models/_phi2lm.py
justinchuby: Should be fixed now

justinchuby (approved):

justinchuby (approved):

PR #2410, opened 2025-06-22
[Rewriter]: fuse successive Relu/Clip nodes

AyoubMDL:
This PR adds the following transformation:
- Relu(Relu(X)) -> Relu
- Relu(Clip(X)) -> Clip
- Clip(Relu(X)) -> Clip
- Clip(Clip(X)) -> Clip


onnxscript/rewriter/fuse_relus_clips_test.py
```diff
@@ -0,0 +1,326 @@
...
+    def clone_model(self, model: ir.Model) -> ir.Model:
+        return ir.from_proto(ir.to_proto(model))
+
+    def run_test(
+        self,
+        base_model: onnx.ModelProto | ir.Model,
+        expected_op_type: str,
+        dtype: str = "float",
+    ):
+        base_model = ir.serde.deserialize_model(base_model)
```
justinchuby:
```suggestion
```

onnxscript/rewriter/fuse_relus_clips_test.py
```diff
@@ -0,0 +1,326 @@
...
+    @property
+    def rng(self):
+        return np.random.default_rng(20250621)
+
+    def clone_model(self, model: ir.Model) -> ir.Model:
+        return ir.from_proto(ir.to_proto(model))
+
+    def run_test(
+        self,
+        base_model: onnx.ModelProto | ir.Model,
```
justinchuby:
```suggestion
        base_model: ir.Model,
```

onnxscript/rewriter/fuse_relus_clips_test.py
```diff
@@ -0,0 +1,326 @@
...
+        model_proto = onnx.parser.parse_model("""
+            < ir_version: 10, opset_import: ["" : 20] >
+            test_model (float[N, 32, 14] X) => (float [N, ?, ?] Y)
+            {
+                x1 = Relu(X)
+                x2 = Relu(x1)
+                Y = Relu(x2)
+            }
+        """)
+        self.run_test(model_proto, expected_op_type="Relu")
```
justinchuby:
```suggestion
        model = ir.from_onnx_text("""
            < ir_version: 10, opset_import: ["" : 20] >
            test_model (float[N, 32, 14] X) => (float [N, ?, ?] Y)
            {
                x1 = Relu(X)
                x2 = Relu(x1)
                Y = Relu(x2)
            }
        """)
        self.run_test(model, expected_op_type="Relu")
```

justinchuby: Here and below, thanks!

justinchuby (commented):
Thank you - I think this can be part of the default rewrite rules

cc @gramalingam

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,179 @@
...
+        if len(node.inputs) > 2:
+            max_clip = node.inputs[2].const_value.numpy()
+
+        return min_clip, max_clip, dtype
+
+    def check(self, context, **kwargs):
+        del context  # Unused
+        check_result = orp.MatchResult()
+
+        # check clip min/max are initializers
```
titaiwangms:
```suggestion
        # check if clip min/max are initializers
```

onnxscript/rewriter/fuse_relus_clips_test.py
```diff
@@ -0,0 +1,320 @@
...
+        """)
+        self.run_test(model, expected_op_type="Clip", dtype=dtype)
+
+    def test_fuse_succesive_clips_no_min(self):
+        model = ir.from_onnx_text("""
+            < ir_version: 10, opset_import: ["" : 20] >
+            test_model (float[N, 32, 14] X) => (float [N, ?, ?] Y)
+            <float max1 = {4.0}, float max2 = {7.0}>
+            {
+                x1 = Clip(X,, max1)
```
titaiwangms, line 314: @justinchuby I didn't know ONNX op accepts None as input. What is this in proto and IR (the min input)?

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,179 @@
...
+        initializers = []
```
titaiwangms: nit: Using initializers as variable name seems a bit confusing here. Something like clip_min_max would be more understandable. Especially I think `is_graph_input()` and `is_initializer()` can express what you are doing here.

onnxscript/rewriter/fuse_relus_clips_test.py
```diff
@@ -0,0 +1,320 @@
...
+                "clip_then_relu",
+                """
+                    x1 = Clip(X, min)
+                    Y = Relu(x1)
+                """,
+                fuse_successive_relu_clip_rule,
+            ),
+        ]
+    )
+    def test_fuse_successive_relu_clip_non_initializers(self, _, nodes, rewrite_rule):
```
titaiwangms: nit: Could you add "successful" or "fail" on the test title so people can know what to expect?

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,179 @@
...
+            second_clip_node = out_second_clip.producer()
+            initializers.extend(
+                [inp for inp in second_clip_node.inputs[1:] if inp is not None]
+            )
+
+        for initializer in initializers:
+            if initializer.is_graph_input():
+                return check_result.fail(f"{initializer.name} is a graph input.")
+
+            if not initializer.is_initializer() or initializer.const_value is None:
```
titaiwangms: Why constants won't work here?

titaiwangms (commented):
Please fix optional lint (it's mainly spelling).

onnxscript/rewriter/fuse_relus_clips.py
titaiwangms: Could you summarize and update the "check logic" in `check()` description?

onnxscript/rewriter/fuse_relus_clips_test.py
justinchuby: in proto it should be `""`. In IR it should be None.

titaiwangms: Yeah, I see what we did in aten_clamp.

titaiwangms: It's surprising that we can provide the second one and leave the first one blank though.

onnxscript/rewriter/fuse_relus_clips.py
AyoubMDL: > Why constants won't work here?

The condition is mainly to check clip min/max are not initializers (inputs). Are you asking if I should also support cases where min and max are Constant nodes ? (is the check I did exclude this case ?)

titaiwangms: Yes, I am curious why we don't include constant nodes here?

AyoubMDL: Do you mean I support or reject Constant nodes ? Also in the rewriter, I supposed that all constant nodes were folded (as ConstantFolding is the first pass applied in the optimizer).

justinchuby: I think the question is this: one would expect that the pattern should include both initialized values and constant nodes. Here the check will reject when the value is produced by a constant node.

AyoubMDL: Maybe I misunderstood const_value attribute, when it is None and the value is initializer, it means that the value is produced by a constant node right ?

justinchuby: That would be inaccurate. A value is an initializer only when it is part of the graph.initializers mapping. You may use `.is_graph_input()` to check if the value is an input. You may use `ir.convenience.get_const_tensor()` to get the tensor value from either an initialized value or a value that comes from a Constant node.

AyoubMDL: Ok I see, thanks. While doing the change, the constant node gets disconnected from the graph but never removed. Should I also handle the removal of Constant nodes ? I thought the rewriter does it by default.

justinchuby: Yes. You can leave it to the rewriter or the final DCE pass, since it's not always safe to remove them during rewrites (it needs to have a single consumer which is disconnected)

AyoubMDL: Added the change. Thanks @titaiwangms for the catch !

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,188 @@
...
+        not graph inputs and are constant values (i.e., provided by Constant nodes or initializers).
+
+        Returns:
+            MatchResult:
+                Success if we need to replace the pattern, Failure otherwise.
+        """
+        del context  # Unused
+        check_result = orp.MatchResult()
+
+        # Check if clip min/max are initializers
```
titaiwangms: Now we need to change this comment.

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,188 @@
...
+            clip_min_max.extend(
+                [inp for inp in second_clip_node.inputs[1:] if inp is not None]
+            )
+
+        for m in clip_min_max:
+            if m.is_graph_input():
+                return check_result.fail(f"{m.name} is a graph input.")
+
+            if ir.convenience.get_const_tensor(m) is None:
+                return check_result.fail(f"{m.name} is not a constant.")
```
titaiwangms, line 108: Thank you for the change. Do you have test cases that are guarding the inputs of the nodes are from constant nodes?

AyoubMDL: I am not sure to understand your question

titaiwangms: Now we have constant node covered, did you also add a test for it?

AyoubMDL: Yes, it was on this commit e30950c5ae72d3a290b56777db042dfe4f9b26e3

titaiwangms:
@justinchuby Is the error "AttributeError: module 'onnx_ir.convenience' has no attribute 'get_const_tensor'" suggesting a newer version of onnx-ir?

justinchuby (approved):

onnxscript/rewriter/fuse_relus_clips.py
justinchuby: @AyoubMDL could you update this?

justinchuby:
@titaiwangms turns out we also want to update the nox file. It needs to stay pinned because we want to test with the lowest supported version of onnx-ir.

onnxscript/rewriter/fuse_relus_clips.py
AyoubMDL: Done (tell me if it is convenient)

titaiwangms:
@AyoubMDL  Looks like there is a case failing.

justinchuby:
Strange - could you help ensure the tensor min_clip has a proper ir.Shape defined? If not there may be a bug in ir.tensor(), or logic in this rewrite rule

titaiwangms (approved):

onnxscript/rewriter/fuse_relus_clips.py
```diff
@@ -0,0 +1,190 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Does the following transformation:
+- Relu(Relu(X)) -> Relu
+- Relu(Clip(X)) -> Clip
+- Clip(Relu(X)) -> Clip
+- Clip(Clip(X)) -> Clip
+"""
+
```
gramalingam, line 9: `from __future__ import annotations` would be helpful

PR #2412, opened 2025-06-23
[DRAFT] Add support for onnx fusions

gramalingam:
* Add basic infrastructure support for fusions targeting ONNX opset 23, with RMSNormalization as one target op.
* Cleanup existing RMSNormalization fusion targetting ORT's contrib op (using pattern-disjunction to simplify rules).

### DIFF
--- a/onnxscript/rewriter/onnx_fusions/rms_normalization.py
+++ b/onnxscript/rewriter/onnx_fusions/rms_normalization.py
@@ -6,16 +6,16 @@ import onnxscript.ir as ir
 from onnxscript.rewriter import _fusion_utils, _ir_utils, pattern
 
 """
-RMS Normalization: This is referred to as SimplifiedLayerNormalization in the ORT codebase.
-See https://github.com/microsoft/onnxruntime/blob/6d9636f07cccdb6e4ac453087ad54c3bc9854d50/onnxruntime/core/graph/contrib_ops/contrib_defs.cc#L2981
+RMS Normalization: ONNX Opset 23 op
+See: https://onnx.ai/onnx/operators/onnx__RMSNormalization.html#l-onnx-doc-rmsnormalization
+
 
 Key points for the fusion optimization:
 * Input and scale are allowed to be of different types.
 * The normalization of the input can be done in a different precision than the input type,
-which is also the precision of reciprocal_rms returned by operation.
+indicated by stash_type.
 * Input (x) must be: float or double or float16 or bfloat16
 * Scale must be: float or double or float16 or bfloat16
-* Normalization precision must be float or double
 """
 
 float_types = [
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/onnx_fusions/rms_normalization.py
```diff
@@ -0,0 +1,80 @@
...
+* Scale must be: float or double or float16 or bfloat16
+"""
+
+float_types = [
+    ir.DataType.FLOAT,
+    ir.DataType.FLOAT16,
+    ir.DataType.BFLOAT16,
+    ir.DataType.DOUBLE,
+]
+fp_float_types = [ir.DataType.FLOAT, ir.DataType.DOUBLE]
```
justinchuby:
```suggestion
fp_float_types = frozenset([ir.DataType.FLOAT, ir.DataType.DOUBLE])
```

onnxscript/rewriter/onnx_fusions/rms_normalization.py
```diff
@@ -0,0 +1,80 @@
...
+* Input (x) must be: float or double or float16 or bfloat16
+* Scale must be: float or double or float16 or bfloat16
+"""
+
+float_types = [
+    ir.DataType.FLOAT,
+    ir.DataType.FLOAT16,
+    ir.DataType.BFLOAT16,
+    ir.DataType.DOUBLE,
+]
```
justinchuby:
```suggestion
float_types = frozenset([
    ir.DataType.FLOAT,
    ir.DataType.FLOAT16,
    ir.DataType.BFLOAT16,
    ir.DataType.DOUBLE,
])
```

onnxscript/_framework_apis/torch_2_9.py
```diff
@@ -0,0 +1,31 @@
...
+    convert_version,
+    get_torchlib_ops,
+    save_model_with_external_data,
+)
+from onnxscript.rewriter import onnx_fusions
+
+
+def optimize(model: ir.Model) -> ir.Model:
+    """Optimize the model."""
+    optimizer.optimize_ir(model)
```
justinchuby: hmm I can see why a torch 2_8 api would be helpful now. Would you like to make this 2_8, and we can cherry pick pytorch?

justinchuby: Or should this be a part of the optimizer?

justinchuby:
I merged the 2_8 apis

onnxscript/_framework_apis/torch_2_9.py
gramalingam: Both doable, I guess. I think there is value in having separate "core optimization", which would be invoked from several places, potentially multiple times, even from within something like onnx_fusion or ort_fusion.

justinchuby (approved):

PR #2416, opened 2025-06-24
[pass] Enable DeduplicateInitializersPass

titaiwangms:


justinchuby:
oh. We should switch to import onnx-ir by now

justinchuby:
No I think it’s because we re exposed the passes under onnxscript ir

justinchuby:
https://github.com/microsoft/onnxscript/pull/2417

onnxscript/optimizer/_optimizer.py
```diff
@@ -54,6 +54,7 @@ def optimize_ir(
         onnxscript.ir.passes.common.CommonSubexpressionEliminationPass(),
         onnxscript.ir.passes.common.LiftConstantsToInitializersPass(),
         onnxscript.ir.passes.common.LiftSubgraphInitializersToMainGraphPass(),
+        onnxscript.ir.passes.common.DeduplicateInitializersPass(),
```
justinchuby:
```suggestion
        onnx_ir.passes.common.DeduplicateInitializersPass(),
```

justinchuby: We can update the above as well

justinchuby (approved):

PR #2417, opened 2025-06-25
Update ir.pass.common alias to import all from onnx_ir

justinchuby:


onnxscript/ir/passes/common/__init__.py
```diff
@@ -1,34 +1,4 @@
...
-    LiftConstantsToInitializersPass,
-    LiftSubgraphInitializersToMainGraphPass,
-    RemoveInitializersFromInputsPass,
-    RemoveUnusedFunctionsPass,
-    RemoveUnusedNodesPass,
-    RemoveUnusedOpsetsPass,
-    ShapeInferencePass,
-    TopologicalSortPass,
-)
+from onnx_ir.passes.common import *
```
justinchuby:
```suggestion
from onnx_ir.passes.common import *  # noqa: F403
```

justinchuby:
Actually bad idea. We should update to use onnx_ir

PR #2420, opened 2025-06-26
Use onnx_ir common passes

justinchuby:
Update the imports to use onnx_ir instead of the alias


titaiwangms (approved):
Should we also delete the passes/redundants in this PR to make sure they are all replaced?

justinchuby:
which ones?

justinchuby:
I kept the onnxscript passes intact so that they are still importable. But since they are just alias to the onnx_ir passes they can be used interchangeably.

Issue #2372, opened 2025-06-11
[torchlib] slice_scatter does not support start and end to be None


titaiwangms:
From https://github.com/microsoft/onnxscript/pull/558/, the comments: https://github.com/microsoft/onnxscript/blob/51ecf47523ef079c53b0e620c62d56d70cfd3871/onnxscript/function_libs/torch_lib/ops/core.py#L7889-L7892 are not implemented, which results in errors when None is unsqueezed later in the code.

PR #2422, opened 2025-06-28
Eliminate unnecessary ScatterND

gramalingam:
Identify ScatterND(data, indices, updates) that can be replaced by Identity(updates). This is generated by the translation of `x[:, ...] = y` in PyTorch. The specific pattern is that the updated indices take the form [[0], ..., [S-1]] for the first dimension, where S is the size of the first dimension of the updated-data tensor. In effect, the scatter-update ends up being an assignment of a new value to the entire tensor.


onnxscript/rewriter/redundant_scatter_nd.py
```diff
@@ -0,0 +1,57 @@
...
+            return False
+        shape = data.shape
+        if not isinstance(shape, ir.Shape):
+            # The data shape is not known, so we cannot apply this rule.
+            return False
+        updated_dim_value = shape[axis_value]
+        transposed_data_shape = transposed_data.shape
+        if not isinstance(transposed_data_shape, ir.Shape):
+            # The transposed data shape is not known, so we cannot apply this rule.
+            return False
```
justinchuby:
```suggestion
        if transposed_data_shape is None:
            # The transposed data shape is not known, so we cannot apply this rule.
            return False
```

onnxscript/rewriter/redundant_scatter_nd.py
```diff
@@ -0,0 +1,57 @@
...
+        # Check that updated-indices represent the full range of the first dimension of the transposed data.
+        # That is: check that the data.shape[axis] matches transposed_data.shape[0].
+        axis_value = ir_utils.get_singleton_value(axis)
+        if not isinstance(axis_value, int):
+            # Cannot determine the axis statically, so we cannot apply this rule.
+            return False
+        shape = data.shape
+        if not isinstance(shape, ir.Shape):
+            # The data shape is not known, so we cannot apply this rule.
+            return False
```
justinchuby:
```suggestion
        if shape is None:
            # The data shape is not known, so we cannot apply this rule.
            return False
```

justinchuby: Do we return some fail objects?

onnxscript/rewriter/redundant_scatter_nd.py
```diff
@@ -0,0 +1,57 @@
...
+            # The first dimension of the transposed data does not match the updated dimension,
+            # so we cannot apply this rule.
+            return False
+        return True
+
+    def rewrite(self, op, updates, **_):
+        return op.Identity(updates)
+
+
+rule = ScatterAll.rule()
```
justinchuby, line 63: Should this rule be enabled by default?

gramalingam:
Can the torchlib implementation of [slice_scatter](https://github.com/microsoft/onnxscript/blob/7b89760d415aeb5ddc9b466dbed61030f4689638/onnxscript/function_libs/torch_lib/ops/core.py#L7877) be optimized to do the same, I wonder? I thought this part was potentially inside pytorch decomposition of `x[:, :, :, s] = ...`, but perhaps it can be done in slice_scatter itself?

onnxscript/rewriter/redundant_scatter_nd.py
gramalingam: The existing form more explicitly indicates that in the code after the if-statement, transposed_data_shape is a Shape object. I find that more helpful.

onnxscript/rewriter/redundant_scatter_nd.py
gramalingam: I think so. Just want to validate the correctness of the optimization first.

onnxscript/rewriter/redundant_scatter_nd.py
justinchuby: I think the current implication when reading is that it can take another type (that is not None), which is not true for this field. Maybe create an assert isinstance after if not None? Although more verbose, may be helpful for readability.

justinchuby:
> Can the torchlib implementation of [slice_scatter](https://github.com/microsoft/onnxscript/blob/7b89760d415aeb5ddc9b466dbed61030f4689638/onnxscript/function_libs/torch_lib/ops/core.py#L7877) be optimized to do the same, I wonder? I thought this part was potentially inside pytorch decomposition of `x[:, :, :, s] = ...`, but perhaps it can be done in slice_scatter itself?

Defininately!

gramalingam:
> The torchlib implementation also seems off: SymInt? start=None, SymInt? end=None means they _can_ be None even though the comment says otherwise. We should improve torchlib for this op. Related #2372

Right. I was puzzled by the comments. Not entirely sure what's happening there.

onnxscript/rewriter/redundant_scatter_nd.py
gramalingam: Ideally, I don't really want an assert afterward. It doesn't matter in this case, but if a third possibility existed hypothetically, I prefer that we skip the optimization in the third case (by going into the then branch), rather than fail at runtime (for some other user who wouldn't know what to do with it). So, that's the style used elsewhere: eg., if we are checking if a value is some scalar constant 1.0, many possibilities exist (it may not be a constant, it may be a constant with non-scalar value, or a different scalar value, etc.). 

I changed it to check `is None`. I added a type-annotation before the condition, which is more or less a static assertion.

onnxscript/rewriter/redundant_scatter_nd.py
```diff
@@ -0,0 +1,65 @@
...
+    return onnxscript.rewriter.MatchResult().fail(*args)
+
+
+class ScatterAll(orp.RewriteRuleClassBase):
+    def pattern(self, op, data, axis, transposed_data, updates):
+        # Construct update-indices spanning an entire axis:
+        shape = op.Shape(data, start=0)
+        dim = op.Gather(shape, axis, axis=0)
+        full_range = op.Range(0, dim, 1)
+        full_range_2d = op.Unsqueeze(full_range, [-1])
```
titaiwangms, line 31: I don't see these ops in the repro: https://github.com/pytorch/pytorch/issues/157289

onnxscript/rewriter/redundant_scatter_nd.py
```diff
@@ -0,0 +1,65 @@
...
+            return fail("Axis value must be a constant integer.", axis)
+        shape: ir.Shape | None = data.shape
+        if shape is None:
+            return fail("Data shape is not statically known.", data)
+        updated_dim_value = shape[axis_value]
+        transposed_data_shape: ir.Shape | None = transposed_data.shape
+        if transposed_data_shape is None:
+            return fail("Transposed data shape is not statically known.", transposed_data)
+        actual_dim_value = transposed_data_shape[0]
+        if updated_dim_value != actual_dim_value:
```
titaiwangms, line 49: Looks like we just need to update the logic in collapse_slices.py: https://github.com/microsoft/onnxscript/blob/7b89760d415aeb5ddc9b466dbed61030f4689638/onnxscript/rewriter/collapse_slices.py#L59-L61 to this line, then it can work better.

onnxscript/rewriter/redundant_scatter_nd.py
titaiwangms: Maybe you can delete https://github.com/microsoft/onnxscript/blob/7b89760d415aeb5ddc9b466dbed61030f4689638/onnxscript/rewriter/collapse_slices.py#L134

gramalingam: I think we can consolidate the rules separately. (I am thinking of trying out Copilot to do it.)

titaiwangms (approved):

onnxscript/rewriter/redundant_scatter_nd.py
gramalingam: May need to make other dimensions symbolic. Otherwise, all of these ops will be constant-folded, and the indices becomes a constant.

onnxscript/rewriter/redundant_scatter_nd.py
gramalingam: I am not sure if we can merge both rules into one. I think we need to check the following conditions:
* Indices must be 2D, of shape [N, 1]
* The values of the indices must be Range(0, N). 

If N is a constant, then the existing rule handles that. But if N is symbolic, then we need to check the ops used to compute the indices.

PR #2423, opened 2025-06-28
Add stacktrace in metadata_props for rewriter created nodes

justinchuby:


onnxscript/ir/_tape.py
```diff
@@ -32,9 +39,18 @@ def _make_node(self, op_type: str, inputs: Sequence[ir.Value], kwargs: dict[str,
             assert isinstance(outputs, int)
             num_outputs = outputs
 
+        metadata_props = {
+            "pkg.onnxscript.rewriter.stacktrace": _get_stacktrace(),
```
gramalingam, line 43: Should this always be added? Or, would it be better to do it only if the user sets some config/environment variable?

justinchuby: I think it's ok as a default? For example torch.export will add metadata as well. If user wants no stack traces they can run a pass, or if there's demand we can add an option to turn it off?

gramalingam: But who's using this? What happens when multiple rewrites happen? Trying to understand what exactly the use-case is and what the users want.

justinchuby: When we look at the graph from pytorch, all nodes can be traced back from source code in the torch model, but it is unclear which rule created the nodes that are added by the rewriter. So I was trying to add this information. A node can only be created once so multiple rewrites should not affect this field

PR #2434, opened 2025-07-03
Improves aten_chunk conversion

xadupre:


titaiwangms (commented):
The original code references on torchscript symbolic function implementation. Could you explain why the op could work on this way? And could you update the code, because g does not make any sense.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1647,29 +1647,12 @@ def aten_choose_qparams_optimized(
...
-
-    remainder = op.Mod(dim_size, num_per_chunk)
-    if remainder > 0:  # type: ignore[operator]
-        # Append the remainder to the [n, n, n, n, ..., r]
-        list_split = op.Concat(list_split, op.Reshape(remainder, neg_1), axis=0)
-
-    return op.SplitToSequence(self, list_split, axis=dim)
+    if chunks == 1:
+        return g.op.Identity(self)
+    return g.op.Split(self, axis=dim, num_outputs=chunks)
```
justinchuby: Thanks - the change looks good except for the use of g. Could you update usage? We can then merge

docs/articles/2023-06-20-introducing-onnx-script/index.md
```diff
@@ -187,8 +187,11 @@ We will cover the new PyTorch ONNX exporter in a separate post with more depth a
 By deeply weaving ONNX Script support into the PyTorch ONNX exporter, we have also made it possible to augment PyTorch model code with specialized ONNX functions as custom operators. We introduced initial support for this in the TorchScript exporter starting with [PyTorch 1.13][torch-onnx-customops] and continue to refine this capability in the new exporter.
 
 ## An End-to-End Example
+
 Let’s look at an example slightly more complicated than GELU. In fact, the following example is adapted directly from the new PyTorch ONNX exporter, implementing support for [`torch.chunk`][torch-chunk], which attempts to split a tensor into the number of specified chunks.
 
+.. TODO: change this example
```
justinchuby: This is a post so may not be a good idea to change it. Maybe we can add a header saying some examples may be obsolete?

xadupre: removed

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1647,29 +1647,12 @@ def aten_choose_qparams_optimized(
...
-
-    remainder = op.Mod(dim_size, num_per_chunk)
-    if remainder > 0:  # type: ignore[operator]
-        # Append the remainder to the [n, n, n, n, ..., r]
-        list_split = op.Concat(list_split, op.Reshape(remainder, neg_1), axis=0)
-
-    return op.SplitToSequence(self, list_split, axis=dim)
+    if chunks == 1:
+        return op.Identity(self)
+    return op.Split(self, axis=dim, num_outputs=chunks)
```
justinchuby, line 1655: I will need to fix the graph builder to return the correct number of outputs

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: So the implementation is correct, but the builder is blocking the pr?

justinchuby: I think so. And not even the builder in pytorch, just how we are doing it in test. I need to double check.

justinchuby:
Moving to a branch in the main repo for easier update

PR #2441, opened 2025-07-09
Improve redundant slice removal

gramalingam:
Improve the optimization for removal of redundant slices. (It doesn't currently handle dynamic shapes.) The optimization is fairly simple, and eliminates the slice when the input and output shapes are same.


justinchuby:
Maybe need to update the test: test_slice_pattern_is_not_matched_when_input_is_dynamic

Issue #2442, opened 2025-07-09
Potential bug with `dtype.is_floating_point`


shuheng-liu:
Hi there,

While using onnxscript to convert a torch model into ONNX format, I encountered an error that, to the best of my guess, likely comes from 

https://github.com/microsoft/onnxscript/blob/f42c2bbfd31edc99a849e0381ae4992da32479de/onnxscript/function_libs/torch_lib/ops/core.py#L6761-L6763
where it tries to infer the type of a tensor. 

With torch 2.6 (and I think many other versions), `dtype.is_floating_point` is a boolean property instead of a callable method. Hence I got the following traceback
```
Traceback (most recent call last):
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/_exporter_legacy.py", line 779, in dynamo_export
    ).export()
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/_exporter_legacy.py", line 556, in export
    onnxscript_graph = fx_interpreter.run(
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/diagnostics/infra/decorator.py", line 146, in wrapper
    ctx.log_and_raise_if_error(diag)
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/diagnostics/infra/context.py", line 355, in log_and_raise_if_error
    raise diagnostic.source_exception
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/diagnostics/infra/decorator.py", line 130, in wrapper
    return_values = fn(*args, **kwargs)
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/fx/fx_onnx_interpreter.py", line 542, in run
    self.run_node(
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/diagnostics/infra/decorator.py", line 146, in wrapper
    ctx.log_and_raise_if_error(diag)
  File "/Users/liushuheng/miniconda3/envs/lerobot/lib/python3.10/site-packages/torch/onnx/_internal/diagnostics/infra/context.py", line 355, in log_and_raise_if_error
[...]

PR #2444, opened 2025-07-10
Remove torchscript graph builder

justinchuby:
Remove torchscript graph builder since the `dynamo_export` api is removed. We no longer support it with the torchscript graph builder, and suggest users to use dynamo=True or downgrade onnxscript.

Do not merge until we are ready to bump to 0.4

Closes https://github.com/microsoft/onnxscript/issues/2442


onnxscript/function_libs/torch_lib/graph_building/__init__.py
```diff
@@ -41,8 +9,16 @@
...
-    TorchScriptGraph,
-    TorchScriptTensor,
-    TorchScriptTracingEvaluator,
-)
+class _RemovedClass:
+    """A onnxscript tensor that wraps a torchscript Value."""
+
+    def __init__(self, *_, **__):
+        raise NotImplementedError(
+            "Support for dynamo_export is dropped since onnxscript 0.4.0. "
```
justinchuby:
```suggestion
            "Support for dynamo_export has been dropped since onnxscript 0.4.0. "
```

titaiwangms (approved):

PR #2445, opened 2025-07-10
[torchlib] Fix sdpa dtype in attn_mask

titaiwangms:
Discovered in benchmark that the op.Where generates fp32 output when the whole model is set to fp16.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2069,8 +2069,8 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
     query_scaled = op.Mul(query, op.Sqrt(scale))
     key_transposed_scaled = op.Mul(key_transposed, op.Sqrt(scale))
     # Turn the Boolean mask to float: attn_mask.masked_fill(not attn_mask, -float('inf'))
-    zero = op.CastLike(op.Constant(value_float=0.0), query_scaled)
-    neg_inf = op.CastLike(op.Constant(value_float=-float("inf")), query_scaled)
+    zero = op.Constant(value=ir.tensor(0.0, dtype=query.dtype))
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -1714,10 +1714,12 @@ def _causal_attention_mask(query: TFloat, key: TFloat) -> TFloat:
     # }
     attn_mask = op.Trilu(attn_mask, upper=0)
     # The causal mask has 0s in the lower triangle and -inf in the upper triangle.
+    zero = op.CastLike(op.Constant(value_float=0.0), attn_mask)
```
justinchuby: attn_mask  is created from line 1713, which should always be float at this point?

titaiwangms: Reverted

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2069,9 +2069,9 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
     query_scaled = op.Mul(query, op.Sqrt(scale))
     key_transposed_scaled = op.Mul(key_transposed, op.Sqrt(scale))
     # Turn the Boolean mask to float: attn_mask.masked_fill(not attn_mask, -float('inf'))
-    attn_mask = op.Where(
-        attn_mask, op.Constant(value_float=0.0), op.Constant(value_float=-float("inf"))
-    )
+    zero = op.CastLike(op.Constant(value_float=0.0), query_scaled)
+    neg_inf = op.CastLike(op.Constant(value_float=-float("inf")), query_scaled)
+    attn_mask = op.Where(attn_mask, zero, neg_inf)
```
justinchuby: Can we create the tensor directly using query.dtype? We can use `op.Constant(value=ir.tensor(...))`

justinchuby (approved):

PR #2461, opened 2025-07-24
Added Causal Mask Pattern Fusion for LongRoPe Models

tadani3:
Modification of the GQA causal mask fusion rule to handle the attention mask fusion for Longrope models such as Phi-4-mini-reasoning. The causal mask modification leads to a result that matches the optimizations made in ModelBuilder.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/gqa.py
+++ b/onnxscript/rewriter/ort_fusions/gqa.py
@@ -7,6 +7,7 @@ from typing import Sequence, Union
 import numpy as np
 import onnx_ir as ir
 
+import onnxscript.onnx_types as _onnx_types
 import onnxscript.rewriter._fusion_utils as _fusion_utils
 from onnxscript.rewriter import _basics, _ir_utils, pattern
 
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
...
+        self._mask_cache = {}
+    
+    def _get_mask_key(self, attention_mask):
+        """
+        Generate a unique key for the mask based on input_ids and past_kv_cache.
+        This is used to cache the mask to avoid recomputation.
+        """
+        return (id(attention_mask))
+    
+    def compute_mask(self, op, attention_mask : _onnx_types.INT64['batch', 'seq_len']):
```
justinchuby: The rewriter doesn't use onnxscript type (yet). Could you instead use a comment to document the shape of the attention_mask?

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
             _outputs=3,
         )
 
+class LongRoPeGQACausalMask(pattern.RewriteRuleClassBase):
```
justinchuby, line 323: Could you use the docstring to document the pattern and its replacement? For the branches A, B, and C, I would consider giving them descriptive names.

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
...
+        mask,
+        input_ids,
+        past_kv_cache_1,
+        past_kv_cache_2,
+        attention_mask,
+        past_seq_length,
+        total_seq_length,
+    ):
+        seq_len = op.Shape(input_ids, end=2, start=1, _outputs=["seq_len"])
+        seq_len_0D = op.Squeeze(seq_len, _outputs=["seq_len_0D"])
```
justinchuby:
```suggestion
        seq_len_0d = op.Squeeze(seq_len, _outputs=["seq_len_0d"])
```

prefer snake case for variable names when possible

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
...
+
+        # EXPAND A/B TO AND
+        mask_expanded_A_sub = op.Sub(mask_expanded_A, 262144, _outputs=["mask_expanded_A_sub"])
+        mask_A_B_greater = op.Greater(mask_expanded_B, mask_expanded_A_sub, _outputs=["mask_A_B_greater"])
+        mask_A_B_greater_bitwise = op.And(True, mask_A_B_greater, _outputs=["mask_A_B_greater_bitwise"])
+        mask_A_B_less = op.LessOrEqual(mask_expanded_B, mask_expanded_A, _outputs=["mask_A_B_less"])
+        mask_A_B_combined = op.And(mask_A_B_greater_bitwise, mask_A_B_less, _outputs=["mask_A_B_combined"])
+        mask_A_B_combined_bitwise = op.And(True, mask_A_B_combined, _outputs=["mask_A_B_combined_bitwise"])
+
+        # EXPAND B/C TO AND
```
justinchuby: I would document the branches in plain English for readers

justinchuby (commented):
Thanks! Could you resolve the lint errors by running `lintrunner -a` and fix any issues?

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
             _outputs=3,
         )
 
+class LongRoPeGQACausalMask(pattern.RewriteRuleClassBase):
+    def __init__(self):
+        super().__init__("LongRoPeGQACausalMask", remove_nodes=False)
+        self._mask_cache = {}
```
justinchuby, line 332: The copilot review is reasonable: the rewrite rule class should be stateless. Is there a different way to do this other than keeping a `self._mask_cache`?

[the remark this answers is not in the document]
justinchuby: We should not maintain a state in the rule.

gramalingam: I think use of state for this purpose is okay? It has been used before for a similar purpose: which is to introduce values that are reused across multiple rewrites. (Now that we have CSE, there is an alternative path, which is to create duplicate copies and then eliminate them via CSE ... but I am not sure it is worth the bother.)

gramalingam: BTW: my GQA fusion doesn't use state, and produces multiple copies (as described above).

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
...
+        # EXPAND BRANCH C
+        mask_shape_C = op.Concat(batch_size, [1], seq_len, total_seq_len_C, axis=0, _outputs=["mask_shape_C"])
+        mask_shape_C_abs = op.Abs(mask_shape_C, _outputs=["mask_shape_C_abs"])
+        batch_size_squeezed = op.Squeeze(batch_size, _outputs=["batch_size_squeezed"])
+        batch_range = op.Range(0, batch_size_squeezed, 1, _outputs=["batch_range"])
+        reshaped_range_C = op.Reshape(batch_range, [-1, 1, 1, 1], allowzero=1, _outputs=["reshaped_range_C"])
+        mask_expanded_C = op.Expand(reshaped_range_C, mask_shape_C_abs, _outputs=["mask_expanded_C"])
+
+        # EXPAND A/B TO AND
+        mask_expanded_A_sub = op.Sub(mask_expanded_A, 262144, _outputs=["mask_expanded_A_sub"])
```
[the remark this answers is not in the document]
gramalingam: Better to make it a pattern-variable, I think ... if I understand right, this is actually a magic sequence-length constant? Perhaps model-specific?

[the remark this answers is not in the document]
gramalingam: On second thoughts, I am guessing this is the window_size, which should become an attribute-parameter to the GQA op.

onnxscript/rewriter/ort_fusions/gqa.py
gramalingam: The following is my understanding: if this is correct, maybe they can be renamed appropriately: I believe that A constructs the kv_range, B constructs the query_range, and C constructs the batch_range. Each constructs the corresponding range as a 4D tensor with 1s in other position (for constructing a final attention-mask of shape [Batch, NumHeads, QueryRange, KVRange] via broadcast). 

I am a bit puzzled that query_range and kv_range look to be the same here, it might be an artifact of this model-usage, I guess.

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
...
+    def __init__(self):
+        super().__init__("LongRoPeGQACausalMask", remove_nodes=False)
+        self._mask_cache = {}
+    
+    def _get_mask_key(self, attention_mask):
+        """
+        Generate a unique key for the mask based on input_ids and past_kv_cache.
+        This is used to cache the mask to avoid recomputation.
+        """
+        return (id(attention_mask))
```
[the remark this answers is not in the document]
gramalingam: If a cache is used, it should be cleaned up like [in this example](https://github.com/microsoft/onnxscript/blob/413f3df16724a0a671757db7191e9e61c1df0b17/onnxscript/rewriter/ort_fusions/cos_sin_cache.py#L105) so that it is not carried over from one graph/model to another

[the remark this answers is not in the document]
gramalingam: And I am not sure if we need to handle np arrays? If the key is either one or two ir.Values, that should be fine ... ir.Values can be used as keys in dictionaries directly, and that should avoid the garbage-collection problem.

onnxscript/rewriter/ort_fusions/gqa.py
justinchuby: My concern is that the states will transfer from model to another if not careful, which is probably not a good idea. Maybe we can have a class managed state dict that will be cleared by the class?

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -354,9 +355,163 @@ def rewrite(
             _outputs=3,
         )
 
+class LongRoPeGQACausalMask(pattern.RewriteRuleClassBase):
+    def __init__(self):
+        super().__init__("LongRoPeGQACausalMask", remove_nodes=False)
+        self._mask_cache = {}
+    
+    def _get_mask_key(self, attention_mask):
```
justinchuby: In general, avoid creating class methods that do not require states from self, and instead make them module-level private functions for testability and clarity.

onnxscript/rewriter/ort_fusions/gqa.py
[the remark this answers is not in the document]
justinchuby: I agree _get_mask_key seems unecessary. We can use the Value objects directly as keys.

onnxscript/rewriter/ort_fusions/gqa.py
```diff
@@ -7,6 +7,7 @@
 import numpy as np
 import onnx_ir as ir
 
+import onnxscript.onnx_types as _onnx_types
```
justinchuby:
```suggestion
```

_onnx_types is incompatible with the rewriter (yet)

tadani3:
@microsoft-github-policy-service agree company="Microsoft"

onnxscript/rewriter/ort_fusions/gqa.py
tadani3: I wasn't sure what the branches referred to but I'll make changes following what Rama is suggesting.

tadani3:
I added a class called Phi4MiniReasoningPostProcessor which uses the ONNX IR to fulfill two tasks:

1. Inserts the cached Cos/Sin values used in the Rotary Embeddings

- The Cos/Sin caches are computed offline following the logic found in Transformer's modeling_phi3.py file.
- The existing pattern that computes Cos and Sin is found in the ONNX model graph and removed.
- An If node containing both sets of caches is inserted to the graph. The selection condition for the two sets of caches is the total_sequence_length, which we obtain from the attention_mask.
- GQA nodes within the graph are updated to have their cos/sin inputs come from the newly inserted caches.

2. Removes Position Ids and remaining child nodes from the model graph.

- Matches the position ids and following nodes that were used in the old Cos/Sin value computation.
- Uses the ONNX IR to remove these nodes from the graph.

Reasoning and Motivation

- These changes allow the version of Phi4-mini-reasoning produced using the DynamoExporter to match the version found in Model Builder

onnxscript/rewriter/phi4_mini_reasoning_post_processor.py
justinchuby, line 1: If you can move this file to a separate PR we can merge the fusion rules. Thanks

gramalingam:
I think this is outdated and not needed?

PR #2469, opened 2025-07-30
[torchlib] Improve aten_chunk conversion

justinchuby:
Simplify implementation for `aten_chunk` and allow it to work on all data types.

Original author: @xadupre 
Updated: Conditionally use the new implementation when torch>=2.7

### DIFF
--- a/tests/function_libs/torch_lib/ops_test_data.py
+++ b/tests/function_libs/torch_lib/ops_test_data.py
@@ -694,7 +694,10 @@ TESTED_TORCHLIB_OPS: tuple[TorchLibOpInfo, ...] = (
         reason="fixme: ORT aborts with zero-dim tensors. https://github.com/microsoft/onnxruntime/issues/16619",
     ),
     TorchLibOpInfo("ceil", core_ops.aten_ceil),
-    TorchLibOpInfo("chunk", core_ops.aten_chunk),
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1647,29 +1648,40 @@ def aten_choose_qparams_optimized(
...
-    list_split = op.Expand(num_per_chunk, op.Reshape(num_chunk, neg_1))
-
-    remainder = op.Mod(dim_size, num_per_chunk)
-    if remainder > 0:  # type: ignore[operator]
-        # Append the remainder to the [n, n, n, n, ..., r]
-        list_split = op.Concat(list_split, op.Reshape(remainder, neg_1), axis=0)
-
-    return op.SplitToSequence(self, list_split, axis=dim)
+if version_utils.torch_older_than("2.7.0"):
+    # PyTorch <2.7 does not support determining the number of outputs for the Split op
```
xadupre, line 1652: I'm wondering if we should just fail in that case. This code won't probably be tested soon.

justinchuby: I think it would be reasonable to maintain support for torch 2.6 for now. It is better to keep the original working (even though untested) than failing I suppose? We can probably drop this when we move the logic to pytorch in 2.9.

PR #2472, opened 2025-07-31
Add Conv-Affine(Mul+Add) and hardswish fusion

### LINKED ISSUE #2468: [Question] rewrite rules: merged into upstream or maintained out of tree
Hi there, onnxscript is super powerful in terms of fusing operators. I've implemented several rewrite rules, could anyone give a guideline on what rewrite rules should be in the onnxscript main repo. Or I just make another private repos holding these rules.

e.g.:
- fused Conv + Mul + Add into Conv
- fused HardSidmoid + Mul into HardSwish

Stonesjtu:
Close #2468 

- Absorbs Affine into Conv:
  - Mul + Add + Conv ==> Conv
  - Conv + Mul + Add ==> Conv
- Fuse HardSwish:
  - Add + Clip + Div ==> HardSigmoid
  - HardSigmoid + Mul ==> HardSwish
  - Add + Clip + Mul + Div ==> HardSwish (Since the order of operator matters, I have to create different rewrite pattern for this)

May not be generic enough, but works for us in `paddleOCRv4` model.

Another question is hardswish is introduced in opset-v14, will onnxscript handles older opset version or rewrite rules take care of this?


onnxscript/rewriter/fuse_conv_affine.py
```diff
@@ -0,0 +1,70 @@
...
+from ._ir_utils import get_singleton_value, get_const_value
+
+
+
+
+
+class _ConvAffineFusionBase(pattern.RewriteRuleClassBase):
+    def check(self, context, x: ir.Value, w: ir.Value, b: ir.Value, scale: ir.Value, offset: ir.Value, conv_out: ir.Value) -> MatchResult:
+        check_result = MatchResult()
+        import ipdb;ipdb.set_trace()
```
xadupre: Can you apply the linter and remove this debugging lines?

xadupre:
We should have a mechanism to enable or disable a rule if the opset does not support the fused operators. But we can manually select the list of rules to apply for now. We should probably opset information in the rule itself. I'll let @gramalingam comment on that.

onnxscript/rewriter/fuse_conv_affine.py
Stonesjtu: lintrunner makes no complaints.

Stonesjtu:
@xadupre CI passed, can you review this PR again?

onnxscript/rewriter/fuse_hardswish.py
```diff
@@ -0,0 +1,129 @@
...
+    def pattern(self, op, x: ir.Value) -> ir.Value:
+        # Floating point matching for 1/6 is not exact, so we use isclose below
+        out = op.HardSigmoid(x, _allow_other_attributes=True, _outputs=["hardsigmoid_out"])
+        out = out * x
+        return out
+
+    def check(self, op, x: ir.Value, hardsigmoid_out: ir.Value) -> MatchResult:
+        check_result = MatchResult()
+        hardsigmoid = hardsigmoid_out.producer()
+        if not np.isclose(hardsigmoid.attributes["alpha"].value, 1 / 6):
```
justinchuby: We need to ensure the attribute exists before accessing it

onnxscript/rewriter/fuse_hardswish.py
```diff
@@ -0,0 +1,129 @@
...
+    def pattern(
+        self,
+        op,
+        x: ir.Value,
+        clip_min: ir.Value,
+        clip_max: ir.Value,
+        bias: ir.Value,
+        divisor: ir.Value,
+    ) -> ir.Value:
+        out = op.Clip(x + bias, clip_min, clip_max) * x
```
justinchuby: Could you explain this pattern? clip_min and clip_max are not used in the rewrite pattern. Is that expected?

onnxscript/rewriter/fuse_hardswish_test.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def test_hardswish_fusion_from_sigmoid(self):
+        model_text = """
+            <ir_version: 8, opset_import: ["" : 18]>
+            hardswish (float[N] x) => (float[N] y) {
+                hardsigmoid_out = HardSigmoid<alpha=0.16666666, beta=0.5>(x)
+                y = Mul(hardsigmoid_out, x)
+            }
+        """
+        model = ir.from_proto(onnx.parser.parse_model(model_text))
```
justinchuby: nit:
```suggestion
        model = ir.from_onnx_text(model_text)
```

onnxscript/rewriter/fuse_hardswish.py
Stonesjtu: Yes it's a expected behavior.

The HardSwish requires constant `clip_min=0` and `clip_max=6`, which is checked in `_HardSigmoidFusionBase.check`, thus we don't need to access the value in `rewrite` anymore.

However I can still assert again in rewrite, but I don't think it's a good coding style since we have a dedicated check function.

justinchuby: Thanks - that makes total sense. Could you add docstrings or comments to each of these rules to make this clear to readers?

Stonesjtu:
@justinchuby Done. I think the test files should also be moved out from the source code dir to `tests/rules/something.py`

justinchuby:
The tests are ok. See: https://github.com/microsoft/onnxscript/wiki/Coding-style#tests

Stonesjtu:
MHA attention fusion tests are failing, should be irrelevant to this PR.

onnxscript/rewriter/rules/common/__init__.py
```diff
@@ -76,6 +79,11 @@
     fuse_batchnorm_into_conv_transpose_rule,
     fuse_batchnorm_into_gemm_rule,
 )
+from onnxscript.rewriter.rules.common._fuse_conv_affine import (
+    affine_conv_fusion_rule,
+    conv_affine_fusion_rule,
+)
+from onnxscript.rewriter.rules.common._fuse_hardswish import fuse_hardswish_rules
```
justinchuby, line 86: Sorry I missed this: Could you expose the individual rules? I will merge this first - feel free to submit a follow up PR @Stonesjtu

justinchuby (approved):

Issue #2435, opened 2025-07-03
[Rewriter] Cannot create an op with a name


AyoubMDL:
Currently, when applying a rewrite rule, I cannot pass a name with the replacement op.
#### Minimal example
```python
import onnx
import onnx_ir as ir

from onnxscript.rewriter import pattern, rewrite


def custom_rewrite(model):
    def target_pattern(op, x):
        return op.Relu(x, _outputs=["relu"])

    def replacement_pattern(op, x, relu):
        ir_node = relu.producer()
        return op.Clip(x, name=ir_node.name)
       """Workaround
        out = op.Clip(x)
        clip_node = out.producer()
        clip_node.name = ir_node.name
        """

    rules = [pattern.RewriteRule(target_pattern, replacement_pattern)]
    model = rewrite(model, rules)
    return model


ir_model = ir.from_onnx_text("""
            < ir_version: 10, opset_import: ["" : 20] >
            test_model (float[N, 16, 16] X) => (float [N, ?, ?] Y)
            {
                Y = Relu(X)
            }
        """)
ir_model.graph[0].name = "relu_1"
ir_model = custom_rewrite(ir_model)
proto = ir.to_proto(ir_model)
onnx.save(proto, "tmp.onnx")
```
Doing this will result in node with attribute **name** 

![Image](https://github.com/user-attachments/assets/5a67eab9-10dd-48fc-9c9a-4a2871079b69)

### Potential fix
Other op parameters (see https://github.com/onnx/ir-py/blob/c7879f7bab341e27ca19ab31b084b449f8a82378/src/onnx_ir/_tape.py#L74) should be extracted just like it is already done for `domain, version, and outputs` https://github.com/microsoft/onnxscript/blob/87baf8ff4b8ef7e75e17bb9383398a10dba8bd91/onnxscript/ir/_tape.py#L25

I can submit a PR if you find the fix convenient

AyoubMDL:
> Can you share why it is important to have a user specified name?
For example, when replacing or removing a node and wanting to retain its original name.

> How do you avoid duplicates?
That's a good point, I didn't think about it. Maybe it will be now user responsibility ?

PR #2474, opened 2025-08-02
[Rewriter] Support specifying node name in rewrites

AyoubMDL:
Allows passing a node name when defining a rewrite. fixes https://github.com/microsoft/onnxscript/issues/2435


justinchuby (approved):

onnxscript/ir/_tape.py
```diff
@@ -26,6 +26,8 @@ def _make_node(self, op_type: str, inputs: Sequence[ir.Value], kwargs: dict[str,
         domain = kwargs.pop("_domain", "")
         version = kwargs.pop("_version", None)
         outputs = kwargs.pop("_outputs", 1)
+        name = kwargs.pop("_name", None)
```
justinchuby, line 39: Could you help document this param and its caveats? Thanks

justinchuby:
Thanks!

> the same approach to the other "op" parameters as well.
Also, should I make the same change in onnx_ir ?

What are the other parameters? What would be the change to onnx ir?

onnxscript/ir/_tape.py
AyoubMDL: Sure, where should I add it ? Right next to the line, or somewhere else?

justinchuby:
I see. The builder class in onnx-ir is not exposed yet. I want to reason about it's api and verify usage in onnxscript first. So far it's only used in the rewriter. (So no need to update onnx-ir for now)

onnxscript/ir/_tape.py
justinchuby: Maybe add a docstring to this class?

justinchuby (approved):

justinchuby:
Thank you!

PR #2478, opened 2025-08-06
Add reproduction test case for incorrect slice rewrite and add potential fix

sebimarkgraf:
This adds a reproduction of the rule introduced in f42c2bbfd31edc99a849e0381ae4992da32479de leading to an incorrect rewrite of the graph.

The original rule does not consider the step parameter, which can influence the result of a `Slice` to be the identity even when input and output shape are equivalent.

The potential fix seems to be to not apply the rule on `step != 1`, therefore the second commit adds this to the original rule implementation.

### DIFF
--- a/onnxscript/rewriter/collapse_slices.py
+++ b/onnxscript/rewriter/collapse_slices.py
@@ -6,6 +6,7 @@ import logging
 
 from onnxscript import ir
 from onnxscript.rewriter._rewrite_rule import RewriteRule, RewriteRuleSet
+from onnxscript.rewriter._ir_utils import is_singleton_value
 
 logger = logging.getLogger(__name__)
 _INT64_MAX = 9223372036854775807
@@ -81,7 +82,7 @@ def _same_shape(op, data: ir.Value, slice_output: ir.Value, steps: ir.Value, **_
     if data.shape is None or slice_output.shape is None:
         return False
 
-    if not (steps.const_value.numpy() == 1).all():
+    if not is_singleton_value(steps, 1):
         return False
 
     return data.shape == slice_output.shape
--- a/onnxscript/rewriter/collapse_slices.py
+++ b/onnxscript/rewriter/collapse_slices.py
@@ -76,10 +76,14 @@ def _potential_redundant_slice(op, data, starts, ends, axes, steps):
     return op.Slice(data, starts, ends, axes, steps, _outputs=["slice_output"])
 
 
-def _same_shape(op, data: ir.Value, slice_output: ir.Value, **_):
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/collapse_slices_test.py
```diff
@@ -100,3 +100,22 @@ def test_slice_equal_dynamic_shape(self):
...
+    def test_slice_equal_dynamic_shape_but_step_reverse(self):
+        model_proto = onnx.parser.parse_model(
+            f"""
+            <ir_version: 7, opset_import: [ "" : 17]>
+            agraph (float[L, M, N] data) => (float[L, M, N] output)
+            {{
+                starts = Constant<value: tensor = int64[1] {{0}}>()
+                ends = Constant<value: tensor = int64[1] {{{9}}}>()
+                axes = Constant<value: tensor = int64[1] {{0}}>()
+                steps = Constant<value: tensor = int64[1] {{-1}}>()
```
sebimarkgraf, line 113: This is the important change to the test case above.

onnxscript/rewriter/collapse_slices.py
```diff
@@ -76,10 +76,14 @@ def _potential_redundant_slice(op, data, starts, ends, axes, steps):
     return op.Slice(data, starts, ends, axes, steps, _outputs=["slice_output"])
 
 
-def _same_shape(op, data: ir.Value, slice_output: ir.Value, **_):
+def _same_shape(op, data: ir.Value, slice_output: ir.Value, steps: ir.Value, **_):
     """Check if the shape of the slice output is the same as the data."""
     if data.shape is None or slice_output.shape is None:
         return False
+
+    if not (steps.const_value.numpy() == 1).all():
```
gramalingam: Suggest `if (steps.const_value is None) or not (steps.const_value.numpy() ...` ... or use `onnxscript.rewriter._ir_utils.is_singleton_value(steps, 1)`

sebimarkgraf: I think the second makes sense?
Not sure how a non singleton value would behave and I would error on the side of caution with optimizations.

sebimarkgraf: Used is_singleton_value. Not sure what the common import conventions are here, so please tell me if I should import differently

justinchuby (approved):

justinchuby:
Could you format the code with `lintrunner f`?

justinchuby:
> Ran `lintrunner f`. I am getting a `Could not find a lintrunner config at: '.lintrunner.private.toml'.`, not sure if that is expected.

That’s expected. Thanks

PR #2479, opened 2025-08-06
Make onnx export SDPA match aten behavior

IlyasMoutawwakil:
This PR makes onnx sdpa export match the behavior of aten sdpa when boolean mask is used.

```python
import onnxruntime as ort
import torch


class ScaledDotProductAttention(torch.nn.Module):
    def forward(self, query, key, value, attn_mask):
        return torch.nn.functional.scaled_dot_product_attention(query, key, value, attn_mask=attn_mask)


model = ScaledDotProductAttention()
attn_mask = torch.ones(2, 4, 8, 8).bool()  # boolean mask for attention
attn_mask[0, 0, 0, :] = False  # masking an entire row (padding token)
query = key = value = torch.randn(2, 4, 8, 16)
output = model(query, key, value, attn_mask)

torch.onnx.export(
    model,
    (query, key, value, attn_mask),
    "scaled_dot_product_attention.onnx",
    input_names=["query", "key", "value", "attn_mask"],
    output_names=["output"],
    opset_version=18,
    dynamo=True, # or False
)
ort_session = ort.InferenceSession("scaled_dot_product_attention.onnx")

np_inputs = {"query": query.numpy(), "key": key.numpy(), "value": value.numpy(), "attn_mask": attn_mask.numpy()}
onnx_outputs = ort_session.run(None, np_inputs)[0]

torch.testing.assert_close(output, torch.tensor(onnx_outputs), equal_nan=True)
```
fails the assertion because the ort model outputs nans.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2076,6 +2076,11 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
         op.Add(op.MatMul(query_scaled, key_transposed_scaled), attn_mask),
         axis=-1,
     )
+    # When using scaled dot product attention with a boolean mask, the softmax operation might return NaN values
+    # due to the presence of -inf in an entire row (padding tokens), resulting in 0/0 (NaN) in the softmax output.
+    # This is because there's no safe/masked softmax imp in ONNX, so we need to handle NaN values explicitly to match
+    # the behavior of PyTorch with boolean masks.
+    attn_weight = op.Where(op.IsNaN(attn_weight), zero, attn_weight)
     attn_weight, _ = op.Dropout(attn_weight, dropout_p)
     return op.MatMul(attn_weight, value)
 


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2076,6 +2076,11 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
         op.Add(op.MatMul(query_scaled, key_transposed_scaled), attn_mask),
         axis=-1,
     )
+    # When using scaled dot product attention with a boolean mask, the softmax operation might return NaN values
+    # due to the presence of -inf in an entire row (padding tokens), resulting in 0/0 (NaN) in the softmax output.
+    # This is because there's no safe/masked softmax imp in ONNX, so we need to handle NaN values explicitly to match
+    # the behavior of PyTorch with boolean masks.
+    attn_weight = op.Where(op.IsNaN(attn_weight), zero, attn_weight)
```
gramalingam, line 2083: Thanks ... overall, this looks reasonable. Just wondering: is there a specification of this behavior in pytorch? I am also curious under what circumstances this arises in practice? If one is not paying attention to any of the key/values, that seems odd.

IlyasMoutawwakil: > If one is not paying attention to any of the key/values, that seems odd.

this happens when sequences are left padded, which is what's done when trying to do batched text generation.

> is there a specification of this behavior in pytorch?

it is discussed in depth in this issue https://github.com/pytorch/pytorch/issues/103749

titaiwangms (approved):

justinchuby:
Do we need to update https://github.com/pytorch/pytorch/blob/main/torch/onnx/_internal/exporter/_torchlib/ops/nn.py as well, or improve specs of the Attention op? @gramalingam @titaiwangms

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2076,6 +2076,11 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
...
     attn_weight, _ = op.Dropout(attn_weight, dropout_p)
```
justinchuby, line 2084: @titaiwangms we should probably conditionally skip this line (even though there is a rewrite rule already)

titaiwangms: OK

gramalingam: If you fix this, can you also please add a reference to https://github.com/pytorch/pytorch/issues/103749 in the comments for the previous line fixing NaN?

titaiwangms: We skip when `dropout_p` is 0?

titaiwangms: https://github.com/microsoft/onnxscript/pull/2482

PR #2481, opened 2025-08-08
[DRAFT] MHA fusion cleanup

gramalingam:
* Cleanup MHA fusion rules by eliminating some redundant rule-variations
* Fix handling of scale attribute in MHA fusion
* Introduce can_match_none attribute-variables in patterns

### DIFF
--- a/onnxscript/rewriter/_pattern_ir.py
+++ b/onnxscript/rewriter/_pattern_ir.py
@@ -76,12 +76,18 @@ class PrefixPattern(StringPattern):
 class AttrPattern(Pattern[ir.Attr]):
     """Base class for an attribute pattern. Matches any attribute value by default."""
 
-    def __init__(self, name: str | None):
+    def __init__(self, name: str | None, *, can_match_none: bool = False):
         self._name = name
+        self._can_match_none = can_match_none
 
     @property
     def name(self) -> str | None:
         return self._name
+
+    @property
+    def can_match_none(self) -> bool:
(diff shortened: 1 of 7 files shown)


onnxscript/rewriter/_pattern_ir.py
```diff
@@ -76,20 +76,28 @@ def __str__(self) -> str:
...
+        return self._can_match_none
+
     def matches(self, attr: ir.Attr) -> bool:
         return True
 
     def __str__(self) -> str:
         return self._name if self._name is not None else "anonymous:" + str(id(self))
 
 
+AttrVar = AttrPattern
```
justinchuby: Could you comment on this alias? Why is it needed?

gramalingam: `AttrVar` is a better name for a pattern-variable meant to be bound against an attribute-value.  `AttrPattern` serves as the base-class for other attribute-patterns as well, such as attribute-constants. The names also mirror those used for ValuePattern and Var. (In the long run, we could make AttrVar a derived class of AttrPattern, but right now there is no need, so an alias works fine.)

justinchuby: I am thinking maybe a derived class is better for documentation, because right now AttrVar will not have the right docstring I assume

gramalingam: Done.

PR #2482, opened 2025-08-08
Add condition to dropout and ref to isnan

titaiwangms:
op.Dropout is only enabled when `dropout_p` is not 0, and added a reference issue discussion about why op.Where and op.IsNaN are needed when attention mask is boolean value.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -2037,7 +2037,8 @@ def _aten_scaled_dot_product_attention_no_mask_onnx(
         op.MatMul(query_scaled, key_transposed_scaled),
         axis=-1,
     )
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2037,7 +2037,8 @@ def _aten_scaled_dot_product_attention_no_mask_onnx(
         op.MatMul(query_scaled, key_transposed_scaled),
         axis=-1,
     )
-    attn_weight, _ = op.Dropout(attn_weight, dropout_p)
+    if dropout_p == 0:
```
justinchuby:
```suggestion
    if dropout_p != 0:
```

typo?

justinchuby (approved):

Issue #2466, opened 2025-07-25
FoldConstantsPass: Why are always_fold_ops restricted to unique consumers?


iksnagreb:
I do not fully understand the following: Why is `always_fold_ops`, which I understand as simply overriding the size limit, restricted to operators which are the single unique consumer of an input? Maybe I have missed this, but I do not see a similar restriction for the usual path where the input is below the size limit? Is it to prevent the model size from increasing by implicitly duplicating large tensors via this override?

https://github.com/microsoft/onnxscript/blob/413f3df16724a0a671757db7191e9e61c1df0b17/onnxscript/optimizer/_constant_folding.py#L1037-L1045

iksnagreb:
Sorry for the delayed answer... I think there is something similar going on here, which took me quite some time to figure out why I cannot get it fold the ConstantOfShape...
https://github.com/microsoft/onnxscript/blob/413f3df16724a0a671757db7191e9e61c1df0b17/onnxscript/optimizer/_constant_folding.py#L1075-L1076
So at least a debug message would be nice to be aware of this behavior (I wonder if there are even more like this?) Currently you even get something like "`DEBUG:onnxscript.optimizer._constant_folding:New constant for value ConstantOfShape_0_output_0 dtype: float32 shape: (1, 1, 15)`" when setting the debug log-level, but it then silently decides to just not apply this new constant...

Options to increase the flexibility or to align this closer to user intent could be another option to the `FoldConstantsPass` besides the `always_fold_ops` to explicitly allow constant folding with potential model size increase, e.g., `allow_size_increase` or even more drastic `allow_bloat`, these could default to false to keep the current behavior. The other option (potentially a breaking change) would be interpret `always_fold_ops` as ALWAYS fold these operators, even if it increases the size - if someone puts an operator there explicitly, it is probably the intention to really always fold it?

gramalingam:
Thanks for bringing up the issue.

(a) Do we need a special behavior for "ConstantOfShape"? I guess, we definitely want a size heuristic (just like any other op): if folding this will increase model size significantly, we don't want to do it. So, the question is: are the general size heuristic options insufficient for "ConstantOfShape"? 

I guess the case to consider is when ConstantOfShape produces a tensor of size < output_limit. 

(b) I wonder if we are complicating it too much: the documentation of the intended behavior seems complicated as it needs to consider the interaction between the different options to control inlining, with input size limit, output size limit, the bloat option, and always-inline list, etc.

[...]

PR #2488, opened 2025-08-13
Update constant folding behavior for large tensors

justinchuby:
Suggested by https://github.com/microsoft/onnxscript/issues/2466, I updated the constant folder logic to 

1. Include an `allow_bloat` option. 
2. When `allow_bloat` and operator is in always fold list, we will fold the value
3. When `ConstantOfShape` is in the always fold list, we will replace it with Constant
4. When `allow_bloat` is False and operator is in always fold list, we will only fold if the large tensor has only one consumer.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -824,6 +824,10 @@ def _merge_shapes(shape1: ir.Shape | None, shape2: ir.Shape | None) -> ir.Shape 
 class FoldConstantsPass(ir.passes.InPlacePass):
     """A pass that folds constant expressions in the model.
 
+    .. NOTE::
+        ``ConstantOfShape`` will not be replaced unless explicitly specified in
+        ``always_fold_ops``.
+
     Attributes:
         shape_inference: Whether to perform shape inference.
         input_size_limit: Maximum size of input tensors to fold.
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -317,12 +317,6 @@ def _get_output(node: ir.Node, index: int) -> ir.Value | None:
     return None
 
 
-def _update_type(value: ir.Value, type: ir.TypeProtocol | None) -> None:
-    if type is not None:
-        # TODO: merge types
-        value.type = type
-
-
 def _get_input_element_type(node: ir.Node, index: int) -> int:
     input = _get_input(node, index)
     if input is not None and input.type is not None:
@@ -834,9 +828,11 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         shape_inference: Whether to perform shape inference.
         input_size_limit: Maximum size of input tensors to fold.
         output_size_limit: Maximum size of output tensors to fold.
-        always_fold_ops: Collection of op types that should always be folded.
+        always_fold_ops: Collection of op types that should always be folded, unless
+            folding the operator will duplicate model weights and allow_bloat is False.
             For ops from the default opset, only op_type is neede (e.g. "Transpose"),
             otherwise specify the domain with ``{domain}::{op_type}``.
+        allow_bloat: If False, the pass will not fold ops that will duplicate model weights.
     """
 
     def __init__(
@@ -846,6 +842,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         input_size_limit: int,
         output_size_limit: int,
         always_fold_ops: Collection[str] = frozenset(["Transpose"]),
+        allow_bloat: bool = False,
     ) -> None:
         self.shape_inference = shape_inference
         self.input_size_limit = input_size_limit
@@ -857,6 +854,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
                 domain = ""
             ops.append((domain, op_type))
         self.always_fold_ops: frozenset[tuple[str, str]] = frozenset(ops)
+        self.allow_bloat = allow_bloat
 
         self._opset_imports: dict[str, int] = {}
         self._counts: dict[str, int] = {}
@@ -896,7 +894,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         input_data = {k: v for k, v in input_data.items() if v is not None}
         if any(t is None for t in input_types.values()):
             logger.debug(
-                "Skipping shape inference for node %s due to missing input type.",
+                "Skipping shape inference for node %r due to missing input type.",
                 node.name,
             )
         else:
@@ -922,7 +920,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
                         output.type = ir.serde.deserialize_type_proto_for_type(inferred_type)
             except Exception as e:
                 logger.debug(
-                    "Skipping shape inference for node %s due to exception: %s",
+                    "Skipping shape inference for node %r due to exception: %s",
                     node.name,
                     e,
                 )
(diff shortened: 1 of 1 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -1007,54 +1009,102 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+                        node,
+                    )
+                else:
+                    log_large_inputs()
+                    return None
+            else:
+                if (node.domain, node.op_type) in self.always_fold_ops and all(
+                    len(input.consumers()) == 1 or (not is_large)
+                    for input, is_large in zip(node.inputs, large_inputs)
+                    if input is not None
```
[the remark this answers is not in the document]
justinchuby: Done

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1007,54 +1009,102 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
```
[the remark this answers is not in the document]
justinchuby: Added an assert

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1007,54 +1009,103 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+                node.name,
+            )
             return None
 
         # Ensure all node inputs are constants
         if any(x.const_value is None for x in node.inputs if x is not None):
             if logger.isEnabledFor(logging.DEBUG):
                 logger.debug(
-                    "Skipping constant folding for node %s because it has non-constant inputs",
+                    "Skipping constant folding for node %r because it has non-constant inputs",
```
gramalingam: nit: this (a node with some non-const input) is so common-place that I don't see a value in this debug-log ... it is just going to overwhelm the log, and if the model is very small, there are much simpler ways to debug. I realize this is pre-existing code ... but wonder if it would be better to drop this entire if-statement in line 1022/1054.

gramalingam: Sorry, I mean drop the one in 1023/1054

justinchuby: Done

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1072,11 +1096,12 @@ def convert(av):
             return None
         if len(node.outputs) == 1 and not isinstance(outputs, (tuple, list)):
             replacement = self.new_constant(node, outputs)
-            if _is_onnx_op(node, "ConstantOfShape") or replacement is None:
+            if replacement is None:
                 return None
             return Replacement(replacement.outputs, [replacement])
         else:
-            logger.warning(
```
justinchuby: Maybe keep warning here

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1072,11 +1096,12 @@ def convert(av):
...
+            # TODO(justinchuby): Enable folding of multiple outputs when allow_bloat is True
```
gramalingam: I think this is a limitation (incomplete implementation) ... not linked specifically to allow_bloat option.

justinchuby:
I will remove the allow bloat option for now. Will create a function arg for user specified rules

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1072,11 +1096,12 @@ def convert(av):
...
+            logger.info(
```
justinchuby: @iksnagreb does this need to be updated in your split PR? I was surprised it worked

iksnagreb: Apparently the registry always takes precedence. If the custom handler finds a match/replacement, it is always accepted. So whenever registering a custom handler for an operator, you are already opting-out of all these restrictions/assumption: large tensors, multi-output, control-flow, not `ConstantOfShape`, etc.

That's why I think some exposed registry/plugin mechanism for constant folding could be quite convenient, see #2507

onnxscript/optimizer/_constant_folding.py
justinchuby: Reverted

justinchuby:
@iksnagreb @gramalingam I removed the previous always_fold option and added a should_fold arg that takes a function. This is a breaking change. Please let me know if it looks ok.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -899,9 +900,10 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         shape_inference: Whether to perform shape inference.
         input_size_limit: Maximum size of input tensors to fold.
         output_size_limit: Maximum size of output tensors to fold.
-        always_fold_ops: Collection of op types that should always be folded.
-            For ops from the default opset, only op_type is neede (e.g. "Transpose"),
-            otherwise specify the domain with ``{domain}::{op_type}``.
+        should_fold: An optional function that takes a node and returns True if
+            the node should be considered for folding.
+            The function should return (1) True to always fold the node, (2) False to
```
gramalingam: nit: may be better to drop the word 'always' and 'never' ... the function returns a True/False value to indicate if this particular node should be folded.

justinchuby: Done

onnxscript/optimizer/_constant_folding_test.py
```diff
@@ -581,6 +581,33 @@ def test_transpose_is_always_folded(self):
...
+        model = ir.from_onnx_text(model_text)
+
+        # ConstantOfShape is not folded by default
+        optimized = self._fold(model)
+        ops = [node.op_type for node in optimized.graph]
+        self.assertEqual(ops, ["ConstantOfShape"])
+
+        # But ConstantOfShape is folded when specified in should_fold
+        optimized = self._fold(
+            model, should_fold=lambda node: node.op_type == "ConstantOfShape"
```
gramalingam: I know this is just a test-case ... but perhaps would be useful to return None for other op-types to illustrate to users how to use this ... that is what users should likely do, just special-case for exceptions, and return None for others to use default behavior ...

gramalingam (approved):

onnxscript/optimizer/_constant_folding_test.py
justinchuby: You are right. Let me change that.

justinchuby: Done

justinchuby:
@gramalingam could you reapprove? Thanks

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1072,62 +1068,102 @@ def process_node(self, node: ir.Node) -> Replacement | None:
...
+            input_tensors = [x.const_value if x is not None else None for x in node.inputs]
+            large_inputs = [
+                tensor is not None and tensor.size > self.input_size_limit
+                for tensor in input_tensors
+            ]
+            if any(large_inputs):
+                # Decide whether to fold large constants
+                assert len(node.inputs) == len(large_inputs)
+                if (node.domain, node.op_type) in _DEFAULT_ALWAYS_FOLD_OPS and all(
+                    len(input.consumers()) == 1 or (not is_large)
```
titaiwangms, line 1134: Could you remind me why small tensor with more than one consumer is acceptable here?

justinchuby: Because folding them will not yield large tensors. E.g. If you have multiple reshape nodes to fold, and the shape tensor is shared, we don't want to prevent folding simply because the shape tensor is used many times.

PR #2492, opened 2025-08-15
Introduce layer-norm fusion

gramalingam:
Introduce layer-norm fusion rules, along with a couple of test cases.

This is just the first version. TO DO:
* We need improved infrastructure for ONNX fusions to handle opset dependence. LayerNorm exists in ONNX since opset 17. For now the fusion rule exists, but it is not automatically called yet (but users can invoke it themselves).
* If users want to use opsets < 17, this could be done as an ORT fusion using ORT contrib op LayerNorm.

### DIFF
--- a/onnxscript/rewriter/onnx_fusions/_layer_norm.py
+++ b/onnxscript/rewriter/onnx_fusions/_layer_norm.py
@@ -34,6 +34,11 @@ fp_float_types = frozenset([ir.DataType.FLOAT, ir.DataType.DOUBLE])
 
 
 class LayerNormFusion(pattern.RewriteRuleClassBase):
+    def __init__(self, name: str, has_bias: bool):
+        super().__init__(name)
+        self._has_bias = has_bias
+        self._stash_dtype: int | None = None
+
     def pattern(self, op, x, scale, bias, epsilon, target_dtype):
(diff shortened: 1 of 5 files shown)


onnxscript/rewriter/onnx_fusions/_layer_norm.py
```diff
@@ -0,0 +1,128 @@
...
+
+float_types = frozenset(
+    [
+        ir.DataType.FLOAT,
+        ir.DataType.FLOAT16,
+        ir.DataType.BFLOAT16,
+        ir.DataType.DOUBLE,
+    ]
+)
+fp_float_types = frozenset([ir.DataType.FLOAT, ir.DataType.DOUBLE])
```
justinchuby:
```suggestion
FLOAT_TYPES = frozenset(
    [
        ir.DataType.FLOAT,
        ir.DataType.FLOAT16,
        ir.DataType.BFLOAT16,
        ir.DataType.DOUBLE,
    ]
)
HIGH_PRECISION_FLOAT_TYPES = frozenset([ir.DataType.FLOAT, ir.DataType.DOUBLE])
```

nit: upper-case for constants

gramalingam: Done. I also gave them more meaningful names.

justinchuby (approved):

PR #2494, opened 2025-08-16
Extend rewriter to handle subgraphs

gramalingam:
This extends the rewriter to apply fusions in nested subgraphs as well. It is currently limited to patterns that completely lie within a single graph (with special exceptions for constants/variables).

### DIFF
--- a/onnxscript/rewriter/_matcher.py
+++ b/onnxscript/rewriter/_matcher.py
@@ -184,10 +184,15 @@ class SimplePatternMatcher(PatternMatcher):
     ) -> bool:
         """Match an IR value against a ValuePattern instance."""
         if value is not None and value.graph is not self._graph_or_function:
-            return self.fail(
-                f"Value {value.name} is not in the graph {self._graph_or_function.name}. "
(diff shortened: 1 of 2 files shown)


justinchuby (approved):

onnxscript/rewriter/_matcher.py
```diff
@@ -352,6 +362,7 @@ def match(
         complications which require careful consideration.
         """
         self._tracer = tracer
+        self._graph_or_function = graph_or_function[0].graph
```
gramalingam, line 365: @justinchuby : this was a quick hack to test this implementation. Would be nice if `ir.Function` exposed a `graph()` property. Basically this should be `graph_or_function if isinstance(graph_or_function, ir.Graph) else graph_or_function._graph`.

justinchuby: Oh I see. We need it for identity comparison. Let me think about this

justinchuby: https://github.com/onnx/ir-py/pull/158

justinchuby (approved):

PR #2495, opened 2025-08-16
Add Erf-based Gelu fusion rule

gramalingam:
Add Erf-based Gelu fusion rule

### DIFF
--- a/onnxscript/rewriter/ort_fusions/gelu.py
+++ b/onnxscript/rewriter/ort_fusions/gelu.py
@@ -6,8 +6,8 @@ import math
 
 from onnxscript.rewriter import _fusion_utils, pattern
 
-_sqrt_two_over_pi = math.sqrt(2.0 / math.pi)
-_sqrt_two = math.sqrt(2.0)
+_SQRT_TWO_OVER_PI = math.sqrt(2.0 / math.pi)
+_SQRT_TWO = math.sqrt(2.0)
 
 
 class GeluTanhFusion(pattern.RewriteRuleClassBase):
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/gelu.py
```diff
@@ -7,6 +7,7 @@
 from onnxscript.rewriter import _fusion_utils, pattern
 
 _sqrt_two_over_pi = math.sqrt(2.0 / math.pi)
+_sqrt_two = math.sqrt(2.0)
```
justinchuby: nit: upper case for constants

gramalingam: Done

onnxscript/rewriter/ort_fusions/gelu.py
```diff
@@ -27,9 +28,23 @@ def rewrite(self, op, x):
...
+        # GELU(x) = 0.5 * x * (1 + erf(x / sqrt(2)))
+        t1 = op.Div(x, _SQRT_TWO)
+        t2 = op.Erf(t1)
+        t3 = op.Add(t2, 1.0)
+        t4 = op.Mul(x, t3)
+        result = op.Mul(t4, 0.5)
+        return result
+
+    def rewrite(self, op, x):
+        return op.Gelu(x, _domain="com.microsoft")
```
justinchuby, line 42: So they are fused to the same op? Is there an approximate option like https://onnx.ai/onnx/operators/onnx__Gelu.html does?

gramalingam: No, they are two different ops in ORT ... the Tanh-based fusion above uses `FastGelu`, while this uses `Gelu`

Issue #2496, opened 2025-08-18
Ensure empty ints attributes are created correctly


justinchuby:
Changes due to https://github.com/onnx/ir-py/pull/148

```pytb
_______________ ShapeOptimizationTest.test_shape_optimization_5 ________________
[gw0] darwin -- Python 3.11.9 /Users/runner/work/onnxscript/onnxscript/.nox/test_onnx_ir_git/bin/python
.nox/test_onnx_ir_git/lib/python3.11/site-packages/parameterized/parameterized.py:620: in standalone_func
    return func(*(a + p.args), **p.kwargs, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/rewriter/ort_fusions/shape_optimization_test.py:58: in test_shape_optimization
    count = shape_optimization.rules.apply_to_model(model)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/rewriter/_rewrite_rule.py:768: in apply_to_model
    count = self._apply_to_graph_or_function(
onnxscript/rewriter/_rewrite_rule.py:659: in _apply_to_graph_or_function
    delta = rule.try_rewrite(
onnxscript/rewriter/_rewrite_rule.py:323: in try_rewrite
    replacement_subgraph = self._replacement_pattern.get_replacement(match)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/rewriter/_rewrite_rule.py:219: in get_replacement
    new_outputs = self._function(context, **match.bindings)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/rewriter/ort_fusions/shape_optimization.py:58: in rewrite
    return op.Constant(value_ints=[])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/ir/_tape.py:23: in <lambda>
    return lambda *args, **kwargs: self._make_node(op_type, args, kwargs)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
onnxscript/ir/_tape.py:36: in _make_node
    value = super().op(
.nox/test_onnx_ir_git/lib/python3.11/site-packages/onnx_ir/_tape.py:92: in op
[...]

PR #2497, opened 2025-08-19
Correctly create empty ints for Constant in rewriter

justinchuby:
Due to changes https://github.com/onnx/ir-py/pull/148, we cannot create an empty list attribute without specifying type because it would be ambiguous.

Fix https://github.com/microsoft/onnxscript/issues/2496

### DIFF
--- a/onnxscript/rewriter/ort_fusions/shape_optimization.py
+++ b/onnxscript/rewriter/ort_fusions/shape_optimization.py
@@ -55,7 +55,7 @@ class ExtractDim(pattern.RewriteRuleClassBase):
         transposed_dims = [dim0, dim2, dim1, dim3]
         sliced_result = transposed_dims[self._start_val : self._end_val]
         if len(sliced_result) == 0:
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/shape_optimization.py
```diff
@@ -55,7 +55,7 @@ def rewrite(self, op, dim0, dim1, dim2, dim3, **_):
         transposed_dims = [dim0, dim2, dim1, dim3]
         sliced_result = transposed_dims[self._start_val : self._end_val]
         if len(sliced_result) == 0:
-            return op.Constant(value_ints=[])
+            return op.Constant(value_ints=ir.AttrInt64s("value_ints", []))
```
justinchuby, line 58: This makes it inconsistent with onnxscript, where we know the attribute type already from the schema. I wonder if there is a way to bridge this?

gramalingam (approved):

PR #2500, opened 2025-08-19
[Rewriter]: add fusion rules for successive Min/Max patterns

AyoubMDL:
This PR adds the following transformation:

- Min(Min(X)) -> Min(X)
- Max(Max(X)) -> Max(X)
- Min(Max(X)) -> Clip(X)
- Max(Min(X)) -> Clip(X)

### DIFF
--- a/onnxscript/rewriter/rules/common/__init__.py
+++ b/onnxscript/rewriter/rules/common/__init__.py
@@ -15,6 +15,10 @@ __all__ = [
     "fuse_batchnorm_into_gemm_rule",
     "fuse_pad_into_conv_integer_rule",
     "fuse_pad_into_conv_rule",
+    "fuse_successive_min_rule",
+    "fuse_successive_max_rule",
+    "fuse_successive_min_max_rule",
+    "fuse_successive_max_min_rule",
     "gemm_to_matmul_add_rule",
     "matmul_add_to_gemm_rule",
     "mul_by_1_rule",
@@ -88,6 +92,12 @@ from onnxscript.rewriter.rules.common._matmul_add_to_gemm import (
     transpose_a_matmul_add_to_gemm_rule,
     transpose_ab_matmul_add_to_gemm_rule,
     transpose_b_matmul_add_to_gemm_rule,
+)
+from onnxscript.rewriter.rules.common._min_max_to_clip import (
+    fuse_successive_max_min_rule,
+    fuse_successive_max_rule,
+    fuse_successive_min_max_rule,
+    fuse_successive_min_rule,
 )
 from onnxscript.rewriter.rules.common._no_op import (
     add_0_rule,
--- a/onnxscript/rewriter/rules/common/__init__.py
+++ b/onnxscript/rewriter/rules/common/__init__.py
@@ -15,10 +15,10 @@ __all__ = [
     "fuse_batchnorm_into_gemm_rule",
     "fuse_pad_into_conv_integer_rule",
     "fuse_pad_into_conv_rule",
-    "fuse_successive_min_rule",
-    "fuse_successive_max_rule",
-    "fuse_successive_min_max_rule",
-    "fuse_successive_max_min_rule",
+    "min_min_rule",
+    "max_max_rule",
+    "min_max_rule",
+    "max_min_rule",
     "gemm_to_matmul_add_rule",
     "matmul_add_to_gemm_rule",
     "mul_by_1_rule",
@@ -94,10 +94,10 @@ from onnxscript.rewriter.rules.common._matmul_add_to_gemm import (
     transpose_b_matmul_add_to_gemm_rule,
 )
 from onnxscript.rewriter.rules.common._min_max_to_clip import (
-    fuse_successive_max_min_rule,
-    fuse_successive_max_rule,
-    fuse_successive_min_max_rule,
-    fuse_successive_min_rule,
+    max_max_rule,
+    max_min_rule,
+    min_max_rule,
+    min_min_rule,
 )
 from onnxscript.rewriter.rules.common._no_op import (
     add_0_rule,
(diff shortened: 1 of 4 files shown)


justinchuby:
Thanks! I think we can enable these patterns by default. Could you share on what models these patterns will appear?

onnxscript/rewriter/min_max_to_clip.py
```diff
@@ -0,0 +1,213 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Does the following transformation:
+- Min(Min(X)) -> Min(X)
+- Max(Max(X)) -> Max(X)
+- Min(Max(X)) -> Clip(X)
+- Max(Min(X)) -> Clip(X)
```
justinchuby: Could you expand the docstring to describe the scalar and number of inputs requirements? I would also add them to the class docstrings.

AyoubMDL:
> Thanks! I think we can enable these patterns by default. Could you share on what models these patterns will appear?

I originally added the rewrite for the `Max(Min) → Clip` fusion to handle cases where I needed to export a Keras model with a bounded ReLU to ONNX. For example, the following model uses `ReLU(max_value=4.0)`:
```python
import keras
import tensorflow as tf
import tf2onnx


def build_model(input_shape=(28, 28, 1), max_relu=4.0):
    inputs = keras.Input(shape=input_shape)
    x = keras.layers.Conv2D(8, (3, 3), padding="same")(inputs)
    outputs = keras.layers.ReLU(max_value=max_relu)(x)
    model = keras.Model(inputs=inputs, outputs=outputs)
    return model


model = build_model()

spec = (tf.TensorSpec((None, 28, 28, 1), tf.float32, name="input"),)

model_proto, _ = tf2onnx.convert.from_keras(
    model,
    input_signature=spec,
    opset=15,
    output_path="model.onnx",
)
```
This will generate the following onnx model which has Max(Min):
image

I added the adder rewrites mainly for generality, but I’m open to remove them if we feel they don’t add value.

onnxscript/rewriter/min_max_to_clip.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        del context  # Not used
+        check_result = MatchResult()
+
+        first_node = out1.producer()
+        second_node = out2.producer()
+
+        # Ensure all inputs except the first are constants
+        for input_ in first_node.inputs[1:] + second_node.inputs[1:]:
+            if input_.is_graph_input():
```
gramalingam: Ideally, this should not be necessary ... it should not be treated as a constant if it is a graph input. I assume that this is to handle initializers that are inputs? We should ensure that the underlying utilities handle it correctly, since this is a common requirement in many rewrite rules. Please let us know if that isn't happening in some scenario.

gramalingam: Specifically, ir.convenience.get_const_tensor used below should handle it.

onnxscript/rewriter/min_max_to_clip.py
```diff
@@ -0,0 +1,258 @@
...
+                return check_result.fail(f"{input_.name} is not a constant.")
+
+            # If scalars are required (Clip fusion), enforce scalar-ness
+            if self.need_scalars and not self._is_scalar(input_.const_value.numpy()):
+                return check_result.fail(f"{input_.name} is not a scalar.")
+
+        if self.need_scalars:
+            # For Clip fusion: check that lower_bound <= upper_bound
+            lower_bound, upper_bound = self.compute_constants(first_node, second_node)
+            if lower_bound[0].numpy() > upper_bound[0].numpy():
```
gramalingam: The Clip op spec allows this and specifies that the result will be the smaller value (specified as max): https://onnx.ai/onnx/operators/onnx__Clip.html#l-onnx-doc-clip

gramalingam: So, I guess we could generalize and handle this case by returning the corresponding constant, I suppose ...

gramalingam (approved):
LGTM ... had just a couple of minor comments/suggestions. Thanks!

onnxscript/rewriter/min_max_to_clip.py
AyoubMDL: Done in https://github.com/microsoft/onnxscript/pull/2500/commits/105b9044aea70ac5e1b0ac3234771ee991e91f12

onnxscript/rewriter/min_max_to_clip.py
AyoubMDL: > The Clip op spec allows this and specifies that the result will be the smaller value (specified as max): https://onnx.ai/onnx/operators/onnx__Clip.html#l-onnx-doc-clip

You are right. That works for Min(Max) pattern but not for Max(Min). Added the change in https://github.com/microsoft/onnxscript/pull/2500/commits/717fa45b5d28cd887dc543f48f8d8d0665833d91

titaiwangms (approved):

onnxscript/rewriter/rules/common/__init__.py
```diff
@@ -15,6 +15,10 @@
     "fuse_batchnorm_into_gemm_rule",
     "fuse_pad_into_conv_integer_rule",
     "fuse_pad_into_conv_rule",
+    "fuse_successive_min_rule",
+    "fuse_successive_max_rule",
+    "fuse_successive_min_max_rule",
+    "fuse_successive_max_min_rule",
```
justinchuby: Following the current convention, I think we can call them
```suggestion
    "min_min_rule",
    "max_max_rule",
    "min_max_rule",
    "max_min_rule",
```

AyoubMDL: Done in e2125dc535f61d9788ea5c51d76cced86a226021

PR #2501, opened 2025-08-20
Support torch_export_patches

titaiwangms:


noxfile.py
```diff
@@ -29,11 +29,11 @@
...
-ONNX = "onnx==1.17"
-ONNX_RUNTIME = "onnxruntime==1.20.1"
-PYTORCH = "torch==2.5.1"
-TORCHVISON = "torchvision==0.20.1"
-TRANSFORMERS = "transformers==4.37.2"
+ONNX = "onnx==1.18"
+ONNX_RUNTIME = "onnxruntime==1.22.1"
+PYTORCH = "torch==2.8.0"
+TORCHVISON = "torchvision==0.23"
+TRANSFORMERS = "transformers==4.55.2"
```
justinchuby, line 36: We want to keep this the lowest versions supported; or create different pipelines that run on different versions?

onnxscript/torch_export_patches/__init__.py
justinchuby, line 1: I would prefer that this is not at the top level because this is not the main responsibility of onnxscript

justinchuby:
What is the plan for phasing the patches out, assuming they are temporary?

PR #2506, opened 2025-08-22
Add constant folding of Split with all constant inputs

iksnagreb:
Adds constant folding of Split with all constant inputs as well as replacing single-output Splits by Identity via the registry. Also adds 5 simple test cases covering both options: Second input specifying the sizes per split as well as nearly even splits into `num_outputs`  via the attribute.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -547,6 +547,47 @@ def sequence_construct(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     return None
 
 
+# Replaces Split operators with all constant inputs by a list of Constant
+# operators
+@register("Split")
+def split(node: ir.Node, op, _):
+    # Replace single output split by Identity(x)
+    if len(node.outputs) == 1:
+        return op.Identity(node.inputs[0])
+
+    # Skip non-constant inputs
+    if (x := ir.convenience.get_const_tensor(node.inputs[0])) is None:
+        return None
+
+    _split = None
+
+    # Option A: Sizes per split
+    if len(node.inputs) == 2:
+        # Skip non-constant splits
+        if (_split := ir.convenience.get_const_tensor(node.inputs[1])) is None:
+            return None
+        # Numpy expects splits as starting indices for each section
+        _split = np.cumsum(_split.numpy()[:-1])
+
+    # Option B: Number of (even) splits
+    if (num_outputs := node.attributes.get("num_outputs")) is not None:
+        # Numpy accepts single integer of (even) splits as well
+        _split = num_outputs.as_int()
+
+    # Hm, something must be terribly wrong...
+    if _split is None:
+        return None
+
+    # Default split axis is 0, according to ONNX operators reference:
+    #   https://onnx.ai/onnx/operators/onnx__Split.html
+    if (axis := node.attributes.get("axis")) is None:
+        axis = ir.Attr("axis", ir.AttributeType.INT, 0)
+
+    # Split constant tensor and wrap a list of Constant operators
+    splits = np.array_split(x.numpy(), _split, axis.as_int())
+    return [op.Constant(value=ir.tensor(x)) for x in splits]
+
+
 @register("Concat")
 def concat(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     """Replace a Concat node with a single input by Identity"""
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -563,18 +563,18 @@ def split(node: ir.Node, op, _):
     # Option A: Sizes per split
     if len(node.inputs) == 2:
         # Skip non-constant splits
-        if (_split := ir.convenience.get_const_tensor(node.inputs[1])) is None:
+        if (split_ := ir.convenience.get_const_tensor(node.inputs[1])) is None:
             return None
         # Numpy expects splits as starting indices for each section
-        _split = np.cumsum(_split.numpy()[:-1])
+        split_ = np.cumsum(split_.numpy()[:-1])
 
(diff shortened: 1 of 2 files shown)


iksnagreb:
@microsoft-github-policy-service agree company="Paderborn University"

onnxscript/optimizer/_constant_folding.py
```diff
@@ -547,6 +547,47 @@ def sequence_construct(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     return None
 
 
+# Replaces Split operators with all constant inputs by a list of Constant
+# operators
+@register("Split")
+def split(node: ir.Node, op, _):
+    # Replace single output split by Identity(x)
```
```suggestion
@register("Split")
def split(node: ir.Node, op, _):
    """Replaces Split operators with all constant inputs by a list of Constant operators."""
    # Replace single output split by Identity(x)
```

nit

onnxscript/optimizer/_constant_folding.py
```diff
@@ -547,6 +547,47 @@ def sequence_construct(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
...
+            return None
+        # Numpy expects splits as starting indices for each section
+        _split = np.cumsum(_split.numpy()[:-1])
+
+    # Option B: Number of (even) splits
+    if (num_outputs := node.attributes.get("num_outputs")) is not None:
+        # Numpy accepts single integer of (even) splits as well
+        _split = num_outputs.as_int()
+
+    # Hm, something must be terribly wrong...
```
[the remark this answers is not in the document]
justinchuby: 😅 or maybe comment on the fact that this is not expected? Maybe add a logger warning for it?

justinchuby:
Thank you!

justinchuby (approved):
LGTM. @gramalingam what do you think?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -547,6 +547,46 @@ def sequence_construct(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
...
+    """Replaces Split operators with all constant inputs by a list of Constant operators."""
+    # Replace single output split by Identity(x)
+    if len(node.outputs) == 1:
+        return op.Identity(node.inputs[0])
+
+    # Skip non-constant inputs
+    if (x := ir.convenience.get_const_tensor(node.inputs[0])) is None:
+        return None
+
+    _split = None
```
```suggestion
    split_ = None
```

tiny nit: prefixed `_` variables are typically unused. But not a big deal either

justinchuby (approved):

onnxscript/optimizer/_constant_folding.py
```diff
@@ -547,6 +547,46 @@ def sequence_construct(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     return None
 
 
+@register("Split")
```
gramalingam, line 550: I think a better, more general, solution would be to fix the restriction in the general constant-folding ([this else branch](https://github.com/microsoft/onnxscript/blob/2cc25021e10c564bdc479961c2721390b3a943ef/onnxscript/optimizer/_constant_folding.py#L1143)) ... so the underlying evaluator evaluates the Split op ... and we just need to bind each output value to the corresponding constant value.

justinchuby: Could you share what needs to be done to enable that part?

justinchuby: @iksnagreb as suggested here, we will instead fix handling for multi-output nodes. Please feel free to contribute or provide your inputs, thanks!

justinchuby (changes requested):
Blocking for now according to reviews, thanks!

onnxscript/optimizer/_constant_folding.py
gramalingam: Good question. What needs to be done is create an appropriate `Replacement` object which wraps the list of new nodes (a list of Constant nodes like in current PR), and a list of values (which should be more or less what is returned by the reference implementation).

I realize that this is lacking even in the current PR down below ... if we don't specify the list of new values (to replace the list of old values produced by the constant-folded node), we are not going to connect these values to the consumers of the old values.

So, if we added a test-case where the outputs of Split are used subsequently (or if we test the graph outputs in the tests below actually refer to the correct split const values), we will likely find that this PR doesn't work. It stems from a limitation in the core constant-folding logic that needs to be fixed.

bas-aarts:
Why special case constant Split layers like this, when a general constant folding optimization already exists?
All that's missing is multi-output support.
replacing
https://github.com/microsoft/onnxscript/blob/ab2b5f42395e2e6f4e32237885b82e06244e97f7/onnxscript/optimizer/_constant_folding.py#L1357-L1361
with
```
        elif len(node.outputs) < len(outputs) and isinstance(node.outputs[0].type, ir.SequenceType):
            return None
        else:
            values: list[ir.Value] = []
            nodes: list[ir.Node] = []
            for index, (_, output) in enumerate(zip(node.outputs, outputs)):
                replacement = self.new_constant(node, output, index)
                if replacement is None:
                    return None
                values.extend(replacement.outputs)
                nodes.append(replacement)
            return Replacement(values, nodes)
```
and
https://github.com/microsoft/onnxscript/blob/ab2b5f42395e2e6f4e32237885b82e06244e97f7/onnxscript/optimizer/_constant_folding.py#L1146-L1148
with
```
    def new_constant(
        self, node: ir.Node, array: np.ndarray | Any, index: int = 0
    ) -> ir.Node | None:
        """Create a new Constant node with the given array as its value."""
        original_value = node.outputs[index]
```
is all you need.
I've been using this for a few months now, and it works for all multi-output layers

onnxscript/optimizer/_constant_folding.py
bas-aarts: see my comment below on how to do this

PR #2513, opened 2025-08-26
[Optimizer] Avoid accessing None value in _process_constant_node

iksnagreb:


onnxscript/optimizer/_constant_folding.py
```diff
@@ -71,6 +71,14 @@ def _process_constant_node(node: ir.Node) -> None:
     if attr_value is None or not isinstance(attr_value, ir.Attr):
         return
 
+    # Even if this is an attribute, the value property might not be set, which
+    # happens e.g. in case of attribute references, i.e., ref_attr_name is set
+    if attr_value.value is None:
+        # For now reject this to prevent TypeError from accessing Nones below
+        # TODO: Is there a way to resolve references? Are references the only
+        #  reason for .value being None?
```
justinchuby: > Are references the only reason for .value being None?

For a valid model, yes I think so

onnxscript/optimizer/_constant_folding.py
```diff
@@ -71,6 +71,14 @@ def _process_constant_node(node: ir.Node) -> None:
     if attr_value is None or not isinstance(attr_value, ir.Attr):
         return
 
+    # Even if this is an attribute, the value property might not be set, which
+    # happens e.g. in case of attribute references, i.e., ref_attr_name is set
+    if attr_value.value is None:
```
justinchuby, line 76: Is there a test you can help add?

onnxscript/optimizer/_constant_folding.py
gramalingam: yes

onnxscript/optimizer/_constant_folding.py
iksnagreb: Hm, not right off the bat, stumbled upon this in the middle of some lengthy debugging. I am not even sure how to properly test this, as there is not really any observable consequence besides "not crashing"?  Maybe some test case could be constructed involving a function with both, a proper attribute embedded within (constant-foldable) and another one referencing an attribute from the outside (not constant-foldable, as long as you don't inline the function)?

onnxscript/optimizer/_constant_folding.py
iksnagreb: Should I then just remove the comment? The only way of resolving references (that I am aware of) would be inlining functions, which is a separate thing distinct from constant folding, in that case the first to-do should be removed as well?

justinchuby: I think so, thanks! Feel free to remove

onnxscript/optimizer/_constant_folding.py
justinchuby: I am thinking you can create a model with a single constant node whose attribute value is empty. Run the pass on the model, and observe that it is unchanged. Would that be good enough?

justinchuby (approved):

PR #2514, opened 2025-08-26
[Optimizer] Fix reinterpretation of strings in _get_numpy_value

iksnagreb:


onnxscript/optimizer/_constant_folding.py
```diff
@@ -278,9 +278,20 @@ def _get_numpy_value(
...
+            # Turn the constant value into a numpy array representation with the
+            # specifics of this conversion handled by the tensor type (might a
+            # yield result which needs to be reinterpreted)
+            array = const_value.numpy()
+            # Make sure strings are converted to object type first (might be
+            # some fixed width string representation which .view cannot
+            # convert, resulting in "TypeError: Cannot change data-type for
+            # array of references.")
+            if const_value.dtype == ir.DataType.STRING:
+                array = array.astype(np.object_)
```
justinchuby: Should we simply skip calling view() for string type? the view call should only be useful for types like bfloat16 and other low precision numeric types.

justinchuby: So then we don't need to cast to np.object_ now, according to the comments above. Is that accurate?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -278,9 +278,21 @@ def _get_numpy_value(
...
         try:
-            # Reinterpret the array with `.view()` because some implementations of
-            # ir.TensorProtocol (e.g. PyTorch<=2.7) do not use ml_dtypes for bfloat16 etc.
-            array = const_value.numpy().view(const_value.dtype.numpy())
+            # Turn the constant value into a numpy array representation with the
+            # specifics of this conversion handled by the tensor type (might a
+            # yield result which needs to be reinterpreted)
+            array = const_value.numpy()
+            # Make sure strings are converted to object type first (might be
+            # some fixed width string representation which .view cannot
```
justinchuby: Is this still relevant?

onnxscript/optimizer/_constant_folding.py
iksnagreb: No, that is still necessary, as sometimes strings are not `np.object_` but a fixed length string, such as `dtype="S5"` denoting a five character byte sequence. STRING should be represented by object, however, at least that is the corresponding numpy type according to https://github.com/onnx/ir-py/blob/main/src/onnx_ir/_enums.py#L264 and I have seen this fail without any reinterpretation.

justinchuby: But this function is only getting the numpy value. When do we require that the type to be object? Does it create a copy?

iksnagreb: Ah, it is not that reinterpretation is really necessary, I am just running into https://github.com/onnx/ir-py/blob/main/src/onnx_ir/_enums.py#L103, so we could simply add a case handling instances of np.bytes_ to that function (np.str_ for unicode strings is already handled). I could just add this to https://github.com/onnx/ir-py/pull/169, then we could simplify this here - no cast necessary for string and view for all non-strings.

justinchuby (approved):

PR #2518, opened 2025-08-26
2301 unify reshape flatten ops

Johansmm:
Following (https://github.com/microsoft/onnxscript/issues/2301), `flatten_to_reshape_rule` rule set is introduced to reduce the following list of operators:
- Reshape ∘ Flatten  -> Reshape
- Flatten ∘ Reshape  -> Reshape

Note to support this changes: 
- `ReshapeReshape` rule is updated to support more cases.
- `Flatten2Reshape` rule is introduced to convert Flatten ops into Reshape when possible.


onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
+        # Shape must be a constant.
+        if (np_shape := ir_utils.get_numpy_value(shape)) is None:
             return check_result.fail("Shape is not a constant.")
-        if shape.const_value.numpy().min() <= 0:
-            return check_result.fail("Shape has non-positive values.")
+        # Convert to array to support assignment destination.
+        self.new_shape = np.array(np_shape, np_shape.dtype)
+
+        # Try to replace {0,-1} values in shape if reshape output is known.
+        if (reshape_output := context.output_values[0].shape) is not None:
```
justinchuby: Would this better be done in constant folding? @gramalingam 

Or maybe not if a user doesn't want to fold other constants?

onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
+            for i, dim in enumerate(reshape_output):
+                if isinstance(dim, int) and dim > 0:
+                    self.new_shape[i] = dim
+
+        # Constraints for shape.
+        self.allowzero = context.nodes[0].attributes.get_int("allowzero", 0)
```
justinchuby:
```suggestion
        self._allowzero = context.nodes[0].attributes.get_int("allowzero", 0)
```

Since these are states, they should be private. I will look into providing a better way to manage the states in a separate issue.

onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
+            return check_result
+        if any(self.new_shape == 0) and any(self.new_shape < 0):
+            return check_result.fail("Shape cannot contain both 0 and -1 dimensions.")
+        elif np.count_nonzero(self.new_shape == 0) > 1:
+            return check_result.fail("Shape cannot contain more than one 0 dimension.")
+
+        # At this point, we can safely replace '0' with '-1'.
+        # Note allowzero is removed since at this point it does not have any effect.
+        self.allowzero = None
+        self.new_shape = np.where(self.new_shape == 0, -1, self.new_shape)
```
justinchuby:
```suggestion
        self._new_shape = np.where(self.new_shape == 0, -1, self.new_shape)
```

onnxscript/rewriter/basic_rules.py
xadupre: Where outside is allowzero used?

onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
+        if self.allowzero == 1 and any(self.new_shape == 0):
+            return check_result
+        if any(self.new_shape == 0) and any(self.new_shape < 0):
+            return check_result.fail("Shape cannot contain both 0 and -1 dimensions.")
+        elif np.count_nonzero(self.new_shape == 0) > 1:
+            return check_result.fail("Shape cannot contain more than one 0 dimension.")
+
+        # At this point, we can safely replace '0' with '-1'.
+        # Note allowzero is removed since at this point it does not have any effect.
+        self.allowzero = None
```
xadupre: A context would be better. This code is not visited if the one if the above conditions is true.

onnxscript/rewriter/basic_rules.py
Johansmm: @xadupre it is '_computed_' in `check()` and used in `rewrite()` function

onnxscript/rewriter/basic_rules.py
Johansmm: @xadupre I do not sure to understand what you mean. I'm not sure if this is clear, but I set None at this point so that when replacing the node, `allowzero` is not serialized and takes the default value (0). What do you mean by 'context would be better' ?

Johansmm:
Last push force with @justinchuby suggestions

onnxscript/rewriter/basic_rules.py
```diff
@@ -311,6 +384,7 @@ def basic_optimization_rules() -> RewriteRuleSet:
             cast_cast_rule,
             cast_identity_rule,
             expand_identity_rule,
+            flatten_to_reshape_rule,
```
titaiwangms: Better specify this rule is order sensitive to reshape_reshape_rule

Johansmm: @titaiwangms How? with a comment?

justinchuby: A comment would be good for now I think

Johansmm: Done

titaiwangms (approved):

onnxscript/rewriter/basic_rules.py
gramalingam: I have a different question (and may be that is your main point too): should this rule be decomposed into two simpler rules ... specifically, the transformation applied to the second Reshape in this pattern can be applied even if it does not follow a first Reshape (at least, that is my reading of this, please correct me otherwise). And then we can have the original rule for composing two Reshapes, improved to allow -1 but not zero in the shape.

As to when/where this transformation is done: maybe it could even be done in the torch exporter itself?

onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
-        if shape.const_value.numpy().min() <= 0:
-            return check_result.fail("Shape has non-positive values.")
+        # Convert to array to support assignment destination.
+        self._new_shape = np.array(np_shape, np_shape.dtype)
+
+        # Try to replace {0,-1} values in shape if reshape output is known.
+        if (reshape_output := context.output_values[0].shape) is not None:
+            for i, dim in enumerate(reshape_output):
+                if isinstance(dim, int) and dim > 0:
+                    self._new_shape[i] = dim
```
gramalingam: Just wondering if it is safer to check for consistency here (between the output shape and shape parameter), in case of incorrectly annotated models.

onnxscript/rewriter/basic_rules.py
```diff
@@ -123,16 +125,37 @@ def pattern(self, op, x, shape_ignored, shape):
...
+
+        # Constraints for shape.
+        self._allowzero = context.nodes[0].attributes.get_int("allowzero", 0)
+        if self._allowzero == 1 and any(self._new_shape == 0):
```
gramalingam: Why do we need the second condition `any(self._new_shape == 0)` here?

Is it just to catch erroneous model/op where there is both 0 and 1 and allow_zero==1?

onnxscript/rewriter/basic_rules.py
Johansmm: I don't think it's necessary. What you're proposing is done by `check_model()`.
I don't think it's the job of the rewriters to verify the consistency of the model, otherwise it would have to be done in all of them!

onnxscript/rewriter/basic_rules.py
Johansmm: The operation of reshape depends on this parameter. If allowzero is zero, ambiguous cases may arise if there is both a 0 and a 1 in the shape, or more than two zeros.

In the case of `allowzero=1`, it is simpler, since if there is at least one zero in the shape, it indicates that the **input is an empty tensor**, so reshaping this is like in numpy and there is no rule that limits it.
E.g. `x = np.ones((0,5,3)); y = np.reshape(x, (0,0,0,15))` is one case supported when `allowzero = 1`.  Here zero **is treated as a static (not dynamic) dimension**, so the subsequent rules do not apply to it.

If I use the suggestion you propose, the rewriter will lose generality. 

Note that this case is verified in detail in the unit tests.

gramalingam: I see. So, it is to enable the replacement of zero by -1, which is valid only if there is at most one zero and no -1, which is used down below

justinchuby:
@Johansmm could you merge from main to resolve conflicts? There was a refactoring done to clean up the name space for all rewrite rules. Thanks

Johansmm:
Last push force rebasing on main

onnxscript/rewriter/rules/common/_basic_rules.py
justinchuby, line 1: You can expose the new rule in https://github.com/microsoft/onnxscript/blob/main/onnxscript/rewriter/rules/common/__init__.py

Johansmm:
Last push force exposing rule

justinchuby (approved):

PR #2521, opened 2025-08-27
Update urls for pytorch documentation

xadupre:
pytorch updated its url.


onnxscript/onnx_opset/_impl/opset18.py
justinchuby, line 1: onnxscript/onnx_opset/_impl is auto generated - we cannot create manual changes to them. I would suggest reverting

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby, line 1: We can still keep the changes here and in the sphinx docs

PR #2522, opened 2025-08-27
[torchlib] Improve handling of SymInt[]

justinchuby:
Previously sizes coming in as `SymInt[]` are first concatenated as INT64 then used. This created inefficiencies where we could not process any static dims from the size list and had to treat the whole shape as dynamic. In aten_expand, this meant we needed to add `Abs` on the shape.

This change updates the functions that take `SymInt[]` such that they are no longer turned into INT64 first. I updated aten_expand to process constant `-1` values so an `Abs` is not required. I also added a helper `merge_dims` to create constants for consecutive constant dims first before concatinating.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/common.py
+++ b/onnxscript/function_libs/torch_lib/ops/common.py
@@ -104,4 +104,8 @@ def merge_dims(dims: Sequence[int | INT64]) -> INT64:
             result_dims.append(current_dim)
     if len(result_dims) == 1:
         return result_dims[0]
-    return op.Concat(result_dims, axis=0)
+
+    # Set the output type to INT64 so op.Concat can be used
+    for dim in result_dims:
+        dim.dtype = ir.DataType.INT64
+    return op.Concat(*result_dims, axis=0)
--- a/onnxscript/function_libs/torch_lib/ops/common.py
+++ b/onnxscript/function_libs/torch_lib/ops/common.py
@@ -5,6 +5,8 @@
 # mypy: disable-error-code="misc,arg-type,type-arg,valid-type,assignment,return-value"
 from __future__ import annotations
 
+from collections.abc import Sequence
+
 import numpy.typing as npt
 import onnx
 
@@ -78,3 +80,28 @@ def constant(
         A constant node.
     """
     return op.Constant(value=ir.tensor(array, dtype=ir.DataType(dtype)))
+
+
+def merge_dims(dims: Sequence[int | INT64]) -> INT64:
+    """Merge consecutive constant dimensions."""
+
+    if not dims:
+        return op.Constant(value_ints=[])
+
+    remaining_dims = list(dims)
+    result_dims = []
+
+    while remaining_dims:
+        current_dim = remaining_dims.pop(0)
+        if isinstance(current_dim, int):
+            merged_dims = [current_dim]
+            # Merge consecutive constant dimensions into a constant node
+            while remaining_dims and isinstance(remaining_dims[0], int):
+                merged_dims.append(remaining_dims.pop(0))
+            result_dims.append(op.Constant(value_ints=merged_dims))
+        else:
+            # A dynamic dimension, just append it
+            result_dims.append(current_dim)
+    if len(result_dims) == 1:
+        return result_dims[0]
+    return op.Concat(result_dims, axis=0)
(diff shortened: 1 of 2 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8293,7 +8291,8 @@ def aten_swapdims(self: TensorType, dim0: int, dim1: int) -> TensorType:
 @torch_op("aten::sym_size.int", trace_only=True)
 def aten_sym_size(self: TensorType, dim: int = 0) -> INT64:
     """sym_size.int(Tensor self, int dim) -> SymInt"""
-    return op.Squeeze(op.Shape(self, end=dim + 1, start=dim))
+
+    return op.Gather(op.Shape(self), dim, axis=0)
```
gramalingam: Is there a difference? If not, retaining the original seems preferable ... a change like this could break patterns used in pattern-matching.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3392,13 +3393,12 @@ def aten_exp2(self: TFloat) -> TFloat:
...
 @torch_op("aten::expand", trace_only=True)
-def aten_expand(self: TTensor, size: TInt, implicit: bool = False) -> TTensor:
+def aten_expand(self: TTensor, size: Sequence[INT64], implicit: bool = False) -> TTensor:
     """expand(Tensor(a) self, SymInt[] size, *, bool implicit=False) -> Tensor(a)"""
-    size = op.Cast(size, to=INT64.dtype)
     # NOTE: PyTorch supports `not changing dim` by -1, but ONNX supports `not changing dim` by 1.
     # To support -1 dim, we need to convert -1 to 1.
-    size = op.Abs(size)
-    return op.Expand(self, size)
+    size = [1 if isinstance(s, int) and s == -1 else s for s in size]
```
gramalingam, line 3402: What can `s` be if not `int`? A symbolic int? Is there an invariant guaranteeing that a symbolic int cannot represent a negative value like -1?

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Which patterns do you have in mind? In particular, this can have less shape calls when sym_size is used on different slices of the same op, which may be preferable? I feel that this combination (shape-gather) is slightly easier to process than shape(end, start)

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: When s is not int it would be INT64 denoting a symint. Since the symints are produced by a shape call (to get a size), I am convinced that we won't see it to be `-1` (which defies the purpose of symints)

onnxscript/function_libs/torch_lib/ops/common.py
```diff
@@ -78,3 +80,35 @@ def constant(
         A constant node.
     """
     return op.Constant(value=ir.tensor(array, dtype=ir.DataType(dtype)))
+
+
+def merge_dims(dims: Sequence[int | INT64]) -> INT64:
```
gramalingam, line 85: I wonder if this is worth it (except in the case when all dims are int constants) ... when I see a concat to produce the shape of a 5-dimensional tensor, I prefer seeing `Concat(dim0, dim1, dim2, dim3, dim4)` instead of `Concat(dim0, dim1, dim2, dim3and4)` ... it is immediately suggestive of a 5-dimensional shape and we can click on dim3 or dim4 as we want, etc.

justinchuby: That makes sense - I can revert if desired

justinchuby: Simplified

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: If dim is int, we should use op.Shape(x, start=, end=...)

xadupre: I vote for the first expression, Gather(Shape) creates 2 allocated tensors, Squeeze(Shape) only one. The first one seems more efficient.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Added a comment

onnxscript/function_libs/torch_lib/ops/core.py

xadupre (approved):

PR #2527, opened 2025-08-28
Increase DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT

justinchuby:
I have seen graphs like `Add(bias, 1)` in gemma3 where bias is an initializer. (Why?) This PR increases the default input limit so these bias initializers can be folded

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -19,7 +19,7 @@ import onnx_ir as ir
 import onnxscript.utils.utils as utils
 from onnxscript.ir import _tape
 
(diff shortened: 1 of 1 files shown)


gramalingam:
Sounds fine to me as long as it doesn't have any other unexpected impact in standard models. The one place where this may be relevant is the ConstantOfShape operator (already discussed in issue/PR elsewhere), because I think it is common to see ConstantOfShape that generate tensors with order of 500 or 1000 values. I think we keep them as ConstantOfShape instead of folding, so there may be no impact there.

xadupre (approved):

PR #2530, opened 2025-08-29
More robust checks for FLOAT8E8M0

justinchuby:
Use hasattr instead of version checks to ensure existence of the FLOAT8E8M0 type.


xadupre (approved):

justinchuby:
@gramalingam could you approve? Thanks

PR #2531, opened 2025-08-29
Refactor rewrite rules into the rewriter.rules namespace

justinchuby:
Organize all rules into a directory that is not with the rewriter infrastructure


gramalingam:
Thanks for doing this, have been thinking about doing this for a while! 

One question: do we need a separate folder, or should this be together with onnx_fusions? There is a clear distinction between ort_fusions and onnx_fusions: but for rules/optimizations that apply to standard onnx, what would the distinction between onnx_fusions and rules be?

justinchuby:
@gramalingam PTAL

gramalingam (approved):

PR #2532, opened 2025-08-29
Add ort-specific passes to ort_fusion

titaiwangms:
There are specific optimization needs from ort shipping models.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/_core.py
+++ b/onnxscript/rewriter/ort_fusions/_core.py
@@ -140,4 +140,19 @@ def optimize_for_ort(
     )
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
+
+    passes = ir.passes.Sequential(
+        # TODO(exporter team): Fold transpose into initializers
+        # Apply the ORT optimization passes.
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -140,4 +140,19 @@ def optimize_for_ort(
     )
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
+
+    passes = ir.passes.Sequential(
+        # TODO(exporter team): Fold transpose into initializers
```
justinchuby: This should be done already

justinchuby (approved):

onnxscript/rewriter/ort_fusions/_core.py
titaiwangms: Could you point me the link?

justinchuby: https://github.com/microsoft/onnxscript/blob/2cc25021e10c564bdc479961c2721390b3a943ef/onnxscript/optimizer/_constant_folding.py#L903

justinchuby (approved):

Issue #2533, opened 2025-09-01
Conversion mismatch for aten::unbind.int


linshokaku:
https://github.com/microsoft/onnxscript/blob/b2d94fe03482f5ed230533bb0b524f1d7ba735fb/onnxscript/function_libs/torch_lib/ops/core.py#L8619-L8624

When split args are specified, `SplitToSequence` disables `keepdims`, resulting in tensor sequences with shapes that differ from those output by `aten::unbind.int`.

cf. https://onnx.ai/onnx/operators/onnx__SplitToSequence.html#attributes

MRADULTRIPATHI:
Hi @microsoft/onnxscript-maintainers 

I’ve implemented a fix for this issue in PR #2534.  

- Updated `aten::unbind.int` to use `SplitToSequence(..., keepdims=False)` (opset18) so the unbound axis is removed correctly, matching PyTorch’s `torch.unbind`.  
- Added an eager-mode test (`test_unbind_matches_torch`) to validate shapes vs PyTorch.  
- Verified locally with pytest → ✅ Passed.  

Kindly review the PR when you get a chance. Thanks a lot

MRADULTRIPATHI:
Hi @justinchuby, @linshokaku 

I’ve updated the implementation and fixed the lint issues for `aten::unbind.int`.  
- Updated core implementation to use `SplitToSequence(..., keepdims=False)` so it matches PyTorch’s `torch.unbind` behavior.  
- Added eager-mode test (`test_unbind_matches_torch`) to validate output shapes vs PyTorch.  
- Fixed all lint/style errors reported by `lintrunner`.  
- Verified locally with:  
  ```bash
  pytest -v docs/test/test_documentation_examples.py

PR #2536, opened 2025-09-02
[torchlib] Modify aten_unbind to use None for split_sizes

justinchuby:
According to https://onnx.ai/onnx/operators/onnx__SplitToSequence.html#summary, `If the argument split is not specified, a default scalar value of 1 is used as the value of split`, and this is the only case when `keepdims` can be set to `0`. 

Fixes https://github.com/microsoft/onnxscript/issues/2533


titaiwangms (approved):
I feel like the wording in ONNX spec for mutually exclusive attribute/input is ambiguous: `Keep the split dimension or not. Default 1, which means we keep split dimension. If input ‘split’ is specified, this attribute is ignored.`

PR #2540, opened 2025-09-04
[torch] Fix incorrect Concat when processing dynamic paddings

kistenklaus:
See issue #2539 for a better explanation. 

I know crazy stuff right =^).


justinchuby (approved):
Good catch, thank you!

titaiwangms:
> Good catch, thank you!

This is a serious bug. Dis we never use this op in LLM? We missed it until now.

PR #2542, opened 2025-09-04
Update SplitToSequence in constant folding

titaiwangms:
`Split` input (SymbolicTensor) could have no const_value, but with shape that gives us information of how many outputs an op.Split should return.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -9,7 +9,7 @@ import dataclasses
 import logging
 import math
 import typing
-from typing import Any, Callable, Iterable, Sequence, Union
+from typing import Any, Callable, Collection, Iterable, Sequence, Union
 
 import numpy as np
 import onnx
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -34,13 +34,6 @@
     }
 )
 
-# A list of ops to always fold regardless of their input size limits, as long as
-# they are the single consumer of the large input tensors
-_DEFAULT_ALWAYS_FOLD_OPS = frozenset(
```
justinchuby, line 39: I think there was a merge conflict. This code should not be removed?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1308,6 +1290,6 @@ def fold_constants(
         shape_inference=onnx_shape_inference,
         input_size_limit=input_size_limit,
         output_size_limit=output_size_limit,
-        should_fold=should_fold,
```
justinchuby, line 1311: This should be reverted

PR #2544, opened 2025-09-04
[Reland] Update SplitToSequence in constant folding

titaiwangms:
Split input (SymbolicTensor) could have no const_value, but with shape that gives us information of how many outputs an op.Split should return.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -818,9 +818,10 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     if split_value is None and split_shape is None:
         return None
 
-    if split_shape is not None:
+    if isinstance(split_shape, tuple) and len(split_shape) == 1:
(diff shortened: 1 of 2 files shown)


onnxscript/optimizer/_constant_folding.py
justinchuby, line 1: Maybe add a test?

titaiwangms: Done

onnxscript/optimizer/_constant_folding.py
```diff
@@ -818,9 +818,10 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     if split_value is None and split_shape is None:
         return None
 
-    if split_shape is not None:
+    if isinstance(split_shape, tuple) and len(split_shape) == 1:
```
justinchuby, line 821: Q: What other types can this be? Is the isinstance necessary or can we just do `split_shape is not None and ...`?

titaiwangms: None

PR #2548, opened 2025-09-08
[torchlib] Fix repeat_interleave when repeats is a symbolic tensor

xadupre:


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7327,9 +7327,18 @@ def aten_repeat_interleave_self_int(
...
+        tile_repeat = op.Constant(value=ir.tensor(tiles, dtype=INT64.dtype))
+    else:
+        # repeats is a symbolic tensor
+        tile_repeat = op.Concat(
+            op.Constant(value=ir.tensor([1] * pos_dim, dtype=INT64.dtype)),
+            op.Reshape(repeats, op.Constant(value=ir.tensor([-1], dtype=INT64.dtype))),
+            op.Constant(value=ir.tensor([1] * (self_rank - pos_dim), dtype=INT64.dtype)),
+            axis=0,
+        )
     tiled = op.Tile(unsqueezed, tile_repeat)
```
justinchuby: Circling back: do you think using Expand would simplify parts of this graph, or would they be similar? (Performance wise and ease of analysis)

tests/function_libs/torch_lib/e2e_ops_tests.py
```diff
@@ -137,6 +137,33 @@ def forward(self, x, ind):
...
+            torch.arange(4, dtype=torch.float32).reshape((2, 2)),
+            torch.arange(6, dtype=torch.float32).reshape((2, 3)),
+        )
+        onnx_program = torch.onnx.export(
+            Model(),
+            inputs,
+            dynamo=True,
+            optimize=False,
+        )
+        onnx_program = torch.onnx.export(
```
justinchuby: Duplicated?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7327,9 +7327,18 @@ def aten_repeat_interleave_self_int(
...
     if self_rank == 1:
         return op.Identity(tiled)
```
justinchuby, line 7349: pos_dim below; rk -> rank;

justinchuby (approved):

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: done

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7384,20 +7384,22 @@ def aten_repeat_interleave_Tensor(
...
         shape_x = op.Shape(self, start=1)
         self = op.Reshape(self, op.Concat(shape_x0, [-1], axis=0))
-    elif rk == 1:
+    elif rank == 1:
         shape_x = None
         self = op.Reshape(self, [-1, 1])
     else:
-        if rk != 2:
-            raise NotImplementedError(f"rank(self)={rk} not implemented for repeat_interleave")
+        if rank != 2:
```
justinchuby, line 7404: Is this logic intended? We have rank==1 and rank>2 covered, so what's left is rank==2 and rank<1?

tests/function_libs/torch_lib/e2e_ops_tests.py
xadupre: removed

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Sorry I see it now

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: changed into Expand

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7332,16 +7332,25 @@ def aten_repeat_interleave_self_int(
...
+        tile_repeat = op.Constant(value=ir.tensor(tiles, dtype=INT64.dtype))
+    else:
+        # repeats is a symbolic tensor
+        tile_repeat = op.Concat(
+            op.Constant(value=ir.tensor([1] * pos_dim, dtype=INT64.dtype)),
+            op.Reshape(repeats, op.Constant(value=ir.tensor([-1], dtype=INT64.dtype))),
+            op.Constant(value=ir.tensor([1] * (self_rank - pos_dim), dtype=INT64.dtype)),
+            axis=0,
+        )
+    tiled = op.Expand(unsqueezed, tile_repeat)
```
justinchuby, line 7347: Is tile_repeat the target shape expected by expand? I feel it should be different from the "repeated copies" expected by Tile but I may be missing something?

justinchuby: Sorry. It’s correct because the dimension is just the number of times it should repeat. So that’s the same for expand and tile. Is that right? Just checking my understanding

justinchuby: @titaiwangms for another review

xadupre: Tile = Expand if the repeated dimensions are equal to 1 otherwise they cannot be exchanged but thats the case here.

PR #2550, opened 2025-09-08
Do not try to fold op.SplitToSequence when split is `None`

titaiwangms:
split is an optional input to op.SplitToSequence.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -784,6 +784,9 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     This allows downstream `SequenceAt` users to be replaced by `split_x` accordingly.
     """
     input = node.inputs[0]
+    if len(node.inputs) == 1:
+        # split is not provided
+        return None
     split = node.inputs[1]
     output = node.outputs[0]
 


onnxscript/optimizer/_constant_folding.py
```diff
@@ -784,6 +784,9 @@ def split_to_sequence(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     This allows downstream `SequenceAt` users to be replaced by `split_x` accordingly.
     """
     input = node.inputs[0]
+    if len(node.inputs) == 1:
+        # split is not provided
```
justinchuby, line 788: When split is not provided it is a scalar 1. Can we use this info?

titaiwangms: Is it? The issue @xadupre brought up was only one input. model id: facebook/sam-vit-base, so it hit indexerror.

justinchuby: I think at least according to the spec: https://onnx.ai/onnx/operators/onnx__SplitToSequence.html#summary

titaiwangms: https://github.com/microsoft/onnxscript/blob/cec5396648fa1aacfd914e6c838642efd8420976/onnxscript/function_libs/torch_lib/ops/core.py#L8732

titaiwangms: Probably came from here

PR #2554, opened 2025-09-09
Fix CI tests for the latest onnx-ir version

justinchuby:
Signed-off-by: Justin Chu <justinchuby@users.noreply.github.com>

### DIFF
--- a/onnxscript/rewriter/_fusion_utils.py
+++ b/onnxscript/rewriter/_fusion_utils.py
@@ -13,19 +13,6 @@ from onnxscript.rewriter._rewrite_rule import RewriteRule, RewriteRuleSet
 Dim = Union[int, ir.SymbolicDim]
 
 
-def _check_shape(bindings: dict[str, Dim], val: ir.Value, shape: Sequence[str]) -> bool:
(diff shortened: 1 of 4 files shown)


titaiwangms (approved):

onnxscript/rewriter/_pattern_ir.py
```diff
@@ -126,7 +126,14 @@ def __init__(self, value: SupportedAttrTypes):
...
 
     def matches(self, attr: ir.Attr) -> bool:
-        return isinstance(attr, ir.Attr) and attr.value == self._value
+        if attr.type in {
+            ir.AttributeType.INTS,
+            ir.AttributeType.FLOATS,
+            ir.AttributeType.STRINGS,
+        }:
+            # Since the type of attr.value is Sequence, we need to convert to the same type for comparison.
+            return tuple(attr.value) == tuple(self._value)
```
xadupre, line 135: should we cast every value in the tuple?

justinchuby: Discussed offline. Will do when needed

PR #2555, opened 2025-09-09
[Rewriter] Implement zero bias removal for Conv operations and related rules

### LINKED ISSUE #2547: Rewrite rule: Remove optional bias
Image

Remove optional bias in conv and other ops when it is all zero.

- Also: Enable more conv fusion rules as default or ort optimization rules.

whyvineet:
Fix #2547

### DIFF
--- a/onnxscript/rewriter/rules/common/_remove_zero_bias.py
+++ b/onnxscript/rewriter/rules/common/_remove_zero_bias.py
@@ -1,0 +1,203 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Remove optional bias when it is all zero from Conv and related operations."""
+
+from __future__ import annotations
+
+import numpy as np
+
+from onnxscript import ir
+from onnxscript.rewriter._basics import MatchResult
+from onnxscript.rewriter._rewrite_rule import RewriteRuleClassBase, RewriteRuleSet
+
+
+class _RemoveZeroBiasBase(RewriteRuleClassBase):
+    """Base class for removing zero bias from operations."""
+
+    def __init__(self, op_type: str):
+        super().__init__(remove_nodes=False)
+        self.op_type = op_type
+
+    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
+        """Remove the bias input from the operation."""
+        return op.op(
+            self.op_type,
+            inputs=[x, w],  # Remove bias input
+        )
+
+    def check(self, context, x: ir.Value, w: ir.Value, b: ir.Value, **_) -> MatchResult:
+        """Check if the bias is present and is all zeros."""
+        del context  # Unused
+        check_result = MatchResult()
+
+        # Check if bias is a constant/initializer
+        if b.const_value is None:
+            return check_result.fail("Bias is not a constant/initializer.")
+
+        # Check if bias is all zeros
+        bias_array = b.const_value.numpy()
+        if not np.allclose(bias_array, 0.0, atol=1e-8):
+            return check_result.fail("Bias is not all zeros.")
+
+        return check_result
+
+
+class RemoveZeroBiasFromConv(_RemoveZeroBiasBase):
+    """Remove zero bias from Conv operations."""
+
+    def __init__(self):
+        super().__init__("Conv")
+
+    def pattern(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
+        return op.Conv(x, w, b, _outputs=["conv_out"])
+
+    def check(self, context, x: ir.Value, w: ir.Value, b: ir.Value, conv_out: ir.Value, **_) -> MatchResult:
+        """Check if the bias is present and is all zeros."""
+        del context  # Unused
+        check_result = MatchResult()
+
+        # Check if bias is a constant/initializer
+        if b.const_value is None:
+            return check_result.fail("Bias is not a constant/initializer.")
+
+        # Check if bias is all zeros
+        bias_array = b.const_value.numpy()
+        if not np.allclose(bias_array, 0.0, atol=1e-8):
+            return check_result.fail("Bias is not all zeros.")
+
+        return check_result
+
+    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value, conv_out: ir.Value) -> ir.Value:
+        """Remove the bias input from the operation."""
+        # Get the Conv node that produced conv_out to access its attributes
+        conv_node = conv_out.producer()
+
+        # Create new Conv with preserved attributes but without bias
+        return op.op(
+            "Conv",
+            inputs=[x, w],  # Remove bias input
+            attributes=conv_node.attributes,
+            domain=conv_node.domain,
+        )
+
+
+class RemoveZeroBiasFromConvTranspose(_RemoveZeroBiasBase):
+    """Remove zero bias from ConvTranspose operations."""
+
+    def __init__(self):
+        super().__init__("ConvTranspose")
+
+    def pattern(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
+        return op.ConvTranspose(x, w, b, _allow_other_inputs=False, _outputs=["conv_out"])
+
+    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value, conv_out: ir.Value) -> ir.Value:
+        """Remove the bias input from the operation."""
+        # Get the ConvTranspose node that produced conv_out to access its attributes
+        conv_node = conv_out.producer()
+
+        # Create new ConvTranspose with preserved attributes but without bias
+        return op.op(
+            "ConvTranspose",
+            inputs=[x, w],  # Remove bias input
+            attributes=conv_node.attributes,
+            domain=conv_node.domain,
+        )
+
+
+class RemoveZeroBiasFromQLinearConv(_RemoveZeroBiasBase):
+    """Remove zero bias from QLinearConv operations."""
+
+    def __init__(self):
+        super().__init__("QLinearConv")
+
+    def pattern(self, op: ir.tape.Tape, x, x_scale, x_zero_point, w, w_scale, w_zero_point,
+                y_scale, y_zero_point, b: ir.Value) -> ir.Value:
+        return op.QLinearConv(
+            x, x_scale, x_zero_point, w, w_scale, w_zero_point,
+            y_scale, y_zero_point, b, _allow_other_inputs=False, _outputs=["conv_out"]
+        )
+
+    def check(self, context, x, x_scale, x_zero_point, w, w_scale, w_zero_point,
+              y_scale, y_zero_point, b: ir.Value, conv_out: ir.Value, **_) -> MatchResult:
+        """Check if the bias (b) is present and is all zeros."""
+        del context  # Unused
+        check_result = MatchResult()
+
+        # Check if bias is a constant/initializer
+        if b.const_value is None:
+            return check_result.fail("Bias is not a constant/initializer.")
+
+        # Check if bias is all zeros
+        bias_array = b.const_value.numpy()
+        if not np.allclose(bias_array, 0.0, atol=1e-8):
+            return check_result.fail("Bias is not all zeros.")
+
+        return check_result
+
+    def rewrite(self, op: ir.tape.Tape, x, x_scale, x_zero_point, w, w_scale, w_zero_point,
+                y_scale, y_zero_point, b: ir.Value, conv_out: ir.Value) -> ir.Value:
+        """Remove the bias input from the operation."""
+        # Get the QLinearConv node that produced conv_out to access its attributes
+        conv_node = conv_out.producer()
+
+        # Create new QLinearConv with preserved attributes but without bias
+        return op.op(
+            "QLinearConv",
+            inputs=[x, x_scale, x_zero_point, w, w_scale, w_zero_point,
+                    y_scale, y_zero_point],  # Remove bias input
+            attributes=conv_node.attributes,
+            domain=conv_node.domain,
+        )
+
+
+class RemoveZeroBiasFromGemm(_RemoveZeroBiasBase):
+    """Remove zero bias from Gemm operations."""
+
+    def __init__(self):
+        super().__init__("Gemm")
+
+    def pattern(self, op: ir.tape.Tape, a: ir.Value, b: ir.Value, c: ir.Value) -> ir.Value:
+        return op.Gemm(a, b, c, _allow_other_inputs=False, _outputs=["gemm_out"])
+
+    def check(self, context, a: ir.Value, b: ir.Value, c: ir.Value, gemm_out: ir.Value, **_) -> MatchResult:
+        """Check if the bias (c) is present and is all zeros."""
+        del context  # Unused
+        check_result = MatchResult()
+
+        # Check if bias is a constant/initializer
+        if c.const_value is None:
+            return check_result.fail("Bias is not a constant/initializer.")
+
+        # Check if bias is all zeros
+        bias_array = c.const_value.numpy()
+        if not np.allclose(bias_array, 0.0, atol=1e-8):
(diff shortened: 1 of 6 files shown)


whyvineet:
> @whyvineet please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

justinchuby:
@AyoubMDL would you be able to help review/critique? Thanks!

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,203 @@
...
+from onnxscript.rewriter._basics import MatchResult
+from onnxscript.rewriter._rewrite_rule import RewriteRuleClassBase, RewriteRuleSet
+
+
+class _RemoveZeroBiasBase(RewriteRuleClassBase):
+    """Base class for removing zero bias from operations."""
+
+    def __init__(self, op_type: str):
+        super().__init__(remove_nodes=False)
+        self.op_type = op_type
```
AyoubMDL: (not mandatory), maybe declare op_type as ClassVar in each subclass instead of declaring the init (something like done here https://github.com/microsoft/onnxscript/blob/821015a652c31381349c5ec7de62b8a21a0fe3cb/onnxscript/rewriter/rules/common/_fuse_batchnorm.py#L109)

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,203 @@
...
+            inputs=[x, w],  # Remove bias input
+        )
+
+    def check(self, context, x: ir.Value, w: ir.Value, b: ir.Value, **_) -> MatchResult:
+        """Check if the bias is present and is all zeros."""
+        del context  # Unused
+        check_result = MatchResult()
+
+        # Check if bias is a constant/initializer
+        if b.const_value is None:
```
AyoubMDL: here you can use `ir.convenience.get_const_tensor` to also handle Constant nodes

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,203 @@
...
+            return check_result.fail("Bias is not a constant/initializer.")
+
+        # Check if bias is all zeros
+        bias_array = b.const_value.numpy()
+        if not np.allclose(bias_array, 0.0, atol=1e-8):
+            return check_result.fail("Bias is not all zeros.")
+
+        return check_result
```
AyoubMDL: This is repeated, you've already done this in the base class

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class RemoveZeroBiasFromConvTranspose(_RemoveZeroBiasBase):
+    """Remove zero bias from ConvTranspose operations."""
+
+    def __init__(self):
+        super().__init__("ConvTranspose")
+
+    def pattern(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
+        return op.ConvTranspose(x, w, b, _allow_other_inputs=False, _outputs=["conv_out"])
```
AyoubMDL: Why you specify `_allow_other_inputs=False` ?

onnxscript/rewriter/rules/common/_remove_zero_bias.py
AyoubMDL, line 1: In the child class you are redefining the rewrite and check methods, I think you can encapsulate that in the base (This is already True for the check), in the rewrite, you'd need to handle the attributes

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
AyoubMDL, line 1: You should also test on the other Ops (Gemm, ConvTranspose...)

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,161 @@
...
+from onnxscript import ir
+from onnxscript.ir import convenience
+from onnxscript.rewriter._basics import MatchResult
+from onnxscript.rewriter._rewrite_rule import RewriteRuleClassBase, RewriteRuleSet
+
+
+class _RemoveZeroBiasBase(RewriteRuleClassBase):
+    """Base class for removing zero bias from operations."""
+
+    op_type: ClassVar
```
AyoubMDL: You can remove this

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
```diff
@@ -0,0 +1,318 @@
...
+    remove_zero_bias_from_conv_transpose_rule,
+    remove_zero_bias_from_gemm_rule,
+    remove_zero_bias_from_qlinear_conv_rule,
+)
+
+
+def test_remove_zero_bias_from_conv():
+    """Test that zero bias is removed from Conv operations."""
+    # Create a simple Conv with zero bias using ONNX parser
+    model_proto = onnx.parser.parse_model(
```
AyoubMDL: you can use directly `ir.from_onnx_text`

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # Check that bias input was removed
+    conv_node = None
+    for node in model.graph:
+        if node.op_type == "Conv":
+            conv_node = node
+            break
+
+    assert conv_node is not None, "Conv node not found"
+    assert len(conv_node.inputs) == 2, f"Expected 2 inputs after optimization, got {len(conv_node.inputs)}"
```
AyoubMDL: I think you can create a helper method as the same code is used for other rules. Also, I think you should also check the inference to compare outputs

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
```diff
@@ -0,0 +1,318 @@
...
+if __name__ == "__main__":
+    test_remove_zero_bias_from_conv()
+    test_conv_with_non_zero_bias_unchanged()
+    test_remove_zero_bias_from_conv_transpose()
+    test_conv_transpose_with_non_zero_bias_unchanged()
+    test_remove_zero_bias_from_gemm()
+    test_gemm_with_non_zero_bias_unchanged()
+    test_remove_zero_bias_from_qlinear_conv()
+    test_qlinear_conv_with_non_zero_bias_unchanged()
+    print("All tests passed!")
```
AyoubMDL: No need to call all the functions here
```suggestion
    unittest.main()
```

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
+        """Remove the bias input from the operation."""
+        return op.op(
+            self.op_type,
+            inputs=[x, w],  # Remove bias input
+        )
```
AyoubMDL: You can remove all the rewrite methods from the subclasses and keep only the one in the base class
```suggestion
    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, **_) -> ir.Value:
        """Remove the bias input from the operation."""
        node = out.producer()
        return op.op(
            self.op_type,
            inputs=[x, w],
            attributes=node.attributes
        )
```

whyvineet: Can you explain this further? While making this change, I'm getting multiple errors...

AyoubMDL: Your base class holds the logic of the rewrite (don't need to redefine the rewrite in the subclasses). Here is an example with your base class and Conv subclass:
```python
class _RemoveZeroBiasBase(RewriteRuleClassBase):
    """Base class for removing zero bias from operations."""

    def rewrite(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, **_) -> ir.Value:
        """Remove the bias input from the operation."""
        node = out.producer()
        return op.op(
            self.op_type,
            inputs=[x, w],
            attributes=node.attributes
        )

    def _check_bias_is_zero(self, bias_value: ir.Value) -> MatchResult:
        """Check if the bias value is present and is all zeros."""
        check_result = MatchResult()

        # Check if bias is a constant/initializer
        bias_tensor = convenience.get_const_tensor(bias_value)
        if bias_tensor is None:
            return check_result.fail("Bias is not a constant/initializer.")

        # Check if bias is all zeros
        bias_array = bias_tensor.numpy()
        if not np.allclose(bias_array, 0.0, atol=1e-8):
            return check_result.fail("Bias is not all zeros.")

        return check_result

    def check(self, context, x: ir.Value, w: ir.Value, b: ir.Value, **_) -> MatchResult:
        """Check if the bias is present and is all zeros."""
        del context  # Unused
        return self._check_bias_is_zero(b)


class RemoveZeroBiasFromConv(_RemoveZeroBiasBase):
    """Remove zero bias from Conv operations."""

    op_type: ClassVar = "Conv"

    def pattern(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
       # name the node output as "out" and use it for other subclasses
        return op.Conv(x, w, b, _outputs=["out"])
```

Tell me if it's clear

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,133 @@
...
+
+    op_type: ClassVar = "Gemm"
+
+    def pattern(self, op: ir.tape.Tape, a: ir.Value, b: ir.Value, c: ir.Value) -> ir.Value:
+        return op.Gemm(a, b, c, _outputs=["out"])
+
+    def check(self, context, a: ir.Value, b: ir.Value, c: ir.Value, **_) -> MatchResult:
+        """Check if the bias (c parameter) is present and is all zeros."""
+        del context  # Unused
+        return self._check_bias_is_zero(c)
```
AyoubMDL: You can remove this (call your inputs x,w,b so it is the same as the others)

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,133 @@
...
+
+class _RemoveZeroBiasBase(RewriteRuleClassBase):
+    """Base class for removing zero bias from operations."""
+
+    def rewrite(self, op: ir.tape.Tape, out: ir.Value, **_) -> ir.Value:
+        """Remove the bias input from the operation."""
+        node = out.producer()
+
+        original_inputs = list(node.inputs)
+        inputs_without_bias = original_inputs[:-1]
```
AyoubMDL: With my suggestion on Gemm, you can remove this. And pass x, w as arguments to the rewrite

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -0,0 +1,133 @@
...
+
+        return op.op(
+            self.op_type,
+            inputs=inputs_without_bias,
+            attributes=node.attributes,
+            domain=node.domain,
```
AyoubMDL: I don't think you need the domain. Is their a reason why you specified it here ?

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
AyoubMDL, line 1: Maybe you can pass some attributes when testing to check that every info is correctly transferred (e.g. stride, transA...)

onnxscript/rewriter/rules/common/_remove_zero_bias.py
whyvineet: instead of hardcoding `x` and `w` parameters, the base class can dynamically handle any number of inputs by filtering out just the bias parameter ('b') and keeping everything else in the new changes made by me.

whyvineet:
@AyoubMDL, is there anything else I should update or improve?

onnxscript/rewriter/rules/common/_remove_zero_bias.py
```diff
@@ -21,12 +21,15 @@ def rewrite(self, op: ir.tape.Tape, out: ir.Value, **_) -> ir.Value:
         """Remove the bias input from the operation."""
         node = out.producer()
 
-        original_inputs = list(node.inputs)
-        inputs_without_bias = original_inputs[:-1]
+        # Filter out the bias parameter and keep all other inputs
+        inputs = []
+        for param_name, param_value in _.items():
```
justinchuby: ?

whyvineet: I understand now... There was just a misunderstanding on my part!

onnxscript/rewriter/rules/common/_remove_zero_bias.py
AyoubMDL: I think you can remove `domain=node.domain` as you are using native onnx ops

whyvineet:
> The implementation looks good to me. I'll let @justinchuby share additional comments

Thanks for the guidance @AyoubMDL! And I don't know how I missed `domain=node.domain` from the last review... I removed it now!

onnxscript/rewriter/rules/common/_remove_zero_bias_test.py
```diff
@@ -0,0 +1,431 @@
...
+from onnxscript.rewriter import testing
+from onnxscript.rewriter.rules.common._remove_zero_bias import (
+    remove_zero_bias_from_conv_rule,
+    remove_zero_bias_from_conv_transpose_rule,
+    remove_zero_bias_from_gemm_rule,
+    remove_zero_bias_from_qlinear_conv_rule,
+)
+
+
+def _apply_rule_and_check_optimization(
```
justinchuby, line 19: This test is a little too complicated to maintain. Please check other tests in the directory and simplify.

justinchuby:
The error is saying the Sequence[int] attributes are supposed to be tuples but the expected values are lists. You can check where you provide the expected outputs.

onnxscript/rewriter/onnxruntime/bfloat16_utils/bfloat16_converter_test.py
```diff
@@ -6,20 +6,23 @@
...
 
 class Bfloat16ConversionTest(unittest.TestCase):
     def setUp(self) -> None:
-        self.v0 = ir.val(name="v0", shape=ir.Shape([2, 3, 4]))
-        self.v0.dtype = ir.DataType.BFLOAT16
-        self.v1 = ir.val(name="v1", shape=ir.Shape([2, 3, 4]))
-        self.v1.dtype = ir.DataType.BFLOAT16
-        self.v2 = ir.val(name="v2", shape=ir.Shape([2, 3, 4]))
-        self.v2.dtype = ir.DataType.BFLOAT16
+        self.v0 = ir.Value(
```
justinchuby: Please revert. ir.val is a convenience function.

whyvineet: It was failing test cases... Okh I'll do so!

justinchuby:
Please use judgement before applying AI edits, and please engage the reviewers if you have clarification questions. It is preferred if you understand what the edits should be before spending time doing them. Have a discussion and lay out your thought process before taking actions. Pure AI edits that don't help with addressing the issues raised in reviews and is a waste of everyone's time.

PR #2566, opened 2025-09-17
[torchlib] Fix aten_gather to correctly handle scalar indices

### LINKED ISSUE #2564: behavior discrepancy when `aten::gather` receives a scalar index
https://github.com/microsoft/onnxscript/blob/ea790222deaada24d4567cd49400b8838a96c31c/onnxscript/function_libs/torch_lib/ops/core.py#L3802-L3822

When a scalar index is provided to aten::gather, the following behavior occurs:

```python
>>> import torch
>>> x = torch.arange(3)
>>> index = torch.tensor(1)
>>> out = torch.ops.aten.gather.default(x, 0, index)
>>> out.shape
torch.Size([])
>>> out
tensor(1)
```

Therefore, when a 0-dimensional tensor is passed, we'll need to implement branching logic for squeeze/unsqueeze operations.

```python
        is_scalar_index = len(index.shape) == 0
        if is_scalar_index:
            index = op.Unsqueeze(index, [0])

        index = op.Cast(index, to=INT64.dtype)
        result = op.GatherElements(self, index, axis=dim)

        if is_scalar_index:
            result = op.Squeeze(result, [0])
```

linshokaku:
Fixes #2564

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3814,11 +3814,15 @@ def aten_gather(
         else:
             return op.Expand(self, op.Shape(index))
 
-    if len(index.shape) == 0:
-        return op.Identity(self)
+    is_scalar_index = len(index.shape) == 0
+    if is_scalar_index:
+        index = op.Unsqueeze(index, [0])
 
     index = op.Cast(index, to=INT64.dtype)
     result = op.GatherElements(self, index, axis=dim)
+
+    if is_scalar_index:
+        result = op.Squeeze(result, [0])
     return result
 
 


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3814,11 +3814,15 @@ def aten_gather(
         else:
             return op.Expand(self, op.Shape(index))
 
-    if len(index.shape) == 0:
-        return op.Identity(self)
+    is_scalar_index = len(index.shape) == 0
```
justinchuby, line 3817: The spec requires "index" to be the same rank as input. Is it guaranteed that when index is a scalar, input will only be scalar or 1d?

linshokaku: Yes. When providing a scalar index to `torch.ops.aten.gather`, the behavior of `self` by rank is as follows:

```python
>>> torch.ops.aten.gather.default(torch.zeros(()), 0, torch.tensor(0))
tensor(0.)
>>> torch.ops.aten.gather.default(torch.zeros(3), 0, torch.tensor(0))
tensor(0.)
>>> torch.ops.aten.gather.default(torch.zeros((3, 3)), 0, torch.tensor(0))
  File "/tmp/fx2onnx_venv/lib/python3.10/site-packages/torch/_ops.py", line 723, in __call__
    return self._op(*args, **kwargs)
RuntimeError: Index tensor must have the same number of dimensions as input tensor
```

justinchuby: Thanks for confirming

justinchuby (approved):

PR #2567, opened 2025-09-18
[rewriter] Remove generic pattern matcher

justinchuby:
It is obsolete and the capability is covered by the simple pattern matcher.


onnxscript/rewriter/generic_pattern_test.py
```diff
@@ -1,607 +0,0 @@
...
-from onnxscript.rewriter import generic_pattern, pattern
-
-FLOAT = onnx.TensorProto.FLOAT
-
-
-@parameterized.parameterized_class(
-    ("matcher_algo",),
-    [
-        (generic_pattern.GenericPatternMatcher,),
-        (pattern.SimplePatternMatcher,),
```
justinchuby, line 27: Maybe we want to move some tests

gramalingam (approved):

PR #2569, opened 2025-09-23
[torchlib] Migrate torchvision implementations

justinchuby:
Adapted from https://github.com/pytorch/vision/blob/main/torchvision/ops/_register_onnx_ops.py. Also migrated tests.


titaiwangms:
Where do we migrate this from? Was there a test we can move to here as well?

gramalingam (approved):

PR #2575, opened 2025-09-24
Record names of contributing values in the constant folding pass

justinchuby:
Record names of contributing values in the constant folding pass to the newly created output as metadata, so that downstream users like Olive can use the info for further manipulations.

This is useful for Olive to identify transposed lora weights in the graph.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -1236,7 +1236,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
     ) -> None:
         logger.debug("Replacing node: %s::%s %s", node.domain, node.op_type, node.name)
 
-        # Record the values that has contributed to the replacement
+        # Record the names of the values that has contributed to the replacement
         _record_contributing_values(node, replacement)
 
         ir.convenience.replace_nodes_and_values(


onnxscript/optimizer/_constant_folding.py
```diff
@@ -914,6 +916,22 @@ def merge_dims(dim1, dim2):
...
+        if input is None:
+            continue
+        folded_from.union(input.meta.get(_FOLDED_FROM_KEY, set()))
+        assert input.name is not None
+        folded_from.add(input.name)
+
+    for new_output in replacement.new_outputs:
+        if new_output is None:
+            continue
+        new_output.meta[_FOLDED_FROM_KEY] = folded_from
```
gramalingam: Does the serializer store `str(folder_from)` in the metadata? Just curious what types of values are allowed in `value.meta[...]`

gramalingam (approved):

onnxscript/optimizer/_constant_folding.py
justinchuby: value.meta can store anything (<str: Any>). It is not serialized. value.metadata_props in contrast is a <str: str> mapping and is serialized.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1203,9 +1231,14 @@ def convert(av):
             )
         return None
 
-    def replace_node(self, node: ir.Node, replacement, root: ir.Graph | ir.Function) -> None:
+    def replace_node(
+        self, node: ir.Node, replacement: Replacement, root: ir.Graph | ir.Function
+    ) -> None:
         logger.debug("Replacing node: %s::%s %s", node.domain, node.op_type, node.name)
 
+        # Record the values that has contributed to the replacement
```
```suggestion
        # Record the names of the values that has contributed to the replacement
```

PR #2578, opened 2025-09-28
Merge output shape with input shape instead of override

wodesuck:
`_constant_folding.cast` override `output.shape` with `input.shape`, that may make a static shape to dynamic shape. Here should use `_merge_shapes` instead.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -491,9 +491,7 @@ def cast(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     # should handle this. Only the optimization to eliminate redundant Cast ops
     # should be needed here.
 
-    input_shape = input.shape
-    if input_shape is not None:
-        output.shape = input_shape.copy()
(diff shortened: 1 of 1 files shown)


justinchuby:
@gramalingam @titaiwangms

justinchuby (commented):
Could you share a concrete example when you see this case?

wodesuck:
```python
import torch
import torch.nn as nn


class Model(nn.Module):
    def forward(self, x):
        return x.new_zeros(x.shape, dtype=torch.int64)


def main():
    model = Model()
    args = torch.rand(4, 4),
    batch = torch.export.Dim("batch")
    dynamic_shapes = {"x": {0: batch}}
    torch.onnx.export(
        model,
        args,
        "model_test.onnx",
        dynamic_shapes=dynamic_shapes,
        dynamo=True,
    )


if __name__ == "__main__":
    main()
```

image

The output shape should be `batch x 4`, but now it's `? x ?`.

justinchuby (approved):
Thanks!

PR #2587, opened 2025-09-30
Extend utilities for checking a scalar value

gramalingam:
Extend the `is_singleton_value` utility to check for singleton values that may be either 0D or 1D tensors.

### DIFF
--- a/onnxscript/rewriter/rules/fusion/_rotary_embedding.py
+++ b/onnxscript/rewriter/rules/fusion/_rotary_embedding.py
@@ -43,10 +43,10 @@ class RotaryEmbedding23Fusion(pattern.RewriteRuleClassBase):
     def check(self, op, x, start1, end1, start2, end2, one1, one2, **_) -> pattern.MatchResult:  # type: ignore[name-defined]
         check_result = pattern.MatchResult()
 
-        if not _ir_utils.is_singleton_value(one1, 1, rank=(0, 1)):
-            return check_result.fail("Unsqueeze axes is not [1] or 1", one1)
-        if not _ir_utils.is_singleton_value(one2, 1, rank=(0, 1)):
-            return check_result.fail("Unsqueeze axes is not [1] or 1", one2)
+        if not _ir_utils.is_singleton_value(one1, 1):
+            return check_result.fail("Unsqueeze axes is not [1]", one1)
+        if not _ir_utils.is_singleton_value(one2, 1):
+            return check_result.fail("Unsqueeze axes is not [1]", one2)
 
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/rules/fusion/_rotary_embedding.py
```diff
@@ -43,20 +43,10 @@ def pattern(self, op, x, freqs, start1, end1, start2, end2, one1, one2):
...
-            )
-
-        if not is_one(one1):
-            return check_result.fail("Unsqueeze axes is not [1]", one1)
-        if not is_one(one2):
-            return check_result.fail("Unsqueeze axes is not [1]", one2)
+        if not _ir_utils.is_singleton_value(one1, 1, rank=(0, 1)):
+            return check_result.fail("Unsqueeze axes is not [1] or 1", one1)
+        if not _ir_utils.is_singleton_value(one2, 1, rank=(0, 1)):
+            return check_result.fail("Unsqueeze axes is not [1] or 1", one2)
```
justinchuby: Unsqueeze must take a 1d axes. I think the previous pattern was due to an incorrect logic from torchlib which was fixed later

onnxscript/rewriter/_ir_utils.py
```diff
@@ -78,23 +78,34 @@ def get_numpy_value(val: ir.Value | None) -> np.ndarray | None:
     return None
 
 
-def get_singleton_value(val: ir.Value | None, rank: int | None = None):
+def get_singleton_value(val: ir.Value | None, rank: int | Sequence[int] | None = None):
```
justinchuby, line 81: changes to the helper lgtm. I think we can simplify the pattern though.

onnxscript/rewriter/rules/fusion/_rotary_embedding.py
gramalingam: Yes, I suspected as much. But, on the other hand, it worked with ORT, right? So, ORT supports both?

onnxscript/rewriter/_ir_utils.py
gramalingam: Simplified the pattern for now. (More cleanup required elsewhere since we have test-cases that use 0D, and also the ORT version of these patterns.)

onnxscript/rewriter/rules/fusion/_rotary_embedding.py
justinchuby: Yeah I suppose ort didn't enforce it

justinchuby (approved):

PR #2588, opened 2025-09-30
Merge input and output shape when removing identity

wodesuck:
Similar with #2578, for this case:

```python
import torch
import torch.nn as nn


class Model(nn.Module):
    def forward(self, x):
        return x.new_zeros(x.shape)


def main():
    model = Model()
    args = torch.rand(4, 4),
    batch = torch.export.Dim("batch")
    dynamic_shapes = {"x": {0: batch}}
    torch.onnx.export(
        model,
        args,
        "model_test.onnx",
        dynamic_shapes=dynamic_shapes,
        dynamo=True,
    )


if __name__ == "__main__":
    main()
```

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -608,6 +608,7 @@ def identity(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     input = node.inputs[0]
     output = node.outputs[0]
     if input is not None and output is not None:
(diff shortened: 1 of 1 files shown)


gramalingam (approved):

gramalingam:
> Should this be added for other ops, eg. CastLike, Abs, etc. ?

All those simplifications rewrite it to `Identity(...)`, so this should automatically work there?

justinchuby:
Turns out we should relay the type as well

Issue #2596, opened 2025-10-02
aten.unbind exports to Sequence ops


yuanyao-nv:
For the following line in model definition:
```
x1, x2 = hidden_states.unflatten(-1, (-1, 2)).unbind(-1)
```
onnxscript translates the aten.unbind op to Sequence ops:

<img width="292" height="290" alt="Image" src="https://github.com/user-attachments/assets/eb3e1398-d4cc-4d7f-a82a-b9f17e23cf71" />

In the case of static shapes, I think it should be possible to translate to a regular Split op instead? Would it be possible to add a rewrite pattern for this case?

The motivating use case comes from the Wan2.2 model definition: https://github.com/huggingface/diffusers/blob/9ae5b6299d3b1a7b0378dc77c3c69baf521587d2/src/diffusers/models/transformers/transformer_wan.py#L109

justinchuby:
@gramalingam @titaiwangms

yuanyao-nv:
Thanks #2597 fixes it.

Issue #2581, opened 2025-09-29
Matching variadic lists of tensors?


iksnagreb:
If I understand correctly it is currently not possible to match a pattern with a variadic list of output tensors, such as Split, and similarly matching variadic inputs, such as Concat, seems only possible via some workarounds (e.g., #2373)? Currently it seems possible to match a Split only if you know the number of outputs in advance and it is not possible to match against the _outputs parameter itself. This would actually allow for some workaround instantiating a set of rules sweeping the _outputs argument with some loop, i.e., explicitly instantiating every variadic version of the operator, but this does not seem sustainable as the ONNX spec allows for up to 2147483647 such inputs or outputs blowing up the number of rewrite rules quite a lot...

So, is there any plan to add proper support for variadic inputs/outputs to the rewriter/pattern matcher in the near future or any pointers on how I could approach implementing this?

justinchuby:
I think you can use _allow_other_input for variations inputs. https://microsoft.github.io/onnxscript/tutorial/rewriter/rewrite_patterns.html#pattern-options

gramalingam:
Sorry for not responding earlier ... this seems to have fallen through the cracks in my email notifications!

One key requirement is that the replacement should supply the values that will replace all of the output-values specified by the pattern. If I understand the proposal, the idea is that the replacement function will look at the matched-nodes and ensure that it returns the correct number of output-values. And the rewriter itself can simply fail if this invariant is not satisfied.

So far, this seems reasonable and doable.

The examples presented above are all cases where this is the final node of the pattern. It is clear what needs to be done in the case where it is the final node. But do we need to support this in intermediate nodes in the pattern? And, if so, how?

If we need to support it in intermediate nodes, that would mean that the pattern indicates the minimum number of expected outputs (which are used in the rest of the pattern). But the extra outputs may be used elsewhere in the graph. In this case, we should fail.

But not sure if there are realistic examples for such a scenario. It would be simpler to support the case where this is allowed only for final node in pattern. Let me know if you think the more general case is worth supporting/handling.

PR #2597, opened 2025-10-02
Simplify aten_unbind when shape is static

justinchuby:
Add static shape handling to aten_unbind function.

Fix https://github.com/microsoft/onnxscript/issues/2596


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8764,6 +8764,11 @@ def aten_type_as(self: TTensor, other: TTensor2) -> TTensor2:
 def aten_unbind(self: TTensor, dim: int = 0) -> Sequence[TTensor]:
     """unbind.int(Tensor(a -> *) self, int dim=0) -> Tensor(a)[]"""
 
+    if isinstance(self.shape[dim], int):
+        # We can create a definitive split op if the input shape is static
+        outputs = op.Split(self, axis=dim, num_outputs=self.shape[dim])
+        return [op.Squeeze(out, [dim]) for out in outputs]
```
titaiwangms, line 8783: I feel like it makes more sense to rewrite it? Although this PR probably works, it's adding another dimension on torchlib (covering both static and dynamic cases). Maybe let torchlib be as dynamic as possible, and we can optimize it after.

justinchuby: How easy is it to create the optimization rules? I am fine either way

kevinch-nv: Looks like `SplitToSequence(self, axis=dim, keepdims=False)` should be generically rewritable to the subgraph if the split axis is known. This can potentially cover more cases with other ops when encountered in the future

justinchuby: @gramalingam for suggestions on the rewrite rule. Is this related to #2581 ?

xadupre: From a computation point of view, it is always better to generate the correct graph rather than producing a graph which needs to be rewritten. Matching a pattern takes time.

xadupre (approved):

justinchuby:
I updated tests and improved compatibility with torch<2.7

onnxscript/function_libs/torch_lib/ops/core.py
titaiwangms: But my understanding is that the original implementation is not incorrect. It's only the op is not preferred because of the backend implementation, which fits the category of rewritten rules. We surely can say it's more convenient to address this way (this PR), but I prefer an established/explicit rule to say when/what we should add support in torchlib, and under what condition we add rewrite rules/constat folding. Otherwise, it's just scattered around. And if we want it to be done in this way, do we consider upstream some other optimizations downstream that are optimized away because of static shapes as well?

justinchuby: I kinda think we should do both. (1) we can measure the complexity of the torchlib implementation. I feel that the complexity of the current implementation is not high for the immediate benefits it brings. If the graph can be significantly simplified because we know some shapes are static, I think we should pursue that in torchlib. When looking at micro-optimizations like this I agree that we should be more careful and decide on a case by case basis. (2) we should still have a rule that will simplify this so that our tooling can handle `SplitToSequence` generally.

PR #2598, opened 2025-10-03
Create initializers with the constant folding pass

justinchuby:
Also Create default values for the constant folding pass

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -935,6 +935,8 @@ def _record_contributing_values(original_node: ir.Node, replacement: Replacement
         assert input.name is not None
         folded_from.add(input.name)
 
+
+
     for new_output in replacement.new_outputs:
         if new_output is None:
             continue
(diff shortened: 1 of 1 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -959,9 +961,9 @@ class FoldConstantsPass(ir.passes.InPlacePass):
     def __init__(
         self,
         *,
-        shape_inference: bool,
-        input_size_limit: int,
-        output_size_limit: int,
+        shape_inference: bool = True,
+        input_size_limit: int = DEFAULT_CONSTANT_FOLD_INPUT_SIZE_LIMIT,
+        output_size_limit: int = DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT,
```
xadupre, line 966: I wonder why you need two limits. Maybe we could set up the second one to None and replace None and the first value if it is still None.

justinchuby:
Replaced by #2650

Issue #2580, opened 2025-09-29
Combine torchlib function overloads


justinchuby:
Combine all overloads so we don't need a dispatcher.

PR #2604, opened 2025-10-03
Consolidate overloads in torchlib

justinchuby:
The goal is to have no overloads and remove the PyTorch dispatcher. Right now there are still the following ops that need to be addressed:

```
Registering private function: aten::as_strided
Registering private function: aten::embedding_bag
Registering private function: aten::embedding_bag.padding_idx
Registering overload for function: aten::index.Tensor
Registering overload for function: aten::_unsafe_index.Tensor
Registering overload for function: aten::index_put
```

I did a bit of cleaning up in tests and torchlib as well.

https://github.com/microsoft/onnxscript/issues/2580

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/nn.py
+++ b/onnxscript/function_libs/torch_lib/ops/nn.py
@@ -1800,7 +1800,7 @@ def aten_scaled_dot_product_attention(
     query: TFloat,
     key: TFloat,
     value: TFloat,
-    attn_mask: Optional[TFloat] = None,
+    attn_mask: Optional[TensorType] = None,
     dropout_p: float = 0.0,
     is_causal: bool = False,
     scale: Optional[float] = None,


justinchuby:
Revert changes to roll

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -1854,6 +1850,11 @@ def aten_scaled_dot_product_attention(
             query, key, value, scale, dropout_p
         )
 
+    if attn_mask.dtype == ir.DataType.BOOL:
```
titaiwangms, line 1853: The function signature still suggests attn_mask in Optional[FLOAT]. Maybe that's the bug you run into?

justinchuby: oh let me update. But I think it's unrelated. The test failing was for onnxscript functions.

titaiwangms: Do we have a test to guard how would the function signature changes in onnxscript side affecting exporter?

justinchuby: No but this change is going to be fine because we have a single implementation (no overloads)

titaiwangms: I forgot - when there is no overload, does dispatcher dispatch it to the only option?

justinchuby: Yes

PR #2606, opened 2025-10-06
Implements aten_index_put if inputs are SymbolicTensor

xadupre:


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4383,6 +4383,11 @@ def aten_index_put(
     See implementation of `torch.onnx.symbolic_opset11.index_put
     <https://github.com/pytorch/pytorch/blob/main/torch/onnx/symbolic_opset11.py#L212>`_.
     """
+    if len(indices) > 1 and any(
+        isinstance(indice, torch.onnx._internal.exporter._tensors.SymbolicTensor)
```
justinchuby: check `not isinstance(index, int)` instead as we should not reference the private class.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4452,14 +4457,110 @@ def _make_reshape_list_broadcastable(reshape_list, values_shape):
...
+                to=INT64.dtype,
+            ),
+            True,
+        )
+
+    rk1s = [(ind is None or len(ind.shape) == 1) for ind in indices]
+    assert all(rk1s) and len(rk1s) == len(x.shape), (
+        f"input_put not implemented for indices={indices}, "
+        f"where rk1s={rk1s}, rank(x)={len(x.shape)}"
+    )
```
justinchuby: Raise NotImplementedError ?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4452,14 +4457,110 @@ def _make_reshape_list_broadcastable(reshape_list, values_shape):
...
+
+    def _mkstride(x, i):
+        if i >= len(x.shape) - 1:
+            return _1dint(1)
+        if i == len(x.shape) - 2:
+            return op.Shape(x, start=i + 1)
+        return op.ReduceProd(op.Shape(x, start=i + 1), keepdims=1)
+
+    shape = [1] * (len(x.shape) + 1)
+    mfixed = []
```
justinchuby: Prefer clear variable names and avoid abbreviations.

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: What are the possible types here? Only int and SymbolicTensor? I prefer to keep SymbolicTensor because I know exactly which type the function is supposed to handle.

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: It is temporary. I'm waiting for the unit test to hit this line or not.

xadupre: removed

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: The variable are used just after. It is short to make the code shorter and esier to read.

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Only int and SymbolicTensor for the exporter, but could be other ir.Value subclasses as well. As the type is an internal type it is not meant for public use, the current usage is not supported and is brittle. 

If preferred you may check for ir.Value instead, but really we just assume a type that has shape and dtype fields

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Any abbreviation adds cognitive load and is not preferred. All names should be thoughtfully created to be precise. Creating variables used right after is a plus but I don’t feel warranting the use of ambiguous names. 

https://google.github.io/styleguide/pyguide.html#316-naming

xadupre: I let you choose the name.

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: Feel free to suggest. I still prefer to keep this one since this is what shows up in the error message.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4229,6 +4229,15 @@ def aten_index_put(
...
     """
+    if (
+        len(indices) > 1
+        and any(
+            isinstance(indice, torch.onnx._internal.exporter._tensors.SymbolicTensor)  # pylint: disable=protected-access
+            for indice in indices
+        )
+        and len(values.shape) == 1
+    ):
+        return _aten_index_put_dynamic(self, indices, values, accumulate=accumulate)
```
gramalingam, line 4234: What happens if the condition is not true?

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4229,6 +4229,15 @@ def aten_index_put(
     See implementation of `torch.onnx.symbolic_opset11.index_put
     <https://github.com/pytorch/pytorch/blob/main/torch/onnx/symbolic_opset11.py#L212>`_.
     """
+    if (
+        len(indices) > 1
+        and any(
+            isinstance(indice, torch.onnx._internal.exporter._tensors.SymbolicTensor)  # pylint: disable=protected-access
+            for indice in indices
```
gramalingam: nit: rename `indice` to `index`

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4229,6 +4229,15 @@ def aten_index_put(
     See implementation of `torch.onnx.symbolic_opset11.index_put
     <https://github.com/pytorch/pytorch/blob/main/torch/onnx/symbolic_opset11.py#L212>`_.
     """
+    if (
+        len(indices) > 1
+        and any(
```
gramalingam, line 4228: Is there a pointer to the real use-cases that show up? I thought each index in indices is supposed to be a 1D tensor? What are the ways in which SymbolicTensors are created?

gramalingam: Is a SymbolicTensor a 1-element tensor version of a SymbolicInt?

justinchuby: SymbolicTensor  is just ir.Value with magic methods defined to support python operators.

justinchuby: Here the symbolic tensors are created from the symints.

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: done

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: We fallback to the previous behavior which is working fine except for this case.

PR #2619, opened 2025-10-09
Assign names for None dims

justinchuby:
When onnx shape inference is run on symbolic input dims, it will not handle the dim name propagation and instead create a None. As long as we rely on the current version onnx shape inference there is not better information we can get.

However, since in the optimizer we also have some custom shape propagator implemented (e.g. for Identity) that will propagate sym dims, we should encode the equivalents for those dimensions as much as possible.

This PR assigns a string to all None dims produced by onnx shape inference, so that the string names can get propagated when possible by the optimizer.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -76,7 +76,7 @@ def _is_onnx_op(node: ir.Node, op_type: str) -> bool:
 
 def _process_constant_node(node: ir.Node) -> None:
     """Sets const_value of output value of a Constant op node."""
-    if node.op_type != "Constant" or node.domain != "":
+    if not _is_onnx_op(node, "Constant"):
         return
     if len(node.attributes) != 1:
         return
@@ -1116,8 +1116,12 @@ class FoldConstantsPass(ir.passes.InPlacePass):
                 self._modified = True
                 # TODO(rama): consider merging type/other info from both values
 
+        # Propagate const_value, and manually find out shape and type
+        # to avoid potentially expensive shape inference on large tensors.
+        if _is_onnx_op(node, "Constant"):
+            _process_constant_node(node)
         # Do incremental shape inference
-        if self.shape_inference and not _is_control_flow_op(node):
+        elif self.shape_inference and not _is_control_flow_op(node):
             self._do_inference(node)
 
         if node.domain not in self._opset_imports:
@@ -1134,6 +1138,10 @@ class FoldConstantsPass(ir.passes.InPlacePass):
                 if isinstance(output, ir.Value):
                     output = [output]
                 return Replacement(output, context.nodes)
+
+        if _is_onnx_op(node, "Constant"):
+            logger.debug("Skipping constant folding for Constant node %r", node.name)
+            return None
 
         if _is_control_flow_op(node):
             logger.info(
@@ -1152,10 +1160,6 @@ class FoldConstantsPass(ir.passes.InPlacePass):
                 node.domain,
                 node.op_type,
             )
-            return None
-
-        if _is_onnx_op(node, "Constant"):
-            _process_constant_node(node)
             return None
 
         if any(x.is_graph_input() for x in node.inputs if x is not None):
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -1220,8 +1220,6 @@ def aten_binomial(
 @torch_op(
     (
         "aten::bitwise_and.Tensor",
-        "aten::bitwise_and.Scalar",
-        "aten::bitwise_and.Scalar_Tensor",
         "_operator::and_",
     ),
     trace_only=True,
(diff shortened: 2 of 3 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -1029,7 +1032,15 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
                         inferred_shape = ir.serde.deserialize_type_proto_for_shape(
                             inferred_type
                         )
-                        output.shape = _merge_shapes(output.shape, inferred_shape)
+                        merged_shape = _merge_shapes(output.shape, inferred_shape)
```
titaiwangms, line 1035: Do you mean `_merge_shapes` propagating sym_dim? It looks like more of filling in shape info if it's a known int.

justinchuby: I think so: https://github.com/microsoft/onnxscript/blob/28a8f561957c46131581bc33c8b43508f41b844f/onnxscript/optimizer/_constant_folding.py#L915-L916

titaiwangms: But in this specific case, I don't see how dim2 can provide meaningful sym_dim, as it's from onnxtype inference?

titaiwangms: I might be wrong.

justinchuby: ~I thought dim2 was the original output dim, which may be from pytorch?~

justinchuby: ~oh dim1 is. Maybe dim1 should be the preferred shape then?~

justinchuby: Yeah you must be right in the dim merging case. However:

If output.shape is None, then we can take inferred_shape; If output.shape is not None, we will keep it. We probably called _merge_shapes here to for a robust logic that is shared.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1038,6 +1049,12 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
                     e,
                 )
 
+    def _new_unknown_dim_name(self) -> str:
+        """Generate a new unique name for an unknown (None) symbolic dimension."""
+        name = f"unknown_{self._unknown_dim_count}"
+        self._unknown_dim_count += 1
```
gramalingam, line 1055: The basic idea is useful. But this doesn't guarantee that the generated dim name will be unique (even though the likelihood of a conflict is low right now). Technically, we will need to identify all symbolic names used in the model first to check for conflicts (eg., like onnx does [here](https://github.com/onnx/onnx/blob/0403b703b441fee8848d50233b2b026898949088/onnx/shape_inference/implementation.h#L41))

onnxscript/optimizer/_constant_folding.py
gramalingam: I think the idea of merging shape information coming from two sources (assuming both to be correct) is a useful one. No point in specializing any further with assumptions about which of the two sources will have what information (because it is not needed). In particular, we can't and need not assume that the existing output.shape comes from pytorch exporter .... it might have been introduced by some optimization rule.

The only special case to be handled is if two different symbolic dims are used in the two different shapes for same dim. For now, we choose the first one. (Ideally, the underlying system would record that the two symbolic dims are meant to be the same ... so that it can be used to globally to use the same one for either one of them. We don't do such things at this point.)

onnxscript/optimizer/_constant_folding.py
justinchuby: I can fix that part. Do you know have an idea why this would break SkipLayerNormFusion?

gramalingam: > I can fix that part. Do you know have an idea why this would break SkipLayerNormFusion?

Will take a look

gramalingam: Strange ... was this working earlier? I noticed failures due to lack of shape information. That makes sense, since the input models don't have shape information. Specifically, these test-cases were generated by converting onnx models into onnxscript models: but the original examples did not include the shape information from original models into the onnxscript models (as [here](https://github.com/microsoft/onnxscript/blob/75b3d42e3416a0c18cf90136d6ca582fa4abc782/onnxscript/rewriter/models/_smollm_1.py#L17)). Later on, the examples were extended so that we stored the value-infos from onnx models also into the corresponding onnxscript test-case (like [here](https://github.com/microsoft/onnxscript/blob/75b3d42e3416a0c18cf90136d6ca582fa4abc782/onnxscript/rewriter/models/_phi2lm.py#L14)) ... but looks like we still have some older test-cases without shape information.

I added a call to shape-inference pass at the beginning of the test-case ... that mostly works, but it still fails in some edge-cases where it looks like we are unable to infer that some dim is "batch-size". 

I am guessing if we update the test-cases to include shape information (as produced by exporters), it should work ... need to think best workaround

justinchuby:
Will instead be resolved by https://github.com/justinchuby/onnx-shape-inference

PR #2622, opened 2025-10-09
Fix constant in constant folding

titaiwangms:
This PR moves the processing of constant ops upward to return before node-level shape type inference (including serialization) and optimizer optimization.

Essentially, avoiding serializing constant ops (potentially large weights in LLMs) reduces the export time in optimize_ir.

Before this PR:

Screenshot 2025-10-09 141403

After this PR:

Screenshot 2025-10-09 141238

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -1103,6 +1103,27 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         # to avoid potentially expensive shape inference on large tensors.
         if _is_onnx_op(node, "Constant"):
             _process_constant_node(node)
+        # Do incremental shape inference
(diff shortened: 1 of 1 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -1099,8 +1099,12 @@ def process_node(self, node: ir.Node) -> Replacement | None:
                 self._modified = True
                 # TODO(rama): consider merging type/other info from both values
 
+        # Propagate const_value, and manually find out shape and type
+        # to avoid potentially expensive shape inference on large tensors.
+        if _is_onnx_op(node, "Constant"):
```
justinchuby, line 1104: I feel that we can standardize this with other node processers, but non blocking for now.

PR #2623, opened 2025-10-10
Merge shapes only in identity op and nodel-level shape inference

titaiwangms:
node-level shape inference covers the forward shape inference, and relying on the logic of constant-folding, we only need `_merge_shapes` in identity op to have backward shape inference.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -496,11 +496,6 @@ def cast(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     if input is None or output is None:
         return None
 
-    # TODO(rama): Parts of the following logic (implementing type/shape inference
-    # for Cast op) should be unnecessary. Generic incremental shape-inference
-    # should handle this. Only the optimization to eliminate redundant Cast ops
-    # should be needed here.
-
     input_dtype = _get_input_element_type(node, 0)
     output_dtype = _get_int_attribute(node, "to", None)
     if output_dtype is not None:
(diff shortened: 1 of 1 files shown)


onnxscript/optimizer/_constant_folding.py
```diff
@@ -904,7 +903,11 @@ def sequence_at(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     return None
 
 
-def _merge_shapes(shape1: ir.Shape | None, shape2: ir.Shape | None) -> ir.Shape | None:
+def _merge_shapes(
+    preferred_shapes: ir.Shape | None, referenced_shapes: ir.Shape | None
```
gramalingam: minor nit: what is the meaning of "referenced" here? What is being referenced?  May be `other_shape` would be more appropriate.

Also, each is a single shape. So, singular `preferred_shape` would be better.

onnxscript/optimizer/_constant_folding.py
```diff
@@ -501,8 +501,6 @@ def cast(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     # should handle this. Only the optimization to eliminate redundant Cast ops
     # should be needed here.
 
-    output.shape = _merge_shapes(output.shape, input.shape)
```
gramalingam, line 504: Should we delete the preceding comment as well?

titaiwangms: Done

onnxscript/optimizer/_constant_folding.py
titaiwangms: Done

PR #2632, opened 2025-10-14
Fusion rule for handling transformers exported models

justinchuby:
When torch.onnx exports a model from transformers with SDPA, it generates a Concat
node to concatenate past_key/value with the new key/value to produce the graph output
for kv cache. This pattern can be fused into the Attention node, which has present_key
and present_value outputs. It is necessary for ONNX Runtime because it requires the outputs
to be produced by the Attention node when past_key and past_value inputs are provided.


justinchuby:
This is still useful when enable_gqa=True

xadupre (approved):

PR #2633, opened 2025-10-15
[passes] Add model-level shape type inference

titaiwangms:
Just for people to know what's the cost for exporting to add shape type inference, and we don't necessarily need to add it.

The export time increases significantly. It would be a huge cost to export. But users would have a more comprehensive model in terms of shape and dtype. I guess this might be the reason we switched to use node-level shape type inference.

Before adding the pass:
Screenshot 2025-10-14 170459

After adding the pass:
Screenshot 2025-10-14 170437

Alternatively, we can add a flag to let users decide if they want to enable model-level shape type inference or not.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -132,7 +132,7 @@ def aten_acosh(self: TFloat) -> TFloat:
     return op.Acosh(self)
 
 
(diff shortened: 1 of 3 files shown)


onnxscript/optimizer/_optimizer.py
```diff
@@ -45,6 +45,9 @@ def optimize_ir(
     passes = [
         ir.passes.PassManager(
             [
+                common_passes.ShapeInferencePass(
```
justinchuby, line 48: Probably not strict_mode because it may fail sometimes.

PR #2635, opened 2025-10-15
[Rewriter]: introduce remove_optional_bias

AyoubMDL:
Fixes https://github.com/microsoft/onnxscript/issues/2547. I've kept the same ops as in https://github.com/microsoft/onnxscript/pull/2555.

### DIFF
--- a/onnxscript/rewriter/rules/common/_remove_optional_bias.py
+++ b/onnxscript/rewriter/rules/common/_remove_optional_bias.py
@@ -52,7 +52,7 @@ class _RemoveOptionalBias(RewriteRuleClassBase):
 class RemoveOptionalBiasFromConv(_RemoveOptionalBias):
     """Remove zero bias from Conv operation."""
 
-    op_type: ClassVar = "Conv"
+    op_type: ClassVar[str] = "Conv"
 
     def pattern(self, op: ir.tape.Tape, x: ir.Value, w: ir.Value, b: ir.Value) -> ir.Value:
         return op.Conv(x, w, b, _outputs=["out"])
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/rules/common/_remove_optional_bias_test.py
```diff
@@ -0,0 +1,236 @@
...
+            float w_scale = {1.5}, uint8 w_zero_point = {123},
+            float y_scale = {1.5}, uint8 y_zero_point = {123}>
+            {
+                Y = QLinearConv(X, x_scale, x_zero_point, W, w_scale, w_zero_point, y_scale, y_zero_point, B)
+            }
+        """,
+            initializers=[w, b],
+        )
+        onnx_checker.CheckerPass(True)(model)
+        return model
```
AyoubMDL, line 222: This gives the following warnings:
```python
Input 'W' of node '' (::QLinearConv:) cannot be found in any scope. The model is invalid but we will still create a new input for the node (current depth: 1)
Input 'B' of node '' (::QLinearConv:) cannot be found in any scope. The model is invalid but we will still create a new input for the node (current depth: 1)
```
This happens because I didn’t specify them in the ONNX text. I’m not sure how to define the initializers properly. I assumed the initializers parameter would handle that.

justinchuby: The initializers parameter will handle that. I just need to fix the warning for it. Feel free to ignore, thanks

onnxscript/rewriter/rules/common/_remove_optional_bias.py
```diff
@@ -0,0 +1,123 @@
...
+        if not np.equal(bias_array, 0.0).all():
+            return check_result.fail("Bias is not all zeros.")
+
+        return check_result
+
+
+class RemoveOptionalBiasFromConv(_RemoveOptionalBias):
+    """Remove zero bias from Conv operation."""
+
+    op_type: ClassVar = "Conv"
```
gramalingam: nit: may be `ClassVar[str]` ?

PR #2641, opened 2025-10-17
Improves implementation of aten_index_put

xadupre:
From #2606.


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4223,6 +4223,30 @@ def aten_index_put(
...
+        len(indices) > 1
+        and any(
+            isinstance(index, torch.onnx._internal.exporter._tensors.SymbolicTensor)  # pylint: disable=protected-access
+            for index in indices
+        )
+        and len(values.shape) == 1
+    ):
+        return _aten_index_put_dynamic(self, indices, values, accumulate=accumulate)
+
+    n_none = [i for i, ind in enumerate(indices) if ind is not None]
```
gramalingam: I guess `n_none` is short for `not_none` ... may be better to call it `not_none`

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4223,6 +4223,30 @@ def aten_index_put(
     See implementation of `torch.onnx.symbolic_opset11.index_put
     <https://github.com/pytorch/pytorch/blob/main/torch/onnx/symbolic_opset11.py#L212>`_.
     """
+    if (
+        len(indices) > 1
+        and any(
+            isinstance(index, torch.onnx._internal.exporter._tensors.SymbolicTensor)  # pylint: disable=protected-access
+            for index in indices
+        )
+        and len(values.shape) == 1
```
gramalingam, line 4242: What is this condition for? I am just trying to understand the assumptions/conditions for this special case.

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: done

onnxscript/function_libs/torch_lib/ops/core.py
xadupre: ONNX does not have similar operator than index_put. So the goal was to convert different cases in different ways. Finding a generic way to convert all the cases may end up with inefficient conversions. So I chose to filter out some cases where I know a simple conversion. At the end of the function, you still have an implementation working for everything left. About this condition, I don't know when SymbolicTensor appears, I just know it appears in some models (Qwen) and in that case, the conversion I implemented works. If the generic case is difficult to implement, let's go through the list of all cases we face. That's the logic I followed.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4302,14 +4339,131 @@ def _make_reshape_list_broadcastable(reshape_list, values_shape):
...
+        )
+
+    shape_x = op.Shape(x)
+    exped = []
+    fixed = []
+    reshape_value_shape2 = []
+    expand_value_shape = []
+    for i, ind in enumerate(indices):
+        if isinstance(ind, torch.onnx._internal.exporter._tensors.SymbolicTensor):  # pylint: disable=protected-access
+            ind.dtype = ir.DataType.INT64
```
gramalingam, line 4409: Why do we need the above line? Just wondering ... shouldn't it already have dtype set?

xadupre: Maybe it is not useful anymore but when I did this PR, it was needed.

PR #2643, opened 2025-10-18
Add a verbose mode to torch api for external data save

justinchuby:
Show progress bar with tqdm when verbose is True.

It will be enabled in PyTorch 2.10

image

### DIFF
--- a/onnxscript/_framework_apis/torch_2_5.py
+++ b/onnxscript/_framework_apis/torch_2_5.py
@@ -85,7 +85,7 @@ def save_model_with_external_data(
     use_tqdm = verbose and importlib.util.find_spec("tqdm") is not None
 
     if use_tqdm:
-        import tqdm
+        import tqdm  # pylint: disable=import-outside-toplevel
 
         with tqdm.tqdm() as pbar:
             total_set = False


onnxscript/_framework_apis/torch_2_5.py
```diff
@@ -63,7 +64,9 @@ def check_model(model: ir.Model) -> None:
     del model  # Unused yet
 
 
-def save_model_with_external_data(model: ir.Model, model_path: str | os.PathLike) -> None:
+def save_model_with_external_data(
+    model: ir.Model, model_path: str | os.PathLike, verbose: bool = False
+) -> None:
```
titaiwangms, line 69: Should we duplicate the api and add verbose in a later version of framework_apis (eg 2_10) if we only want this after 2.10?

justinchuby: I think this is easier and is done in a bc compatible way. But if you think duplicating makes it more obvious I can do that too

justinchuby: Will merge first. LMK if you would like to see a follow up, thanks!

PR #2644, opened 2025-10-19
Add support for traced if statements in onnxscript script

gramalingam:
Extend the onnxscript converter to handle conditionals that are evaluated at script-time (that is, in the style of trace-mode). This makes it easier to define parametric scripts that can be used to generate variations of a pattern: for example, like the many variations of the [SDPA pattern test cases](https://github.com/microsoft/onnxscript/blob/8a94ad646440f462dd9ae1de6b303fe5f7b7f564/onnxscript/rewriter/ort_fusions/sdpa_test.py#L40).

This supports just a very basic version, where the if-condition is an outer-scope variable, like below:
```py
   if outer_scope_variable:
      ...
   else:
      ...
```
For such cases, the script will just include the then or else branch as appropriate, without generating an if-node.

Also: introduce an analyzer class to encapsulate analysis information, and avoid updates to AST node.

TODO: some simple extension may be useful (perhaps allow any expression in the if-condition that does not contain local variables).

### DIFF
--- a/onnxscript/_internal/analysis.py
+++ b/onnxscript/_internal/analysis.py
@@ -98,7 +98,7 @@ class AstAnalyzer:
         Returns:
             The constant boolean value of the if-statement condition, or None if not constant.
         """
-        return self._constant_if_condition.get(if_stmt, None)  # type: ignore[return-value]
+        return self._constant_if_condition.get(if_stmt, None)
 
     def assigned_vars(self, stmt: ast.stmt | list[ast.stmt]) -> Set[str]:
         """Return the set of all variables that may be assigned to in an execution of input stmt


onnxscript/_internal/analysis_test.py
```diff
@@ -248,5 +253,42 @@ def f(x):
         self.assert_assigned_vars(f, {"x", "y"})
 
 
+class ConstantIfAnalysisTest(unittest.TestCase):
+    def test_constant_ifs(self):
+        cond1 = True
+        cond2 = False
+
+        def f(x):
+            if cond1:
```
gramalingam, line 262: One design question for consideration here: when should we consider "tracing" the if-then-else (emitting only the then/else branch instead of an onnx If op node). Obviously, we need the condition to be evaluable at script time. The choice is whether we should require users to explicitly indicate this or whether we should do this implicitly.

In the explicit mode, we would require users to indicate via some syntax like `if onnxscript.eval(cond1):` to indicate that `cond1` should be evaluated at script time for a tracing mode translation. In the implicit mode, we just see if the condition can be evaluated or not.

While the difference is not as significant here (implicit mode seems fine), it could be more important when we get to loops. There the question is whether the loop should be unrolled or not ... but even if a loop can be unrolled, a user might not want the loop to be unrolled, and so may prefer to exercise control over the behavior. In a loop, we might support a syntax like `for var in onnxscript.unroll(iterator):`.

gramalingam: I guess we could make it a user-controllable option to the converter with a particular default value. But it would still be useful to find out what users prefer as the default setting.

titaiwangms: Just for my education, is this design question for future reference? Or the current PR is implementing it in an implicit way? (Looks like right now we only consider the condition being outer-scope variable.)

I remember in torchscript that users use torch.jit.script() to specify the scripted blocks. But in general, tracing is preferred to scripting. What is the case that users might not want the loop being unrolled even it's fixed?

titaiwangms (approved):

justinchuby (approved):

onnxscript/_internal/analysis_test.py
gramalingam: The current PR is implementing it in an implicit way. But if we want to something a bit differently, it would be good to do it sooner than later, before users get used to one way of doing it.

As for loops: you can ask Kunal or other users. Originally, Kunal wanted unrolled loops, but more recently he was also asking for a loop (without unrolling it) in some context.

gramalingam: Yes, in general, looks like people prefer tracing .. and we will have a tracing mode and a script mode (with user selecting it). This question is about preferences withing script-mode: should the switch to trace-mode be explicitly required?

PR #2646, opened 2025-10-20
Unsqueeze unbatched input of avg_pool

wodesuck:
Onnx's `AveragePool` require input shape as `N,C,H,W`, but torch accept both `N,C,H,W` and `C,H,W`. Unsqueeze if input is unbatched, just like what `max_pool` does.

### DIFF
--- a/tests/function_libs/torch_lib/e2e_ops_tests.py
+++ b/tests/function_libs/torch_lib/e2e_ops_tests.py
@@ -242,12 +242,12 @@ class TorchLibe2eTest(unittest.TestCase):
         class Model(torch.nn.Module):
             def forward(self, x2d, x3d, x4d, x5d):
                 return (
-                    torch.nn.functional.avg_pool1d(x2d, 2),
-                    torch.nn.functional.avg_pool1d(x3d, 2),
-                    torch.nn.functional.avg_pool2d(x3d, 2),
-                    torch.nn.functional.avg_pool2d(x4d, 2),
-                    torch.nn.functional.avg_pool3d(x4d, 2),
-                    torch.nn.functional.avg_pool3d(x5d, 2),
+                    torch.nn.functional.avg_pool1d(x2d, 2),  # pylint: disable=not-callable
+                    torch.nn.functional.avg_pool1d(x3d, 2),  # pylint: disable=not-callable
+                    torch.nn.functional.avg_pool2d(x3d, 2),  # pylint: disable=not-callable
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -114,6 +114,34 @@ def _adjust_attributes_of_avg_pool(
...
+def _aten_avg_pool_onnx(
+    self: TFloat,
+    kernel_shape: Sequence[int],
+    strides: Sequence[int],
+    pads: Sequence[int],
+    ceil_mode: bool,
+    count_include_pad: bool,
+    unbatched_rank: int,
+) -> TFloat:
+    self_rank_is_unbatched_rank = len(self.shape) == unbatched_rank
```
titaiwangms: unbatched_rank seems to be len(kernel_shape)+1? Only check it with equal or not is a bit ambiguous.

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -134,16 +162,7 @@ def aten_avg_pool1d(
...
-        self,
-        ceil_mode=ceil_mode,
-        count_include_pad=count_include_pad,
-        kernel_shape=kernel_shape,
-        pads=pads,
-        strides=strides,
-    )
-
-    return result
+    return _aten_avg_pool_onnx(self, kernel_shape, strides, pads, ceil_mode, count_include_pad, 2)
```
titaiwangms: I wonder expand_size+1 would be more clear than hard coded 2.

onnxscript/function_libs/torch_lib/ops/nn.py
titaiwangms: Or add assertion would do too

titaiwangms (commented):
Thank you. Minor comments and please check this page for lint: https://github.com/microsoft/onnxscript#coding-style

onnxscript/function_libs/torch_lib/ops/nn.py
wodesuck: I use `len(kernel_shape) + 1` instead of `unbatched_rank` here. `_aten_max_pool_onnx` could do the same thing.

wodesuck:
@titaiwangms Pylint says "torch.nn.functional.avg_pool1d is not callable", that's not true. I have run lintrunner locally without wrong, don't known why it still blame.

PR #2650, opened 2025-10-20
Create initializers not constant nodes in constant folding pass

titaiwangms:
Partially From #2598

This provides a better optimized graph after constant folding in terms of the number of nodes, which is better for debugging.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 dependencies = [
   "ml_dtypes",
   "numpy",
-  "onnx_ir>=0.1.10,<2",  # Expect onnx_ir to have a breaking change in 2.0. If not, extend this range.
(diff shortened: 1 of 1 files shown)


titaiwangms:
Need https://github.com/onnx/ir-py/pull/227

noxfile.py
```diff
@@ -41,7 +41,7 @@
     "packaging",
     "protobuf",
 )
-ONNX_IR = "onnx_ir==0.1.10"
+ONNX_IR = "onnx_ir==0.1.12"
```
justinchuby, line 44: Please update https://github.com/microsoft/onnxscript/blob/45b51898dc605c2ce3e800e23aeca05e25cacb4c/pyproject.toml#L30 as well

justinchuby (approved):
Thanks!

titaiwangms:
> Just making sure: is the replaced initializer immediately freed (if there is no users)? We need to do that to avoid memory usage jumps.

I will check.

PR #2653, opened 2025-10-23
Support math trunc

titaiwangms:
Fix https://github.com/pytorch/pytorch/issues/166110

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -8536,6 +8536,8 @@ def aten_trunc(self: TFloat) -> TFloat:
 @torch_op("math::trunc", trace_only=True)
 def python_math_trunc(self: TFloat) -> TInt:
     """trunc(Tensor self) -> Tensor"""
+    # NOTE: This is used in SymInt/SymBool/SymFloat context, so
+    # we don't expect overflow to happen here.
     return op.Cast(self, to=INT64.dtype)
 
 


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8533,6 +8533,14 @@ def aten_trunc(self: TFloat) -> TFloat:
     return op.Floor(op.Abs(self)) * op.Sign(self)
 
 
+@torch_op("math::trunc", trace_only=True)
```
justinchuby, line 8536: Possible to use a single cast to do this?

titaiwangms: Sorry I don't understand.

justinchuby: A single cast to INT64 should do the truncation

gramalingam: Are you saying that the discussion referenced below (onnx issue 4588) is not relevant now?

Surprisingly, the onnx Cast op documentation does not seem to say what it does for float to int conversion.

justinchuby: Given that we don't expect overflows for symfloats, I think this is ok. onnx#4588 is more rigorous and the impl is used for the aten op.

Cast op: there was a spec clarification issue I remember

titaiwangms: DOne

Issue #2561, opened 2025-09-13
Do not use -inf in boolean masks?


justinchuby:
Should we set https://github.com/microsoft/onnxscript/blob/a70ee8d0905f563c840bbd5338595e9ac6b1b5b4/onnxscript/function_libs/torch_lib/ops/nn.py#L2141

to use float.lowest()?

Ref: https://github.com/microsoft/onnxruntime/pull/25966#pullrequestreview-3219259522

PR #2654, opened 2025-10-26
Fix attention mask to use float_lowest instead of -inf and add NaN-safe softmax handling

Aniketsy:
#2561 

- Use lowest representable float value instead of -inf for attention masks.
- Add NaN-safe handling and a unit test for softmax with all masked positions.

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thankyou !


tests/common/testutils.py
```diff
@@ -101,3 +102,9 @@ def test_onnxruntime_rewrite(
                     f"Failed for model {model_name} and output {i} with rtol={rtol} and atol={atol}\n{e}"
                 )
                 raise
+
+def test_softmax_with_all_inf_mask():
```
justinchuby: Please remove the test. It does not belong here

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2048,6 +2049,9 @@ def _aten_scaled_dot_product_attention_no_mask_onnx(
         attn_weight, _ = op.Dropout(attn_weight, dropout_p)
     return op.MatMul(attn_weight, value)
 
+def float_lowest(dtype):
```
justinchuby: ~This lgtm. Please make the function private by prefixing with `_`~ See comment below

onnxscript/function_libs/torch_lib/ops/nn.py
```diff
@@ -2078,7 +2082,7 @@ def _aten_scaled_dot_product_attention_bool_mask_onnx(
     key_transposed_scaled = op.Mul(key_transposed, op.Sqrt(scale))
     # Turn the Boolean mask to float: attn_mask.masked_fill(not attn_mask, -float('inf'))
     zero = op.Constant(value=ir.tensor(0.0, dtype=query.dtype))
-    neg_inf = op.Constant(value=ir.tensor(-float("inf"), dtype=query.dtype))
+    neg_inf = op.Constant(value=ir.tensor(float_lowest(query.dtype)), dtype=query.dtype)
```
justinchuby: Actually, you can use query.dtype.min directly because it is an implemented method in ir.DataType: https://onnx.ai/ir-py/api/generated/onnx_ir.DataType.html#onnx_ir.DataType.min

justinchuby:
If this PR is facilitated by an AI, please disclose its usage.

Aniketsy:
I went through the questions you mentioned and yes, i used AI assistance to help add the unit test.

Issue #2616, opened 2025-10-09
Adding support for integrating arbitrary ModelProto in onnxscript


chapman73:
Hey team, 

I'm looking to submit my attempt at resolving this issue: https://github.com/microsoft/onnxscript/issues/1882 

I'm opening this as an _issue_ first, because I think the solution may be quite large/with various design decisions which (IMO) non-trivial. Though also happy to start with a large PR (so theres a bird's eye view) and then determine which parts are appropriate/not appropriate. 

---

# What (more or less works) -- Updating Docs? Adding a util function?

The docs more or less address how this is achieved here: https://microsoft.github.io/onnxscript/auto_examples/06_plot_model_local_funs.html

example code that works:

```py
import numpy as np
import onnxruntime
import pandas as pd
from onnxscript import FLOAT, script
from onnxscript import opset17 as op
from onnxscript.values import Opset
from skl2onnx import to_onnx
from sklearn.linear_model import LogisticRegression
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import StandardScaler


pandas_dataset = pd.DataFrame(
    {
        "age": np.random.randint(18, 60, 100),
        "height": np.random.rand(100) * 100 + 100,
        "weight": np.random.rand(100) * 100 + 40,
    }
)

multi_classifier_target = pd.Series(np.random.choice([0, 1, 2, 3], 100))
# Create the pipeline
pipe = Pipeline([("scaler", StandardScaler()), ("linear", LogisticRegression())])

# Fit the pipeline to the data
pipe.fit(pandas_dataset, multi_classifier_target)

# local = CustomOpset("local", 1)  # see below in the next section
local = Opset("local", 1)

X = pandas_dataset.copy()
scaler_step = pipe.steps[0][1]
linear_step = pipe.steps[1][1]

scaler_model = to_onnx(scaler_step, X.values.astype(np.float32), target_opset={"ai.onnx.ml": 2, "": 17})
linear_model = to_onnx(
    linear_step, scaler_step.transform(X.values.astype(np.float32)), options={"zipmap": False}, target_opset={"ai.onnx.ml": 2, "": 17}
)


# custom func that convers model proto to function proto -- convert_model_proto_to_function_proto
scaler_function_proto = convert_model_proto_to_function_proto(scaler_model, "local", "scaler")
linear_function_proto = convert_model_proto_to_function_proto(linear_model, "local", "linear")


@script(local, default_opset=op)
[...]

PR #2655, opened 2025-10-27
Add utility function for converting model protos to function proto

chapman73:
This addresses the first part of: https://github.com/microsoft/onnxscript/issues/2616

It tries to convert arbitrary model protos to function protos. Since function protos don't have a concept of initializers, they are made to be constants. This approach should address a large number of model protos (including 'models' from skl2onnx)

Happy for any particular feedback. 

Note that this does not address the second part of https://github.com/microsoft/onnxscript/issues/2616 , running the script with the operator embedded will not work.

### DIFF
--- a/onnxscript/utils/model_proto_to_function_proto.py
+++ b/onnxscript/utils/model_proto_to_function_proto.py
@@ -48,7 +48,15 @@ def convert_model_proto_to_function_proto(
     """
     model = _initializers_to_constants(model)
     model_ir = onnx_ir.serde.deserialize_model(model)
+    doc_string = model.doc_string if model.doc_string else ""
+    graph = model_ir.graph
+
     function_ir = onnx_ir.Function(
-        domain=domain, name=name, graph=model_ir.graph, attributes={}
+        domain=domain, name=name, graph=graph, attributes={}
     )
+
+    # set metadata
+    function_ir.doc_string(doc_string)
+    function_ir.metadata_props = graph.metadata_props  # I believe the docs suggest directly modifying the attr?
+    # function_ir.value_infos(graph.value_infos)  # theres no setter defined?
     return onnx_ir.to_proto(function_ir)


onnxscript/utils/model_proto_to_function_proto.py
```diff
@@ -0,0 +1,54 @@
...
+) -> onnx.FunctionProto:
+    """Converts an arbitrary ModelProto to a FunctionProto.
+
+    Since function protos don't support initializers (or rather it does not make sense in the context of a function)
+    we need to convert them to constants first.
+    """
+    model = _initializers_to_constants(model)
+    model_ir = onnx_ir.serde.deserialize_model(model)
+    function_ir = onnx_ir.Function(
+        domain=domain, name=name, graph=model_ir.graph, attributes={}
```
gramalingam: Copy opset_import over?

gramalingam: Also: doc_string, metadata_props?

gramalingam: And value_info over from the graph ... we would also need to copy the value_info for graph inputs/outputs over

chapman73: Updated here: https://github.com/microsoft/onnxscript/pull/2655/commits/aadc8815c8028a8524da099ea4fdd5a3591764cf

Though not sure about how to set the value_info? or if metadata_props is done correctly. If I understand correctly does a setter need to exist in the upstream onnx_ir part?

onnxscript/utils/model_proto_to_function_proto.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+import onnx
+import onnx_ir
+from onnx import helper
+
+
+def _initializers_to_constants(model: onnx.ModelProto) -> onnx.ModelProto:
```
justinchuby, line 8: Sorry for the delay: I would prefer that you create the initializers to constants logic in onnx ir

onnxscript/utils/model_proto_to_function_proto.py
justinchuby: @gramalingam all of the information is carried over by assigning the graph: model_ir.graph.

PR #2659, opened 2025-10-28
[version converter] Fix DFT opset 20

titaiwangms:
Fixes https://github.com/pytorch/pytorch/issues/148687

Axis is actually the third input of DFT.


justinchuby (approved):
Thank you - I completely forgot about this and thought we needed an implementation in pytorch.

justinchuby (commented):
Possible to create a e2e model to guard this?

Issue #2660, opened 2025-10-29
[torchlib] Filter out `(0,)` shaped tensors in aten_concat


justinchuby:
According to PyTorch doc: `All tensors must either have the same shape (except in the concatenating dimension) or be a 1-D empty tensor with size (0,).` We can simply filter out the values with size (0,) when converting.

cc @titaiwangms

PR #2661, opened 2025-10-29
[torchlib] Fix concat when input tensor has shape `(0,)`

justinchuby:
Filter out size-0 tensors. When there is only one input, create an identity op instead of a concat op.

Fix https://github.com/microsoft/onnxscript/issues/2660

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -1540,7 +1540,7 @@ def aten_cat(tensors: Sequence[TTensor], dim: int = 0) -> TTensor:
         if tensor is None:
             continue
         # Remove empty tensors
-        if tensor.shape is not None and 0 in tensor.shape:
+        if tensor.shape == (0,):
             continue
         filtered_tensors.append(tensor)
     assert filtered_tensors, "aten::cat received all None or empty tensors"


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1534,8 +1534,18 @@ def aten_cat_complex(tensors: Sequence[TTensor], dim: int = 0) -> TTensor:
...
 
-    # Remove None tensors
-    tensors = [tensor for tensor in tensors if tensor is not None]
+    filtered_tensors = []
+    for tensor in tensors:
+        # Remove None tensors
+        if tensor is None:
+            continue
+        # Remove empty tensors
+        if tensor.shape is not None and 0 in tensor.shape:
```
gramalingam: Should this be restricted to the 1D-case with shape (0,) as you showed in the documentation? Just wondering ... consider concatenating (5, 0) with (10, 0) along axis=0 to get (15, 0) tensor ... I think this might be legal and correct in both pytorch and onnx.

justinchuby: Thanks - good catch

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1534,8 +1534,18 @@ def aten_cat_complex(tensors: Sequence[TTensor], dim: int = 0) -> TTensor:
...
     """cat(Tensor[] tensors, int dim=0) -> Tensor"""
 
-    # Remove None tensors
-    tensors = [tensor for tensor in tensors if tensor is not None]
+    filtered_tensors = []
+    for tensor in tensors:
+        # Remove None tensors
+        if tensor is None:
+            continue
+        # Remove empty tensors
```
titaiwangms, line 1542: I wonder if we have a higher-level mechanism to filter empty tensors (when it's single tensor)? I can see this might be exception as it's in sequence.

justinchuby: Can you say more on higher-level mechanism? I wasn't sure what it was referring to

titaiwangms: I forget what does exporter do when it sees an empty tensor?

justinchuby: It will pass it over. Won't do anything

titaiwangms: And why is it a problem when it's in concat?

PR #2665, opened 2025-10-29
Migrate ir_builder to use onnx_ir (take 2)

justinchuby:
Continuation of https://github.com/microsoft/onnxscript/pull/2457

### DIFF
--- a/onnxscript/_converter.py
+++ b/onnxscript/_converter.py
@@ -5,13 +5,14 @@
 from __future__ import annotations
 
 import ast
+from collections import defaultdict
 import dataclasses
 import logging
-from collections import defaultdict
 from typing import (
     Any,
     Dict,
     List,
+    Mapping,
     NoReturn,
     Optional,
     Sequence,
@@ -20,11 +21,13 @@ from typing import (
 )
 
 import onnx_ir as ir
+from onnxscript.ir import _schemas
 
 import onnxscript
 from onnxscript import onnx_types, sourceinfo, values
 from onnxscript import type_annotation as ta
 from onnxscript._internal import analysis, ast_utils, autocast
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


onnxscript/_converter.py
```diff
@@ -80,167 +82,212 @@ def ignore(cond, msg):
...
+
+def set_sourceinfo(value: ir.Value, info: sourceinfo.SourceInfo):
+    """Set the source information for an ONNX value."""
+    value.meta[_SOURCEINFO_FIELD] = info
+
+
+def is_base_type_bool(attr: ir.Attr) -> bool:
+    """Check if the attribute is a boolean type."""
+    # FIXME: Add meta to attributes
+    attr.meta[_SOURCEINFO_FIELD]
```
gramalingam, line 140: ? Why is this related to SOURCE_INFO?

onnxscript/_converter.py
```diff
@@ -80,167 +82,212 @@ def ignore(cond, msg):
 }
 
 
-class Variable:
-    """Represents an ONNX variable.
+_CASTABLE_FIELD = "pkg.onnxscript.converter.castable"
```
gramalingam, line 85: nit: why not just call these _CASTABLE and _SOURCEINFO ?

onnxscript/_converter.py
```diff
@@ -80,167 +82,212 @@ def ignore(cond, msg):
...
+    value.meta[_SOURCEINFO_FIELD] = info
+
+
+def is_base_type_bool(attr: ir.Attr) -> bool:
+    """Check if the attribute is a boolean type."""
+    # FIXME: Add meta to attributes
+    attr.meta[_SOURCEINFO_FIELD]
+
+@dataclasses.dataclass
+class ASTMeta:
```
gramalingam, line 143: This is no longer needed

onnxscript/_converter.py
```diff
@@ -80,167 +82,212 @@ def ignore(cond, msg):
...
+    """
+
+    # For liveness analysis,
+    live_out: set[str] | None = None
+    live_in: set[str] | None = None
+
+
+class _ValueEnvironment:
+    def __init__(self, converter: Converter):
+        self._py_var_name_to_ir_values: dict[str, ir.Value] = {}
```
gramalingam, line 156: Not sure I see the value in splitting this map into three maps?

onnxscript/_converter.py
```diff
@@ -80,167 +82,212 @@ def ignore(cond, msg):
...
+                "default_opset must be specified in script for functions "
+                "that do not contain any use of an ONNX opset."
+            )
 
-        # States initialized by `_init_function_translation`
-        self._outer: List[irbuilder.IRFunction] = []
-        self._current_fn: irbuilder.IRFunction = None
+        # TODO(justinchuby): Update ir version to be user defined
+        # TODO(justinchuby): Maybe just store a list of functions
+        self._model = ir.Model(ir.Graph((), (), nodes=()), ir_version=10)
```
gramalingam, line 263: nit (can be fixed later too): not sure we need to create an ir.Model ahead of time ... eg., if this is just intended to be used as a function/FunctionProto

onnxscript/_converter.py
gramalingam: Also: I wonder about the use "pkg" ... while a conflict seems unlikely, it would be good to set and follow some conventions for the metadata key strings to avoid conflicts between different tools metadata. The ONNX spec says use reverse domain name, just like for custom domain names for opsets: https://github.com/onnx/onnx/blob/main/docs/IR.md#other-metadata 
Of course, too bad we dont have onnxscript.ai domain, but we could in ONNX standard reserve names like that.

justinchuby: I think `pkg.` is a legitimate reverse domain (style) name?

gramalingam: It is. I just mean that we don't own that domain name.

PR #2668, opened 2025-10-30
Clear initializers in constant folding pass

justinchuby:
Clear unused initializers on the fly to prevent memory usage jump due to intermediate folded tensors.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -1256,6 +1256,11 @@ class FoldConstantsPass(ir.passes.InPlacePass):
 
         # Record the names of the values that has contributed to the replacement
         _record_contributing_values(node, replacement)
+
+        # Obtain the list of non-None inputs to the node before it is cleared by
+        # replace_nodes_and_values to check for unused initializers later.
+        node_inputs = [v for v in node.inputs if v is not None]
+
         ir.convenience.replace_nodes_and_values(
             root, node, [node], replacement.new_nodes, node.outputs, replacement.new_outputs
         )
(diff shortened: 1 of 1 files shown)


VERSION
```diff
@@ -1 +1 @@
-0.5.5
```
titaiwangms, line 1: Is this intended

justinchuby: Yes, because I didn't want to redo the git tag. We will just bump the version and keep 0.5.5 unreleased

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1256,10 +1256,20 @@ def replace_node(
...
+        # replace_nodes_and_values to check for unused initializers later.
+        node_inputs = [v for v in node.inputs if v is not None]
+
         ir.convenience.replace_nodes_and_values(
             root, node, [node], replacement.new_nodes, node.outputs, replacement.new_outputs
         )
 
+        if isinstance(root, ir.Graph):
+            # The old node should now be detached from the graph
+            assert node.graph is None
```
gramalingam, line 1270: Do we need this assertion? This might be true in the current implementation ... but just wondering if we might ever want to replace [oldnode] by [newnode1, oldnode, newnode2], for example ... not sure of replace_nodes_and_values supports such a scenario ... but the logic here seems to be robust even in that case. But the assert would fail in that case. Anyway ... just a minor suggestion.

Also: in principle, this logic could go into `replace_nodes_and_values`, can't it?

gramalingam (approved):

onnxscript/optimizer/_constant_folding.py
justinchuby: > this logic could go into replace_nodes_and_values, can't it?

Good point - I agree. But also replace_nodes_and_values is more general. So checking value users may be more expensive there.

> replace [oldnode] by [newnode1, oldnode, newnode2]

Not sure? replace_nodes_and_values will disconnect the old nodes (which breaks all the value connections) so I don't think it will work?

Issue #2546, opened 2025-09-05
[torchlib] Implement RNN operators


justinchuby:
- lstm 

```yml
- func: lstm.input(Tensor input, Tensor[] hx, Tensor[] params, bool has_biases, int num_layers, float dropout, bool train, bool bidirectional, bool batch_first) -> (Tensor, Tensor, Tensor)
  tags: nondeterministic_seeded
```

- gru

```yml
- func: gru.input(Tensor input, Tensor hx, Tensor[] params, bool has_biases, int num_layers, float dropout, bool train, bool bidirectional, bool batch_first) -> (Tensor, Tensor)
  tags: nondeterministic_seeded
```

Test models 

- https://github.com/pytorch/pytorch/issues/120626
- https://github.com/pytorch/pytorch/issues/123089

utsab345:
I can work on this , Please assigned it to me.

PR #2674, opened 2025-11-02
feat: implement LSTM and GRU operators for torchlib

ombrdr47:
Implement aten_lstm and aten_gru operators to enable torch.onnx.export for PyTorch LSTM and GRU layers. This addresses issue #2546.

Key features:
- Full support for multi-layer RNNs (num_layers > 1)
- Bidirectional support (forward and backward directions)
- Handles both biased and non-biased configurations
- batch_first parameter support with automatic transposition
- Dropout support between layers (nondeterministic seeded)
- Proper gate reordering for ONNX compatibility:
  * LSTM: PyTorch [i,f,g,o] -> ONNX [i,o,f,g]
  * GRU: PyTorch [r,z,n] -> ONNX [z,r,n]

Implementation details:
- Uses ONNX LSTM/GRU operators with proper parameter formatting
- Handles weight matrix transposition and reshaping
- Correctly concatenates biases using op.Concat
- Processes each layer independently with proper state management
- Returns outputs in PyTorch-compatible format

Closes: #2546

Also resolves:
- pytorch/pytorch#120626 (GRU export)
- pytorch/pytorch#123089 (LSTM export)  
- pytorch/pytorch#164834 (GRU dynamo export)

### DIFF
--- a/onnxscript/rewriter/__init__.py
+++ b/onnxscript/rewriter/__init__.py
@@ -16,6 +16,7 @@ __all__ = [
     "RewriterContext",
     "MatchingTracer",
     "MatchStatus",
+    "RULE_NAME_TAG",
 ]
 
 import onnx
@@ -25,6 +26,7 @@ from onnxscript import ir
 from onnxscript.rewriter import pattern
 from onnxscript.rewriter._basics import MatchContext, MatchingTracer, MatchResult, MatchStatus
 from onnxscript.rewriter._rewrite_rule import (
+    RULE_NAME_TAG,
     RewriterContext,
     RewriteRule,
     RewriteRuleClassBase,
--- a/onnxscript/rewriter/_rewrite_rule.py
+++ b/onnxscript/rewriter/_rewrite_rule.py
@@ -25,6 +25,11 @@ T = TypeVar("T")
 
 RewriterContext = _tape.Builder
 
+# TODO(rama): Standardize metadata property keys. May be worth standardizing at ONNX level for
+# source/producer metadata.
+
+RULE_NAME_TAG = "pkg.onnxscript.rewriter.rule_name"
+
 
 @dataclasses.dataclass
 class ReplacementSubgraph:
@@ -718,6 +723,13 @@ class RewriteRuleSet:
                     print("++++Replacement Nodes++++")
                     _ir_utils.display_nodes(delta.new_nodes)
                     print("++++End Replacement Nodes++++")
+
+                # Capture rewrite rule name as metadata.
+                # TODO(rama): This is just a basic version. We may wish to compose "source" metadata
+                # from multiple rules in future.
+                if rule.name:
+                    for n in delta.new_nodes:
+                        n.metadata_props[RULE_NAME_TAG] = rule.name
 
                 convenience.replace_nodes_and_values(
                     graph_or_function,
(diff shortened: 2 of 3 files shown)


ombrdr47:

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4991,6 +4991,356 @@ def aten_lstm_mps_backward(
     raise NotImplementedError()
 
 
+@torch_op("aten::lstm", trace_only=True)
```
```suggestion
@torch_op("aten::lstm.input", trace_only=True)
```

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4991,6 +4991,356 @@ def aten_lstm_mps_backward(
...
+    
+    # Handle batch_first for output
+    if batch_first:
+        # Convert from [seq, batch, features] to [batch, seq, features]
+        current_input = op.Transpose(current_input, perm=[1, 0, 2])
+    
+    return current_input, final_h, final_c
+
+
+@torch_op("aten::gru", trace_only=True)
```
```suggestion
@torch_op("aten::gru.input", trace_only=True)
```

justinchuby: Could you put this function with the `aten_g...` functions so that they are sorted alphabetically?

ombrdr47: Done! Moved `aten_gru` to appear after `aten_ger`.

ombrdr47:
> Feel free to add commit the tests and I can take a look

Thanks! I've added all 6 tests and pushed them. The LSTM tests all pass, but the GRU tests are failing. The outputs don't match PyTorch within tolerance even though the gate reordering looks correct.
All changes are commited. Thank you!

tests/function_libs/torch_lib/e2e_ops_tests.py
```diff
@@ -302,6 +302,90 @@ def forward(self, x):
...
+            def __init__(self):
+                super().__init__()
+                self.gru = torch.nn.GRU(input_size=10, hidden_size=20, num_layers=3, batch_first=True)
+
+            def forward(self, x):
+                return self.gru(x)
+
+        model = GRUModel()
+        x = torch.randn(5, 3, 10)  # (batch, seq, input_size)
+        onnx_program = torch.onnx.export(model, (x,), dynamo=True, verbose=False)
```
titaiwangms, line 406: Could you also add some cases to test the dynamism? I think batch_size and sequence length should at least be dynamic?

For example (I might be wrong on axis names), 
```python
onnx_program = torch.onnx.export(
    model, 
    (x,),
    dynamic_shapes=({0: "batch", 1: "seq_len"}) 
    dynamo=True,
    verbose=False
)
```

justinchuby:
I think it's ok to skip dynamic shapes testing for now if the error comes from torch.export, thanks!

titaiwangms:
> I think it's ok to skip dynamic shapes testing for now if the error comes from torch.export, thanks!

Have not seen this error message for a while. Is the CI using old torch? (before oblivious backed size?)

justinchuby:
> > I think it's ok to skip dynamic shapes testing for now if the error comes from torch.export, thanks!
> 
> Have not seen this error message for a while. Is the CI using old torch? (before oblivious backed size?)

Right we do test with 2.7

PR #2679, opened 2025-11-03
Keep creating constants when constants are folded inside ir.Function

### LINKED ISSUE #2673: Constant-folding behavior inside functions
With the recent change to the constant-folding (see #2650) I started to encounter problems with constant folding within functions: The model checker complains about topological sorting and onnxruntime complains about "Input 0 expected to have type but instead is null" afterwards, but trying to narrow this down it seems the root cause is an initializer created by constant folding is missing from the graph. The following model should be a minimal example to reproduce this:
```
<
   ir_version: 9,
   opset_import: ["this" : 1, "" : 19]
>
model (float x) => (float return_val) {
   [n0] return_val = this.function (x)
}
<
  domain: "this",
  opset_import: ["" : 19]
>
function (x) => (return_val)
{
   [n0] tmp = Constant <value_int: int = 1> ()
   [n1] tmp_0 = Cast <to: int = 1> (tmp)
   [n2] return_val = Sub (tmp_0, x)
}
```

Up until including v0.5.4 this folds (correctly?) to the following, where a single constant node replaces the cast operator:
```
<
   ir_version: 9,
   opset_import: ["this" : 1, "" : 19]
>
model (float x) => (float return_val) 
   <float "this::function/tmp_0">
{
   [n0] return_val = this.function (x)
}
<
  domain: "this",
  opset_import: ["" : 19]
>
function (x) => (return_val)
{
   [node_Constant_0] tmp_0 = Constant <value: tensor = float tmp_0 {1}> ()
   [n2] return_val = Sub (tmp_0, x)
}
```

Since v0.5.5, however, the resulting model looks as follows, where the temporary has no producer and also no initializer (which I would expect according to the recent changes):
```
<
   ir_version: 9,
   opset_import: ["this" : 1, "" : 19]
>
model (float x) => (float return_val) {
   [n0] return_val = this.function (x)
}
<
  domain: "this",
  opset_import: ["" : 19]
>
function (x) => (return_val)
{
   [n2] return_val = Sub (tmp_0, x)
}
```
The model checker raises `ValidationError: Nodes in a function must be topologically sorted, however input 'tmp_0' of node: Name: n2 OpType: Sub is neither output of any previous nodes nor input of the function.` The following reproduces the two outputs, depending on the onnxscript version:
```python
import onnx
import onnx_ir as ir

import onnxscript.optimizer as optimizer
from onnxscript.optimizer import _constant_folding

model = """
<
   ir_version: 9,
   opset_import: ["this" : 1, "" : 19]
>
model (float x) => (float return_val) {
   [n0] return_val = this.function (x)
}
<
  domain: "this",
  opset_import: ["" : 19]
>
function (x) => (return_val)
{
   [n0] tmp = Constant <value_int: int = 1> ()
   [n1] tmp_0 = Cast <to: int = 1> (tmp)
   [n2] return_val = Sub (tmp_0, x)
}
"""
model = ir.from_onnx_text(model)
_constant_folding.fold_constants(model)
optimizer.remove_unused_nodes(model)
print(ir.to_onnx_text(model))
onnx.checker.check_model(ir.serde.serialize_model(model))
```

Is this a bug, or, as #2650 was labeled as a breaking change, expected behavior and I need to adjust my models somehow?

titaiwangms:
Fixes #2673

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -1132,7 +1132,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
 
         return initializer
 
-    def process_node(self, node: ir.Node, is_function: bool = False) -> Replacement | None:
+    def process_node(self, node: ir.Node, is_function: bool) -> Replacement | None:
         """Process a node and return a Replacement if the node can be replaced."""
         for i, value in enumerate(node.inputs):
             sym_value = self._state.get_sym_value(value)
@@ -1285,7 +1285,7 @@ class FoldConstantsPass(ir.passes.InPlacePass):
         if outputs is None:
             return None
         if len(node.outputs) == 1 and not isinstance(outputs, (tuple, list)):
-            # TODO: We don't support initializers in functions yet, so we need to create Constant nodes
+            # We don't support initializers in functions, so we need to create Constant nodes
             if is_function:
                 replacement = self.new_constant(node, outputs)
                 if replacement is None:


onnxscript/optimizer/_constant_folding.py
```diff
@@ -1252,6 +1298,12 @@ def convert(av):
         if outputs is None:
             return None
         if len(node.outputs) == 1 and not isinstance(outputs, (tuple, list)):
+            # TODO: We don't support initializers in functions yet, so we need to create Constant nodes
```
titaiwangms: I suppose this is the main reason?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,6 +1050,52 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
...
+
+        logger.debug(
+            "New constant for value %s dtype: %s shape: %s",
+            irvalue.name,
+            value.dtype,
+            value.shape,
+        )
+
+        attributes = ir.convenience.convert_attributes({"value": tensor})
+        node = ir.Node("", "Constant", inputs=[], attributes=attributes, num_outputs=1)
```
```suggestion
        node = ir.Node("", "Constant", inputs=[], attributes=(ir.AttrTensor("value", tensor),))
```

Simplify like this?

onnxscript/optimizer/_constant_folding.py
```suggestion
            # We don't support initializers in functions so we need to create Constant nodes
```

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,6 +1050,51 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
...
+                irvalue.name,
+                type(value),
+            )
+            return None
+
+        tensor = ir.tensor(value)
+        tensor.name = irvalue.name
+        irvalue.const_value = tensor
+
+        if value.size > self.output_size_limit:
```
justinchuby: Is this logic the same as new_initializer? Can it be shared?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,6 +1050,51 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
                     e,
                 )
 
+    def new_constant(self, node: ir.Node, value) -> ir.Node | None:
```
justinchuby: Possible to create a docstring to document the arguments? Also value is missing a type annotaion

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,6 +1050,51 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
...
+            "New constant for value %s dtype: %s shape: %s",
+            irvalue.name,
+            value.dtype,
+            value.shape,
+        )
+
+        node = ir.Node("", "Constant", inputs=[], attributes=(ir.AttrTensor("value", tensor),))
+        return node
+
     def new_initializer(self, node: ir.Node, array) -> ir.Value | None:
```
justinchuby: Would be great to document the arguments here

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1252,6 +1297,12 @@ def convert(av):
         if outputs is None:
             return None
         if len(node.outputs) == 1 and not isinstance(outputs, (tuple, list)):
+            # TODO: We don't support initializers in functions yet, so we need to create Constant nodes
```
justinchuby: Is this still a todo? What is the expected fix?

justinchuby (approved):
Thanks - minor comments. Would be good to get another eye from @gramalingam

onnxscript/optimizer/_constant_folding.py
titaiwangms: Why won't we support initializers in functions?

onnxscript/optimizer/_constant_folding.py
titaiwangms: Done

onnxscript/optimizer/_constant_folding.py
titaiwangms: Done

onnxscript/optimizer/_constant_folding.py
titaiwangms: Done

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,46 +1050,82 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
...
+    def new_constant(self, node: ir.Node, array: np.ndarray | Any) -> ir.Node | None:
+        """Create a new Constant node with the given array as its value."""
+        original_value = node.outputs[0]
+
+        tensor = self._prepare_folded_tensor(node, original_value.name, array)
+        if tensor is None:
+            return None
+
+        # previous node output value is updated
+        original_value.const_value = tensor
```
titaiwangms: @justinchuby This is from the original. I am wondering if this is redundant, since the node will be removed anyway?

justinchuby: Ah - you are right

titaiwangms: Removed

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1099,7 +1132,7 @@ def new_initializer(self, node: ir.Node, array) -> ir.Value | None:
 
         return initializer
 
-    def process_node(self, node: ir.Node) -> Replacement | None:
+    def process_node(self, node: ir.Node, is_function: bool = False) -> Replacement | None:
```
gramalingam: nit: why do we need a default-value for is_function? Is this called from any other place?

onnxscript/optimizer/_constant_folding.py
```diff
@@ -1050,46 +1050,79 @@ def get_type(value: ir.Value) -> onnx.TypeProto | None:
...
+        node = ir.Node("", "Constant", inputs=[], attributes=(ir.AttrTensor("value", tensor),))
+        return node
+
+    def new_initializer(self, node: ir.Node, array: np.ndarray | Any) -> ir.Value | None:
+        """Create a new initializer value with the given array as its value."""
+        original_value = node.outputs[0]
+
+        tensor = self._prepare_folded_tensor(node, original_value.name, array)
+        if tensor is None:
+            return None
```
gramalingam, line 1117: nit: the first 4 lines of new_constant and new_initializer are the same. Wonder why not just do it in one place (before the `if is_function:` down below?

titaiwangms: hmm I think the current way is fine, as it makes the logic in `is_function` clearer.

onnxscript/optimizer/_constant_folding.py
titaiwangms: Removed

onnxscript/optimizer/_constant_folding.py
titaiwangms: Removed TODO

gramalingam (approved):

PR #2680, opened 2025-11-04
Avoid initializer name collision in _fuse_batchnorm.py

titaiwangms:
Fixes https://github.com/pytorch/pytorch/issues/166797

The original naming collides when there are multiple matched patterns sharing the same parent node. This PR changes the naming to depend on their own Conv weight name, which should be non-duplicated identifier.

~~NOTE: I don't know if my understanding is correct. It seems x is an input of the pattern, which x.name + "_bias" collides with `max_pool` bias (see the pic in the original issue)? If we check the output model after _fuse_batchnorm.py, the bias would be correct with a name `val_17` (the name may be collided and given by NameAuthority?). However, when the following rule _remove_optional_bias tries to fetch the bias, it would see all zero for some reasons.~~

### DIFF
--- a/onnxscript/rewriter/rules/common/_fuse_batchnorm.py
+++ b/onnxscript/rewriter/rules/common/_fuse_batchnorm.py
@@ -69,7 +69,8 @@ class _FuseBatchNormBase(RewriteRuleClassBase, ABC):
         else:
             original_bias = np.zeros_like(input_mean)
             # Use inbound input 1 (should be weight) to derive a name for the bias
-            # to avoid name collision on initializer creation.
+            # to avoid name collision on initializer creation when there are multiple patterns
+            # sharing the same parent nodes.
             bias_name = inbound_node.inputs[1].name + "_bias"
         fused_bias = ir.tensor((original_bias - input_mean) * scale_factor + beta)
 
--- a/onnxscript/rewriter/rules/common/_fuse_batchnorm_test.py
+++ b/onnxscript/rewriter/rules/common/_fuse_batchnorm_test.py
@@ -253,6 +253,64 @@ class FuseBatchnormTest(unittest.TestCase):
         # No changes were applied as W is a graph input
         self.assertEqual(count, 0)
 
+    def test_fuse_batchnorm_does_not_collide_names_with_same_parent_node(self):
+        model_proto = onnx.parser.parse_model("""
+            < ir_version: 7, opset_import: ["" : 17] >
+            test_model (float[N, 32, 14, 16] X) => (float [N, ?, ?, ?] Y1, float [N, ?, ?, ?] Y2)
+            {
+                X1 = MaxPool<kernel_shape=[3,3]>(X)
+                X2 = Conv(X1, W1)
+                Y1 = BatchNormalization(X2, gamma_64, beta_64, input_mean_64, input_var_64)
+                X3 = Conv(X1, W2)
+                Y2 = BatchNormalization(X3, gamma_256, beta_256, input_mean_256, input_var_256)
+            }
+        """)
+        initializers = [
+            onnx.numpy_helper.from_array(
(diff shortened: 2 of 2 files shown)


onnxscript/rewriter/rules/common/_fuse_batchnorm.py
```diff
@@ -68,7 +68,9 @@ def rewrite(self, op, x: ir.Value, inbound_out: ir.Value, batchnorm_out: ir.Valu
             bias_name = inbound_node.inputs[2].name
         else:
             original_bias = np.zeros_like(input_mean)
-            bias_name = x.name + "_bias"
```
justinchuby, line 71: Curious why the original would collide?

titaiwangms: It's hard to tell. Let me see if I can dig more.

AyoubMDL: It is when there are multiple consumers. In the issue, there are two match patterns with the same parent, something like this 
image
I think it's good to take the weights name or the output name of the Conv

titaiwangms: Thanks! I will add a test case.

titaiwangms: Done

gramalingam: Sounds like:
1. When we create initializers, we need a way to create new unique names (at least unique among initializers, to begin with). I wonder if we can expose an IR method that will change an initializer's name to be unique (instead of throwing an exception) when adding it?
2. In this example, it should work fine since the first inputs are different. What happens if the first inputs are the same in two nodes? Granted, it won't happen in realistic examples. But it could still cause an exception, I think?

titaiwangms: > When we create initializers, we need a way to create new unique names (at least unique among initializers, to begin with). I wonder if we can expose an IR method that will change an initializer's name to be unique (instead of throwing an exception) when adding it?

Agree. In the created case, from what I saw, we don't even see an exception, the fused bias from pattern 1 and 2 co-exists with the same names.

> 2. In this example, it should work fine since the first inputs are different. What happens if the first inputs are the same in two nodes? Granted, it won't happen in realistic examples. But it could still cause an exception, I think?

In terms of Conv node, the example is problematic when bias is fused from op.BatchNorm to op.Conv because they will share the same naming.

justinchuby (approved):

onnxscript/rewriter/rules/common/_fuse_batchnorm.py
titaiwangms: Merge for now

titaiwangms: Feel free to continue the discussion

onnxscript/rewriter/rules/common/_fuse_batchnorm.py
```diff
@@ -68,7 +68,10 @@ def rewrite(self, op, x: ir.Value, inbound_out: ir.Value, batchnorm_out: ir.Valu
...
+            # Use inbound input 1 (should be weight) to derive a name for the bias
+            # to avoid name collision on initializer creation when there are multiple patterns
+            # sharing the same parent nodes.
+            bias_name = inbound_node.inputs[1].name + "_bias"
```
xadupre, line 74: add one output name into the name, that way it will be unique

onnxscript/rewriter/rules/common/_fuse_batchnorm.py
titaiwangms: https://github.com/onnx/ir-py/issues/246

PR #2682, opened 2025-11-04
Merge metadata props in rewriter

gramalingam:
Introduce basic infrastructure for merging metadata props (for use in rewriter/optimizer etc.) A basic version added to rewriter.

TODO:
* Allow user control over this: should this be configurable at the level of a RewriteRuleSet? Or, perhaps at a global level (given that ORT fusions uses a number of rewrite-rule-sets for various reasons)?
* This [line](https://github.com/microsoft/onnxscript/blob/1a27df145b7ec03da7d316a38c2cb005cf0a45b7/onnxscript/rewriter/ort_fusions/_core.py#L148) should also be factored out or made user-controllable in some fashion. Otherwise, the metadata gets lost anyway.

### DIFF
--- a/onnxscript/rewriter/__init__.py
+++ b/onnxscript/rewriter/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from typing import Sequence, TypeVar, Union
 
 __all__ = [
+    "merge_metadata",
     "pattern",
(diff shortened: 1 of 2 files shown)


justinchuby (approved):
Can have a test?

PR #2683, opened 2025-11-05
Implement SDPA via MHA

gramalingam:
Implement SDPA via MHA. This handles the case when earlier fusion rules do not map larger patterns containing SDPA into MHA or GQA or Attention (from ORT contrib ops). It implements SDPA via MHA.

### DIFF
--- a/onnxscript/rewriter/ort_fusions/sdpa_via_mha.py
+++ b/onnxscript/rewriter/ort_fusions/sdpa_via_mha.py
@@ -14,6 +14,7 @@ Dim = Union[int, ir.SymbolicDim]
 
 class SDPAImplementation(pattern.RewriteRuleClassBase):
     def pattern(self, op, query, key, value, key_format):
+        """Pattern matches any call to SDPA. See sdpa.py for documentation on the SDPA op."""
         return op.SDPA(
             query,
             key,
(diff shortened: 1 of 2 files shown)


onnxscript/rewriter/ort_fusions/sdpa_via_mha.py
```diff
@@ -7,43 +7,56 @@
...
 
 from onnxscript.rewriter import _fusion_utils, pattern
+from onnxscript.rewriter._basics import MatchFailureError
 
 Dim = Union[int, ir.SymbolicDim]
 
 
 class SDPAImplementation(pattern.RewriteRuleClassBase):
-    def pattern(self, op, query, key, value):
+    def pattern(self, op, query, key, value, key_format):
```
justinchuby, line 16: Create a docstring for these params?

gramalingam: Thinking ... [ok, I picked that from chatgpt :-)] ... I suspect what you are asking for is a "spec" for the SDPA (the internal op used by fusion, which does not correspond to any ONNX standard op or ORT contrib op, though it is a close approximation to ONNX's Attention). Is that right? Because that op is used across multiple fusions and files, the sdpa.py file might be the right place for that. Because the 1-liner docstring for these params would be that they correspond to the corresponding inputs and attribute of the SDPA op ... which doesn't say anything more than what the reader can see in the next few lines.

gramalingam: Ok, added documentation as discussed above.

PR #2684, opened 2025-11-05
[torchlib] Fix unbind.int if num_outputs=1

sebimarkgraf:
This fixes the issue of 
```
  return [op.Squeeze(out, [dim]) for out in outputs]
                                              ^^^^^^^
TypeError: 'SymbolicTensor' object is not iterable
```
when trying to export LSTM modules in `torch`.
This also already appeared in torch issues in https://github.com/pytorch/pytorch/issues/126339

The core seems to be the changes in #2597.
To my understanding the split returns a single `SymbolicTensor` instead of a sequence when `dim=1`.
The fix implemented here is the casting of the return type to a list.

I struggled with writing a test that reproduces this nicely in here, any guidance on that would be welcome.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -8958,9 +8958,10 @@ def aten_unbind(self: TTensor, dim: int = 0) -> Sequence[TTensor]:
         # We can create a definitive split op if the input shape is static
         # Only torch>=2.7 supports correctly generating the correct number of outputs for Split
         num_outputs = self.shape[dim]
-        outputs = op.Split(self, axis=dim, num_outputs=num_outputs)
-        if num_outputs == 1:
-            outputs = [outputs]
+        if num_outputs != 1:
+            outputs = op.Split(self, axis=dim, num_outputs=num_outputs)
+        else:
+            outputs = [self]
 
         return [op.Squeeze(out, [dim]) for out in outputs]
 


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -8957,7 +8957,11 @@ def aten_unbind(self: TTensor, dim: int = 0) -> Sequence[TTensor]:
     if isinstance(self.shape[dim], int) and not version_utils.torch_older_than("2.7"):
         # We can create a definitive split op if the input shape is static
         # Only torch>=2.7 supports correctly generating the correct number of outputs for Split
-        outputs = op.Split(self, axis=dim, num_outputs=self.shape[dim])
+        num_outputs = self.shape[dim]
+        outputs = op.Split(self, axis=dim, num_outputs=num_outputs)
+        if num_outputs == 1:
+            outputs = [outputs]
```
```suggestion
        outputs = op.Split(self, axis=dim, num_outputs=num_outputs)
        if not isinstance(outputs, Sequence):
            outputs = [outputs]
```

gramalingam:
> If num_outputs=1, should we still use a Split node, or is Identity ok?

Seems worthwhile emitting the optimized version without Split ... but it would still need the Squeeze ... no need for even Identity if we handle the two cases completely separately.

onnxscript/function_libs/torch_lib/ops/core.py
```suggestion
        if num_outputs != 1:
            outputs = op.Split(self, axis=dim, num_outputs=num_outputs)
        else:
            outputs = [self]
```

justinchuby (approved):

xadupre (approved):

PR #2685, opened 2025-11-05
Add option to clear metadata in ort fusion

gramalingam:
Add option to clear metadata in ort fusion. Otherwise, we lose all metadata after running ORT fusion

### DIFF
--- a/onnxscript/rewriter/ort_fusions/_core.py
+++ b/onnxscript/rewriter/ort_fusions/_core.py
@@ -129,6 +129,7 @@ def optimize_for_ort(
             Typically it identifies the Execution Provider (EP) to optimize for.
             If None, the default configuration will be used.
         debug: If debug is True, enable pattern matching tracer for debugging.
+        clear_metadata: If True, clear metadata and doc strings from the model.
 
     Returns:
         A tuple containing:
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -115,6 +115,7 @@ def optimize_for_ort(
     config_name: str | None = None,
     *,
     debug: bool = False,
+    clear_metadata: bool = False,
```
gramalingam, line 118: @titaiwangms : can we change the calls to specify this flag to be true, when desired? Hopefully we have only two or three places where we call this.

We could make the default-value True, but it seems to me for devs/experimenters False is better. But that's an option too

justinchuby: Defaulting to false sounds preferable to me

onnxscript/rewriter/ort_fusions/_core.py
```diff
@@ -142,10 +143,12 @@ def optimize_for_ort(
     # Apply the ORT pattern rewrite rules.
     rewrite(model, ORT_PATTERN_REWRITE_RULES)
 
+    if clear_metadata:
+        common_passes.ClearMetadataAndDocStringPass()(model)
```
justinchuby: We should do it at the end because the other passes may create metadata?

onnxscript/rewriter/ort_fusions/_core.py
titaiwangms: This was set to True because builder.py (model shipping) needs it. I guess it's fine as long as it's exposed to public. I don't think any published code are relying this ort_fusion and this default value.

onnxscript/rewriter/ort_fusions/_core.py
titaiwangms: Yeah, I think it's better we have this in the end, in case metadata is needed in future passes as well.

gramalingam: Done

PR #2686, opened 2025-11-07
[torchlib] Fix mod on SymInt

justinchuby:
Fix the error: `<class 'AttributeError'>: 'int' object has no attribute 'dtype'` by splitting the implementation out for `operator.*` ops.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -3767,6 +3767,192 @@ def aten_ger(self: TensorType, vec2: TensorType) -> TensorType:
     raise NotImplementedError()
 
 
+@torch_op("aten::gru.input", trace_only=True)
+def aten_gru(
+    input: TFloat,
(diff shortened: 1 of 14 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -3749,6 +3749,12 @@ def aten_ge(self: TTensor, other: TTensor) -> BOOL:
     return op.GreaterOrEqual(self, other)
 
 
+@torch_op("_operator::ge", trace_only=True)
+def operator_ge(self: TTensor, other: TTensor) -> BOOL:
```
titaiwangms, line 3753: Should this be TInt?

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: This is more lenient, which I think is ok

Issue #2687, opened 2025-11-07
repeat_interleave is missing support for output_size kwarg


yuanyao-nv:
When converting a graph with repeat_interleave, I'm hitting the following error
```
<class 'TypeError'>: aten_repeat_interleave_self_int() got an unexpected keyword argument 'output_size'
```
Looks like onnxscript is missing support for the argument?
https://github.com/microsoft/onnxscript/blob/70e751ae4a630eb444f80bfd4af435a94ca747f3/onnxscript/function_libs/torch_lib/ops/core.py#L7512-L7515

PR #2691, opened 2025-11-08
Add missing output_size kwarg to repeat_interleave

yuanyao-nv:
Fixes https://github.com/microsoft/onnxscript/issues/2687


justinchuby (approved):

yuanyao-nv:
> Does it affect the implementation?

It should not. It's only for possible optimizations, see https://docs.pytorch.org/docs/stable/generated/torch.repeat_interleave.html

PR #2693, opened 2025-11-09
[torchlib] Fix linspace implementation for int64

### LINKED ISSUE #854: [torchlib] linspace results do not match with PyTorch when dtype is int64

Aravind-11:
## Description
Fixes #854 - linspace now correctly handles int64 dtype

## Changes
- Modified `aten_linspace` to compute in floating-point then cast to target dtype
- This matches PyTorch's behavior and fixes integer division precision loss

## Testing
Manually verified: `linspace(0, 10, 5, dtype=int64)` now produces correct output `[0, 2, 5, 7, 10]`

## Questions
Where should I add automated test cases for this fix? Happy to add tests wherever you suggest!

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -5517,6 +5517,7 @@ def aten_linspace(
     # are truncated before computing the linspace
     dtype = ir.DataType(dtype)
     if dtype.is_integer():
+        # Use double precision for computation to match PyTorch's internal precision
         compute_dtype = ir.DataType.DOUBLE
         # Cast to integer dtype first (truncation), then to compute dtype
         start_int = op.Cast(start, to=dtype)  # Truncate to int32/int64


Aravind-11:
> ## Description
> Fixes #854 - linspace now correctly handles int64 dtype
> 
> ## Changes
> * Modified `aten_linspace` to compute in floating-point then cast to target dtype
> * This matches PyTorch's behavior and fixes integer division precision loss
> 
> ## Testing
> Manually verified: `linspace(0, 10, 5, dtype=int64)` now produces correct output `[0, 2, 5, 7, 10]`
> 
> ## Questions
> Where should I add automated test cases for this fix? Happy to add tests wherever you suggest!

Who can review : @justinchuby

justinchuby:
Thanks. Could you unskip the tests: https://github.com/microsoft/onnxscript/blob/c1bfdfc5baf2dabf09191ad2ab7603ea21cfe614/tests/function_libs/torch_lib/ops_test_data.py#L770

Aravind-11:
> Thanks. Could you unskip the tests:
> 
> https://github.com/microsoft/onnxscript/blob/c1bfdfc5baf2dabf09191ad2ab7603ea21cfe614/tests/function_libs/torch_lib/ops_test_data.py#L770

Thank you for reviewing! Done.

Aravind-11:
> > Thanks. Could you unskip the tests:
> > https://github.com/microsoft/onnxscript/blob/c1bfdfc5baf2dabf09191ad2ab7603ea21cfe614/tests/function_libs/torch_lib/ops_test_data.py#L770
> 
> Thank you for reviewing! Done.

hi @justinchuby , are there any updates on this that you could let me know. Thank you!

Aravind-11:
> Thanks. In CI we only run cpu tests so we should be ok.

Got it. Thank you. Could you please approve the tests ?

Aravind-11:
I tested both precisions, and float64 is required for correctness.
With float32, we get precision errors at certain indices:

```python
# Index 21: float32 gives 0.999999761... → truncates to 0
#           float64 gives 1.000000000000... → truncates to 1 
```

Aravind-11:
Hi!, anything blocking merge ? lmk if i need to change anything. thank you!

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5117,24 +5117,22 @@ def aten_linspace(
...
         # Cast to integer dtype first, then to compute dtype
         # This ensures truncation happens before computation
         start_int = op.Cast(start, to=dtype)
         end_int = op.Cast(end, to=dtype)
-        start_f = op.Cast(start_int, to=compute_dtype)
-        end_f = op.Cast(end_int, to=compute_dtype)
+        start = op.Cast(start_int, to=compute_dtype)
+        end = op.Cast(end_int, to=compute_dtype)
     else:
-        # For float dtypes, cast directly to compute dtype
```
justinchuby: @Aravind-11 there is one issue: We don't always want to use the DOUBLE type. When the input is float, it would be preferable to use a smaller type for performance. Is there a way to use DOUBLE only when needed?

Aravind-11: Suggested Comment:

Hi @justinchuby, the current implementation already uses DOUBLE only for integer output dtypes. For float outputs, it uses the target dtype directly: 

```python
if ir.DataType(dtype).is_integer():
    compute_dtype = ir.DataType.DOUBLE  # Only for int32/int64
else:
    compute_dtype = dtype  # Uses float16/float32/float64 directly
```
However, I want to confirm: should we also use higher precision for float16 outputs? PyTorch might compute linspace in float32 internally even when the output is float16. I can test this and adjust if needed.

Let me know if there's a specific case you're concerned about!

justinchuby: Given that there are test errors, I wonder which part went wrong? For float16 I don't mind using float32 if that's what's required, but I would be careful to make sure only when it's necessary.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -5517,12 +5517,12 @@ def aten_linspace(
     # are truncated before computing the linspace
     dtype = ir.DataType(dtype)
     if dtype.is_integer():
-        # Use double precision for computation to match PyTorch's internal precision
```
justinchuby: I think we can keep this comment

Aravind-11: got it .

justinchuby (approved):

PR #2697, opened 2025-11-10
add converter for aten::sym_storage_offset

xadupre:


titaiwangms (approved):
ExportedProgram will complain if we put this in https://github.com/pytorch/pytorch/blob/f6a79b2a4ac92a9077fc03cb124e3bc0a0a28730/torch/onnx/_internal/exporter/_fx_passes.py#L33

PR #2703, opened 2025-11-12
Use IR and inplace output in relace_functions

titaiwangms:
Use IR and inplace

### DIFF
--- a/onnxscript/utils/replace.py
+++ b/onnxscript/utils/replace.py
@@ -4,14 +4,11 @@
 
 from typing import Sequence
 
-import onnx
 import onnx_ir as ir
 import onnx_ir.passes.common as common_passes
 
 
-def replace_functions(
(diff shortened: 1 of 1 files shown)


onnxscript/utils/replace.py
```diff
@@ -4,14 +4,11 @@
...
 
-import onnx
 import onnx_ir as ir
 import onnx_ir.passes.common as common_passes
 
 
-def replace_functions(
-    model: onnx.ModelProto, functions: Sequence[onnx.FunctionProto]
-) -> onnx.ModelProto:
+def replace_functions(irmodel: ir.Model, irfunctions: Sequence[ir.Function]) -> None:
```
gramalingam, line 11: I would suggest two functions: create a new one `replace_functions_inplace`, which is like this one (with IR inputs updated in-place), and update the existing `replace_functions`  to call the inplace one in the middle ... some might need one or the other

PR #2705, opened 2025-11-14
Remove usage of python typing for programmatic annotation of tensor dtype/shapes.

### LINKED ISSUE #2666: Remove usage of python typing for programmatic annotation of tensor dtype/shapes
Python typing is a little fragile across python versions

Aniketsy:
Fix #2666 

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thankyou !


gramalingam:
Hmmm ... I am not sure I understand the issue here. Justin can elaborate what he meant in #2666 ... but I think he is referring to the general use of type annotation in onnxscript. Would like to understand the issue better

Aniketsy:
@justinchuby Could you please provide more details about the issue and correct me if i’m misunderstanding it?

PR #2706, opened 2025-11-15
Migrate onnxscript converter to use onnx ir

gramalingam:
Migrate onnxscript converter to use onnx ir.

### DIFF
--- a/onnxscript/values.py
+++ b/onnxscript/values.py
@@ -657,31 +657,7 @@ class OnnxFunction(Op, Generic[_P, _R]):
         Returns:
             An instance of :class:`onnx.ModelProto`.
         """
-        value_infos = (
-            [
-                onnx.helper.make_value_info(name, type.to_type_proto())
-                for name, type in value_infos.items()
-            ]
-            if value_infos
-            else None
-        )
-
-        graph = self.function_ir.to_graph_proto()
-        if value_infos:
-            graph.value_info.extend(value_infos)
-        if io_types is not None:
-            for input in graph.input:
-                if not input.HasField("type"):
-                    input.type.CopyFrom(io_types.to_type_proto())
-            for output in graph.output:
-                if not output.HasField("type"):
-                    output.type.CopyFrom(io_types.to_type_proto())
-        if input_types is not None:
-            for input, type in zip(graph.input, input_types):
-                input.type.CopyFrom(type.to_type_proto())
-        if output_types is not None:
-            for output, type in zip(graph.output, output_types):
-                output.type.CopyFrom(type.to_type_proto())
+        # Identify functions to include in the model
         if functions is None:
             sub_functions = self.function_ir.get_called_functions()
             functions = sub_functions.values()
@@ -696,7 +672,8 @@ class OnnxFunction(Op, Generic[_P, _R]):
 
             functions = [to_proto(f) for f in functions]
 
-        opsets = self.function_ir.opset_imports.copy()
+        # Determine opset imports
+        opsets = self.function_ir.graph.opset_imports
 
         for proto in functions:
             if proto.domain not in opsets:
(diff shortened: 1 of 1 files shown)


onnxscript/values.py
```diff
@@ -23,17 +26,30 @@
...
 from onnxscript._internal import ast_utils, deprecation
 from onnxscript.ir import _schemas
+from onnxscript.onnx_types import ONNXType
 
 _R = TypeVar("_R")
 _P = ParamSpec("_P")
 
 
+def select_ir_version(version: int, domain: str = "") -> int:
+    """Selects a suitable ONNX ir_version for a given opset version."""
```
justinchuby, line 43: Probably default the minimal version to a more recent one, like v10, such that the generated model is compatible with newer tool chains (metadata)? This can also be a good default to push for adoption

onnxscript/values.py
```diff
@@ -23,17 +26,30 @@
...
+    if domain == "":
+        domain = "ai.onnx"
+    if (domain, version) not in onnx.helper.OP_SET_ID_VERSION_MAP:
```
justinchuby, line 46: I would consider adding inline noqas for TID251 so we allow the helper module in a case by case basis. There is also a possibility that it is not needed.

onnxscript/values.py
```diff
@@ -603,7 +624,103 @@ def to_model_proto(self, **kwargs):
...
+                opset_version if opset_version is not None else onnx.defs.onnx_opset_version()
+            )
+
+        if "ir_version" not in kwargs:
+            kwargs["ir_version"] = select_ir_version(opsets[""])
+        opset_imports = [
+            onnx.helper.make_opsetid(domain, version) for domain, version in opsets.items()
+        ]
+
+        return onnx.helper.make_model(
```
justinchuby: Curious why we do this instead of converting from an IR Model?

onnxscript/values.py
gramalingam: This (choosing a default IR version) has been an annoying problem in onnx repo at least. Ideal is probably to not have a default, forcing the users to select one. Except that that would likely break some unit test-cases and backward compat.

gramalingam: If we consider its usage in evaluator, to evaluate a node using ORT: we would need some logic like this ... eg., the op might use a type that was introduced in IR version > 10, and then 10 won't work ... but when we have the op schema (opset version), we can figure out the minimum IR version required. Perhaps we can force it to be max(10, version selected as above), but not sure about just plain 10.

justinchuby: I think a `min(10, version)` makes sense, as we don't need a lower IR version

onnxscript/values.py
gramalingam: Not sure I understand. The IR now internally uses function/graph (in IR form), but does not maintain any ir.Model. We are generating the GraphProto from the Graph/Function IR, and wrapping a ModelProto around it.

justinchuby: I meant that we can create an ir model first, then serialize it to a model proto. This way we can avoid a copy to the ModelProto when it is initialized from the graph proto.

onnxscript/values.py
```diff
@@ -47,7 +47,8 @@ def select_ir_version(version: int, domain: str = "") -> int:
         return max(
             v for k, v in onnx.helper.OP_SET_ID_VERSION_MAP.items() if k[0] == "ai.onnx"
         )
-    return onnx.helper.OP_SET_ID_VERSION_MAP[domain, version]
+    required_min_version = onnx.helper.OP_SET_ID_VERSION_MAP[domain, version]
+    return max(required_min_version, 10)
```
justinchuby, line 51: I still think this should be min. Apparently we may need a version higher than 10 if we encounter a new dtype, but there is no point using a version that is older than 10

onnxscript/values.py
justinchuby: We can also avoid using onnx.helper this way.

onnxscript/values.py
gramalingam: ? ... if we need to use IR version 12 for some new dtype, then required_min_version = 12. max(12, 10) is 12, and we use 12. min(12, 10) is 10 ... what am I missing?

onnxscript/values.py
gramalingam: That's true. But the `functions` input parameter complicates it a bit, since it is apparently allowed to contain a FunctionProto ... not sure if this is used by anyone. We may want clean this up also. But meanwhile it requires either a bit more  complex logic or double-conversion to IR and back. Let me see what I can do.

gramalingam: I guess I can append the functions after creating the ModelProto.

onnxscript/values.py
justinchuby: My bad 😅 I don't know what I was thinking

PR #2708, opened 2025-11-17
Provide inplace replacement util

gramalingam:
Redo https://github.com/microsoft/onnxscript/pull/2703, thanks to Titai.

### DIFF
--- a/onnxscript/utils/replace.py
+++ b/onnxscript/utils/replace.py
@@ -9,6 +9,29 @@ import onnx_ir as ir
 import onnx_ir.passes.common as common_passes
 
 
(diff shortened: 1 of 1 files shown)


onnxscript/utils/replace.py
```diff
@@ -9,19 +9,17 @@
 import onnx_ir.passes.common as common_passes
 
 
-def replace_functions(
-    model: onnx.ModelProto, functions: Sequence[onnx.FunctionProto]
-) -> onnx.ModelProto:
+def replace_functions_inplace(irmodel: ir.Model, irfunctions: Sequence[ir.Function]) -> None:
```
justinchuby, line 12:
```suggestion
def replace_functions_inplace(model: ir.Model, functions: Collection[ir.Function]) -> None:
```

Consider consistent names

PR #2713, opened 2025-11-21
Don't constant fold Quantize/DequantizeLinear nodes by default

ruro:
I added support for exporting `QuantizeLinear`/`DequantizeLinear` nodes (from `fake_quantize_per_*_affine` torch operators) in a previous PR.

Unfortunately, the current default onnxscript optimizer settings tend to automatically remove any weight quantization. This is because the `Weight -> QDQ -> ...` pattern looks like it can be just constant folded to `QDQ(Weight) -> ...`.

I believe that this behavior is not desirable, since the presence of `QDQ` nodes in the graph is what allows inference engines to run the supported computations using quantized data types. So the purpose of `QDQ` nodes is to hold the relevant quantization "metadata". As such, they normally shouldn't be constant folded.

I have extended the existing logic in `FoldConstantsPass` that was used to exclude `ConstantOfShape` from constant folding.

I haven't found any tests verifying this behavior for `ConstantOfShape` and I'm not sure, how to set up such a unit test, so I have left this code untested for now. If adding tests is mandatory, please give me a hint on where should I add such a test and what would be the best way to check/assert that the optimized graph matches the expectations (hopefully without reinventing the wheel or manually introspecting the `ir.Model` object).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -45,7 +45,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v3
+      uses: github/codeql-action/init@v4
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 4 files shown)


titaiwangms (approved):
@justinchuby @gramalingam @xadupre Merging this as it seems pretty reasonable to me and the fix is simple. We can re-visit this PR if it obviously breaks anything and it's spotted in the future (observing from benchmark perhaps?).

PR #2719, opened 2025-11-26
  Fix aten_unbind for torch >= 2.7 dynamo export

afshin-paydar:
Fix aten_unbind for torch >= 2.7 dynamo export

## Problem

When exporting PyTorch models to ONNX with `dynamo=True` on torch >= 2.7, the `aten_unbind` operation fails with:

```
TypeError: 'SymbolicTensor' object is not iterable
```

This occurs because `op.Split(self, axis=dim, num_outputs=num_outputs)` returns a single `SymbolicTensor` object rather than an iterable sequence during dynamo export. The subsequent list comprehension `[op.Squeeze(out, [dim]) for out in outputs]` attempts to iterate over this non-iterable object, causing that error.

This affects models using LSTM and other operations that internally call `unbind`, preventing successful ONNX export.

## Solution
Replace the `Split` op approach with explicit `Slice` operations:
  - For each output along the unbind dimension, create an individual `Slice` operation to extract one element
  - Apply `Squeeze` to remove the size-1 dimension
  - Collect all results in a list

## Test coverage:
  - num_outputs = 1: test_unbind_size_one
  - num_outputs = 2: test_unbind_with_lstm, test_unbind_dim1 (size 2 along some dims) 
  - num_outputs = 3: test_unbind_dim0, test_unbind_dim1 
  - num_outputs = 4: test_unbind_negative_dim 
  - Negative dimensions: test_unbind_negative_dim 

## Related Issues
Fixes pytorch/pytorch#168969

### DIFF
--- a/tests/function_libs/torch_lib/e2e_ops_tests.py
+++ b/tests/function_libs/torch_lib/e2e_ops_tests.py
@@ -596,6 +596,36 @@ class TorchLibe2eTest(unittest.TestCase):
         onnx_program = torch.onnx.export(model, (tokens, h, c), dynamo=True, verbose=False)
         _testing.assert_onnx_program(onnx_program)
 
+    def test_unbind_dynamic_dim0(self):
+        """Test unbind with dynamic dimension 0 - triggers SplitToSequence"""
+
+        class UnbindModel(torch.nn.Module):
+            def forward(self, x):
+                tensors = torch.unbind(x, dim=0)
+                return sum(tensors)
+
+        model = UnbindModel()
+        x = torch.randn(3, 4, 5)
+        onnx_program = torch.onnx.export(
+            model, (x,), dynamo=True, verbose=False, dynamic_shapes=({0: "batch_size"},)
+        )
+        _testing.assert_onnx_program(onnx_program)
+
+    def test_unbind_dynamic_dim1(self):
+        """Test unbind with dynamic dimension 1 - triggers SplitToSequence"""
+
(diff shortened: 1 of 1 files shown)


titaiwangms (commented):
Can you add your test as well? https://github.com/microsoft/onnxscript/blob/main/tests/function_libs/torch_lib/e2e_ops_tests.py

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -9201,15 +9201,22 @@ def aten_unbind(self: TTensor, dim: int = 0) -> Sequence[TTensor]:
     """unbind.int(Tensor(a -> *) self, int dim=0) -> Tensor(a)[]"""
 
     if isinstance(self.shape[dim], int) and not version_utils.torch_older_than("2.7"):
```
justinchuby: Can remove the version check because the change is compatible with older torch versions too

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -9201,15 +9201,22 @@ def aten_unbind(self: TTensor, dim: int = 0) -> Sequence[TTensor]:
...
-        # We can create a definitive split op if the input shape is static
-        # Only torch>=2.7 supports correctly generating the correct number of outputs for Split
+        # For torch>=2.7 with static shapes, use explicit Slice operations
+        # to avoid issues with Split returning a non-iterable SymbolicTensor
```
```suggestion
```

PR #2727, opened 2025-12-12
[torchlib] Fix and implement overloads for aten::remainder

justinchuby:
Previously the Scalar_Tensor overload will fail because the first arg will be a scalar which does not have dtype.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -7657,11 +7657,8 @@ def aten_refine_names(self: TensorType, names: Sequence[str]) -> TensorType:
     raise NotImplementedError()
 
 
-@torch_op("aten::remainder.Tensor", trace_only=True)
-def aten_remainder(self: TTensor, other: TTensor) -> TTensor:
-    """remainder.Tensor(Tensor self, Tensor other) -> Tensor"""
-
-    if self.dtype.is_integer():
+def _aten_remainder(self: TTensor, other: TTensor, integer: bool) -> TTensor:
+    if integer:
         return op.Mod(self, other)
 
     # TODO(justinchuby): Improve fp16 precision by following the logic in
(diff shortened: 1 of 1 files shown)


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7673,6 +7673,36 @@ def aten_remainder(self: TTensor, other: TTensor) -> TTensor:
...
+        return op.Mod(self, other_tensor)
+
+    # a - a.div(b, rounding_mode="floor") * b
+    rounded_quotient = op.Floor(op.Div(self, other_tensor))
+
+    return op.Sub(self, op.Mul(rounded_quotient, other_tensor))
+
+
+@torch_op("aten::remainder.Scalar_Tensor", trace_only=True)
+def aten_remainder_scalar_tensor(self: TTensor, other: TTensor) -> TTensor:
```
```suggestion
def aten_remainder_scalar_tensor(self: float, other: TTensor) -> TTensor:
```

titaiwangms (approved):

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -7673,6 +7673,36 @@ def aten_remainder(self: TTensor, other: TTensor) -> TTensor:
     return op.Sub(self, op.Mul(rounded_quotient, other))
 
 
+@torch_op("aten::remainder.Scalar", trace_only=True)
+def aten_remainder(self: TTensor, other: float) -> TTensor:
+    """remainder.Scalar(Tensor self, Scalar other) -> Tensor"""
+
+    other_tensor = ir.tensor(other, dtype=self.dtype)
+
+    if self.dtype.is_integer():
```
xadupre: I think onnx Mod supports float as well but i don't remember why we should not use it here.

justinchuby: Hmm good question

justinchuby: The rounding behavior is different: https://docs.pytorch.org/docs/stable/generated/torch.remainder.html

justinchuby: I think we should expand https://onnx.ai/onnx/operators/onnx__Mod.html support for float when fmod=0

justinchuby: Created https://github.com/onnx/onnx/issues/7520. Resolving for now

PR #2732, opened 2025-12-16
Fix unused initializer check

### LINKED ISSUE #2728: Optimizer bug on cond
The optimizer in 0.5.6 and main produces a wrong model when running on the attached model that has a foldable graph output. Discovered through a torch exporter test case.

Before:

Image

After:

Image

## Models
[models.zip](https://github.com/user-attachments/files/24143889/models.zip)

cc @xadupre @titaiwangms

gramalingam:
The check for removing unused initializers in the constant-folder was not handling the case where the initializer is a graph output.

Fix #2728


justinchuby:
@copilot is there a test you can create?

PR #2741, opened 2025-12-20
[Rewriter] Extend list of supported commutative operations

iksnagreb:


onnxscript/rewriter/_pattern_ir.py
```diff
@@ -899,12 +899,26 @@ def num_outputs(self) -> int:
         return len(self._outputs)
 
     def commute(self) -> Sequence[GraphPattern]:
+        # List all commutative elementwise (binary) operators for which we
```
gramalingam, line 902: This is fine ... but, please note that the existing implementation can lead to an exponential increase in the number of patterns that are checked, with efficiency implications. (The underlying implementation needs to be improved: for example, now that there is support for disjunctive (OR) patterns, it may be useful to use that instead of creating many patterns.)

justinchuby: Should we make `commute` rewrite the pattern into disjunctive patterns instead?

iksnagreb: Are there still doubts regarding the performance implications? Could I do anything to help, though I don't think I could contribute a complete rework of this mechanism...

We really need this to move forward, but as our rewrite patterns can become quite complex, manually writing out all commutative variations is infeasible and this mechanism comes in really handy.

justinchuby: @gramalingam

fpjentzsch:
+1 This is a blocker for our project.

PR #2742, opened 2025-12-22
Add op test for torch.unique_consecutive

### LINKED ISSUE #2695: [torchlib] Add op test to torch.unique_consecutive
We should check if common_invocation_method.py has the op test.

cnaples79:
## Summary
- Adds OpInfo-based test for `torch.unique_consecutive` operator
- Addresses issue #2695

## Changes
- Added `sample_inputs_unique_consecutive` function in `extra_opinfo.py` that filters samples from `common_methods_invocations.sample_inputs_unique` to only include cases with `dim=None` (as the implementation has limited dim support)
- Added `OpInfo` entry for `ops.aten.unique_consecutive` with `integral_types` dtypes (matching the implementation which only supports int32/int64)
- Added `TorchLibOpInfo` entry in `ops_test_data.py` to enable the test

## Testing
The test follows the same pattern as other unique operator tests (`_unique`, `_unique2`, `unique_dim`) and reuses PyTorch's standard test infrastructure.

Fixes #2695

### DIFF
--- a/tests/function_libs/torch_lib/extra_opinfo.py
+++ b/tests/function_libs/torch_lib/extra_opinfo.py
@@ -2270,6 +2270,16 @@ def sample_inputs_unique_dim(op_info, device, dtype, requires_grad, **kwargs):
             yield sample
 
 
+def sample_inputs_unique_consecutive(op_info, device, dtype, requires_grad, **kwargs):
+    for sample in common_methods_invocations.sample_inputs_unique(
+        op_info, device, dtype, requires_grad, **kwargs
+    ):
+        # unique_consecutive only supports dim=None or (dim=0 with rank=1)
+        # So filter out samples with dim != None
+        if sample.kwargs.get("dim") is None:
+            yield sample
+
+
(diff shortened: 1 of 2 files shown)


justinchuby:
Thanks - looks like there are some errors

justinchuby:
Thanks — but this OpInfo currently exposes a genuine lowering bug rather than just adding coverage, which is why the matrix is red. For `return_inverse=True` with multi-dim or scalar inputs, `torch.unique_consecutive` returns `inverse` shaped like the ORIGINAL input, but `aten_unique_consecutive` (~core.py L10031) computes `inverse` on the FLATTENED x and never reshapes it back; scalar `()` inputs (reshaped to `(1,)`) also mismatch torch's `()` inverse. Two ways forward: (a) fix `aten_unique_consecutive` to reshape `inverse` back to the input shape and handle rank-0, then keep the broad sampler; or (b) narrow the sampler to rank-1 inputs / add an `xfail` for `return_inverse` on multi-dim+scalar. Either gets CI green.

krajani21:
Since PR #2742 has been inactive, I'd like to pick this up and implement the fix suggested by @justinchuby.

PR #2745, opened 2025-12-23
Add github codespace support

oxygen-dioxide:
GitHub codespace is an online Linux environment with a vscode web UI. It is useful for developers who don't have a Linux device but want to test and debug their code on Linux.


justinchuby:
Is there a reason the config is needed? codespace can be launched directly I think.

oxygen-dioxide:
> Is there a reason the config is needed? codespace can be launched directly I think.

With this config, when a user launches codespace, required softwares and vscode extensions will be automatically installed. Users won't need to manually install them.

PR #2746, opened 2025-12-24
feat: modify aten_bilinear from einsum to matmul

### LINKED ISSUE #2573: Implement aten_bilinear
Implement the aten_binear function at https://github.com/microsoft/onnxscript/blob/e67eeefc8bc2b120bab79a8d04f303690ddc4bc0/onnxscript/function_libs/torch_lib/ops/core.py#L1192

Operator schema: `bilinear(Tensor input1, Tensor input2, Tensor weight, Tensor? bias=None) -> Tensor` 

Docs:

```
Applies a bilinear transformation to the incoming data: :math:`y = x_1^T A x_2 + b`.

    Args:
        in1_features: size of each first input sample, must be > 0
        in2_features: size of each second input sample, must be > 0
        out_features: size of each output sample, must be > 0
        bias: If set to ``False``, the layer will not learn an additive bias.
            Default: ``True``

    Shape:
        - Input1: :math:`(*, H_\text{in1})` where :math:`H_\text{in1}=\text{in1\_features}` and
          :math:`*` means any number of additional dimensions including none. All but the last dimension
          of the inputs should be the same.
        - Input2: :math:`(*, H_\text{in2})` where :math:`H_\text{in2}=\text{in2\_features}`.
        - Output: :math:`(*, H_\text{out})` where :math:`H_\text{out}=\text{out\_features}`
          and all but the last dimension are the same shape as the input.

    Attributes:
        weight: the learnable weights of the module of shape
            :math:`(\text{out\_features}, \text{in1\_features}, \text{in2\_features})`.
            The values are initialized from :math:`\mathcal{U}(-\sqrt{k}, \sqrt{k})`, where
            :math:`k = \frac{1}{\text{in1\_features}}`
        bias:   the learnable bias of the module of shape :math:`(\text{out\_features})`.
                If :attr:`bias` is ``True``, the values are initialized from
                :math:`\mathcal{U}(-\sqrt{k}, \sqrt{k})`, where
                :math:`k = \frac{1}{\text{in1\_features}}`
```

fw7th:
#### What this does
- Replaces Bilinear Transformation's einsum implementation with matmul

#### Why
- Performance increase from MatMul over Einsum

#### Testing
- Existing tests pass for function implementation

Resolves #2573

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -1208,29 +1208,32 @@ def aten_bilinear(
     # bias shape: (out_features) - optional
     # output shape: (..., out_features)
 
+    # input1 and input2 must have identical batch dimensions
     # Use MatMul to compute the bilinear transformation
-    batch_size = op.Shape(input1, start=0, end=1)
+    batch_size = op.Shape(input1, start=0, end=-1)
     input1_shape = op.Shape(input1, start=-1)
     input2_shape = op.Shape(input2, start=-1)
     output_shape = op.Shape(weight, start=0, end=1)
     neg_1 = op.Constant(value_ints=[-1])
 
-    # (O, H1, H2) -> (H1, O, H2) where O = output_shape, H1 = input1_shape, H2 = input2_shape
+    # (out_features, in1_features, in2_features) -> (in1_features, out_features, in2_features)
     W_permute = op.Transpose(weight, perm=[1, 0, 2])
 
-    # (H1, O, H2) -> (H1, O * H2)
+    # (in1_features, out_features, in2_features) -> (in1_features, out_features * in2_features)
     W_flat = op.Reshape(
         W_permute,
         op.Concat(input1_shape, op.Mul(output_shape, input2_shape), axis=0),
     )
 
-    # (B, H1) @ (H1, O*H2) -> (B, O*H2)
+    # (..., in1_features) @ (in1_features, out_features * in2_features) -> (..., out_features * in2_features)
     tmp = op.MatMul(input1, W_flat)
 
-    # (B, O*H2) -> (B, O, H2)
+    # (..., out_features * in2_features) -> (..., out_features, in2_features)
     tmp = op.Reshape(tmp, op.Concat(batch_size, output_shape, input2_shape, axis=0))
 
-    # (B, H2) -> (B, H2, 1) -> (B, O, H2) @ (B, H2, 1) -> (B, O, 1) -> (B, O)
+    # (..., in2_features) -> (..., in2_features, 1)
+    #   -> (..., out_features, in2_features) @ (..., in2_features, 1)
+    #   -> (..., out_features, 1) -> (..., out_features)
     result = op.Squeeze(op.MatMul(tmp, op.Unsqueeze(input2, neg_1)), neg_1)
 
     if bias is not None:


justinchuby (approved):
Thanks. @xadupre or @gramalingam for another review

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1208,12 +1208,31 @@ def aten_bilinear(
...
+    )
+
+    # (B, H1) @ (H1, O*H2) -> (B, O*H2)
+    tmp = op.MatMul(input1, W_flat)
+
+    # (B, O*H2) -> (B, O, H2)
+    tmp = op.Reshape(tmp, op.Concat(batch_size, output_shape, input2_shape, axis=0))
+
+    # (B, H2) -> (B, H2, 1) -> (B, O, H2) @ (B, H2, 1) -> (B, O, 1) -> (B, O)
+    result = op.Squeeze(op.MatMul(tmp, op.Unsqueeze(input2, neg_1)), neg_1)
```
justinchuby: Output has shape `(..., out_features)`. The final output shape should be `op.Concat(op.Shape(input1, start=0, end=-1), op.Shape(weight, start=0, end=1))`

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1208,12 +1208,31 @@ def aten_bilinear(
...
+    W_flat = op.Reshape(
+        W_permute,
+        op.Concat(input1_shape, op.Mul(output_shape, input2_shape), axis=0),
+    )
+
+    # (B, H1) @ (H1, O*H2) -> (B, O*H2)
+    tmp = op.MatMul(input1, W_flat)
+
+    # (B, O*H2) -> (B, O, H2)
+    tmp = op.Reshape(tmp, op.Concat(batch_size, output_shape, input2_shape, axis=0))
```
justinchuby, line 1232: Instead of using batch_size, I think you should use `op.Shape(input1, start=0, end=-1)` to get all of the leading shapes.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1208,12 +1208,31 @@ def aten_bilinear(
...
+    W_permute = op.Transpose(weight, perm=[1, 0, 2])
+
+    # (H1, O, H2) -> (H1, O * H2)
+    W_flat = op.Reshape(
+        W_permute,
+        op.Concat(input1_shape, op.Mul(output_shape, input2_shape), axis=0),
+    )
+
+    # (B, H1) @ (H1, O*H2) -> (B, O*H2)
+    tmp = op.MatMul(input1, W_flat)
```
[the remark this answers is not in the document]
fw7th: This is incorrect no? Since matmul would auto broadcast for trailing dimensions.

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -1208,12 +1208,31 @@ def aten_bilinear(
     # bias shape: (out_features) - optional
     # output shape: (..., out_features)
 
-    # Use Einsum to compute the bilinear transformation
-    # "...i,oij,...j->...o" means:
-    # - input1[..., i] * weight[o, i, j] * input2[..., j] -> output[..., o]
-    result = op.Einsum(input1, weight, input2, equation="...i,oij,...j->...o")
+    # Use MatMul to compute the bilinear transformation
+    batch_size = op.Shape(input1, start=0, end=1)
```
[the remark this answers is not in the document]
fw7th: I see, my mistake there, fully own it

onnxscript/function_libs/torch_lib/ops/core.py
fw7th: @justinchuby, forgive me if I'm incorrect, but the output should be: (..., out_features), I believe that's how it's implemented in pytorch, and your specifications.

onnxscript/function_libs/torch_lib/ops/core.py
fw7th: Yes, this was the fix, thank you Justin.

fw7th:
Hey @justinchuby! I fixed the batch-dim issue by switching to Shape(input1, start=0, end=-1) so all leading dims are preserved (the op requires identical batch shapes anyway).

Tests still intermittently fail due to the existing RNG seed issue (same as the upsample_bilinear2d tests), which seems unrelated to this change.

To better guard against regressions here, would you like me to add a test case for bilinear with batch_dim > 1? I can do that in a follow-up PR if that’s preferable.

onnxscript/function_libs/torch_lib/ops/core.py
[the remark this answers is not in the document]
fw7th: The batch dimensions from tmp (which comes from input1) and the batch dimensions from input2, are the same, this follows the implementation of the bilinear transformation.

fw7th: oh wait, we're both saying the same thing lmao

onnxscript/function_libs/torch_lib/ops/core.py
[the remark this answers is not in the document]
justinchuby: You are right. matmul should broadcast.

PR #2747, opened 2025-12-26
Fix unicode issue

gramalingam:
This copilot-suggested fix enabled converting a recent onnx model to onnxscript.

### DIFF
--- a/onnxscript/backend/onnx_export.py
+++ b/onnxscript/backend/onnx_export.py
@@ -394,7 +394,7 @@ class _Exporter:
                 attributes.append((at.name, at.ref_attr_name))
                 continue
             value = _attribute_value(at)
-            if isinstance(value, str, bytes):
+            if isinstance(value, (str, bytes)):
                 attributes.append((at.name, repr(value)))
                 continue
             if isinstance(value, np.ndarray):


onnxscript/backend/onnx_export.py
```diff
@@ -152,7 +152,10 @@ def _translate_value_info(value_info: ValueInfoProto) -> str:
 
 def _to_str(s):
     if isinstance(s, bytes):
-        return s.decode("utf-8")
+        try:
+            return s.decode("utf-8")
+        except UnicodeDecodeError:
+            return s.decode("latin1")  # or "cp1252" or other fallback
```
justinchuby: I have seen cases where the string attributes iare used to store binary data (e.g. in the custom Tokenizer op, even though this is invalid onnx). Is this trying to address those cases? If so, I would simply preserve the b-string to avoid accidental modifications.

Related: https://github.com/onnx/ir-py/pull/184

gramalingam: I don't know. All I know is that this enabled me to convert a recent model (Word Fluency) into onnxscript so that I could see the control-flow structure of the model in a compact readable format. I don't mind alternatives if they allow me to do this.

justinchuby: Right - that's the same model I looked at. Making a suggestion

justinchuby: Can you try the current change?

onnxscript/backend/onnx_export.py
```diff
@@ -391,7 +394,7 @@ def _translate_attributes(self, node):
                 attributes.append((at.name, at.ref_attr_name))
                 continue
             value = _attribute_value(at)
-            if isinstance(value, str):
+            if isinstance(value, str, bytes):
                 attributes.append((at.name, f"{value!r}"))
```
justinchuby:
```suggestion
                attributes.append((at.name, repr(value)))
```

onnxscript/backend/onnx_export.py
```diff
@@ -391,8 +394,8 @@ def _translate_attributes(self, node):
                 attributes.append((at.name, at.ref_attr_name))
                 continue
             value = _attribute_value(at)
-            if isinstance(value, str):
-                attributes.append((at.name, f"{value!r}"))
+            if isinstance(value, str, bytes):
```
justinchuby:
```suggestion
            if isinstance(value, (str, bytes)):
```

PR #2752, opened 2025-12-29
Cleanup SymbolValue

gramalingam:
More cleanup (in the migration of converter to onnx IR):

* Eliminate some redundant info/logic (class Dynamic)
* Cleanup AttrRef and SymbolValue.

### DIFF
--- a/onnxscript/_internal/converter.py
+++ b/onnxscript/_internal/converter.py
@@ -1059,7 +1059,7 @@ class Converter:
                     typeinfo = None
                 if typeinfo is not None:
                     set_type_info(t, typeinfo)
-                var = values.Dynamic(t, values.DynamicKind.Intermediate, info, typeinfo)
+                var = values.SymbolValue(t, info)
                 self._bind(lhs, var)
             elif isinstance(lhs, ast.Tuple):
                 # Assignments of the form "x, y, z = op.SomeOp(...)"
@@ -1082,9 +1082,7 @@ class Converter:
                 for x, output in zip(lhs.elts, outputs):
                     self._bind(
                         x.id,
-                        values.Dynamic(
-                            output, values.DynamicKind.Intermediate, self._source_of(x)
-                        ),
+                        values.SymbolValue(output, self._source_of(x)),
                     )
             else:
                 self.fail(lhs, f"Unsupported construct in LHS of assignment: '{type(lhs)!r}'")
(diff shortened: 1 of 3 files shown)


onnxscript/_internal/converter.py
```diff
@@ -354,7 +354,7 @@ def _to_onnx_attr_ref(
             msg = f"Unsupported attribute type {pytype!r}."
             fail(info.msg(msg) if info else msg)
         attr_type = ir.AttributeType(ta.pytype_to_attrtype(pytype))
-        return ir.Attr(attrname, attr_type, None, val.value)
+        return ir.Attr(attrname, attr_type, None, val.value.name)
```
justinchuby:
```suggestion
        return ir.Attr(attrname, attr_type, value=None, ref_attr_name=val.value.name)
```

would help readability

onnxscript/_internal/converter.py
```diff
@@ -536,7 +536,7 @@ def _translate_attr(
             val = self._lookup(expr.id, self._source_of(expr))
             if isinstance(val, values.AttrRef):
                 attr_type = ir.AttributeType(ta.pytype_to_attrtype(val.typeinfo))
-                attr_ref = ir.Attr(attr_name, attr_type, None, val.value)
+                attr_ref = ir.Attr(attr_name, attr_type, None, val.value.name)
```
justinchuby:
```suggestion
                attr_ref = ir.Attr(attr_name, attr_type, None, ref_attr_name=val.value.name)
```

onnxscript/_internal/converter.py
```diff
@@ -1117,6 +1131,7 @@ def ret(exp, i, suffix):
             preferred_name = f"return_val{suffix}"
             return_var = self._translate_expr(exp, preferred_name)  # TODO(rama)
             val = self._lookup(return_var.name, self._source_of(exp), False)
+            # TODO(rama): Can we avoid using val.kind here? And rely on the ir.Value methods like is_graph_input()?
```
gramalingam: @justinchuby : I was looking into eliminating and simplifying some of the extra information currently tracked by the converter. One of this is this values.DynamicKind ... I remember that in some other work I was doing earlier, I found out that methods like is_graph_input on ir.Value works fine when we deserialize a model. But when constructing/building an IR from scratch, it seemed trickier to use, since it was set at some later stage of the construction. So: when is it safe to use those methods? I guess if it has been added as an input to some graph, then it is guaranteed to return true?

justinchuby: As long as the value is added to the graph input/output list, the methods will work

PR #2753, opened 2025-12-31
[torchlib] Fix `aten__native_batch_norm_legit_functional`

justinchuby:
Fix aten__native_batch_norm_legit_functional where the running mean/var were returned without creating a new value, making the graph invalid.

Fixes https://github.com/pytorch/pytorch/issues/171471

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6893,7 +6893,7 @@ def _aten_native_batch_norm_inference_onnx(
     # https://github.com/pytorch/pytorch/blob/a44f8894fa6d973693aab44a3dda079a168b05c1/torch/_decomp/decompositions.py#L1475
     running_mean_fp32 = op.Cast(running_mean, to=FLOAT.dtype)
     invstd = op.Cast(invstd, to=FLOAT.dtype)
-    return norm, running_mean_fp32, invstd, running_mean, running_var
+    return norm, running_mean_fp32, invstd, op.Identity(running_mean), op.Identity(running_var)
 
 
 # TODO: This op is using duplicated code from aten_native_batch_norm,


onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -6893,7 +6893,7 @@ def _aten_native_batch_norm_inference_onnx(
     # https://github.com/pytorch/pytorch/blob/a44f8894fa6d973693aab44a3dda079a168b05c1/torch/_decomp/decompositions.py#L1475
     running_mean_fp32 = op.Cast(running_mean, to=FLOAT.dtype)
     invstd = op.Cast(invstd, to=FLOAT.dtype)
-    return norm, running_mean_fp32, invstd, running_mean, running_var
+    return norm, running_mean_fp32, invstd, op.Identity(running_mean), op.Identity(running_var)
```
gramalingam, line 6896: Do you know what was happening? I assume running_mean/var were valid values already ... but presumably some other requirement was violated (like an input-value cannot be output-value)?

In other words, what is the requirement for torchlib functions from a dev's perspective? Are they required to never return an input-value as an output-value without wrapping in an Identity? Seems like something that the underlying infrastructure could take care of without burdening the torchlib developer.

gramalingam (approved):

onnxscript/function_libs/torch_lib/ops/core.py
justinchuby: Here the running_mean and running_var are treated as mutable buffers in the pytorch model. It is an initializer directly as a graph output.

This only happens with the training graph. So we did not really see the case in our testing.

From torchlib's perspective, yes an input should not be returned directly as output. It is probably true that we can detect that externally and wrap the output with an identity.

PR #2768, opened 2026-01-13
Remove ParamSchema

justinchuby:
Remove ParamSchema and replace it with the new OpSignature.

## BC breaking

The `param_schemas()` methods are removed from Ops and ONNXFuntions.

### DIFF
--- a/onnxscript/_internal/param_manipulation.py
+++ b/onnxscript/_internal/param_manipulation.py
@@ -135,7 +135,7 @@ def tag_arguments_with_signature(
             if fill_defaults:
                 default_value = param.default
                 # Extract value from Attr object if it's an AttributeParameter
-                if isinstance(param, _schemas.AttributeParameter) and param.default:
+                if isinstance(param, _schemas.AttributeParameter):
                     default_value = param.default.value
                 tagged_kwargs[param.name] = (default_value, param)
         elif param.required:


onnxscript/_internal/values.py
```diff
@@ -147,130 +147,17 @@ def _prepare_inputs(self, _: onnx.defs.OpSchema, *inputs):
...
-            schemas.append(_param_schema_from_function_ir_input(arg))
-        elif isinstance(arg, ir.Attr):
-            # attr
-            schemas.append(_param_schema_from_function_ir_attr(arg))
-        else:
-            raise TypeError(f"Unknown input/attr type {type(arg)} from FunctionIR.")
-
-    return tuple(schemas)
-
-
```
gramalingam, line 273: I see that the class OnnxFunction still has both an `op_signature` and `op_schema` ... do we need both? If so, any explanation (of the relationship/difference between them etc.) would help.

gramalingam: I understand that function signatures from python can have more information than what an OpSchema in ONNX captures. I am trying to understand what we should compute and store. For example, I think perhaps there might be scope for cleaning up and unifying some of the (possibly duplicated) logic [here](https://github.com/microsoft/onnxscript/blob/3adee71b45f3e95e143c35e71fb71dfa002db730/onnxscript/_internal/converter.py#L1435). It could be done as a separate PR, just clarifying where my question is coming from

justinchuby: I agree schema is probably not needed for a function. I left those untouched to focus on the param schema first. I can sent a follow up to clean up the schema as well

onnxscript/_internal/evaluator.py
```diff
@@ -287,16 +288,16 @@ def eval_function(
         adapted_args: list[ExtendedModeValue] = []
         adapted_kwargs: dict[str, ExtendedModeValue] = {}
         has_array = False
-        for arg, param_schema in tagged_args:
-            if param_schema.is_input:
+        for arg, param in tagged_args:
+            if isinstance(param, _schemas.Parameter):
```
titaiwangms, line 292: I don't recall why we name this Parameter? Is it the same as Input? https://github.com/microsoft/onnxscript/blob/3adee71b45f3e95e143c35e71fb71dfa002db730/onnxscript/ir/_schemas.py#L103

onnxscript/_internal/param_manipulation_test.py
```diff
@@ -63,10 +63,27 @@ class TestSeparateInputAttributesFromArguments(unittest.TestCase):
...
+                _schemas.Parameter(
+                    name="a", type_constraint=type_constraint, required=True, variadic=False
+                ),
+                _schemas.AttributeParameter(
+                    name="b", type=_schemas.ir.AttributeType.INT, required=True, default=None
+                ),
+                _schemas.AttributeParameter(
+                    name="c",
+                    type=_schemas.ir.AttributeType.FLOAT,
+                    required=False,
```
titaiwangms, line 82: Looks like "required" is related to default. Should we set something like required has to be True when default is None in Parameter?

titaiwangms (approved):

onnxscript/_internal/param_manipulation_test.py
justinchuby: Default is None does not imply required=True. For example, Pad can have a padding attribute that, when not provided, defaults to 0. In the representation this is None (because it is data dependent (on input Rank))

onnxscript/_internal/evaluator.py
justinchuby: This is how onnx calls it https://onnx.ai/onnx/api/defs.html#onnx.defs.OpSchema.FormalParameter a "FormalParameter"

Issue #2736, opened 2025-12-17
[Optimizer] Fold Shape -> {Slice, Gather} -> Concat -> Reshape


justinchuby:
In `Shape -> {Slice, Gather} -> Concat -> Reshape`, when the dynamic axis coming from Shape is a single axis and the rest of the dimensions are not `-1`, we can replace the shape for Reshape to have a static `-1` axis, and eliminate the nodes for dynamic shape evaluation.

OYCN:
On our side, we handle the `Shape -> {Slice, Gather} -> {AnyOp} -> Concat -> Reshape` pattern using the following passes, as long as there is only one dynamic dimension.

One thing worth noting is that the `{AnyOp}` stage is often present in real-world models, so supporting it can significantly improve the coverage of this optimization.

```
PassManager:
  `const_folding & shape_inference` 
  `fuse_shape_gather`,
  `fuse_shape_slice`,
  `const_folding & shape_inference ` 
  `fuse_concat_reshape`,
```

PR #2769, opened 2026-01-15
Fix Issue #2736

P-Mahi10:
### Fix Issue #2736

This change extends constant folding for `Reshape` to eliminate dynamic shape
subgraphs of the form `Shape → Gather → Concat → Reshape` when the target shape
contains exactly one dynamic dimension. The optimizer rewrites the reshape to
use a static `-1` dimension, relying on ONNX Reshape semantics to infer the
dynamic value from the input element count. This removes runtime shape
evaluation and enables further canonicalization.

### **Key Changes**

- **`onnxscript/optimizer/_constant_folding.py`**  
  Updated the `reshape` optimizer to detect reshape targets with exactly one
  dynamic dimension and replace the dynamic shape computation with a static
  shape containing `-1`.

- **`onnxscript/optimizer/_constant_folding_test.py`**  
  Added a unit test to verify that dynamic shape subgraphs (`Shape`, `Gather`,
  `Concat`) are eliminated and that the reshape is either rewritten with `-1`
  or folded away entirely.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -484,6 +484,48 @@ def reshape(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
     if shape_value is None or input_shape is None:
         return _propagate_shape_value(node, op, state)
 
+    """Replace dynamic reshape shapes with a static -1 dimension when safe.
+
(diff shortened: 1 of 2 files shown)


P-Mahi10:

PR #2770, opened 2026-01-15
[torchlib] Fix irfft

simonbyrne:
This fixes the onnxscript export for the irfft function. 

Fixes https://github.com/onnx/onnx/issues/5920, and adds support to the changes in https://github.com/onnx/onnx/pull/7574 and https://github.com/microsoft/onnxruntime/pull/27028.

Most of the diff is due to the onnx_opset generated code changes from https://github.com/onnx/onnx/issues/5920. That can be removed if you would prefer.

### DIFF
--- a/tests/function_libs/torch_lib/ops_test_data.py
+++ b/tests/function_libs/torch_lib/ops_test_data.py
@@ -463,6 +463,9 @@ TESTED_TORCHLIB_OPS: tuple[TorchLibOpInfo, ...] = (
         fft_ops.aten__fft_c2r,
         tolerance={torch.complex64: (3e-3, 1.8e-4)},
         complex=True,
+    ).xfail(
+        matcher=lambda sample: True,
+        reason="Requires ONNX with IRFFT support (onesided=True, inverse=True)",
     ),
     TorchLibOpInfo(
         "ops.aten._fft_r2c",  # Custom from extra_opinfo


onnxscript/function_libs/torch_lib/ops/fft.py
```diff
@@ -123,7 +123,7 @@ def aten__fft_c2r(
         dft_length=last_dim_size,
         axis=dimension,
         inverse=True,
-        onesided=False,
+        onesided=True,
```
justinchuby: Is there a way to implement this with how the dft op is defined currently? I think the current implementation has issues with normalization, but I couldn't figure out what went wrong. You expertise is appreciated!

Given the onnx update is not merged and implemented yet, we are not able to incorporate the change to torchlib at the moment as we treat it as UB (unless the current behavior is actually correct in runtimes?).

simonbyrne: I think the current implementation may be UB itself: from what I can tell, it treats the `n//2+1` length matrix as if it was actually length `n`. Most of the time this means that you will be accessing unset memory and get small values (e.g. `1e-300`), so it is equivalent to roughly scaling most of the factors by 1/2 (except for the 0th frequency, so the offset remains the same).

simonbyrne: Actually, I take that back: since it passes the fft length, it won't be UB, but it still will zero pad (and hence give the wrong answer).

simonbyrne: It would be possible to implement the correct behavior given what we have currently (you would need to extend the array and reverse/conjugate the values), but it would be suboptimal. If the patch to ONNX is accepted, what would be the timeline to getting it updated here?

justinchuby: When the patch is accepted, depending on consensus it may or may not need to be in a new opset. If it is in the new opset, we need to update pytorch code. It will be accepted here if it is just a clarification of the old opset.

onnxscript/onnx_opset/_impl/opset1.py
justinchuby, line 1: Could you exclude changes to the onnx_opset/ directory?

onnxscript/function_libs/torch_lib/ops/fft.py
```diff
@@ -123,7 +123,7 @@ def aten__fft_c2r(
         dft_length=last_dim_size,
         axis=dimension,
         inverse=True,
```
justinchuby: Could you also help add support for len(dim) != 1 cases?

simonbyrne: `rfft` / `irfft` only support transforms along a single dimension (mathematically, a multidimensional FFT is equivalent to applying FFTs to each dimension sequentially: so you can do an RFFT for 1 dimension, but then you will have complex outputs, so the remainder would then need to be regular FFTs).

justinchuby: Yes. That makes sense. I think what you described is the pytorch behavior (creating a loop to run subsequent ifft on the rest of the dims will do)

onnxscript/function_libs/torch_lib/ops/fft.py
```diff
@@ -111,26 +109,34 @@ def aten__fft_c2r(
...
-        inverse=True,
-    )
+    for idx, dimension in enumerate(dim):
+        # Adjust dimension if we unsqueezed at the beginning
+        dimension += unsqueeze_first_dim
+
+        if idx < len(dim) - 1:
+            transformed = op.DFT(transformed, axis=dimension, inverse=True, onesided=False)
+        else:
+            # last operation is one-sided, transform to real
```
justinchuby, line 122: Should the one-sided inverse be performed first?

simonbyrne: No it has to be performed last (since it does the conversion back from complex to real).

simonbyrne: we do them in the opposite order of the forward operation.

PR #2771, opened 2026-01-15
Replace op_schema with op_signature

justinchuby:
This pull request refactors how operator schemas are handled throughout the autocast, converter, and evaluator modules. The main change is replacing direct usage of `OpSchema` with a new `_schemas.OpSignature` abstraction, leading to more consistent and modular code when dealing with operator signatures, especially for input casting and evaluation. Several related methods are renamed and refactored for clarity and encapsulation.

## Important changes

- The `Evaluator` interface now defines `eval_op` on onnx ops. The old `eval` was removed in favor of a more flexible `eval_op`. The exporter's eval will continue to function with a compatible logic in `class Op`
- `op_schema` properties from Functions are removed

**Operator signature abstraction and autocast refactor:**

* Replaced usage of `OpSchema` with `_schemas.OpSignature` in `onnxscript/_internal/autocast.py`, updating all relevant function signatures and internal logic to use the new abstraction. This includes changing how input parameters are filtered and type constraints are accessed. 

**AST Converter integration:**

* Updated the converter (`onnxscript/_internal/converter.py`) to pass `op_signature` instead of `op_schema` to autocast functions, ensuring compatibility with the new signature abstraction. 

**Evaluator refactor and encapsulation:**

* Refactored the evaluator (`onnxscript/_internal/evaluator.py`) to use `_adapt_inputs`, `_adapt_attributes`, and `_adapt_outputs` methods, encapsulating the logic for adapting inputs/outputs and removing unused or redundant methods. Now, operator signatures are consistently adapted from `OpSchema` using `_schemas.OpSignature`. 

Addtionally:

* Clean up typing annotation utilities
* Now supply IR attrs directly when creating attributes to avoid proto serialization and loss of subgraph references

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -37,27 +37,17 @@ include_patterns = [
     'onnxscript/**/*.pyi',
 ]
 exclude_patterns = [
-    'tests/**', # Skip linting test files for speed
-    # FIXME: Fix typing annotations in these files
-    'examples/custom_op_expansion.py',
-    'onnxscript/_internal/converter_test.py',
-    'onnxscript/_internal/converter.py',
-    'onnxscript/_internal/evaluator_test.py',
-    'onnxscript/_internal/evaluator.py',
-    'onnxscript/onnx_types.py',
-    'onnxscript/**/*_test.py', # Skip linting test files for speed
-    'onnxscript/function_libs/torch_lib/ops/**', # Operators typing do not play well with mypy
-    'onnxscript/optimizer/_legacy/evaluator.py',  # FIXME
-    'onnxscript/optimizer/_legacy/constant_folding.py',  # FIXME
-    'onnxscript/rewriter/onnxruntime/transformers/fastgelu.py',  # FIXME
-    'onnxscript/rewriter/onnxruntime/instance_to_group_normalization.py',  # FIXME
-    'onnxscript/rewriter/ort_fusions/models/*.py',  # onnxscript code
-    'onnxscript/rewriter/ort_fusions/models/_phi2lm.py',  # onnxscript code
-    'onnxscript/rewriter/ort_fusions/models/_phi4lm.py',  # onnxscript code
-    'onnxscript/rewriter/ort_fusions/_rotary_embedding_models.py',  # onnxscript code
-    'onnxscript/rewriter/onnxruntime/transformers/multihead_attention.py',  # FIXME
-    'onnxscript/tools/function_unittest_producer.py',  # FIXME
-    'onnxscript/rewriter/onnxruntime/transformers/layernorm.py',  # FIXME
+    # 'onnxscript/_internal/converter.py',
+    # 'onnxscript/_internal/evaluator.py',
+    # 'onnxscript/**/*_test.py', # Skip linting test files for speed
+    # 'onnxscript/**/tools/**/*.py',
+    # 'onnxscript/function_libs/torch_lib/ops/**', # Operators typing do not play well with mypy
+    # 'onnxscript/onnx_opset/**/*.py',  # auto-generated code
+    # 'onnxscript/onnx_types.py',
+    # 'onnxscript/rewriter/ort_fusions/_rotary_embedding_models.py',  # onnxscript code
+    # 'onnxscript/rewriter/ort_fusions/models/*.py',  # onnxscript code
+    # 'onnxscript/rewriter/models/*.py',  # onnxscript code
+    '**/*.py', # Temporarily disable mypy
 ]
 command = [
     'python',


justinchuby:
I thought it was but looks not. Will update

onnxscript/_internal/converter.py
```diff
@@ -1124,13 +1121,9 @@ def ret(exp, i, suffix):
...
-            if self.returntype is None:
-                t = None
-            else:
-                t = self.returntype[i]
-            self._current_fn.outputs.append(
-                make_value(return_var.name, t, self._source_of(stmt))
-            )
+            if self.returntype is not None:
+                set_type_info(return_var, self.returntype[i])
+            self._current_fn.outputs.append(return_var)
```
justinchuby, line 1112: Fixed a bug here

onnxscript/_internal/autocast.py
```diff
@@ -140,12 +126,15 @@ def cast_inputs(
...
-    if op_schema is None:
+    if op_signature is None:
         # Either an error or a custom op.
         # No checks/casts in this case.
         return tuple(cast(x, None) for x in args)
 
-    expected_inputs = op_schema.inputs
+    # Filter to get only input parameters (not AttributeParameters)
+    expected_inputs = [
+        param for param in op_signature.params if isinstance(param, _schemas.Parameter)
```
gramalingam: minor suggestion: I guess OpSignature could expose an inputs property that returns this value.

onnxscript/_internal/evaluator.py
```diff
@@ -123,20 +124,24 @@ def _unwrap_tensors_in_kwargs(kwargs: Mapping[str, Any]) -> dict[str, Any]:
...
-        """Evaluates an ONNX op.
+        """Evaluates an Op.
 
         Args:
-            schema: The OpSchema of the operator to evaluate.
-            inputs: The ONNX inputs to the op.
-            attributes: The ONNX attributes to the op.
+            op: The Op to evaluate.
+            args: The positional arguments to the op.
+            kwargs: The keyword arguments to the op.
```
gramalingam, line 144: Does this have a body (definition)? Looks like the next function eval_function comes right after this comment, unless the formatting is confusing me

onnxscript/_internal/type_annotation.py
```diff
@@ -215,82 +181,3 @@ def get_return_types(typeinfo: type | Sequence[type]) -> Sequence[type]:
     if typing.get_origin(typeinfo) is tuple:
         return typing.get_args(typeinfo)
     return (typeinfo,)
-
-
-def pytype_to_type_strings(pytype: TypeAnnotationValue) -> list[str]:
```
gramalingam, line 220: Are these unused? (I have a vague memory of something like this also being used in the so called "exporter" that converts an onnx model back to a string representation of an onnxscript program that produces that onnx model ... but looking at this function, this might not be the one.)

onnxscript/_internal/values.py
```diff
@@ -518,49 +380,60 @@ def _to_model_proto(
...
+                    ir_functions.append(func.function_ir)
+                else:
+                    raise TypeError(
+                        f"functions must be a sequence of "
+                        f"ir.Function, onnx.FunctionProto, or OnnxFunction, "
+                        f"not {type(func)!r}."
+                    )
+
+        # Duplicate the graph to create the model
+        main_graph = self.function_ir.graph.clone()
```
gramalingam, line 385: I wonder if this is necessary. I guess this is being used to handle the functions down below better. But, in practice, I don't think I have seen uses where functions are passed in, but we do call to_model_proto commonly.

Aren't we making the common case more expensive (cloning the graph before serializaing it instead of serializing it) in order to handle a usage I have not seen?

I guess two other options are: (a) Handle the special-case where no functions are specified more efficiently, or (b) Break compatibility and even drop the functions parameter or design a better API.

onnxscript/_internal/values.py
```diff
@@ -518,49 +380,60 @@ def _to_model_proto(
...
         # Determine opset imports
-        opsets = self.function_ir.graph.opset_imports
+        opset_imports = main_graph.opset_imports
+
+        for func in ir_functions:
+            domain = func.domain
+            if domain is not None and domain not in opset_imports:
+                opset_imports[domain] = func.meta.get("opset_version", 1)
```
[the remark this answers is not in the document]
justinchuby: No

onnxscript/_internal/values.py
justinchuby: to_model_proto is called very rarely, so I don't think this will be costly. I cloned because the graph was subsequently modified (with updated opset imports etc.)

I do agree we should do (b). But that can be independent of copying.

onnxscript/_internal/type_annotation.py
justinchuby: Yes. They are unused.

onnxscript/_internal/evaluator.py
justinchuby: The docstring is the body. This is a protocol so there is no other logic in the method

onnxscript/_internal/autocast.py
justinchuby: Done

onnxscript/_internal/values.py
gramalingam: I call to_model_proto all the time ... admittedly that is not production code, but creating test models for various testing purposes. But: I think it is better to get the API right before trying to fix any internal implementation issue, especially if it is going to affect the implementation choice. In this case, the API change can make the copy unnecessary. (If and when we expand the scope of onnxscript to define entire models, I think conversion to Model (proto or IR) will become more important.

justinchuby: Still, copying the IR object is cheap. I don't think it will make any material runtime difference but the duplication makes the process more robust.

justinchuby: Removed clone and the functions parameter.

PR #2775, opened 2026-01-21
Remove creation of duplicate value

gramalingam:
Remove creation of duplicate value when processing variables defined within a loop.


justinchuby:
Thanks!

PR #2776, opened 2026-01-21
Remove duplicated return_var

justinchuby:
A duplicated ir.Value for return_var was unintentionally generated.

### DIFF
--- a/onnxscript/_internal/converter.py
+++ b/onnxscript/_internal/converter.py
@@ -1112,8 +1112,8 @@ class Converter:
 
         def ret(exp, i, suffix):
             preferred_name = f"return_val{suffix}"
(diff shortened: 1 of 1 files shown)


onnxscript/_internal/converter.py
```diff
@@ -1112,8 +1112,8 @@ def check_num_outputs(n):
 
         def ret(exp, i, suffix):
             preferred_name = f"return_val{suffix}"
-            return_var = self._translate_expr(exp, preferred_name)  # TODO(rama)
```
justinchuby, line 1115: @gramalingam could you comment on what the todo was?

gramalingam: Good question! I don't remember right now. But I think it is ok to remove it

PR #2777, opened 2026-01-22
Fix loop body value creation

justinchuby:
Create and use ir.Value in loop subgraph inputs and bodies to avoid unintentional duplication of ir.Values (which led to invalid IR graphs).

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/prims.py
+++ b/onnxscript/function_libs/torch_lib/ops/prims.py
@@ -772,12 +772,18 @@ def prims_sub(self: TTensor, other: TTensor) -> TTensor:
     return op.Sub(self, other)
 
 
+@torch_op("prims::sum", trace_only=True)
 def prims_sum(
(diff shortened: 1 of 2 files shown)


onnxscript/_internal/converter.py
```diff
@@ -1220,8 +1224,16 @@ def _translate_loop_stmt(self, loop_stmt: ast.For | ast.While) -> None:
...
             python_loop_var_name = "infinite_loop"
             o_loop_bound = None
-            i_cond_var = ir.Value(name=test.id)  # TODO(Rama)
-            cond_while = ir.Value(name=test.id)  # TODO(Rama)
+            i_cond_var = make_value(
+                self.generate_unique_name(test.id),
+                onnx_types.BOOL,
+                self._source_of(loop_stmt),
+            )
+            cond_while = make_value(
```
justinchuby: @gramalingam could you double check if the initialization of cond_while is correct?

justinchuby: Apparently copilot suggested a fix

PR #2780, opened 2026-01-25
Optimize aten::min/max.dim with TopK op

### LINKED ISSUE #76344: [ONNX] Use topk to export max(keepdim) to onnx
### Model 1:

```python
class MinTest(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        top_probs, top = x.max(1, keepdim=True)
        return top_probs, top

```

image

### Model 2:

```python
class MinTest(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        top_probs, top = x.topk(1, dim=1)
        return top_probs, top
```

image

### Export:

``` python
m = MinTest().eval()

x = torch.rand(1,146,768)

torch.onnx.export(m, x, "/trt/output/min.onnx", verbose=True, input_names=['x'], output_names=["top_probs", "top"], opset_version=14)

```

There are two issues with exporting "Max(1,keep_dim=True)"

1. It is not efficient (to call both max and argmax when they could be combined)
2. onnxruntime CUDA EP does and extra copy to memory for argmax (opening another issue for onnxruntime, to link here)

### Versions

Collecting environment information...
PyTorch version: 1.10.2
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.2 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: version 3.21.0
Libc version: glibc-2.10

Python version: 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 06:08:21)  [GCC 9.4.0] (64-bit runtime)
Python platform: Linux-5.4.0-1074-azure-x86_64-with-debian-bullseye-sid
Is CUDA available: True
CUDA runtime version: 11.4.48
GPU models and configuration: GPU 0: Tesla V100-PCIE-16GB
Nvidia driver version: 470.103.01
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.2.2
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.2.2
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] numpy==1.21.6
[pip3] torch==1.10.2
[pip3] torchlars==0.1.2
[pip3] torchvision==0.11.3
[conda] blas                      1.0                         mkl
[conda] cudatoolkit               11.3.1               h2bc3f7f_2
[conda] ffmpeg                    4.3                  hf484d3e_0    pytorch
[conda] mkl                       2021.4.0           h8d4b97c_729    conda-forge
[conda] mkl-service               2.4.0            py37h402132d_0    conda-forge
[conda] mkl_fft                   1.3.1            py37h3e078e5_1    conda-forge
[conda] mkl_random                1.2.2            py37h219a48f_0    conda-forge
[conda] numpy                     1.21.6                   pypi_0    pypi
[conda] pytorch                   1.10.2          py3.7_cuda11.3_cudnn8.2.0_0    pytorch
[conda] pytorch-mutex             1.0                        cuda    pytorch
[conda] torchlars                 0.1.2                    pypi_0    pypi
[conda] torchvision               0.11.3               py37_cu113    pytorch

danielhumanmod:
Fix https://github.com/pytorch/pytorch/issues/76344

## Context

As mentioned in the issue, `torch.max(dim=...)` can be optimized with TopK to replace the current ReduceMax and ArgMax implementation. This optimization reduces redundant input scans and avoids potential performance overhead in certain execution providers (e.g., ONNX Runtime CUDA EP https://github.com/microsoft/onnxruntime/issues/11348).

In additional, given the `torch.min(dim=...)` has the similar pattern with max, I also apply this optimization to it.

## Verification
Successfully passed existing OpInfo consistency tests:
- pytest tests/function_libs/torch_lib/ops_test.py 
- pytest tests/function_libs/torch_lib/e2e_ops_tests.py


justinchuby (changes requested):
Thanks for creating the PR. Reading it again it seems like topk is more general than ReduceMax and ArgMax. From a node count perspective this may be fewer nodes, but I wonder if the original is easier to optimize with.

danielhumanmod:
Hey @justinchuby ,I’ve added a new rewrite rule to optimize this case based on our previous discussion. Whenever you have a moment, I’d appreciate your thoughts on it. Thanks!

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk_test.py
```diff
@@ -0,0 +1,543 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
danielhumanmod: Not sure what's wrong with this line, ` lintrunner -a` already passed locally

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,297 @@
...
+        - reduce_op_type: Property returning the name of the Reduce op (e.g., "ReduceMax")
+        - arg_op_type: Property returning the name of the Arg op (e.g., "ArgMax")
+        - largest: Property returning 1 for Max operations, 0 for Min operations
+    """
+
+    @property
+    @abstractmethod
+    def reduce_op_type(self) -> str:
+        """Return the name of the Reduce operation"""
+        ...
```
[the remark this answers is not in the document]
titaiwangms: I am not sure if an empty line would be more preferred: https://github.com/microsoft/onnxscript/blob/0206a98ec3f6326dd9fe42f20e40c423a09eb7ff/onnxscript/rewriter/rules/common/_fuse_batchnorm.py#L39

At lease codeQL is not happy about this. cc @justinchuby

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+from abc import abstractmethod
+
+import numpy as np
+import onnx_ir as ir
+
+from onnxscript.rewriter._basics import MatchResult
+from onnxscript.rewriter._rewrite_rule import RewriteRuleClassBase, RewriteRuleSet
+
+
+class FuseReduceArgToTopKBase(RewriteRuleClassBase):
```
titaiwangms:
```suggestion
class _FuseReduceArgToTopKBase(RewriteRuleClassBase):
```

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+            MatchResult: Success if the pattern can be fused, Failure otherwise.
+        """
+        del context
+        check_result = MatchResult()
+
+        reduce_node = reduce_val.producer()
+        arg_node = arg_idx.producer()
+
+        if reduce_node is None or arg_node is None:
+            return check_result.fail("Cannot find producer nodes.")
```
titaiwangms: Could this happen? I thought these two are originated from the pattern?

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # Step 1: Get keepdims attribute from both nodes
+        reduce_keepdims_attr = reduce_node.attributes.get("keepdims")
+        arg_keepdims_attr = arg_node.attributes.get("keepdims")
+
+        # ONNX default: keepdims = 1 for both Reduce and Arg operations
+        reduce_keepdims = (
+            reduce_keepdims_attr.as_int() if reduce_keepdims_attr is not None else 1
+        )
+        arg_keepdims = arg_keepdims_attr.as_int() if arg_keepdims_attr is not None else 1
```
titaiwangms: Formatting

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+        if reduce_node is None or arg_node is None:
+            return check_result.fail("Cannot find producer nodes.")
+
+        # Step 1: Get keepdims attribute from both nodes
+        reduce_keepdims_attr = reduce_node.attributes.get("keepdims")
+        arg_keepdims_attr = arg_node.attributes.get("keepdims")
+
+        # ONNX default: keepdims = 1 for both Reduce and Arg operations
+        reduce_keepdims = (
+            reduce_keepdims_attr.as_int() if reduce_keepdims_attr is not None else 1
```
titaiwangms: @justinchuby is there a more clean way to get the default? (just curious)

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+
+    @property
+    @abstractmethod
+    def reduce_op_type(self) -> str:
+        """Return the name of the Reduce operation"""
+        ...
+
+    @property
+    @abstractmethod
+    def arg_op_type(self) -> str:
```
titaiwangms: I wonder why not we just use node.op_type to get this property in the check? I think Node class is pretty well-defined already.

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+
+        # Step 2: Check if keepdims match
+        if reduce_keepdims != arg_keepdims:
+            return check_result.fail(
+                f"keepdims mismatch: {self.reduce_op_type} has {reduce_keepdims}, "
+                f"{self.arg_op_type} has {arg_keepdims}."
+            )
+
+        # Step 3: Get axes from Reduce operation
+        # In opset 18+, axes is an input; in opset 13-17, it's an attribute
```
titaiwangms: I wonder if we would be interested in only supporting opset 18+ here to reduce the complexity? (we have version converter) It's just the matter whether we see the rule will be applied standalone or not I guess?

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+        if select_last_index_attr is not None and select_last_index_attr.as_int() != 0:
+            return check_result.fail(
+                f"{self.arg_op_type} has select_last_index=1, which is not supported by TopK."
+            )
+
+        # Step 7: Normalize axes if rank is known (handle negative indices)
+        input_x = reduce_node.inputs[0]
+        rank = len(input_x.shape) if input_x.shape is not None else None
+
+        def normalize_axis(axis: int, rank: int | None) -> int:
```
titaiwangms: Moving this out and making it a private function would be preferrable.

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
```diff
@@ -0,0 +1,296 @@
...
+        # TopK always returns the first occurrence in case of ties
+        select_last_index_attr = arg_node.attributes.get("select_last_index")
+        if select_last_index_attr is not None and select_last_index_attr.as_int() != 0:
+            return check_result.fail(
+                f"{self.arg_op_type} has select_last_index=1, which is not supported by TopK."
+            )
+
+        # Step 7: Normalize axes if rank is known (handle negative indices)
+        input_x = reduce_node.inputs[0]
+        rank = len(input_x.shape) if input_x.shape is not None else None
```
titaiwangms: I wonder if symbolic shape could work on this case? @justinchuby

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
danielhumanmod: Thanks for the feedback, yah that would be a better & cleaner option

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
danielhumanmod: We can remove this, I am a little bit too cautious here

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
danielhumanmod: That makes sense to remove, I see this rule should be mostly used in pipeline, thanks for the suggestion!

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk_test.py
```diff
@@ -0,0 +1,543 @@
...
+        # In opset 18+, axes must be passed as the second input to ReduceMax
+        base_model = ir.from_onnx_text(f"""
+            < ir_version: 10, opset_import: ["" : 18] >
+            test_model (float[N, 32, 14, 17] X) => (float{output_shape_str} max_val, int64{output_shape_str} max_idx)
+            <int64[1] axes = {{{axis}}}>
+            {{
+                max_val = ReduceMax<keepdims={keepdims}>(X, axes)
+                max_idx = ArgMax<axis={axis}, keepdims={keepdims}>(X)
+            }}
+        """)
```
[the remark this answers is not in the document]
danielhumanmod: Added in latest comit, thanks AI

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
titaiwangms: @justinchuby What do you think?

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
justinchuby: I think there should be a get_int method

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
justinchuby: Only opset 18+ is fine

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
justinchuby: Can you elaborate?

onnxscript/rewriter/__init__.py
```diff
@@ -61,6 +62,7 @@
     *_fuse_pad_into_conv.rules,
     *_fuse_batchnorm.rules,
     *_remove_optional_bias.rules,
+    *_fuse_reduce_arg_to_topk.rules,
```
justinchuby:
```suggestion
```

Do not expose as default unless it is measured to be performant

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
titaiwangms: Skipping none of shape means this does not support dynamic at the moment. But symbolic inference should be able to handle the __eq__

onnxscript/rewriter/__init__.py
danielhumanmod: Cool, will remove. Also, curious about how we verify these rewrite rules - Do we have a specific benchmark suite or performance test

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
danielhumanmod: good point, updated in latest commit, thanks for the suggestions folks

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
danielhumanmod: Ohh actually it is a very good catch, I will use `shape.rank()` instead to ensure to support both static and symbolic shape, thanks a bunch!

danielhumanmod:
Hey team I solved all the pending comments, appreciate if you could take another look when you have time, thanks! cc @justinchuby @titaiwangms

onnxscript/rewriter/rules/common/_fuse_reduce_arg_to_topk.py
titaiwangms: Sorry @danielhumanmod Can you add a NOTE/comment somewhere that says the rule is only for opset 18+. Since now it's not for default rewrite rules, it could be used standalone for other users.

titaiwangms (approved):
Thank you. an unblocking comment

justinchuby (approved):
Unblocking but I would need to review it more closely next week. Thanks

danielhumanmod:
> Unblocking but I would need to review it more closely next week. Thanks

Hey @justinchuby gently ping this PR, appreciate if you could take another look when you have time, thanks!

justinchuby (approved):

justinchuby (commented):
Thanks for the work on this PR — the base/subclass architecture mirrors `_FuseBatchNormBase` cleanly, the test suite is thorough with good failure-reason assertions, and the decision to expose-but-not-default is architecturally sound. Here are the remaining issues that need attention before merge:

## 🔴 Blocking

### 1. Rules not exported from `common/__init__.py`
@justinchuby explicitly asked to *"expose the rule in `onnxscript/rewriter/rules/common/__init__.py`"* — but the PR diff adds only two files and makes no change to any `__init__.py`. The rules are unreachable without importing from the private `_fuse_reduce_arg_to_topk` module directly, which violates the project convention that `_*` modules are internal.

Please add to `common/__init__.py`:
```python
from onnxscript.rewriter.rules.common._fuse_reduce_arg_to_topk import (
    fuse_reduce_max_argmax_to_topk_rule,
    fuse_reduce_min_argmin_to_topk_rule,
)
```
(See naming note below on the `fuse_` prefix.)

### 2. Missing dtype guard — bfloat16 inputs produce invalid graphs
`ReduceMax`/`ReduceMin` support `bfloat16` in opset 18, but `TopK` does not. With no dtype check, the rule can fire on a bfloat16 model and emit an invalid `TopK` node that fails ONNX checker validation — a silent correctness regression.

Please add to `check()` after obtaining the input:
```python
_TOPK_SUPPORTED_DTYPES = frozenset({
    ir.DataType.FLOAT16, ir.DataType.FLOAT, ir.DataType.DOUBLE,
    ir.DataType.INT8, ir.DataType.INT16, ir.DataType.INT32, ir.DataType.INT64,
    ir.DataType.UINT8, ir.DataType.UINT16, ir.DataType.UINT32, ir.DataType.UINT64,
})
if x.dtype is not None and x.dtype not in _TOPK_SUPPORTED_DTYPES:
    return check_result.fail(f"Input dtype {x.dtype} is not supported by TopK")
```

## 🟠 High

### 3. Open maintainer thread — opset 18+ NOTE still unresolved
@titaiwangms has an open (unresolved) review thread requesting a visible NOTE that these rules are opset 18+ only, since they can be used standalone. The module docstring mentions it in the `Constraints:` block, but users who encounter `FuseReduceMaxArgMaxToTopK` or `FuseReduceMinArgMinToTopK` in isolation won't see it. Please add a one-liner to each public class docstring, e.g.:
```
# NOTE: Requires opset 18+. Apply a version converter before using on older models.
```
This will also allow @titaiwangms to resolve that thread.

### 4. No explicit opset guard — only coincidental protection
The opset 18+ constraint is enforced incidentally (pre-18 `ReduceMax` puts axes in an attribute rather than an input, so the pattern simply fails to match). A manually constructed pre-18 model with axes-as-input could bypass this. Consider an explicit check, or at minimum document the coincidental guard in a comment.

## 🟡 Medium

### 5. `except Exception` too broad
The `except Exception` block in `check()` masks real bugs (`MemoryError`, unexpected `AttributeError`, etc.). Please narrow to `except (ValueError, TypeError)`.

### 6. `_normalize_axis` defined but not used in `rewrite()`
`rewrite()` contains inline axis normalization logic that duplicates the `_normalize_axis()` static method on the same class. This is fragile — if either copy is updated, they silently diverge. Please have `rewrite()` call `self._normalize_axis()`.

### 7. Rule naming breaks `fuse_*_rule` convention
Every rule in `common/__init__.py` uses a verb-first name: `fuse_batchnorm_into_conv_rule`, `cast_cast_rule`, `collapse_slice_rule`. The new rules are noun-first (`reduce_max_argmax_to_topk_rule`), which breaks autocomplete and grep discoverability for users searching `fuse_*`. Please rename to `fuse_reduce_max_argmax_to_topk_rule` and `fuse_reduce_min_argmin_to_topk_rule`.

## 🔵 Minor

- **`select_last_index_attr` → `select_last_index`**: The `_attr` suffix implies an `ir.Attr` object; it holds an `int`. Should match the style of adjacent variables (`reduce_keepdims`, `arg_axis`, etc.).
- **Step-numbered comments**: `# Step 1` … `# Step 7` (and `# Step 2b`) describe *what* the next line does, not *why*. The `_fuse_batchnorm.py` reference has zero step comments. Keep only the comments that explain non-obvious reasons (e.g., `# ONNX default: axis=0 for ArgMax/ArgMin`).
- **`del context` should be `del context  # Unused`** — matches `_fuse_batchnorm.py` style.
- **`rng` property creates a new RNG on every access**: Properties are expected to return a consistent object. A module-level constant `_RNG = np.random.default_rng(20260127)` avoids the subtle behavior difference between `self.rng.shuffle(a); self.rng.shuffle(b)` vs `rng = self.rng; rng.shuffle(a); rng.shuffle(b)`.
- **`sorted=1` with `k=1` is redundant**: A single-element result is trivially sorted; omitting `sorted` (or using `sorted=False`) is cleaner.
- **`TestFuseReduceMaxArgMaxToTopK` missing class docstring** — the `Min` test class has one; the `Max` class doesn't.
- **Linting**: Several RUFF/format and EDITORCONFIG issues remain open from the automated scan. Please run `lintrunner -a` and commit the result.
- **Missing test coverage** (8 lines per Codecov): the pre-opset-18 fallback path and scalar-axes edge case are not tested.

---

Overall: the logic is correct and the architecture is clean. The two blocking issues (missing export and bfloat16 guard) are small, targeted fixes. Happy to re-review once those are addressed.

justinchuby (commented):
**Addendum — one additional medium-severity finding:**

### 8. Potential `TypeError` in `rewrite()` — `_normalize_axis` not called

`rewrite()` computes `x.shape.rank() + axis` directly when `axis < 0`:

```python
if axis < 0 and x.shape is not None:
    axis = x.shape.rank() + axis
```

If `x.shape` is not `None` but `x.shape.rank()` returns `None` (unknown-rank tensor with a non-`None` shape object), this crashes with a `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`. The `_normalize_axis()` static method on the same class already handles this case correctly by guarding on `rank is not None`. `rewrite()` should call `self._normalize_axis(axis, x.shape.rank() if x.shape is not None else None)` instead of duplicating the logic inline. This is both a DRY violation and a latent crash bug.

danielhumanmod:
Hey @justinchuby I solve the feedback above, appreciate another review when you have a chacne, thanks!

justinchuby:
The fusion rule is correct — `TopK(k=1, axis=dim, largest=1` for max / `0` for min) returns both values and indices matching `torch.min/max(dim)`; keepdim handled via Squeeze when `keepdims==0`; it correctly bails when `select_last_index==1` (TopK can't replicate last-index tie-breaking); indices are int64 and there's a dtype allowlist guard. Tests numerically validate both outputs across keepdims×axis plus negative/failure cases. The red `py311-torch-nightly` jobs fail on the same upstream torch-nightly breakage seen on main (not this PR). LGTM pending approval.

Issue #2714, opened 2025-11-21
Unintuitive behavior: Returned input aliases generate Identity nodes with generic names instead of preserving Python variable names


patmjen:
# Description
If a function returns a reassigned version of an input, the name of the output will get a generic name, i.e., `return_val` or `return_val_[index]` for multiple returns. This, to me, is unintuitive because it makes it harder to deduce the names of the generated outputs. 

Consider this example case:

```python
import onnx
from onnxscript import script, FLOAT
from onnxscript import opset20 as op

@script(default_opset=op)
def foo(x: FLOAT) -> FLOAT:
    y = x
    return y

onnx.checker.check_model(foo.to_model_proto())
print('Input:', [x.name for x in foo.to_model_proto().graph.input])
print('Output:', [x.name for x in foo.to_model_proto().graph.output])
```
This will print:
```
Input: ['x']
Output: ['return_val']
```
However, if I replace `y = x` with `y = x + 0` or `y = op.Identity(x)`, it will print `Output: ['y']`. I think it would be more intuitive if the generated output name was `y` since that matches the "intent" of the code. This will also match the behavior of using the `y = op.Identity(x)` which is what is "inserted" by the converter anyway.

# Cause
The reason this happens comes from the internal `ret` function in `Converter._translate_return_stmt`:
https://github.com/microsoft/onnxscript/blob/9dbf6858dbe01a10abda78312a4c2cfcd7e54a93/onnxscript/converter.py#L1070-L1088

In the case of `y = x`, `self._translate_expr(exp, preferred_name)` will correctly deduce that `y` just refers to `x` so `return_var` will be `'x'`. Therefore, `val.kind` will be an `Input` so a copy, i.e., an `Identity` op, will be emitted but it will use the generic `preferred_name` as the target.

# Possible fix
An "easy" fix in this case could be to check if `exp` is an `ast.Name` and then use `exp.id` as the suggested name to `emit_copy`. This would result in an `Identity` op being generated with `y` as the output, so the model output is as expected (i.e., it matches the "intent" of the code). Since `emit_copy` still calls `generate_unique_name`, it should still be safe.

I fully realize this may not cover all cases and that it might be a personal opinion that this is unintuitive. Also, this fix suggestion does not cover the case of `y = op.Identity(x); z = y; return z` where the "intent" may be to have an output with name `'z'`. So I would fully accept this not being changed.

[...]

PR #2788, opened 2026-02-02
Fix: preserve Python variable name for returned input aliases

Aniketsy:
#2714


onnxscript/_internal/converter.py
```diff
@@ -1101,11 +1101,14 @@ def ret(exp, i, suffix):
                 if val.value.is_graph_input():
                     # In ONNX, a graph-input cannot be an output of the graph.
                     # We need to insert a copy.
-                    return_var = self._emit_copy(return_var, preferred_name)
+                    suggested_name = preferred_name
+                    if isinstance(exp, ast.Name):
+                        suggested_name = exp.id
```
gramalingam, line 1113: A unit test would make it clear what this does. It sounds reasonable to me, but would be good to see it illustrated with an example. Specifically, if this is returning an input called X, the output cannot also be called X. Would it get renamed into something like "X_1" ?

justinchuby:
@Aniketsy could you address the comments?

Aniketsy:
> @Aniketsy could you address the comments?

yes sure, sorry i missed the review comment, i'll update with this thanks for ping

PR #2789, opened 2026-02-02
Fix _to_model_proto after protobuf update

xadupre:


onnxscript/_internal/values.py
```diff
@@ -449,7 +449,11 @@ def _to_model_proto(
                 output.type.CopyFrom(type.to_type_proto())
 
         for k, v in kwargs.items():
-            setattr(model_proto, k, v)
+            if k == "functions":
```
justinchuby, line 452: The functions argument was removed. Was it used anywhere?

xadupre: You can look into the unittest. ``def to_model_proto(self, **kwargs):``, the signature is not very explicit about what is allowed and what is not.

justinchuby: There isn’t a test failing? I thought I removed them all. We can update the signature to reflect the api change. (It was mentioned in the release notes)

PR #2791, opened 2026-02-04
Remove function value error in version converter

### LINKED ISSUE #2790: Version conversion failing due to functions in the model
After https://github.com/onnx/ir-py/pull/321, the onnx-ir nightly CIs in onnxscript starts to fail: https://github.com/microsoft/onnxscript/actions/runs/21643673831/job/62389942477, where the version converter starts to complain that there are non-inlined functions. We need to check

1. Are there unused functions, or are the functions ligitimate? (I assume there are unused ones)
2. We may need to also update the version converter to support functions because of https://github.com/pytorch/pytorch/pull/174283

## Example error

```pytb
.nox/test_onnx_ir_git/lib/python3.11/site-packages/onnx_ir/passes/_pass_infra.py:250: in call
    pass_result = pass_(model)
                  ^^^^^^^^^^^^
.nox/test_onnx_ir_git/lib/python3.11/site-packages/onnx_ir/passes/_pass_infra.py:127: in __call__
    result = self.call(model)
             ^^^^^^^^^^^^^^^^
onnxscript/version_converter/__init__.py:77: in call
    raise ValueError(
E   ValueError: The model contains functions. The version conversion pass does not support functions. Please use `common_passes.InlinePass` to inline the functions before applying this pass (_ConvertVersionPassRequiresInline).
```

titaiwangms:
Fix #2790 

This pull request makes a targeted change to the version converter in `onnxscript`. The main update removes the restriction that prevented models containing functions from being processed by the version conversion pass.

Version conversion support update:

* Removed the check that raised an error when the input model contained functions, allowing the version conversion pass to process such models without requiring prior inlining. (`onnxscript/version_converter/__init__.py`)

### DIFF
--- a/onnxscript/version_converter/__init__.py
+++ b/onnxscript/version_converter/__init__.py
@@ -38,7 +38,7 @@ class ConvertVersionPass(ir.passes.InPlacePass):
         self.target_version = target_version
         self.fallback = fallback
         self.convert_pass = ir.passes.Sequential(
-            _ConvertVersionPassRequiresInline(
+            _ConvertVersionPass(
                 target_version=target_version,
                 fallback=fallback,
             ),
@@ -51,7 +51,7 @@ class ConvertVersionPass(ir.passes.InPlacePass):
         return self.convert_pass(model)
 
 
-class _ConvertVersionPassRequiresInline(ir.passes.InPlacePass):
+class _ConvertVersionPass(ir.passes.InPlacePass):
     """Convert the model to the specified ONNX opset version.
 
     This pass leverages the onnxscript version converter to convert the model. If
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -274,10 +274,10 @@ class _VersionConverter:
         if attr.is_ref():
             return
         if attr.type == ir.AttributeType.GRAPH:
-            self.visit_graph(attr.as_graph())
+            self.visit_graph_or_function(attr.as_graph())
         elif attr.type == ir.AttributeType.GRAPHS:
             for graph in attr.as_graphs():
-                self.visit_graph(graph)
+                self.visit_graph_or_function(graph)
 
     def visit_node(
         self,
@@ -303,8 +303,8 @@ class _VersionConverter:
             self._default_metadata_merger.copy_merged_metadata([node], replacement.new_nodes)
             self.replace_node(node, replacement, root)
 
-    def visit_graph(self, graph: ir.Graph) -> None:
-        for node in graph:
+    def visit_graph_or_function(self, graph_or_function: ir.Graph | ir.Function) -> None:
+        for node in graph_or_function:
             if node.domain != "":
                 continue
             node_version = node.version or self._default_onnx_opset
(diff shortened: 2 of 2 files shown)


onnxscript/version_converter/__init__.py
```diff
@@ -73,12 +73,6 @@ def __init__(self, target_version: int, fallback: bool) -> None:
         self.fallback = fallback
 
     def call(self, model: ir.Model) -> ir.passes.PassResult:
-        if model.functions:
-            raise ValueError(
```
justinchuby, line 77: I think there are two issues: (1) we should run the (new) version converter on the functions too. and (2) we may probably still fail the pass if we fall back to the c++ version converter

titaiwangms: > (1) we should run the (new) version converter on the functions too

This is back to the priority concern. I don't see a lot of people asking about this, so I think we should just let it go for now.

> (2) we may probably still fail the pass if we fall back to the c++ version converter

How does one use C++ version converter? Would it still have issue if our current guarantee on exporting does not include passes?

justinchuby: (1) If that's the case, then failing may make sense. It is ok to run unused elimination otherwise. The issue with this is what do we do when we have exporter preserved functions in the model? For the new version converter I think it should be as easy as looping over the functions and run conversion on the function graphs.

(2) The c++ converter is triggered when it falls back. Could you elaborate on your question?

titaiwangms: My point is that (1) this assertion does not make sense when we decide to allow functions in our models, so we should remove it. But to support functions with all our passes, (2) we can wait.

justinchuby: Agreed

titaiwangms: https://github.com/microsoft/onnxscript/issues/2792

onnxscript/version_converter/__init__.py
titaiwangms, line 47: @justinchuby I think you were the one composing these passes. Does this still make sense to you? This is going to change the fact that the export step at version conversion no longer return an inline model. https://github.com/pytorch/pytorch/blob/fbf54b01bf1ee569e631311492b3dc28c8afc4df/torch/onnx/_internal/exporter/_core.py#L1564

justinchuby: That's ok? We will inline in the optimizer

onnxscript/version_converter/_version_converter.py
```diff
@@ -329,9 +329,34 @@ def visit_graph(self, graph: ir.Graph) -> None:
...
+                )
+            for from_version in range(node_version, self._target_version):
+                try:
+                    self.visit_node(node, function, from_version, up_conversion=True)
+                except VersionConverterError as e:
+                    logger.warning(
+                        "Skipping version conversion for node %s due to exception: %s",
+                        node.op_type,
+                        e,
+                    )
```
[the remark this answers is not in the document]
titaiwangms: It should be covered. A test is added.

onnxscript/version_converter/__init__.py
justinchuby: Looks like the name of the class can be updated

onnxscript/version_converter/_version_converter.py
```diff
@@ -329,9 +329,34 @@ def visit_graph(self, graph: ir.Graph) -> None:
                         e,
                     )
 
+    def visit_function(self, function: ir.Function) -> None:
```
justinchuby: Is this the exact same as visit_graph? Consider combining the logic?

titaiwangms: Is `graph_or_function` preferred in this regard? Having them in separate functions also makes sense.

justinchuby: No strong opinion, but graph_or_function works

justinchuby (approved):

gramalingam:
Unfortunately, there is a problem with generalizing the version-converter to work for functions. The issues is that for nodes inside a Function, there may be attributes that are attribute-references ... and if the version-converter logic needs the values of these attributes, then they may not be able to handle nodes where there are attribute-references instead of actual attribute-values.

At the least, we need logic in the version-converter (for any op, where we check an attribute-value) to catch the case where it is an attribute-reference and raise an NotImplemented error. For example, see the logic [here](https://github.com/microsoft/onnxscript/blob/2b2618e0820c9cff029953c2624f219dc4e3bade/onnxscript/version_converter/_version_converter.py#L160) ... here we should not use the default value of zero if it is an attribute-reference; instead we should copy the attribute-reference. Some adjustment to the logic is required, along with test-cases.

PR #2805, opened 2026-02-12
Add converter for `aten::_grouped_mm.default`

### LINKED ISSUE #2795: Missing converter for OpOverload(op='aten._grouped_mm', overload='default')
```python
import torch

class Model(torch.nn.Module):
    def forward(self, a, b):
        return torch.nn.functional.grouped_mm(
            a.to(torch.bfloat16), b.transpose(-2, -1).to(torch.bfloat16)
        ).to(torch.float32)

G, M, N, K = 4, 16, 32, 64
a = torch.randn(G, M, K, device="cpu", dtype=torch.float32)
b = torch.randn(G, N, K, device="cpu", dtype=torch.float32)
model = Model()
expected = model(a, b)
ep = torch.export.export(model, (a, b))
names = [str(n.target) for n in ep.graph.nodes]
assert "aten._grouped_mm.default" in names
epo = torch.onnx.export(
    model, (a, b), dynamic_shapes=({0: "G", 1: "M", 2: "K"}, {0: "G", 1: "N", 2: "K"})
)
```

```
<class 'torch.onnx._internal.exporter._errors.DispatchError'>: No ONNX function found for <OpOverload(op='aten._grouped_mm', overload='default')>. Failure message: No decompositions registered for the real-valued input
```

Issue met with ``microsoft/OptiMind-SFT``.

Sid-V5:
Implemented the converter for `aten::_grouped_mm.default` to address #2795.

## Changes

- Added `aten_grouped_mm` function in `onnxscript/function_libs/torch_lib/ops/core.py`

## Implementation Details

The `aten::_grouped_mm` operator performs grouped matrix multiplication. This implementation handles the **batch/dense mode** (when `offs` is `None`), where groups are implicit in the batch dimension:

- `self: (G, M, K)`, `mat2: (G, K, N)` → result: `(G, M, N)`
- Uses `op.MatMul` for the core computation
- Supports optional `bias` addition via `op.Add`
- Supports optional `out_dtype` casting via `op.Cast`

The **offset-based mode** (when `offs` is provided) raises `NotImplementedError`, as it requires segment-level matrix multiplications that are not directly expressible with standard ONNX operators.

## Testing

The function follows the same patterns as other converters in `core.py` (e.g., `aten_bmm`, `aten_mm`) and uses the `@torch_op` decorator for automatic registration.

Fixes #2795


Sid-V5:

onnxscript/function_libs/torch_lib/ops/core.py
```diff
@@ -4503,6 +4503,27 @@ def aten_grid_sampler_3d_backward(
...
+    mat2: TFloat,
+    offs: Optional[TInt] = None,
+    bias: Optional[TFloat] = None,
+    out_dtype: Optional[int] = None,
+) -> TFloat:
+    """_grouped_mm(Tensor self, Tensor mat2, *, Tensor? offs=None, Tensor? bias=None, int? out_dtype=None) -> Tensor"""
+
+    # If offs is None, it uses the "dense" / "batch" mode where groups are implicit in the batch dimension.
+    # self: (G, M, K), mat2: (G, K, N) -> (G, M, N)
+    # TODO: Implement sparse mode when offs is not None.
```
justinchuby: Could you raise a not implemented error?

tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -3144,4 +3172,12 @@ def __init__(self):
         sample_inputs_func=sample_inputs_roi_pool,
         supports_out=False,
     ),
+    opinfo_core.OpInfo(
+        "ops.aten._grouped_mm",
+        aten_name="_grouped_mm",
+        op=_mock_grouped_mm,
```
justinchuby: Why not use 
```suggestion
        op=torch.ops.aten._grouped_mm,
```

?

justinchuby:
Two items. (1) The CONFLICTING flag is trivial — a `git merge-tree` against main resolves cleanly (insertions into non-overlapping regions), so a rebase clears it. (2) Real correctness issue: the `offs` argument is silently ignored — the code has only a `# TODO` and falls through to `MatMul` (~core.py L4515-4523), but the PR description says offset mode "raises NotImplementedError." With `offs` provided (the ragged/MoE case) this emits silently-incorrect results instead of failing fast; please add an explicit `if offs is not None: raise NotImplementedError(...)`. Also, the OpInfo currently compares against a hand-written `torch.matmul`-based mock rather than the real `torch._grouped_mm`, so it's near-tautological — worth strengthening toward real-op parity and covering bias / out_dtype.

justinchuby:
@Sid-V5 would you like to update this PR?

Sid-V5:
@justinchuby 
Rebased onto main to resolve the conflict. 
updated core.py to raise NotImplementedError if offs is not None, and updated the OpInfo tests to cover bias/out_dtype and use the real torch._grouped_mm when available

gramalingam (approved):

justinchuby:
Could you fix lint?

tests/function_libs/torch_lib/extra_opinfo.py
justinchuby: Also this

justinchuby: @Sid-V5

justinchuby:
@Sid-V5 just my inline reviews and it should be ready to merge after that

justinchuby (approved):

tests/function_libs/torch_lib/extra_opinfo.py
```diff
@@ -3211,4 +3266,12 @@ def sample_inputs_masked_scatter(op_info, device, dtype, requires_grad, **kwargs
         sample_inputs_func=sample_inputs_masked_scatter,
         supports_out=False,
     ),
+    opinfo_core.OpInfo(
+        "ops.aten._grouped_mm",
+        aten_name="_grouped_mm",
+        op=_mock_grouped_mm,
```
justinchuby: As commented, please change this to 
```suggestion
        op=torch.ops.aten._grouped_mm,
```

Then in TorchLibOpInfo, you may skip the test if `not hasattr(torch.ops.aten, "_grouped_mm")`

justinchuby (changes requested):
Nit as commented

justinchuby (approved):

justinchuby:
Could you fix lint? Thanks

justinchuby (approved):

PR #2806, opened 2026-02-12
Raise version converter error when function attribute is RefAttr

titaiwangms:
Regarding the concern from https://github.com/microsoft/onnxscript/pull/2791#issuecomment-3887765687, this PR specifically raises when function has RefAttr.

---
This pull request adds a new validation to the version converter to ensure nodes with reference attributes (`RefAttr`) are not supported, and introduces a corresponding unit test to verify that the version converter raises an error in such cases.

Validation for unsupported reference attributes:

* Added a check in `visit_graph_or_function` in `onnxscript/version_converter/_version_converter.py` to raise a `VersionConverterError` if any node has a reference attribute (`RefAttr`), as these are not currently supported by the version converter.

Testing and error handling:

* Added a new unit test `test_version_convert_raises_on_function_node_with_ref_attribute` in `onnxscript/version_converter/_version_converter_test.py` to ensure that converting a model containing a function node with a reference attribute raises the appropriate error and includes a clear message.

### DIFF
--- a/onnxscript/version_converter/_version_converter_test.py
+++ b/onnxscript/version_converter/_version_converter_test.py
@@ -506,14 +506,11 @@ class VersionConverterMetadataMergeTest(unittest.TestCase):
         )
 
         target_version = 20
-        with self.assertRaises(
+        with self.assertRaisesRegex(
             version_converter._version_converter.VersionConverterError,  # pylint: disable=protected-access
-        ) as ctx:
-            version_converter.convert_version(model, target_version=target_version)
-        self.assertRegex(
-            str(ctx.exception),
             "has ref attribute, which is not supported by version converter",
-        )
+        ):
+            version_converter.convert_version(model, target_version=target_version)
 
 
 class VersionConverter25to26Test(unittest.TestCase):


onnxscript/version_converter/_version_converter.py
```diff
@@ -310,6 +310,11 @@ def visit_graph_or_function(self, graph_or_function: ir.Graph | ir.Function) ->
             node_version = node.version or self._default_onnx_opset
             if node_version is None:
                 raise VersionConverterError(f"Node {node} has no version.")
+            # RefAttr is not supported by adapters for now.
+            if any(attr.is_ref() for attr in node.attributes.values()):
+                raise VersionConverterError(
+                    f"Node {node} has ref attribute, which is not supported by version converter."
```
justinchuby:
```suggestion
                    f"Node '{node!r}' has ref attribute, which is not supported by version converter."
```

nit

onnxscript/version_converter/_version_converter_test.py
```diff
@@ -455,6 +448,73 @@ def test_metadata_is_copied_to_multiple_replacement_nodes(self):
...
+        )
+
+        target_version = 20
+        with self.assertRaises(
+            (
+                version_converter._version_converter.VersionConverterError,  # pylint: disable=protected-access
+                ir.passes.PassError,
+            )
+        ) as ctx:
+            version_converter.convert_version(model, target_version=target_version)
```
justinchuby: It is an issue that PassError is raised. I realized this in the shape inferencer as well. I think we should find a way to move the main logic out of the pass, and make the pass a simple wrapper around the logic so VersionConverterError can be properly raised by `convert_version`

onnxscript/version_converter/_version_converter_test.py
```diff
@@ -455,6 +448,73 @@ def test_metadata_is_copied_to_multiple_replacement_nodes(self):
...
+        # Check the error message, unwrapping PassError if needed
+        error = ctx.exception
+        if isinstance(error, ir.passes.PassError) and error.__cause__ is not None:
```
justinchuby: We should use assertRegex, after the error is properly raised

titaiwangms: Done

onnxscript/version_converter/_version_converter_test.py
titaiwangms: I moved it out. We should improve the ability of error expression in sequential if we still want to rely on it.

onnxscript/version_converter/_version_converter_test.py
```diff
@@ -455,6 +455,66 @@ def test_metadata_is_copied_to_multiple_replacement_nodes(self):
...
+            opset_imports={"": 18, "pkg.custom": 1},
+        )
+        model = ir.Model(
+            main_graph,
+            ir_version=8,
+            functions=[function],
+        )
+
+        target_version = 20
+        with self.assertRaises(
```
justinchuby:
```suggestion
        with self.assertRaisesRegex(
```

onnxscript/version_converter/_version_converter_test.py
```diff
@@ -455,6 +455,66 @@ def test_metadata_is_copied_to_multiple_replacement_nodes(self):
...
+            version_converter._version_converter.VersionConverterError,  # pylint: disable=protected-access
+        ) as ctx:
+            version_converter.convert_version(model, target_version=target_version)
+        self.assertRegex(
+            str(ctx.exception),
+            "has ref attribute, which is not supported by version converter",
+        )
```
justinchuby:
```suggestion
```

onnxscript/version_converter/_version_converter_test.py
justinchuby: + "has ref attribute, which is not supported by version converter"

onnxscript/version_converter/__init__.py
```diff
@@ -37,18 +37,26 @@ def __init__(self, target_version: int, fallback: bool = False) -> None:
...
             common_passes.RemoveUnusedFunctionsPass(),
             common_passes.RemoveUnusedOpsetsPass(),
         )
 
     def call(self, model: ir.Model) -> ir.passes.PassResult:
-        return self.convert_pass(model)
+        # Run the conversion pass outside of Sequential so that errors
+        # (e.g. VersionConverterError) propagate directly without being
+        # wrapped in PassError.
+        result = self._convert_pass(model)
```
titaiwangms, line 54: Do I need to pass result for InplacePass?

justinchuby: I don't think so. Either way is fine

PR #2807, opened 2026-02-13
Improve Reshape and Slice folding

justinchuby:
- Fold the input to reshape if from the output shape we know it can be folded.
- Eliminate no-op slice if the input shape is the same as output shape.


xadupre (approved):

onnxscript/rewriter/rules/common/_materialize_reshape_shape.py
```diff
@@ -0,0 +1,65 @@
...
+        output = context.output_values[0]
+        if output.shape is None:
+            return check_result.fail("Output shape is not known.")
+
+        dims = list(output.shape)
+        sym_count = sum(1 for d in dims if not isinstance(d, int))
+
+        if sym_count == 0:
+            self._new_dims = [int(d) for d in dims]
+        elif sym_count == 1:
```
gramalingam: Nit: We can omit the first branch and generalize this condition to `sym_count <= 1` since the logic below handles both cases correctly.

onnxscript/rewriter/rules/common/_materialize_reshape_shape.py
```diff
@@ -0,0 +1,65 @@
...
+            self._new_dims = [-1 if not isinstance(d, int) else int(d) for d in dims]
+        else:
+            return check_result.fail(
+                f"Output shape has {sym_count} symbolic dims, cannot materialize."
+            )
+
+        # Preserve allowzero attribute from original node
+        self._allowzero = context.nodes[0].attributes.get_int("allowzero", 0)
```
gramalingam: This may not be correct. I think we may need to make allowzero=1 in the rewritten node

gramalingam: Let me rephrase. May be it is correct. But verifying it is correct seems less trivial. It seems easier if allowzero=1.

justinchuby: I thought about this and thought the current is correct. Let me look at this again

onnxscript/rewriter/rules/common/_materialize_reshape_shape.py
justinchuby: Fixed

onnxscript/rewriter/rules/common/_materialize_reshape_shape.py
justinchuby: Set to allowzero=1

gramalingam (approved):

PR #2799, opened 2026-02-10
Fix version converter regression

titaiwangms:
This pull request improves the ONNX version converter by enabling direct opset version conversion for models containing functions, and ensures that version conversion now correctly applies to function nodes and their subgraphs. The changes also update the internal APIs to handle both models and functions, and expand test coverage for these scenarios.

Enhancements to version conversion logic:

* Removed the requirement to inline functions before version conversion, allowing `_ConvertVersionPass` to process models with functions directly (`onnxscript/version_converter/__init__.py`). [[1]](diffhunk://#diff-e2693e79e7e645175e987e44ba3432063a21003bf4aa50119c1a786925f6e042L41-R41) [[2]](diffhunk://#diff-e2693e79e7e645175e987e44ba3432063a21003bf4aa50119c1a786925f6e042L55-R54) [[3]](diffhunk://#diff-e2693e79e7e645175e987e44ba3432063a21003bf4aa50119c1a786925f6e042L76-L81)
* Updated the internal API (`_set_onnx_opset_version`) to support both models and functions, so opset imports are set correctly for functions as well as models (`onnxscript/version_converter/_version_converter.py`).

Support for function and subgraph conversion:

* Modified traversal logic to process nodes inside functions and subgraphs within control flow nodes, ensuring that all relevant nodes are version-converted (`onnxscript/version_converter/_version_converter.py`). [[1]](diffhunk://#diff-b6c70f90bafaee79b30e43c90bc0fd5192fb3de7ccc4cf9d48a209798dd775faL277-R280) [[2]](diffhunk://#diff-b6c70f90bafaee79b30e43c90bc0fd5192fb3de7ccc4cf9d48a209798dd775faL306-R307) [[3]](diffhunk://#diff-b6c70f90bafaee79b30e43c90bc0fd5192fb3de7ccc4cf9d48a209798dd775faL324-R324) [[4]](diffhunk://#diff-b6c70f90bafaee79b30e43c90bc0fd5192fb3de7ccc4cf9d48a209798dd775faL334-R337)

Expanded test coverage:

* Added tests to verify that nodes inside functions and subgraphs within control flow nodes (e.g., `If` branches) are correctly version-converted, and that opset imports are updated for functions (`onnxscript/version_converter/_version_converter_test.py`).

PR #2813, opened 2026-02-17
Add back inliner to version converter

titaiwangms:
Basically, version converter only supports inlined model. In terms of torch onnx exporter, we should just remind/warn users that if their models includes local-function. We just don't support it.

series of version converter PRs:
https://github.com/microsoft/onnxscript/pull/2791 (Removed assertion of no local function included)
https://github.com/microsoft/onnxscript/pull/2799 (Missing opset import)
https://github.com/microsoft/onnxscript/pull/2806 (Raise for RefAttr)

Maybe we should just try revert https://github.com/microsoft/onnxscript/pull/2791?

cc @gramalingam @justinchuby

### DIFF
--- a/onnxscript/version_converter/__init__.py
+++ b/onnxscript/version_converter/__init__.py
@@ -37,7 +37,9 @@ class ConvertVersionPass(ir.passes.InPlacePass):
         super().__init__()
         self.target_version = target_version
         self.fallback = fallback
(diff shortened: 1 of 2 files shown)


titaiwangms:
I will just revert the commits

justinchuby (approved):

PR #2814, opened 2026-02-17
First version of GraphBuilder

gramalingam:
First version of GraphBuilder.

TODO later:
* Integrate schema-based separation of inputs and attributes
* Align behaviors of script-mode and the new trace-mode as much as possible
  - With respect to conversion of python constants into ONNX tensors
  - With respect to conversion of python values into ONNX attributes
  - With respect to support for special kwargs like _domain and _version (to be added to script mode)
  - Consider supporting op.builder and op.builder.opset(...) in script mode
* Integrate calls to script-mode functions from trace-mode (and vice-versa)

### DIFF
--- a/onnxscript/_internal/builder.py
+++ b/onnxscript/_internal/builder.py
@@ -249,7 +249,7 @@ class GraphBuilder:
         for i, x in enumerate(inputs):
             if i < len(expected_inputs):
                 expected = expected_inputs[i]
-            elif (
+            elif expected_inputs and (
                 expected_inputs[-1].option == onnx.defs.OpSchema.FormalParameterOption.Variadic
             ):
                 expected = expected_inputs[-1]


onnxscript/_internal/_inference.py
```diff
@@ -0,0 +1,103 @@
...
+            #     "This may lead to incorrect constant folding.",
+            #     val.name,
+            # )
+            return None
+        assert isinstance(array, np.ndarray)
+        return array
+    return None
+
+
+def _do_onnx_inference(node: ir.Node) -> None:
```
justinchuby, line 54: Add a brief docstring for this function?

justinchuby, line 54: Looks like this can be reused by the optimzer?

onnxscript/_internal/_inference.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def infer_outputs(node: ir.Node) -> None:
+    try:
+        _do_onnx_inference(node)
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        # Inference is best-effort: failures (missing schema, unknown types,
+        # serialization errors, etc.) are recorded but must not crash the builder.
+        # TODO: compose with any existing error
+        node.metadata_props["inference_error"] = str(e)
```
justinchuby: I would assign a scoped key e.g.
```suggestion
        node.metadata_props["pkg.onnxscript.shape_inference_error"] = str(e)
```

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+
+from typing import Any, Callable, Sequence
+
+import onnx
+import onnx_ir as ir
+
+import onnxscript._internal._inference as inference
+import onnxscript.optimizer
+
+_DTYPE_SUFFIX_MAP: dict[ir.DataType, str] = {
```
justinchuby: Use `dtype.short_name()`? https://onnx.ai/ir-py/api/generated/onnx_ir.DataType.html#onnx_ir.DataType.short_name

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        return self._graph
+
+    def initializer(
+        self, tensor: ir.TensorProtocol, name: str | None = None, *, qualify: bool = True
+    ) -> ir.Value:
+        if name is None:
+            name = tensor.name
+        if qualify:
+            name = self.qualify_name(name)
+        tensor.name = name
```
justinchuby: It's ok to leave the tensor name untouched because the serializer only looks at the value name

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        shape = ir.Shape((d if isinstance(d, int) else d.value) for d in tensor.shape.dims)
```
justinchuby:
```suggestion
        shape = ir.Shape(tensor.shape)
```

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        return value
+
+    @staticmethod
+    def _type_suffix(element_type: type) -> str:
+        """Return a short type suffix for naming constants based on Python type."""
+        if element_type is int:
+            return "i64"
+        if element_type is float:
+            return "f32"
+        return ""
```
justinchuby: I would avoid static methods. Instead, create helper functions in the module level: https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+
+    @staticmethod
+    def _dtype_suffix(dtype: ir.DataType) -> str:
+        """Return a short type suffix for naming constants based on ir.DataType."""
+        return _DTYPE_SUFFIX_MAP.get(dtype, dtype.name.lower())
+
+    @staticmethod
```
justinchuby: Static methods -> module level functions

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        if isinstance(value, (int, float, bool, str)):
+            return f"const_{value}_{type_suffix}" if type_suffix else f"const_{value}"
+        # Sequence: use up to 2 elements in the name
+        if len(value) <= 2:
+            vals = ",".join(str(v) for v in value)
+        else:
+            vals = f"{value[0]},{value[1]},..."
+        return f"const_[{vals}]_{type_suffix}" if type_suffix else f"const_[{vals}]"
+
+    def _input_to_ir_value(self, value, like_type: ir.Value | None = None) -> ir.Value:
```
justinchuby: Add a brief docstring?

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        if version is not None:
+            try:
+                return onnx.defs.get_schema(op_type, version, domain)
+            except onnx.defs.SchemaError:
+                pass
+        return None
+
+    def _partition_inputs_attributes(
+        self,
+        schema: onnx.defs.OpSchema | None,
```
justinchuby, line 217: I would use the op signature instead because they type representation is easier to work with

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+    def context_name(self) -> str:
+        return self._context_stack[-1] if self._context_stack else ""
+
+    def qualify_name(self, name: str) -> str:
+        """Prepend the current hierarchical context prefix to the given name."""
+        prefix = self.context_name()
+        return f"{prefix}.{name}" if prefix else name
+
+
+class OpBuilder:
```
justinchuby, line 354: Create a brief docstring?

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        if "_domain" not in kwargs:
+            kwargs["_domain"] = self._domain
+        if self._version is not None and "_version" not in kwargs:
+            kwargs["_version"] = self._version
+        return self._builder.call_op(op_type, inputs, kwargs)
+
+    def __getattr__(self, op_type: str) -> Callable:
+        return lambda *args, **kwargs: self._call_op(op_type, args, kwargs)
+
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
```
justinchuby, line 378: In the future it would be good to code gen either sub classes or pyi files so we get a good intellicode experience

justinchuby:
Adding to the list: it would be nice to be able to call an existing model graph/function as an operator.

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+
+        # Context stack to manage hierarchical naming. Each module/layer can push a new context, and pop it when done.
+        # The current context is used as a prefix for naming values and nodes.
+        # This allows us to generate names like "layer1.attention.query"
+        self._context_stack: list[str] = [""]
+
+        # Cache for constant initializers (scalars and sequences), keyed by (value, dtype).
+        # This avoids creating duplicate initializers for the same constant
+        # and allows sharing them across different layers/contexts.
+        self._constant_cache: dict[tuple[Any, ir.DataType | None], ir.Value] = {}
```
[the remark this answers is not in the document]
gramalingam: I think the current solution is acceptable.

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+    def push_module(self, module: str) -> None:
+        current = self.context_name()
+        if module:
+            new_context = f"{current}.{module}" if current else module
+        else:
+            new_context = current
+        self._context_stack.append(new_context)
+
+    def pop_module(self) -> None:
+        self._context_stack.pop()
```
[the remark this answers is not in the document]
gramalingam: Done.

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+        return self._builder.call_op(op_type, inputs, kwargs)
+
+    def __getattr__(self, op_type: str) -> Callable:
+        return lambda *args, **kwargs: self._call_op(op_type, args, kwargs)
+
+    def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
+        return self._builder.initializer(tensor, name)
+
+    def call(self, function, *args, **kwargs):
+        return self._builder.call(function, *args, **kwargs)
```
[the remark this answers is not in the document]
gramalingam: Done

onnxscript/_internal/builder.py
gramalingam: Done

onnxscript/_internal/builder.py
gramalingam: Does the serializer set the tensor name from the value name? If so, that's great, no need to worry here.

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,353 @@
...
+
+        def adapt(x, typevar: str | None) -> ir.Value | None:
+            if x is None:
+                return None
+            if typevar is None:
+                return self._input_to_ir_value(x)
+            type_like = type_bindings.get(typevar) if typevar is not None else None
+            return self._input_to_ir_value(x, type_like)
+
+        return [adapt(x, typevar) for x, typevar in args_typevars]
```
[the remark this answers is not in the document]
gramalingam: done

onnxscript/_internal/builder.py
justinchuby: It does

onnxscript/_internal/builder.py
```diff
@@ -11,6 +11,20 @@
 import onnxscript._internal._inference as inference
 import onnxscript.optimizer
 
+# A permissable value for an op input, which can be converted to an ir.Value.
+VALUE_LIKE = (
```
justinchuby: We need to use Union to support python 3.9, when the type is evaluated.

onnxscript/_internal/builder.py
```diff
@@ -11,6 +11,20 @@
 import onnxscript._internal._inference as inference
 import onnxscript.optimizer
 
+# A permissable value for an op input, which can be converted to an ir.Value.
```
justinchuby: permissible

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,367 @@
...
+
+def _constant_name(value: int | float | bool | str | Sequence, type_suffix: str) -> str:
+    """Generate a descriptive name for a constant value."""
+    if isinstance(value, (int, float, bool, str)):
+        return f"const_{value}_{type_suffix}" if type_suffix else f"const_{value}"
+    # Sequence: use up to 2 elements in the name
+    if len(value) <= 2:
+        vals = ",".join(str(v) for v in value)
+    else:
+        vals = f"{value[0]},{value[1]},..."
```
justinchuby: The commas in the name feel a bit strange. There are likely going to be a few of tensors like this in a model (shapes etc.). Should we just allow it to take a generic (value independent) name?

onnxscript/_internal/builder.py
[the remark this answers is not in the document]
justinchuby: @gramalingam I think this is relavent. I think it's better to leave the tensor name untouched.

onnxscript/_internal/_inference.py
gramalingam: yes, likely.

onnxscript/_internal/builder.py
gramalingam: I don't like the commas either. Prefer standard identifiers. But for a shape with 2 or 3 dims would it be worth seeing the actual values? May be using underscore instead of comma etc.? 

But generic name may be good for now ... we can leave refinements to later.

onnxscript/_internal/builder.py
gramalingam: Ok, let me do it in the next phase when implementing this logic.

xadupre (approved):

onnxscript/_internal/builder.py
```diff
@@ -0,0 +1,379 @@
...
+    def push_module(self, module: str) -> None:
+        """Push a new naming context onto the stack (e.g. a layer or module name)."""
+        current = self.context_name()
+        if module:
+            new_context = f"{current}.{module}" if current else module
+        else:
+            new_context = current
+        self._context_stack.append(new_context)
+
+    def pop_module(self) -> None:
```
xadupre, line 338: I would make this a context manager.

justinchuby: This is likely going to be used in the module interface (not directly by the user). So the current design is easier to support the use case, I think.

justinchuby (approved):

onnxscript/_internal/builder.py
gramalingam: We will add the context-manager on top of this.

PR #2819, opened 2026-02-20
Add onnxscript.nn module with Module and Parameter classes

justinchuby:
This pull request introduces a new PyTorch-like module interface for building ONNX graphs, enabling users to define reusable neural network components and manage parameters in a structured way. The main changes add the `onnxscript.nn` package, expose its API, and implement core classes for module and parameter management.

Addition of ONNX neural network module interface:

* Added new `onnxscript.nn` package and exposed it in the main `onnxscript` API, allowing users to access neural network module functionality. (`onnxscript/__init__.py`, [[1]](diffhunk://#diff-a1562cf8b37bd59e756ae3802f64fb4b0712845c4dd5a747d7db2a4212f1bfa5R8) [[2]](diffhunk://#diff-a1562cf8b37bd59e756ae3802f64fb4b0712845c4dd5a747d7db2a4212f1bfa5L130-R131)
* Introduced `Module` and `Parameter` classes in `onnxscript/nn/_module.py` and `onnxscript/nn/_parameter.py`, providing a PyTorch-like interface for defining ONNX graph modules, registering parameters, and managing module hierarchies. (`onnxscript/nn/_module.py`, [[1]](diffhunk://#diff-26fd1ef9987845a4613d3d4b2e8d0c6d276b4946ba1a6a3e5e77b6e156e224beR1-R206); `onnxscript/nn/_parameter.py`, [[2]](diffhunk://#diff-94ffcabca05fa631e1902fbd4a2e967c00f9db0dca95baae367daef60cd8a7e1R1-R66)
* Created `onnxscript/nn/__init__.py` to expose `Module` and `Parameter` as the public API of the new package. (`onnxscript/nn/__init__.py`, [onnxscript/nn/__init__.pyR1-R9](diffhunk://#diff-db47492cd4dbb1a882423614592cb391e31abdec6b5aff85d6f1d2e2716507b4R1-R9))

Core module and parameter functionality:

* `Module` class supports automatic registration of parameters and child modules, implements methods for iterating over parameters/modules, and provides `state_dict`/`load_state_dict` for parameter serialization/deserialization, mirroring PyTorch's API. (`onnxscript/nn/_module.py`, [onnxscript/nn/_module.pyR1-R206](diffhunk://#diff-26fd1ef9987845a4613d3d4b2e8d0c6d276b4946ba1a6a3e5e77b6e156e224beR1-R206))
* `Parameter` class subclasses `ir.Value`, allowing direct use in ONNX ops and supporting initialization, realization, and representation of parameter tensors. (`onnxscript/nn/_parameter.py`, [onnxscript/nn/_parameter.pyR1-R66](diffhunk://#diff-94ffcabca05fa631e1902fbd4a2e967c00f9db0dca95baae367daef60cd8a7e1R1-R66))

### DIFF
--- a/onnxscript/nn/_parameter.py
+++ b/onnxscript/nn/_parameter.py
@@ -35,7 +35,7 @@ class Parameter(ir.Value):
         data: ir.TensorProtocol | None = None,
     ) -> None:
         if data is not None:
-            if data.dtype != dtype:
+            if dtype is not None and data.dtype != dtype:
                 raise ValueError(
                     f"Data type of provided data ({data.dtype}) does not match the specified dtype ({dtype})."
                 )


onnxscript/nn/_module.py
```diff
@@ -0,0 +1,206 @@
...
+                object.__setattr__(value, "_name", name)
+            self._modules[name] = value
+            object.__setattr__(self, name, value)
+        else:
+            object.__setattr__(self, name, value)
+
+    def __call__(self, op: OpBuilder, *args: Any, **kwargs: Any) -> Any:
+        builder: GraphBuilder = op.builder
+        module_name = self._name or ""
+        builder.push_module(module_name)
```
gramalingam: nit: may be use `if self._name: builder.push_module(self._name)` and same for pop?

onnxscript/nn/_module.py
```diff
@@ -0,0 +1,206 @@
...
+        """Return a dictionary mapping parameter names to their tensor data.
+
+        Mirrors ``torch.nn.Module.state_dict()``. Keys use dot-separated
+        hierarchical names (e.g. ``"layer1.weight"``). Values are the
+        ``const_value`` of each parameter (``None`` if uninitialized).
+        """
+        result: dict[str, ir.TensorProtocol | None] = {}
+        for name, param in self._parameters.items():
+            full_name = f"{prefix}.{name}" if prefix else name
+            result[full_name] = param.const_value
```
gramalingam, line 154: This could be None, right? ... I guess I am just looking for how it will be used, or rather, how load_state_dict will be used in practice with real trained weights coming from somewhere else ... I suppose that will come later.

onnxscript/nn/_parameter.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if self.name is None:
+            raise ValueError(
+                "Parameter._realize() called on a Parameter without a name. "
+                "Ensure the Parameter is attached to a Module attribute or otherwise "
+                "initialized with a name before realization."
+            )
+        if self.name:
+            self.name = builder.qualify_name(self.name)
+        builder.graph.initializers[self.name] = self  # type: ignore[index]
```
gramalingam: Does the IR already support initializers without a value? Can it be serialized?

onnxscript/nn/_module.py
gramalingam: Just to be explicit here (independent of the implementation of push/pop) about the treatment when no name is specified. (Alternatively, we can make name a required parameter in the constructor ... not sure if people will want modules without names/hierarchical names)

justinchuby: Done, added the conditional

onnxscript/nn/_module.py
justinchuby: Do you mean param.const_value can be None? I think that's ok. (not sure what None means yet but seems ok)

onnxscript/nn/_parameter.py
justinchuby: Yes. It will not be serialized if empty. The idea is the const_values will be filled in later, after the model is built.

justinchuby:
Updated how scoped names are constructed. @gramalingam

onnxscript/nn/_parameter.py
```diff
@@ -0,0 +1,72 @@
...
+        dtype: Data type of the parameter. Defaults to FLOAT.
+        name: Name for the parameter. If None, the attribute name from
+            the parent Module is used.
+        data: Optional initial tensor data. If provided, the initializer
+            will carry this as its const_value.
+    """
+
+    def __init__(
+        self,
+        shape: Sequence[int | ir.SymbolicDim | None],
```
justinchuby: We an accept a str too

onnxscript/_internal/builder.py
```diff
@@ -80,10 +80,10 @@ def __init__(self, graph: ir.Graph) -> None:
 
         self._op_builder = self.opset("", opset_version)
 
-        # Context stack to manage hierarchical naming. Each module/layer can push a new context, and pop it when done.
-        # The current context is used as a prefix for naming values and nodes.
-        # This allows us to generate names like "layer1.attention.query"
-        self._context_stack: list[str] = [""]
+        # Module scope stack. Each entry is (name, class_name) where name is
```
gramalingam, line 83: Ok for now. Would like to ensure that the strategy works for more complex cases (eg., modules calling functions calling modules, etc.) ...

gramalingam: It might just require fixing the comments to illustrate the different usage scenarios

onnxscript/_internal/builder.py
```diff
@@ -322,33 +324,80 @@ def call_op(
             version=version,
             name=node_name,
         )
+
+        # Attach scope metadata to the node
+        node.metadata_props["namespace"] = self._build_namespace()
+        node.metadata_props["pkg.onnxscript.class_hierarchy"] = repr(self._scope_classes())
```
gramalingam, line 330: The high level concept is fine. But, in terms of the details of the naming pattern, are we trying to maintain some compatibility? Just noting that the existing ModelBuilder naming style and Exporter naming style may not be exactly the same, which would mean that we can accommodate one or the other, may not be both.

onnxscript/_internal/builder.py
```diff
@@ -322,33 +324,80 @@ def call_op(
...
 
-    def push_module(self, module: str) -> None:
-        """Push a new naming context onto the stack (e.g. a layer or module name)."""
-        current = self.context_name()
-        if module:
-            new_context = f"{current}.{module}" if current else module
-        else:
-            new_context = current
-        self._context_stack.append(new_context)
+    def push_module(self, module: str, class_name: str = "") -> None:
```
gramalingam, line 337: Ok for now ... but may be worth thinking whether some abstraction here [that is, pushing an abstract object here which exposes methods like module() and class_name()] would lead to better extensibility in the future ... eg., if we ever want to capture more metadata in a node.

onnxscript/_internal/builder.py
justinchuby: Sounds good

onnxscript/_internal/builder.py
justinchuby: This is more of the exporter style. Current model builder does not contain metadata. The node and value naming styles are closer to current model builder.

onnxscript/_internal/builder.py
justinchuby: Sounds good.

PR #2820, opened 2026-02-20
Allow GraphBuilder to call script functions

gramalingam:
* This extension allows calls to scripted functions within traced (GraphBuilder) calls.
* As part of this extension, allow the use of OpBuilder within scripted function
* Also fix the default value naming strategy to include node number to ensure uniqueness (in common cases).

TODO later: Consider integration of standard execution of script functions with execution of OpBuilder calls, as well as other extensions within script-mode to ensure both modes are seamless and uniform

### DIFF
--- a/onnxscript/_internal/_inliner.py
+++ b/onnxscript/_internal/_inliner.py
@@ -38,10 +38,11 @@ def instantiate(
 
     def rename(node: ir.Node) -> None:
         if prefix:
-            node.name = prefix + (node.name or "")
+            if node.name:
+                node.name = prefix + node.name
             for output in node.outputs:
-                if output is not None:
-                    output.name = prefix + (output.name or "")
+                if output is not None and output.name:
+                    output.name = prefix + output.name
 
     cloner = Cloner(
         attr_map=attributes,
--- a/onnxscript/_internal/builder.py
+++ b/onnxscript/_internal/builder.py
@@ -1,5 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
+"""Graph builder for constructing ONNX IR graphs imperatively.
+
+This module provides imperative builders for constructing ONNX IR graphs with automatic
+constant promotion, type casting, and shape inference. The GraphBuilder class enables
+programmatic construction of graphs with proper scoping, constant management, and node
+creation. The OpBuilder class provides dynamic op dispatching via attribute access.
+"""
 
 from __future__ import annotations
 
@@ -10,7 +17,7 @@ import onnx_ir as ir
 
 import onnxscript._internal._inference as inference
 import onnxscript.optimizer
-from onnxscript._internal import _inliner as inliner
+from onnxscript._internal import _inliner
 
 # A permissible value for an op input, which can be converted to an ir.Value.
 VALUE_LIKE = Union[
@@ -362,7 +369,7 @@ class GraphBuilder:
         else:
             for output in function_ir.outputs:
                 output_renaming[output.name] = self._qualify_value_name(output.name)
-        nodes, outputs = inliner.instantiate(function_ir, args, kwargs)
+        nodes, outputs = _inliner.instantiate(function_ir, args, kwargs)
         if _prefix:
             self.push_module(_prefix)
         for node in nodes:
@@ -487,6 +494,19 @@ class OpBuilder:
         _prefix: str = "",
         **kwargs,
     ):
+        """Call a function and inline it into the graph.
+
+        Args:
+            function: The function to call (ir.Function or onnxscript.OnnxFunction).
+            *args: Positional arguments to pass to the function.
+            _outputs: Optional sequence of output names. If provided, must match the
+                number of function outputs.
+            _prefix: Optional prefix for module scoping (e.g., "layers.0").
+            **kwargs: Keyword arguments to pass to the function.
+
+        Returns:
+            The output value(s) from the function call.
+        """
         return self._builder.call(
             function, *args, _outputs=_outputs, _prefix=_prefix, **kwargs
         )


onnxscript/_internal/builder.py
```diff
@@ -9,6 +9,7 @@
 import onnx_ir as ir
 
 import onnxscript._internal._inference as inference
+import onnxscript._internal._inliner as inliner
```
justinchuby:
```suggestion
from onnxscript._internal import _inliner
```

onnxscript/_internal/builder.py
```diff
@@ -182,12 +183,13 @@ def _adapt_outputs(
...
                 raise ValueError(f"Number of outputs must be non-negative, got {outputs}")
+            count = self.graph.num_nodes()
             if outputs == 1:
-                name = f"{op_type}_output" if op_type else "output"
+                name = f"{op_type}_n{count}_output" if op_type else f"n{count}_output"
                 return [ir.Value(name=self.qualify_name(name))]
             else:
                 names = [
-                    f"{op_type}_output{i}" if op_type else f"output{i}" for i in range(outputs)
+                    f"{op_type}_n{count}_output{i}" if op_type else f"n{count}_output{i}" for i in range(outputs)
```
justinchuby: There may be some conflicts: I updated the naming patterns in https://github.com/microsoft/onnxscript/pull/2819

gramalingam: Undid the changes

onnxscript/_internal/_inliner.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from __future__ import annotations
+
+from typing import Mapping, Sequence
+
+import onnx_ir as ir
+from onnx_ir._cloner import Cloner
```
justinchuby, line 9: Should we integrate this into onnx-ir?

onnxscript/_internal/builder.py
```diff
@@ -10,6 +10,7 @@
 
 import onnxscript._internal._inference as inference
 import onnxscript.optimizer
+from onnxscript._internal import _inliner as inliner
```
justinchuby:
```suggestion
from onnxscript._internal import _inliner
```

onnxscript/_internal/builder.py
```diff
@@ -334,6 +335,49 @@ def call_op(
...
+        *args,
+        _outputs: Sequence[str] | None = None,
+        _prefix: str = "",
+        **kwargs,
+    ):
+        if isinstance(function, ir.Function):
+            function_ir = function
+        elif isinstance(function, onnxscript.OnnxFunction):
+            function_proto = function.to_function_proto()
+            function_ir = ir.serde.deserialize_function(function_proto)
```
justinchuby, line 357: I think we can take the ir function from OnnxFunction directly?

onnxscript/_internal/builder.py
```diff
@@ -334,6 +335,49 @@ def call_op(
...
+                    f"number of function outputs {len(function_ir.outputs)}."
+                )
+            for output, name in zip(function_ir.outputs, _outputs):
+                output_renaming[output.name] = self._qualify_value_name(name)
+        else:
+            for output in function_ir.outputs:
+                output_renaming[output.name] = self._qualify_value_name(output.name)
+        nodes, outputs = inliner.instantiate(function_ir, args, kwargs)
+        if _prefix:
+            self.push_module(_prefix)
```
justinchuby, line 374: push_module now takes two args, I think

justinchuby: Sorry. Keeping it empty is ok.

onnxscript/_internal/builder.py
```diff
@@ -426,3 +478,15 @@ def __getattr__(self, op_type: str) -> Callable:
...
         return self._builder.initializer(tensor, name)
+
+    def call(
+        self,
+        function,
+        *args,
+        _outputs: Sequence[str] | None = None,
+        _prefix: str = "",
+        **kwargs,
+    ):
```
justinchuby, line 496: Create a docstring? Note that for example it will raise if the number of outputs doesn't match the number of outputs defined by the function.

onnxscript/_internal/builder.py
gramalingam: I think it has a default value? The tests work

onnxscript/_internal/_inliner.py
gramalingam: Yes, that would be worth doing

onnxscript/_internal/builder.py
gramalingam: Done. A minor nit: I think these should be just review comments not "commit suggestion" since it will break unless the reference to inliner is also renamed ... I think a "commit suggestion" should guarantee that it makes sense on a standalone basis.

onnxscript/_internal/builder.py
gramalingam: I don't think we expose this functionality yet?

onnxscript/_internal/builder.py
gramalingam: done

onnxscript/_internal/builder.py
gramalingam: The reason for that is we need to figure out IR ownership ... ideally we should expose a readonly function IR to avoid external caller from modifying the IR. Anyway, that is an issue we need to resolve separately.

onnxscript/_internal/builder.py
```diff
@@ -487,6 +494,19 @@ def call(
         _prefix: str = "",
         **kwargs,
     ):
+        """Call a function and inline it into the graph.
+
+        Args:
+            function: The function to call (ir.Function or onnxscript.OnnxFunction).
+            *args: Positional arguments to pass to the function.
```
justinchuby, line 501: Does it support attributes? Would it be better to just take inputs and attributes as tuple and dict?

gramalingam: Yes. The idea is to expose an API that is as close as possible to the op-calling API. Thus, *args and **kwargs are handled the same way whether we call an op or a function (except for special kwargs like _prefix).

gramalingam: I think we can later add a core API that uses inputs as a tuple and attributes as a dict, but that would be a lower-level API

justinchuby: Sounds good, thanks

PR #2823, opened 2026-02-24
Add nn.Sequential: callable ModuleList that chains forward calls

justinchuby:
Usage PR: https://github.com/justinchuby/onnx-genai-models/pull/19

nn.Sequential mirrors torch.nn.Sequential — children are registered
with numeric keys and forward() passes each child's output as the
next child's input.

Key design decisions:
- Subclasses ModuleList for registration, indexing, iteration
- Overrides _set_name to keep children with simple '0', '1' names
  (avoids double-prefixing since __call__ already pushes the parent name)
- Raises RuntimeError on empty Sequential

This enables matching HF diffusers naming conventions (e.g.
nn.Sequential(SiLU(), Linear(...)) producing 'mod.1.weight') without
needing preprocess_weights renames.

### DIFF
--- a/docs/tutorial/builder/graph_builder.md
+++ b/docs/tutorial/builder/graph_builder.md
@@ -406,3 +406,110 @@ def build_linear(op, x, weight, bias_value):
 
 This pattern keeps function signatures simple while preserving access to the
 full builder API when needed.
+## Calling Script Functions from OpBuilder
+
+The `OpBuilder` provides a `call()` method to inline `@script`-decorated ONNX functions directly into the builder's graph. This enables composition of both imperative (builder) and declarative (`@script`) code within a single graph.
+
+### Basic function inlining
+
+Define an ONNX script function and then call it through `op.call()`:
+
+```python
+from onnxscript import script, opset23 as op23
+
+# Define a reusable script function
+@script(default_opset=op23)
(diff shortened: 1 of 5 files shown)


onnxscript/nn/_sequential.py
```diff
@@ -0,0 +1,64 @@
...
+        for key, child in self._modules.items():
+            child._set_name(key)
+
+    def forward(self, op: _builder.OpBuilder, *args: Any, **kwargs: Any) -> Any:
+        """Run each child module sequentially, passing output to the next."""
+        if len(self) == 0:
+            raise RuntimeError("Cannot call forward on an empty Sequential container")
+        for i, module in enumerate(self):
+            if i == 0:
+                args = (module(op, *args, **kwargs),)
```
gramalingam: I assume this is mimicking torch's behavior here (the distinction between 1st and rest wrt kwargs)?

gramalingam: It's a little bit asymmetric and unintuitive.

justinchuby: Yeah strange. I need to double check this

justinchuby: Updated to match pytorch behavior

PR #2828, opened 2026-02-25
Add input() and add_output() methods to GraphBuilder

justinchuby:
This pull request adds new methods to the `GraphBuilder` class to simplify the creation and management of graph inputs and outputs, and introduces corresponding unit tests to ensure their correct behavior. The changes improve the usability and reliability of the graph-building API.

**Enhancements to the GraphBuilder API:**

* Added a new `input` method to `GraphBuilder` for creating and registering graph input values with support for specifying name, dtype, shape, type, constant value, and metadata properties.
* Added a new `add_output` method to `GraphBuilder` to append an output value to the graph and optionally rename it.

### DIFF
--- a/onnxscript/_internal/builder_test.py
+++ b/onnxscript/_internal/builder_test.py
@@ -560,6 +560,43 @@ class GraphBuilderTest(unittest.TestCase):
         self.assertEqual(len(graph.inputs), 1)
         self.assertIs(graph.inputs[0], value)
 
+    def test_input_with_const_value_registers_initializer(self):
+        """Test that GraphBuilder.input registers initializer when const_value is provided."""
+        graph = ir.Graph(
+            name="test_model",
+            inputs=[],
+            outputs=[],
+            nodes=[],
+            opset_imports={"": _default_opset_version},
+        )
+        graph_builder = builder.GraphBuilder(graph)
+
+        const_tensor = ir.tensor([1.0, 2.0], dtype=ir.DataType.FLOAT, name="const_data")
+        value = graph_builder.input("const_input", const_value=const_tensor)
+
+        self.assertEqual(len(graph.inputs), 1)
+        self.assertIs(graph.inputs[0], value)
+        self.assertIn("const_input", graph.initializers)
+        self.assertIs(graph.initializers["const_input"], value)
+        self.assertIs(value.const_value, const_tensor)
+
+    def test_input_without_const_value_does_not_register_initializer(self):
+        """Test that GraphBuilder.input does not register initializer without const_value."""
+        graph = ir.Graph(
+            name="test_model",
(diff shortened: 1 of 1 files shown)


onnxscript/_internal/builder.py
```diff
@@ -152,6 +152,52 @@ def initializer(
...
+    ) -> ir.Value:
+        """Create an input to the graph and return the corresponding ir.Value.
+
+        Args:
+            name: The name of the value.
+            dtype: The data type of the TensorType of the value. This is used only when type is None.
+            shape: The shape of the value.
+            type: The type of the value. Only one of dtype and type can be specified.
+            const_value: The constant tensor that initializes the value. Supply this argument
+                when you want to create an initializer. The type and shape can be obtained from the tensor.
```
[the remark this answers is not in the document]
justinchuby: We should update it to register initializer as well

onnxscript/_internal/builder.py
```diff
@@ -152,6 +152,52 @@ def initializer(
         self._graph.register_initializer(value)
         return value
 
+    def input(
+        self,
+        name: str,
+        dtype: ir.DataType | None = None,
```
gramalingam, line 158: I was thinking about this too. I think if we could accommodate something like `FLOAT['N', 1024]` as a way of compactly specifying type and shape, it would help. Like it is done [here](https://github.com/microsoft/onnxscript/blob/ef2bc22dea5e12726a5ad6286aab2bcc537eccb3/onnxscript/_internal/builder.py#L100).

justinchuby: I was hesitant to pass in generic-looking type objects around. The behavior of generic type classes tend to be a bit unstable across different python versions (where things are stored, how data can be accessed, when something is evaluated, etc.). So I am not preferring it for now.

onnxscript/_internal/builder.py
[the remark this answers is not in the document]
justinchuby: Done

onnxscript/_internal/builder.py
gramalingam: What is unstable? Maybe we can fix it. Or support something like `(dtype, ('N', 1024))`

gramalingam: In summary: the suggestions are
* a combined TypeAndShape would be useful and more compact (in some settings, not necessarily all).
* we could support the legacy onnxscript notation by adding a method to convert it to an ir.TypeAndShape(), and allowing objects that support a method called "toTypeAndShape" ... I currently have a to_ir method, but that name is too generic to be used safely for this purpose, but a name like "toTypeAndShape" should be reasonable.

gramalingam (approved):

PR #2835, opened 2026-03-03
Use logging instead of print in RewritePass

tgolsson:
Makes it possible to filter this out using the regular logging mechanics in Python.

### DIFF
--- a/onnxscript/rewriter/__init__.py
+++ b/onnxscript/rewriter/__init__.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 from __future__ import annotations
 
+import logging
 from typing import Sequence, TypeVar, Union
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


onnxscript/rewriter/__init__.py
```diff
@@ -82,7 +85,8 @@ def __init__(
     def call(self, model: ir.Model) -> ir.passes.PassResult:
         count = self.rules.apply_to_model(model)
         if count:
-            print(f"Applied {count} of general pattern rewrite rules.")
+            logger.info("Applied %s of general pattern rewrite rules.", count)
```
justinchuby, line 88: Is there a config to display this by default? Could you do that?

tgolsson: What do you mean by "display by default"? Libraries should not print to stdout by default. I don't care about these logs, they're a secondary effect of torch using onnxscript internally and I'd like them gone - using logging lets me and any other user configure this on our end. So those who care can have them and those who do not can disable them.

justinchuby: Sounds reasonable, thanks.

PR #2837, opened 2026-03-03
Implement schema-based input/attribute partitioning in GraphBuilder

gramalingam:
- Convert onnx.defs.OpSchema to ir.schemas.OpSignature via from_op_schema and delegate to separate_input_attributes_from_arguments
- Add allow_extra_args parameter to separate_input_attributes_from_arguments for rejecting unexpected positional arguments (default True for compat)
- Builder uses strict mode: allow_extra_kwargs=False, allow_extra_args=False
- Refactor _build test helper: accept TypeSpec, optional trace_function, return ir.Graph directly
- Add comprehensive tests for input/attribute partitioning

### DIFF
--- a/onnxscript/_internal/builder_test.py
+++ b/onnxscript/_internal/builder_test.py
@@ -32,7 +32,7 @@ def _resolve_type_spec(spec: builder.TypeSpec) -> ir.TypeAndShape:
     if isinstance(spec, ir.TypeAndShape):
         return spec
     if isinstance(spec, type) and issubclass(spec, TensorType):
-        return spec.to_ir()
+        return spec.to_ir_type_and_shape()
     raise TypeError(f"Expected ir.TypeAndShape or a TensorType subclass, got {type(spec)!r}.")
 
 


onnxscript/_internal/builder.py
```diff
@@ -255,9 +255,16 @@ def _partition_inputs_attributes(
         inputs: Sequence[ir.Value | ir.TensorProtocol],
         kwargs: dict[str, Any],
     ) -> tuple[Sequence[ir.Value | ir.TensorProtocol], dict[str, Any]]:
-        # Not implemented yet
-        del schema
-        return inputs, kwargs
+        if schema is None:
```
justinchuby, line 362: The Op object has the signature cached. Would it be possible to use that?

gramalingam: The OpBuilder does not, AFAIK. Are you thinking of the Op used in script functions? Currently, this is a light-weight wrapper object, so there's not value in caching here. If we want caching (to avoid going into C++?), we would need a cache somewhere else to enable better reuse

PR #2843, opened 2026-03-05
Rewriter: fix fuse_relu_clip with None max

### LINKED ISSUE #2836: fuse_relus_clips optimization crashes on Clip without an explicit max value
converting the attached ONNX model with the following script:

```
from onnxscript import ir
from onnxscript.optimizer import optimize

ir_model = ir.load("input.onnx")
optimized_model = optimize(ir_model)
ir.save(optimized_model, "output.onnx")
```

crashed the optimizer:

```
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnx_ir/passes/_pass_infra.py", line 250, in call
    pass_result = pass_(model)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnx_ir/passes/_pass_infra.py", line 127, in __call__
    result = self.call(model)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/__init__.py", line 83, in call
    count = self.rules.apply_to_model(model)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/_rewrite_rule.py", line 802, in apply_to_model
    count = self._apply_to_graph_or_function(
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/_rewrite_rule.py", line 669, in _apply_to_graph_or_function
    delta = rule.try_rewrite(
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/_rewrite_rule.py", line 324, in try_rewrite
    replacement_subgraph = self._replacement_pattern.get_replacement(match)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/_rewrite_rule.py", line 220, in get_replacement
    new_outputs = self._function(context, **match.bindings)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/rules/common/_fuse_relus_clips.py", line 39, in rewrite
    min_clip, max_clip = self.compute_clip_min_max(first_clip_node, second_clip_node)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/rules/common/_fuse_relus_clips.py", line 152, in compute_clip_min_max
    min_clip, max_clip, dtype = self.extract_min_max(first_clip_node)
  File "/local/mnt/workspace/onnxscript/lib/python3.10/site-packages/onnxscript/rewriter/rules/common/_fuse_relus_clips.py", line 77, in extract_min_max
    max_clip = node.inputs[2].const_value.numpy()
AttributeError: 'NoneType' object has no attribute 'const_value'
```

This happens because the Clip has no explicit max value, but relies on the default

[input.onnx.zip](https://github.com/user-attachments/files/25722464/input.onnx.zip)

```
ml_dtypes         0.5.4
onnx              1.20.1
onnx-ir           0.1.16
onnxscript        0.6.2
```

AyoubMDL:
Fixes https://github.com/microsoft/onnxscript/issues/2836


justinchuby (approved):

onnxscript/rewriter/rules/common/_fuse_relus_clips.py
```diff
@@ -74,7 +74,9 @@ def extract_min_max(self, node: ir.Node):
                 min_clip = min_input.const_value.numpy()
 
         if len(node.inputs) > 2:
-            max_clip = node.inputs[2].const_value.numpy()
+            max_clip = node.inputs[2]
+            if max_clip is not None:
+                max_clip = max_clip.const_value.numpy()
```
gramalingam, line 79: I don't see a check here whether max_clip.const_value is None. It is possible that it is being checked elsewhere before this method if called, but, even so, it would be better if all related logic is in one place (checking if an input is present, whether it is a constant, etc.). Overall, it feels like the structure of this implementation turns what should be a mathematically simple rule (easy to read and validate) into something opaque. 

I realize that this is mostly a problem with the pre-existing rule, but it just makes updates like this harder to review.

AyoubMDL: I agree with the general point, but I think this case is a bit special. Normally we would perform these checks in the check method and fail the pattern if they don’t pass. However, in this pattern the `max/min` values can be None, and the rewrite can still proceed in that case. That’s why the logic was isolated in `extract_min_max` method instead of being handled entirely in check.
That said, I’m open to suggestions if you think there’s a cleaner way to structure this.

bas-aarts:
What is the reason the PR cannot be merged? It's a rathe simple fix.

justinchuby (approved):

PR #2851, opened 2026-03-13
Add an option to not inline a function when building the graph

justinchuby:
- Introduced distinct `call` and `call_inline` methods in `GraphBuilder` and `OpBuilder` to differentiate between creating a single function call node (`call`) and inlining a function's body directly into the graph (`call_inline`). The `call` method now registers the function in the builder, while `call_inline` does not.


onnxscript/_internal/builder.py
```diff
@@ -543,16 +549,18 @@ def call_op(
 
     def call(
         self,
-        function,
+        function: ir.Function | onnxscript.OnnxFunction,
         *args,
         _outputs: Sequence[str] | None = None,
         _prefix: str = "",
+        _inline: bool = True,
```
gramalingam: Should we enforce that _prefix is "" if _inline is False?

onnxscript/_internal/builder.py
```diff
@@ -505,13 +511,13 @@ def call_op(
         self,
         op_type: str,
         inputs: Sequence[ir.Value | ir.TensorProtocol],
-        kwargs: dict[str, Any],
+        kwargs: dict[str, ir.Value | ir.TensorProtocol],
+        /,
+        domain: str = "",
+        version: int | None = None,
```
xadupre, line 806: do we expect different versions of the same opset?

onnxscript/_internal/builder.py
```diff
@@ -672,27 +721,51 @@ def version(self) -> int | None:
...
+            op_type, inputs, kwargs, domain=domain, version=version, outputs=outputs
+        )
 
     def __getattr__(self, op_type: str) -> Callable:
         return lambda *args, **kwargs: self._call_op(op_type, args, kwargs)
 
     def initializer(self, tensor: ir.TensorProtocol, name: str | None = None) -> ir.Value:
         return self._builder.initializer(tensor, name)
 
+    def functions(self) -> dict[ir.OperatorIdentifier, ir.Function]:
```
xadupre, line 1052: no property ?

justinchuby: Good catch thanks

onnxscript/_internal/builder.py
justinchuby: We this is supported when building an ONNX IR model, even though we don't do this in practice

onnxscript/_internal/builder.py
```diff
@@ -543,7 +549,45 @@ def call_op(
...
+            domain=function.domain,
+            name=self._qualify_node_name(function.name),
+        )
+        # Attach scope metadata to the node
+        node.metadata_props["namespace"] = self._build_namespace()
+        node.metadata_props["pkg.onnxscript.class_hierarchy"] = repr(self._scope_classes())
+        node.metadata_props["pkg.onnxscript.name_scopes"] = repr(self._scope_names())
+
+        self.add_node(node)
+        self._functions[function.identifier()] = function
```
xadupre: the same function can be called multiple times, so the

xadupre: I guess two different functions cannot have the same identifier.

justinchuby: Yes - that's true. Should we raise if redefined?

justinchuby: Will address in a follow up if needed

onnxscript/_internal/builder.py
```diff
@@ -447,6 +447,8 @@ def __init__(self, graph: ir.Graph, parent: GraphBuilder | None = None) -> None:
         if parent is None:
             self._constant_cache: dict[tuple[Any, ir.DataType | None], ir.Value] = {}
 
+        self._functions: dict[ir.OperatorIdentifier, ir.Function] = {}
+
```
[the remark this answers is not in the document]
gramalingam: Agree with this suggestion

onnxscript/_internal/builder_test.py
```diff
@@ -1040,6 +1040,157 @@ def test_none_input_with_custom_domain(self):
...
+        # Should produce a single function call node
+        nodes = list(op.builder.graph)
+        self.assertEqual(len(nodes), 1)
+        self.assertEqual(nodes[0].op_type, "simple_add")
+        self.assertIsInstance(result, ir.Value)
+
+        # Function should be registered
+        self.assertEqual(len(op.builder.functions), 1)
+
+    def test_call_inline_produces_more_nodes_than_call(self):
```
gramalingam: I am not sure what this test does that is not already tested independently (for call and call_inline).

onnxscript/_internal/builder_test.py
```diff
@@ -1040,6 +1040,157 @@ def test_none_input_with_custom_domain(self):
...
+
+        node = nodes[0]
+        self.assertEqual(node.op_type, "mul_add_relu")
+        self.assertEqual(list(node.inputs), [x, y])
+
+        # The result should be a single ir.Value
+        self.assertIsInstance(result, ir.Value)
+        self.assertIs(result, node.outputs[0])
+
+    def test_call_registers_function(self):
```
gramalingam, line 1067: What happens if we have two calls to same function? May be worth adding a test for that.

onnxscript/_internal/builder.py
```diff
@@ -833,7 +839,51 @@ def call_op(
...
     def call(
         self,
-        function,
+        function: ir.Function | onnxscript.OnnxFunction,
+        *args,
+        _outputs: int | Sequence[str | ir.Value] | None = None,
+        **kwargs,
+    ):
+        """Call a function as a single function node."""
+        if isinstance(function, ir.Function):
```
gramalingam, line 847: May be worth thinking through this: do we want/need to create a copy of the function? Or, leave that to the caller to do? Maybe we shouldn't allow OnnxFunction here? (Just start with a simpler, more explicit, interface.) Also worth thinking about multiple calls to same function (and how we check for identity/equality of function ... so letting the caller deal with that might simplify things here).

justinchuby: I think the current assumption is that the functions cannot be different if they have the same identifier. Will think more deeply about this. For now I removed OnnxFunction.

justinchuby: Actually retained OnnxFunction to support the scripted functions.

gramalingam: Just to clarify my concerns: if they have same identifier, do we assume that the functions are same, or can we check they are the same? If we want to check, is there an efficient pointer-equality check? Is there some convention that would enable to do this (check via an efficient id-equality check not requiring structural equality chec). Also, the question about should we clone script function's ir. 

Ok if we want to work through this separately.

onnxscript/_internal/builder_test.py
justinchuby: Removed

onnxscript/_internal/builder.py
justinchuby: I think we need to assume that they are the same - but checking is ok too if desired. But I would like to do it as a follow up after the release. Were you proposing to copy or not copy the script function? I understand you suggested to copy it? In that case I am adding a copy.

onnxscript/_internal/builder_test.py
justinchuby: Done

PR #2887, opened 2026-04-14
fix(optimizer): move subgraph initializers to main graph when inlining If branches

### LINKED ISSUE #2810: Constant Folding removes 'axes' initializer form Squeeze layer, resulting in incorrect ONNX model
converting the attached ONNX model with the following script:

```
from onnxscript import ir
from onnxscript.optimizer import optimize

ir_model = ir.load("input.onnx")
optimized_model = optimize(ir_model)
ir.save(optimized_model, "output.onnx")
```
produces an incorrect ONNX model. 
axes initializer '17' in Squeeze layer 'If_11_Squeeze_13' is missing.
The constant folder is causing this.

The If layer in the input model is the default way that a squeeze layer is exported from Torch, making this a common pattern.

[input.onnx.zip](https://github.com/user-attachments/files/25318319/input.onnx.zip)

```
ml_dtypes         0.5.4
onnx              1.20.1
onnx-ir           0.1.16
onnxscript        0.6.2

```

jaytiwarihub:
## Fixes #2810

When `FoldConstantsPass` inlines an `If` branch (because its condition is a known constant), initializers belonging to the branch subgraph were silently abandoned. This caused downstream nodes that referenced those initializers — most visibly `Squeeze` nodes whose `axes` tensor had been folded in a prior pass — to have dangling inputs, producing an invalid ONNX model that fails schema validation. 

### Cause 
The `if_op` partial evaluator correctly moves the branch's nodes into the main graph via `Replacement`, but it never touched the branch's initializers. Initializers are stored on the `ir.Graph` object itself, not on individual nodes, so they were invisible to the node-migration logic. After inlining, those values existed nowhere in the main graph.

### Fix
Added a helper `_move_initializers_to_graph(src, dst)` that migrates all initializers from a subgraph to the destination graph before the `Replacement` is returned, with name-collision handling.

### Testing
Added two new tests to `_constant_folding_test.py`:

* **`test_fold_if_cond_with_subgraph_initializer`**: Reproduces the exact two-pass scenario from the bug report. The first pass folds a constant inside the branch into a subgraph initializer; the second pass inlines the branch and verifies the initializer is successfully migrated to the main graph and that the model passes `onnx.checker.check_model`.
* **`test_fold_if_cond_with_subgraph_initializer_name_collision`**: Verifies that name collisions between subgraph and main-graph initializers are handled gracefully via unique suffixing, preventing data loss or crashes during the migration.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -1,8 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
-
 # NOTE: This will eventually replace the existing constant_folding.py and evaluator.py files.
-
 from __future__ import annotations
 
 __all__ = [
@@ -81,8 +79,6 @@ def _is_onnx_op(node: ir.Node, op_type: str) -> bool:
 # The API below works only for non-control-flow ops (ops without any graph-attributes).
 # This currently used ONNX's reference implementation. But we could also
 # use ORT's implementation if we want to.
-
-
 def _process_constant_node(node: ir.Node) -> None:
     """Sets const_value of output value of a Constant op node."""
     if not _is_onnx_op(node, "Constant"):
@@ -126,7 +122,6 @@ def _process_constant_node(node: ir.Node) -> None:
 
 def basic_constant_propagation(nodes: Iterable[ir.Node]) -> None:
     """Performs basic constant propagation for a sequence of nodes.
-
     Just marks the output values of Constant op nodes with their const_value.
     """
     for node in nodes:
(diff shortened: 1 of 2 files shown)


jaytiwarihub:

onnxscript/optimizer/_constant_folding.py
```diff
@@ -562,6 +554,27 @@ def size(node: ir.Node, op, state: OptimizerState) -> ReturnValue:
...
+    so that the inlined nodes can still reference them; failing to do so leaves the
+    references dangling and produces an invalid model.
+    """
+    counter: dict[str, int] = {}
+    for name in list(src.initializers):
+        initializer = src.initializers.pop(name)
+        # Ensure name uniqueness in the destination graph
+        new_name = name
+        while new_name in dst.initializers:
+            counter[name] = counter.get(name, 0) + 1
```
gramalingam, line 571: nit: not sure the dictionary helps. We can just use a local variable counter (reinitialized for every initializer).

justinchuby:
Thanks - closing in favor of https://github.com/microsoft/onnxscript/pull/2825

PR #2910, opened 2026-05-07
Unify BuilderBase and GraphBuilder: shared call_op with feature flags

gramalingam:
## Summary

Extends the `BuilderBase` / `TapeBuilder` infrastructure (from PR #2905) so that `GraphBuilder` inherits from `BuilderBase`, sharing a single `call_op()` implementation gated by feature flags.

### Class rename and API merge
- Renamed `OpBuilderBase` → `BuilderBase` (no backward-compat alias needed — not yet released)
- Merged the separate `op()` and `_make_node()` methods into a single `op(op_type, *args, **kwargs)` ergonomic API
- Promoted `_create_op` to `BuilderBase.call_op()` — the single core node-creation method

### Feature flags (`BuilderFeature` enum)
- `SCHEMA_PARTITION` — partition positional args into inputs vs attributes using op schema
- `CAST_INPUTS` — promote scalars to tensors, match type variables
- `CAST_ATTRIBUTES` — cast attributes using schema info (stub for future)
- `INFER_SHAPES` — run shape inference after node creation
- `CONSTANT_PROPAGATION` — run constant propagation after node creation
- `FULL` = all of the above (used by GraphBuilder)
- `NONE` = no extra processing (used by TapeBuilder for rewriter/optimizer/version-converter)

### GraphBuilder inherits from BuilderBase
- `GraphBuilder(BuilderBase)` with `features=BuilderFeature.FULL`
- `call_op()` is inherited directly — no override needed
- Default hook implementations (`_get_schema`, `_partition_inputs_attributes`, `_cast_inputs`, `_cast_attributes`) live in `BuilderBase`
- GraphBuilder only overrides hooks needing graph-specific state:
  - `_input_to_ir_value` — constant promotion with caching
  - `_generate_node_name` / `_adapt_outputs` — qualified naming
  - `_post_create_node` — constant propagation + shape inference
  - `_annotate_node` — scope metadata

### Exports
- `BuilderFeature` exported from `onnxscript` package

## Testing
- All 559 rewriter/optimizer/version-converter tests pass
- All 90 GraphBuilder tests pass
- All 15 TapeBuilder/context tests pass

### DIFF
--- a/onnxscript/_internal/tape_builder.py
+++ b/onnxscript/_internal/tape_builder.py
@@ -258,6 +258,9 @@ class BuilderBase(abc.ABC):
         This avoids initializer-name collisions when the builder is used
         inside the rewriter/optimizer.
 
+        When *dtype* is None, ``ir.tensor`` infers the dtype from the Python
+        value (e.g. float → FLOAT, int → INT64).
+
         GraphBuilder overrides this with a cache-based initializer strategy.
         """
         tensor = ir.tensor(value, dtype=dtype)
@@ -392,13 +395,13 @@ class BuilderBase(abc.ABC):
                 name=name,
             )
         else:
-            num_outputs = len(outputs) if isinstance(outputs, Sequence) else outputs
+            assert isinstance(outputs, int), f"Expected int, got {type(outputs)}"
             node = ir.Node(
                 domain,
                 op_type,
                 args,
                 attributes=attrs,
-                num_outputs=num_outputs,
+                num_outputs=outputs,
                 version=version,
                 name=name,
             )
--- a/onnxscript/_internal/tape_builder.py
+++ b/onnxscript/_internal/tape_builder.py
@@ -33,13 +33,6 @@ from onnxscript._internal import param_manipulation
 
 UsedOpsets = set[tuple[str, Optional[int]]]
 
-# Mapping from Python scalar types to their default ONNX DataType,
-# used when no schema-based type binding is available.
-_PYTHON_TYPE_TO_DTYPE: dict[type, ir.DataType] = {
-    int: ir.DataType.INT64,
-    float: ir.DataType.FLOAT,
-}
-
 
 def _dtype_suffix(dtype: ir.DataType) -> str:
     """Return a short type suffix for naming constants based on ir.DataType."""
@@ -323,7 +316,7 @@ class BuilderBase(abc.ABC):
 
         _inference.infer_outputs(node)
 
-    def _annotate_node(self, node: ir.Node) -> None:
+    def _annotate_node(self, node: ir.Node) -> None:  # noqa: B027
         """Attach metadata to a node after creation.
 
         Default: no-op. Override to add scope/namespace annotations.
--- a/onnxscript/_internal/tape_builder.py
+++ b/onnxscript/_internal/tape_builder.py
@@ -4,16 +4,17 @@
 
 This module defines:
 
-- ``BuilderBase``: Abstract base class for building ONNX IR nodes via a
-  dynamic dispatch interface (``op.Relu(x)``, ``op.op("Relu", x)``,
-  ``op.initializer(...)``).
+- ``BuilderBase``: Abstract base class for building ONNX IR nodes.
+  Provides the core node-creation pipeline (``call_op``) and an
+  initializer creation API.
   Subclasses implement the storage strategy by overriding ``_add_node``,
   ``_add_initializer``, and ``_record_opset``.
 
 - ``TapeBuilder``: Concrete subclass backed by simple lists.  Engines
   (rewriter, optimizer, version converter) create an instance, pass it to a
   rule or evaluator, and harvest the accumulated nodes / initializers / opsets
-  after it returns.
+  after it returns.  Provides the user-facing ``op()`` method and
+  ``__getattr__``-based dynamic dispatch (``op.Relu(x)``).
 
 - ``BuilderFeature``: Flag enum controlling optional processing steps
   (schema partitioning, input casting, shape inference, etc.).
@@ -64,26 +65,12 @@ class BuilderFeature(enum.Flag):
     SCHEMA_AWARE = SCHEMA_PARTITION | CAST_INPUTS | CAST_ATTRIBUTES
     FULL = SCHEMA_AWARE | INFER_SHAPES | CONSTANT_PROPAGATION
 
-    @property
-    def any_schema_feature(self) -> bool:
-        """True if any schema-dependent feature is enabled."""
-        return bool(
-            self
-            & (
-                BuilderFeature.SCHEMA_PARTITION
-                | BuilderFeature.CAST_INPUTS
-                | BuilderFeature.CAST_ATTRIBUTES
-            )
-        )
-
 
 class BuilderBase(abc.ABC):
     """Abstract base class for building ONNX IR nodes.
 
-    Supports two creation operations:
-
-    1. **Op creation** — ``op.op("Relu", x)`` or ``op.Relu(x)`` (syntactic sugar).
-    2. **Initializer creation** — ``op.initializer(tensor, name=...)``.
+    Provides the core node-creation pipeline via :meth:`call_op` and an
+    initializer creation API via :meth:`initializer`.
 
     Subclasses must implement the three protected methods that define where
     created nodes and initializers are stored:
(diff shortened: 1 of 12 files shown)


onnxscript/rewriter/rules/common/_remove_optional_bias.py
```diff
@@ -19,8 +19,8 @@ def rewrite(self, op, out: ir.Value, **_) -> ir.Value:
 
         return op.op(
             self.op_type,
-            inputs=node.inputs[:-1],
-            attributes=node.attributes,
+            *node.inputs[:-1],
```
justinchuby, line 22: Is this a breaking change?

gramalingam: yes

titaiwangms:
# Review

Synthesized from a multi-reviewer pass (correctness, adversarial, readability). In-scope findings only, sorted by severity.

## �� Critical / Bug

**1. `GraphBuilder.op` is a `@property`, shadowing `BuilderBase.op()` method — LSP violation**
`builder.py` / `tape_builder.py`. `BuilderBase.op()` is a method (`b.op("Relu", x)`); `GraphBuilder.op` is a property returning an `OpBuilder` (`b.op.Relu(x)`). The calling convention silently flips between base and subclass. Any code typed against `RewriterContext = BuilderBase` and handed a `GraphBuilder` will fail at runtime. Rename the property (e.g. `ops` / `default_op`) or document the difference loudly in `BuilderBase`.

**2. `OpBuilder._call_op` drops `_name`**
`builder.py`. Pops `_domain`, `_version`, `_outputs` but not `_name`. So `b.op.Add(x, y, _name="foo")` either raises an unexpected-kwarg error (schema-aware ops) or creates a bogus attribute named `_name` (custom ops). Pop `_name` and forward as `name=` to `builder.call_op(...)`.

## 🟠 Major

**3. Silent breakage of the old `op.op(inputs=..., attributes=..., domain=..., name=...)` signature**
`tape_builder.py`. The new signature ignores those kwargs and routes them into `**kwargs` as ONNX attributes — producing a wrong graph with no error. Internal rules were migrated, but third-party rewrite rule authors will silently lose. The PR body says "not yet released, no compat alias needed" — please double-check that's true for the public rewriter/version-converter extension API before merging. If there's any external surface, either add a compat shim (detect `inputs=`/`attributes=` and remap with a `DeprecationWarning`) or a CHANGELOG migration note.

**4. `OpBuilderBase` removed with no alias**
Same "not released" rationale. Trivial to keep `OpBuilderBase = BuilderBase` as a one-line transitional alias.

**5. Missing tests for the new `BuilderFeature` matrix and multi-output contract**
Only default `TapeBuilder()` behavior is exercised. Add focused tests for partial flag combinations (e.g. `SCHEMA_PARTITION` only, `CAST_INPUTS` only) and for the new multi-output return shape (`op.op("Split", x, _outputs=3)` returning `Sequence[ir.Value]` vs the old single-`ir.Value` contract).

## 🟡 Minor

**6. Name-determinism shift.** `_cast_inputs` now runs before `_adapt_outputs` / `_generate_node_name`, so promoted CastLike/Constant sub-nodes advance `graph.num_nodes()` before the outer node names itself. Old: `Add_0` / `Add_node_0`. New: `Add_1` / `Add_node_1`. Won't corrupt graphs but may break snapshot tests or debug-name expectations. Either snapshot the count up-front (pass it to both hooks) or accept-and-document the new contract.

**7. Recursive `CastLike` loses opset context.** `_input_to_ir_value` calls `self.call_op("CastLike", ..., domain="", version=None)` instead of going through the graph-bound op builder. For graphs targeting older opsets, schema lookup will miss. Suggest a `_make_cast_like(value, like_type)` hook so `GraphBuilder` can supply its ambient version.

**8. `_domain` / `_version` / `_outputs` / `_name` underscore convention is undocumented.** Add one sentence to `op()`'s docstring explaining the prefix exists to avoid clashing with ONNX attribute names passed via `**kwargs`. Non-obvious to a first-time rule author.

**9. Duplicated helpers.** `_dtype_suffix` and `_constant_name` are essentially identical in `builder.py` and `tape_builder.py`. Import from one source to avoid drift.

**10. `BuilderFeature.any_schema_feature` is a redundant shadow of `SCHEMA_AWARE`.** Single call site; inline `features & SCHEMA_AWARE` and delete the property to eliminate drift risk if a new schema flag is later added to `SCHEMA_AWARE`.

## ✅ Praise

The feature-flag pipeline is a good architectural direction — it centralizes schema/cast/inference/storage that was duplicated across builders, and the hook-based extension points are well-chosen. Section dividers in `BuilderBase` and the "Override in `GraphBuilder` to …" pattern in hook docstrings are excellent for onboarding. The 11-step numbered pipeline in `call_op` is easy to follow.

## Follow-up suggestions (out of scope)

- Rename `tape_builder.py` since it now hosts `BuilderBase` as its primary export.
- Restore the `BuilderBase` class docstring to call out `__getattr__` dispatch as a distinct creation path (the bundled "two operations" wording hides it).
- Consider inlining `OpBuilder._call_op` (3-line body, naming-collision with `BuilderBase.call_op`).

## Recommendation

Request changes, focused on **#1 and #2** (real bugs). Then re-discuss **#3/#4** (compat policy) and **#5** (tests). Rest can land in follow-ups.

onnxscript/_internal/tape_builder.py
```diff
@@ -62,92 +119,349 @@ def _record_opset(self, domain: str, version: int | None) -> None:
...
+    def _promote_constant(self, value: Any, dtype: ir.DataType | None) -> ir.Value:
+        """Convert a Python constant into an ir.Value via a Constant node.
+
+        Creates a ``Constant`` op node whose output carries the tensor value.
+        This avoids initializer-name collisions when the builder is used
+        inside the rewriter/optimizer.
+
+        GraphBuilder overrides this with a cache-based initializer strategy.
+        """
+        tensor = ir.tensor(value, dtype=dtype)
```
xadupre, line 266: dtype can be None, the docstring does not say what it does in that case.

onnxscript/_internal/tape_builder.py
```diff
@@ -62,92 +119,349 @@ def _record_opset(self, domain: str, version: int | None) -> None:
...
+                domain,
+                op_type,
+                args,
+                attributes=attrs,
+                outputs=output_values,
+                version=version,
+                name=name,
+            )
+        else:
+            num_outputs = len(outputs) if isinstance(outputs, Sequence) else outputs
```
xadupre: I think it worth asserting it is an int. The annotation says it can be string as well. Maybe it is handled before.

onnxscript/_internal/tape_builder.py
```diff
@@ -62,92 +119,349 @@ def _record_opset(self, domain: str, version: int | None) -> None:
...
+        # 9. Store
         self._add_node(node)
         self._record_opset(domain, version)
-        return node.outputs[0]
+
+        # 10. Post-creation hooks (inference, const-prop)
+        if features & BuilderFeature.CONSTANT_PROPAGATION:
+            self._constant_propagation(node)
+        if features & BuilderFeature.INFER_SHAPES:
+            self._infer_shapes(node)
```
xadupre, line 420: Do you need to infer the shape if it became a constant? (If it is a big one, let's avoid serializing)

gramalingam:
Addressed comments from Titai, except for 3, 4, and 5. 3 and 4 are as planned, so not an issue. 5 is doable, but not important right now, I think. Will consider it later.

onnxscript/_internal/tape_builder.py
gramalingam: Good point. The underlying inference method should do this. (I will look into it, either in this PR or next.)

gramalingam: Let me do that as a separate PR, since it is a bit involved, and orthogonal to current PR.

onnxscript/_internal/tape_builder.py
gramalingam: Done

justinchuby (approved):

onnxscript/_internal/tape_builder.py
gramalingam: Ok, done, this part was simple enough to include here.

justinchuby (approved):

PR #2922, opened 2026-05-21
fix: invoke NameFixPass after graph-modifying operations in rewriter, version-converter, and optimizer

gramalingam:
## Problem

TapeBuilder creates values with explicit names that can clash with existing graph value names when nodes are inserted via replace_nodes_and_values. NameAuthority silently registers duplicate names without conflict checking, causing ValueError when the clashing value is an initializer.

## Fix

Invoke NameFixPass conditionally (only when changes were actually made) at the end of the three graph-modifying callers:

- RewriteRuleSet.apply_to_model -- runs when count > 0
- _VersionConverter.visit_model -- runs when self._modified is True (set in replace_node)
- FoldConstantsPass.call -- runs when self._modified is True

Each call site includes an explanatory comment and the _modified check is encapsulated within the class that owns it.

## Tests

Added name-clash regression tests to all three test suites:

- NameClashAfterRewriteTest in onnxscript/rewriter/pattern_test.py
- NameClashAfterConversionTest in onnxscript/version_converter/_version_converter_test.py
- NameClashAfterFoldTest in onnxscript/optimizer/_constant_folding_test.py

Each test triggers the operation, injects a duplicate value name to simulate what TapeBuilder can produce via NameAuthority, and asserts all value names are unique after the operation.

## Known limitation

There is a pre-existing terminal NameFixPass in the optimizer pipeline (_optimizer.py), so optimize_ir may call NameFixPass up to 5 times on a 2-iteration run. This is acceptable overhead and is tracked as a follow-up to investigate deduplication.

### DIFF
--- a/onnxscript/optimizer/_constant_folding.py
+++ b/onnxscript/optimizer/_constant_folding.py
@@ -22,6 +22,7 @@ import numpy as np
 import onnx
 import onnx.reference.ops
 import onnx_ir as ir
+import onnx_ir.passes.common as ir_passes_common
 
 import onnxscript.utils.utils as utils
 from onnxscript._internal.tape_builder import BuilderBase, TapeBuilder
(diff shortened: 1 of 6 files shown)


justinchuby (approved):

justinchuby:
@gramalingam is this still needed?

gramalingam:
> @gramalingam is this still needed?

Merged it, thanks. It might be worthwhile at some point to pin down the invariants we want to guarantee (like name uniqueness) to reduce redundant passes, but this seems safer right now

PR #2935, opened 2026-06-14
[torchlib] Fix aten_conv3d generating a 2D bias when bias is missing

### LINKED ISSUE #2931: aten::conv3d onnx export bug when bias is missing
This issue has not repro case, it just points out the issue.

aten::conv3d has the following code when bias is missing:
https://github.com/microsoft/onnxscript/blob/c72e810b5806f4ace9d747db8ec798387899d04a/onnxscript/function_libs/torch_lib/ops/core.py#L2128-L2132

This code is copy-pasted for the aten::conv3d_complex code:
https://github.com/microsoft/onnxscript/blob/c72e810b5806f4ace9d747db8ec798387899d04a/onnxscript/function_libs/torch_lib/ops/core.py#L2173-L2177

and it therefore generates a 2D bias, which is incorrect for non-complex conv3d. This will cause downstream issues.

the aten::conv1d and aten:::conv2d onnx export is correct, and that code should be used instead:
https://github.com/microsoft/onnxscript/blob/c72e810b5806f4ace9d747db8ec798387899d04a/onnxscript/function_libs/torch_lib/ops/core.py#L2038-L2042

javierdejesusda:
The `bias is None` branch of `aten_conv3d` was copy-pasted from `aten_conv3d_complex`, so it builds
the bias shape with `Concat(weight_dim_0, [2])`, producing a rank-2 bias `[out_channels, 2]`. ONNX
`Conv` requires a 1D bias `[out_channels]`. `aten_conv1d` and `aten_conv2d` already use the correct
pattern.

```diff
-        bias_shape = op.Concat(weight_dim_0, op.Constant(value_ints=[2]), axis=0)
+        bias_shape = op.Expand(weight_dim_0, op.Constant(value_ints=[1]))
```

`aten_conv3d_complex` is left unchanged - its rank-2 bias is correct for the real/imag pair.

Adds `test_conv3d_without_bias_produces_1d_bias`, which asserts the synthesized Conv bias is rank 1.
It fails (`2 != 1`) before the fix and passes after. (`assert_onnx_program` alone misses this because
the rank-2 bias broadcasts numerically.)

Fixes #2931

### DIFF
--- a/tests/function_libs/torch_lib/e2e_ops_tests.py
+++ b/tests/function_libs/torch_lib/e2e_ops_tests.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import math
 import unittest
 
+import onnx
 import parameterized
 
 # TODO(pytorch/pytorch#129279): Migrate these tests to the PyTorch repo
@@ -1003,6 +1004,32 @@ class TorchLibe2eTest(unittest.TestCase):
                 expected = model(*inputs)
                 got = onnx_program.call_reference({"x": inputs[0]})
                 torch.testing.assert_close(expected, got[0])
+
+    def test_conv3d_without_bias_produces_1d_bias(self):
+        class Model(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.conv = torch.nn.Conv3d(3, 4, kernel_size=2, bias=False)
+
+            def forward(self, x):
+                return self.conv(x)
+
+        onnx_program = torch.onnx.export(
+            Model().eval(), (torch.randn(1, 3, 8, 8, 8),), dynamo=True, optimize=False
+        )
+        _testing.assert_onnx_program(onnx_program)
+
+        # The bias synthesized for a bias-less conv must be 1D ([out_channels]) to
+        # match the ONNX Conv spec. See https://github.com/microsoft/onnxscript/issues/2931.
+        inferred = onnx.shape_inference.infer_shapes(onnx_program.model_proto, data_prop=True)
+        shape_ranks = {
+            value_info.name: len(value_info.type.tensor_type.shape.dim)
+            for value_info in inferred.graph.value_info
+            if value_info.type.tensor_type.HasField("shape")
+        }
+        conv_nodes = [node for node in inferred.graph.node if node.op_type == "Conv"]
+        self.assertEqual(len(conv_nodes), 1)
+        self.assertEqual(shape_ranks[conv_nodes[0].input[2]], 1)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


javierdejesusda:

tests/function_libs/torch_lib/e2e_ops_tests.py
```diff
@@ -1004,6 +1005,32 @@ def forward(self, x):
...
+        # match the ONNX Conv spec. See https://github.com/microsoft/onnxscript/issues/2931.
+        inferred = onnx.shape_inference.infer_shapes(onnx_program.model_proto, data_prop=True)
+        shape_ranks = {
+            value_info.name: len(value_info.type.tensor_type.shape.dim)
+            for value_info in inferred.graph.value_info
+            if value_info.type.tensor_type.HasField("shape")
+        }
+        conv_nodes = [node for node in inferred.graph.node if node.op_type == "Conv"]
+        self.assertEqual(len(conv_nodes), 1)
+        self.assertEqual(shape_ranks[conv_nodes[0].input[2]], 1)
```
justinchuby, line 1032: Ok to review this part as assert_onnx_program will already run the model and validate the shapes.

tests/function_libs/torch_lib/e2e_ops_tests.py
```diff
@@ -5,6 +5,7 @@
 import math
 import unittest
 
+import onnx
```
justinchuby, line 8:
```suggestion
```

tests/function_libs/torch_lib/e2e_ops_tests.py
javierdejesusda: Before I drop it: the synthesized bias is all zeros, so `assert_onnx_program` passes even with the rank-2 bias. The conv output is numerically identical, so only the rank check fails (`2 != 1`) before the fix, which is the part that actually pins this regression. If you'd rather rely on `assert_onnx_program` for path coverage here, I'll remove the block and the `onnx` import to match the convention in the suite.

justinchuby:
The fix is right — the old `Concat(weight_dim_0, [2])` produced a rank-2 `[out_channels, 2]` bias (copy-pasted from the complex variant); switching to `Expand(weight_dim_0, [1])` yields the correct 1-D `[out_channels]` bias, matching the `aten_conv1d`/`aten_conv2d` pattern. The complex variant is correctly left as `Concat[2]`, and the bias-present path is untouched. The regression test asserting Conv `input[2]` rank==1 is the right guard (`assert_onnx_program` alone misses it since a rank-2 bias broadcasts numerically). The 2 red jobs (py312 / py311-onnx-ir-git) are infra runner cancellations ("received a shutdown signal"), not your change — a re-run should green them. LGTM.

bas-aarts:
What is the reason the PR cannot be merged? It's a rather simple fix. Thanks.

tests/function_libs/torch_lib/e2e_ops_tests.py
justinchuby: I see. In that case could you drop onnx and use onnx-ir apis for the comparison?

justinchuby:
Looks like there is a merge conflict. @javierdejesusda could you update the branch?

PR #2936, opened 2026-06-15
Fix incorrect rank handling in aten_repeat_interleave_self_int ONNX export #2932

### LINKED ISSUE #2932: bug in onnx export of aten_repeat_interleave_self_int results in incorrect code
running this script:

```
import onnx
import torch

class MyModule(torch.nn.Module):
    def forward(self):
        return torch.repeat_interleave(torch.tensor([2]), 2, 0)

model = MyModule()
model.eval()
with torch.no_grad():
    torch.onnx.export(model, (), optimize=False).save("output.onnx")
onnx.shape_inference.infer_shapes_path("output.onnx", strict_mode=True)
```

results in the following error:

`onnx.onnx_cpp2py_export.shape_inference.InferenceError: [ShapeInferenceError] Inference error(s): (op_type:Identity, node name: node_repeat_interleave): [ShapeInferenceError] Inferred shape and existing shape differ in rank: (2) vs (1)`

the issue is this line of code https://github.com/microsoft/onnxscript/blob/c72e810b5806f4ace9d747db8ec798387899d04a/onnxscript/function_libs/torch_lib/ops/core.py#L8314.
This line checks the original rank, and not the current rank. It should therefore be:
`    if self_rank == 0:
`

```
onnx                          1.20.0
onnx-ir                       0.2.1
onnxscript                    0.7.0
torch                         2.11.0

```

PratikWayase:
## Summary
This PR fixes a `ShapeInferenceError` that occurs during the ONNX export of `torch.repeat_interleave` on 1D tensors. Additionally, it implements full support for the `dim=None` case, which was previously raising a `NotImplementedError`.

## Problem
When exporting a model using `torch.repeat_interleave` on a 1D tensor, the exporter incorrectly returned an `Identity` node with a rank of 2 instead of 1. This caused a strict shape inference failure:
```text
[ShapeInferenceError] Inferred shape and existing shape differ in rank: (2) vs (1)
```
## Solution

- **Handled `dim=None`**: Modified the logic in `core.py` to flatten the input tensor (`op.Reshape(self, [-1])`) and set `dim = 0` and `self_rank = 1` when `dim` is `None`. This allows the existing expansion logic to process it correctly.
- **Removed faulty shortcut**: Deleted the incorrect `if self_rank == 1: return op.Identity(tiled)` block. This block was returning a tensor with an inflated rank. The code now correctly falls through to the `final_shape` calculation and `op.Reshape` to ensure the output rank perfectly matches the input rank.
- **Updated Tests**: Removed the `.skip()` markers in `ops_test_data.py` that were intentionally bypassing tests for the `dim=None` case. This enables full test coverage for this newly supported scenario.

## Testing

- ✅ Verified the original reproducer script from #2932 now passes strict shape inference without errors.
- ✅ Ran `pytest tests/function_libs/torch_lib/ops_test.py -k "repeat_interleave" -v`, resulting in **4 passed, 0 failed**. The previously skipped `dim=None` test cases now execute and pass successfully.

## Related Issue

Fixes #2932

### DIFF
--- a/tests/function_libs/torch_lib/ops_test.py
+++ b/tests/function_libs/torch_lib/ops_test.py
@@ -324,38 +324,5 @@ common_device_type.instantiate_device_type_tests(
     TestOutputConsistencyFullGraph, globals(), only_for=["cpu", "cuda"]
 )
 
-def test_repeat_interleave_dim_none_regression():
-    """Regression test for repeat_interleave shape inference error when dim is None.
-    
-    Previously, this would raise:
-    InferenceError: Inferred shape and existing shape differ in rank: (2) vs (1)
-    """
-    import torch
-    import onnx
-    import tempfile
-    import os
-
-    class MyModule(torch.nn.Module):
-        def forward(self, x):
-            return torch.repeat_interleave(x, 2)
-
-    model = MyModule()
-    model.eval()
-    x = torch.tensor([2])
-    
-    with tempfile.TemporaryDirectory() as tmpdir:
-        onnx_path = os.path.join(tmpdir, "model.onnx")
-        
-        # Use standard torch.onnx.export which is compatible across PyTorch 2.x versions
-        torch.onnx.export(
-            model,
-            (x,),
-            onnx_path,
-            opset_version=18,
-        )
-        
-        #  raise an InferenceError if the fix is not applied
-        onnx.shape_inference.infer_shapes_path(onnx_path, strict_mode=True)
-
 if __name__ == "__main__":
     unittest.main()
--- a/tests/function_libs/torch_lib/ops_test.py
+++ b/tests/function_libs/torch_lib/ops_test.py
@@ -324,5 +324,38 @@ common_device_type.instantiate_device_type_tests(
     TestOutputConsistencyFullGraph, globals(), only_for=["cpu", "cuda"]
 )
 
+def test_repeat_interleave_dim_none_regression():
+    """Regression test for repeat_interleave shape inference error when dim is None.
+    
+    Previously, this would raise:
+    InferenceError: Inferred shape and existing shape differ in rank: (2) vs (1)
+    """
+    import torch
+    import onnx
+    import tempfile
+    import os
+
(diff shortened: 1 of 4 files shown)


tests/function_libs/torch_lib/ops_test.py
```diff
@@ -324,5 +324,38 @@ def test_complex_output_match_opinfo_(
     TestOutputConsistencyFullGraph, globals(), only_for=["cpu", "cuda"]
 )
 
+def test_repeat_interleave_dim_none_regression():
```
justinchuby: Please follow the rest of the tests in the e2e file for adding a test.

justinchuby:
The rank fix is correct — removing the `if self_rank == 1: return op.Identity(tiled)` early-return (which returned a rank-2 result and caused the (2)-vs-(1) shape-inference error) lets it fall through to the final_shape Concat+Reshape, giving the correct rank-1 output. Traced dim=None and dim=0/1 cases for 1D/2D all match `torch.repeat_interleave`. The removed `.skip(dim is None)` entries now run real OpInfo numerical parity, plus the new e2e `test_repeat_interleave_int_dim_none`. The 3 red CI jobs are all `py311-torch-nightly` failing on `test_output_match_opinfo__logit_*` — an unrelated torch-nightly numerical flake (this PR doesn't touch logit; same fails on main). Minor non-blocking nit: the docstring (~core.py L8278) lost its blank lines so the `.. code-block:: python` RST is now malformed — worth restoring. LGTM otherwise.

bas-aarts:
What is the reason this PR is not progressing? It's a rather simple fix. Thanks.

PratikWayase:
> What is the reason this PR is not progressing? It's a rather simple fix. Thanks.

Thanks for following up. The remaining blocker appears to be unrelated CI failures (torch-nightly) and canceled CI runs, not the implementation itself. If a maintainer could rerun the affected checks, that would help move this PR forward.

justinchuby (approved):

PR #2941, opened 2026-06-18
Fix for version converter not applying applicable conversion

### LINKED ISSUE #2938: Version converter does not apply applicable conversion
For ONNX Ops, which only have a few operator versions available, the version upgrade does not work reliably:

Example:
Assume that the input graph is in version 17 (specific versions are not important, this is an issue for all versions) and should be converted to version 18. The graph contains a node, which only has operator version 13 and 18 (Again, specific versions are not important, just used for the example.), like [ReduceLogSumExp](https://onnx.ai/onnx/operators/onnx__ReduceLogSumExp.html). Therefore, the operator in the graph is still in version 13, because v18 is not available in a graph of version 17.
A conversion function for this operator would now implement a conversion from 13 to 18, because these are the closest available versions of the operator and be registered as `@register("ReduceLogSumExp", node_version=13, up_conversion=True)` like these other conversion functions: https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L157

If the version converter now runs with a model of version 17 as input, the converter tries to look up a conversion function here: https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L255
where the `from_version` is set to the model version in this line
https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L313
because `node.version` is usually not set for most ONNX models.
As a result, the registry tries to look up a conversion from v17 to v18, which does not exist, because there is no v17 version of the Op, which causes the conversion not to run.

Possible Solution:
One possible solution could be to switch the conversion function lookup from the source version to the target version. For each target version, there should exist at most one conversion function. If a conversion function exists, then the operator needs to be upgraded and if there is no conversion function, then the old operator from the previous version is left in the model.

LinusJungemann:
This PR fixes #2938.
Done by changing the way in which the registration of converter functions works. 
The version in the decorator represents the target version now.
Because there can only ever be one converter function with same target version and direction combincation, this lookup is unique and will correctly lookup the conversion from the last supported op version, even if multiple model versions are skipped between op versions.

### DIFF
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -86,10 +86,10 @@ class AdapterRegistry:
         self,
         domain: str,
         opname: str,
-        original_version: int,
+        target_version: int,
         up_conversion: bool = True,
     ) -> AdapterFunction | None:
-        adapter_func = self.op_adapters.get((domain, opname, original_version, up_conversion))
+        adapter_func = self.op_adapters.get((domain, opname, target_version, up_conversion))
         if adapter_func is not None:
             return adapter_func
         return None
@@ -97,7 +97,7 @@ class AdapterRegistry:
     def register(
         self, opname: str, domain: str = "", node_version=None, up_conversion=True
     ) -> Callable[[AdapterFunction], AdapterFunction]:
-        """Register an adapter based on the domain, operator type, node version and whether to upgrade/downgrade node version"""
+        """Register an adapter based on the domain, operator type, target node version and whether the target version is a result of an up-conversion or down-conversion."""
 
         def decorator(function: AdapterFunction) -> AdapterFunction:
             @functools.wraps(function)
@@ -154,7 +154,7 @@ def _get_str_attribute(node: ir.Node, name: str, default: str | None = None) -> 
 # Opset 19 -> 20
 
 
-@register("DFT", node_version=19, up_conversion=True)
+@register("DFT", node_version=20, up_conversion=True)
 def dft_19_20(node: ir.Node, op):
     input = node.inputs[0]
     dft_length = node.inputs[1] if len(node.inputs) > 1 else None
@@ -167,7 +167,7 @@ def dft_19_20(node: ir.Node, op):
     return None
 
 
-@register("GridSample", node_version=19, up_conversion=True)
+@register("GridSample", node_version=20, up_conversion=True)
 def gridsample_19_20(node: ir.Node, op):
     x = node.inputs[0]
     grid = node.inputs[1]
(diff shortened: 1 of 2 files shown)


LinusJungemann:
@microsoft-github-policy-service agree company="Paderborn Center for Parallel Computing"

justinchuby:
I don't think this is the right fix, as from target version we have just the same information. Each function corresponds to one opset pair.

The core issue here is the model supplied is opset 17, which is not supported by this converter. We can add support for that.

justinchuby (changes requested):
As commented

LinusJungemann:
This is might be part of the problem, but I think the underlying problem is still how the converters are selected independent of opset version:
Take the [Acos](https://onnx.ai/onnx/operators/onnx__Acos.html) operation as an example:
If we have an model with onnx opset 19 (which is supported by the version converter already), then the Acos version will be v7. 
If we now upgrade the model to onnx opset 22, then this loop will loop from 19 to 21:
https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L330
`node_version` will not be set to 7, because `node_version` will be set to `self._default_onnx_opset`, because `node.version` is not set for a default onnx model. 
`self._default_onnx_opset` is set by
https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L341
which in turn is set to the model version:
https://github.com/microsoft/onnxscript/blob/5989b563c44f09be794ba6cad636e56e6f021ace/onnxscript/version_converter/_version_converter.py#L35

If we now have our conversion function registered like this
`@register("Acos", node_version=7, up_conversion=True)`
then the version converter loop will never find it, because we start at 19. This is wrong, because operator version 7 belongs to model version 7-21. We have to directly upgrade from operator version 7 to operator version 22.
So the upgrade has to only trigger if the version to which we want to upgrade is 22. Because all other upgrades are noops as operator version 7 is still the newest one.

In my opinion, the easiest way to achieve this behavior is not to look at the source version, but the target version. Then we get for an up_conversion registered for opset 22 automatically support for all model source opsets from 7-21 (or 18-21 if the old opset bound for the converter is kept).

I tested the change by implementing an version converter for ReduceLogSumExp in onnx-passes: https://github.com/iksnagreb/onnx-passes/pull/4

An alternative approach would probably be to register the nodes not based on operator version, but based on the last model opset, which they support before an upgrade. So for the example Acos Op, this would then be `@register("Acos", needs_upgrade_after_opset=21, up_conversion=True)` or something along this line...

justinchuby:
Thanks for your analysis! Will take a closer look tomorrow

Issue #2945, opened 2026-06-22
Bug for reshape->reshape pattern which include -1


OYCN:
repro script:
```
"""Minimal reproduction of onnxscript ReshapeReshape rewrite bug.

When two Reshape->Reshape chains share the same shape constant containing -1,
the ReshapeReshape rule resolves -1 to different concrete values per match,
but creates new initializers reusing the original name (name=shape.name).

Both rewrites write to graph.initializers["shared_shape"], so only the last
one survives. NameFixPass then renames the orphaned Value to "shared_shape_1"
to ensure uniqueness, but does NOT register it as an initializer. The result:
a Reshape node references "shared_shape_1" which is neither a graph input nor
an initializer nor a node output -> toposort validation failure.

Bug location: onnxscript/rewriter/rules/common/_basic_rules.py
    class ReshapeReshape.rewrite():
        new_shape = op.initializer(ir.Tensor(self._new_shape, name=shape.name))
                                                              ^^^^^^^^^^^^^^
    Reusing the shared input name causes both rewrites to clobber the same
    initializer dict entry. The loser becomes an unregistered dangling Value.

Fix: generate a unique name instead of reusing the shared input name, e.g.:
        new_shape = op.initializer(ir.Tensor(self._new_shape))
"""

import logging

import numpy as np
import onnx
import onnx_ir as ir
from onnxscript import rewriter
from onnxscript.optimizer import _constant_folding, common_passes

# Suppress onnx_ir deserialization warnings (known from_onnx_text limitation:
# initializers are attached after parsing, so their names aren't in scope yet).
logging.getLogger("onnx_ir").setLevel(logging.ERROR)

MODEL_TEXT = """\
<ir_version: 9, opset_import: ["" : 21]>
test (float[2, 3] input1, float[2, 6] input2) => (float[2, 3] out1, float[2, 6] out2) {
    mid1 = Reshape (input1, shape_mid_a)
    mid2 = Reshape (input2, shape_mid_b)
    out1 = Reshape (mid1, shared_shape)
    out2 = Reshape (mid2, shared_shape)
}
"""


def build_model() -> ir.Model:
    """Two Reshape->Reshape chains sharing shape=[2, -1].

    input1 [2,3] -> Reshape([6]) -> Reshape([2,-1]) -> out1 [2,3]  (-1 resolves to 3)
    input2 [2,6] -> Reshape([12]) -> Reshape([2,-1]) -> out2 [2,6] (-1 resolves to 6)
[...]

PR #2946, opened 2026-06-22
fix: prevent initializer name collision in rewrite rule registration

OYCN:
## Summary

> https://github.com/microsoft/onnxscript/issues/2945

Fix a bug where multiple rewrite rule matches producing initializers with the same name would silently clobber each other, resulting in a dangling `ir.Value` reference and a toposort validation failure on serialization.

## Problem

When two `Reshape→Reshape` chains share the same shape initializer (e.g. containing `-1`), the `ReshapeReshape` rule fires twice, each time calling `op.initializer(ir.Tensor(..., name=shape.name))`. Both produce an initializer with the same name but different resolved values.

The registration logic in `_rewrite_rule.py` had two independent for-loops:

```python
# Loop 1: check for duplicates (dead code — `continue` only skips this loop)
for initializer in delta.new_initializers:
    if initializer.name in initializers:
        continue

# Loop 2: unconditionally overwrite
for initializer in delta.new_initializers:
    initializers[initializer.name] = initializer
```

The second loop always overwrote the dict entry, so the first registered Value lost its dict reference. `NameFixPass` would then rename it to ensure uniqueness, but never re-registered it as an initializer — leaving it as a dangling reference that fails `onnx.checker`.

## Fix

Merge into a single loop. On name conflict with a different Value object, auto-generate a unique suffix (`_1`, `_2`, ...) and register under the new name. Since node inputs hold object references (not name strings), the rename propagates correctly on serialization.

## Test

Added `test_reshape_reshape_shared_shape_no_clobbering` covering the shared-shape scenario. Asserts that all shape Values are registered initializers and the serialized model passes `onnx.checker.check_model(full_check=True)`.

## Related

Also filed a separate issue/PR in [onnx-ir](https://github.com/onnx/ir-py/pull/449) for `NameFixPass` not registering renamed Values as initializers — that's an independent bug exposed by the same scenario.

### DIFF
--- a/onnxscript/rewriter/_rewrite_rule.py
+++ b/onnxscript/rewriter/_rewrite_rule.py
@@ -712,10 +712,21 @@ class RewriteRuleSet:
                     initializers = graph_or_function.initializers
                     for initializer in delta.new_initializers:
                         if initializer.name in initializers:
+                            existing = initializers[initializer.name]
+                            if existing is initializer:
+                                # Same Value already registered, skip.
+                                continue
+                            # Name conflict with a different Value: generate a unique name.
+                            base_name = initializer.name
+                            counter = 1
+                            while f"{base_name}_{counter}" in initializers:
+                                counter += 1
                             if verbose:
-                                print(f"Initializer {initializer.name} already exists.")
-                            continue
-                    for initializer in delta.new_initializers:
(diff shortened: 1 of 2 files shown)


OYCN:

PR #2961, opened 2026-07-12
Fix optional input positions in converter

### LINKED ISSUE #2007: Translation of inputs to Resize, when using 'sizes' argument, drops required inputs
Hi,

I found a bug happening when translating a Resize op that uses a 'sizes' argument.
The required input arguments 'rois' and 'scales' are being dropped, since they are None, but they are "required empty" by onnxruntime.

The strange thing is that the translation when running 'from function' is OK, but exporting the model misses the inputs.

Here is the minimal working example:
```python
import numpy as np
import onnx
import onnxruntime as ort
from onnxscript import FLOAT
from onnxscript import opset20 as op
from onnxscript import script


@script()
def resize(X: FLOAT[800, 600]) -> FLOAT[512, 512]:
    return op.Resize(X, sizes=[512, 512])


# checker is OK
onnx.checker.check_model(resize.to_model_proto())

# This works - skipped arguments `roi` and `scales` get correctly translated to "empty tensors"
print("From function - OK")
X = np.eye(800, 600, dtype=np.float32)
Y = resize(X)

# This doesn't - the `roi` and `scales` get omitted, which makes `sizes` a second argument, while it needs to be fourth.
try:
    print("\n\n\nBug - not OK \n==================")
    session = ort.InferenceSession(resize.to_model_proto().SerializeToString())
    Y = session.run(None, {"X": X})
except Exception as e:
    print(e)
    print("==================")

# workaround - add two empty arguments for `roi` and `scales`
args = resize.function_ir.stmts[-1].args
resize.function_ir.stmts[-1].args = (args[0], "", "", args[1])

# Now it's OK
print("\nWith workaround - OK")
session = ort.InferenceSession(resize.to_model_proto().SerializeToString())
Y = session.run(None, {"X": X})

``` 

And the output of it:
```
From function - OK

Bug - not OK 
==================
[ONNXRuntimeError] : 10 : INVALID_GRAPH : This is an invalid model. Type Error: Type 'tensor(int64)' of input parameter (const) of operator (Resize) in node (n1) is invalid.
==================

With workaround - OK
```

I'm using onnxscript=0.1.0.dev20250108

And lastly - THANKS - onnxscript is such a great idea for writing pre/post-processing functions!

the0cp:
## Summary

- Preserve empty input positions when a later optional ONNX input is supplied
- Allow the converter to consume normalized `None` placeholders
- Add generic argument-partitioning tests
- Add a Resize conversion and ONNX Runtime regression test

Fixes #2007


the0cp:
Hi @justinchuby, could you please review this PR or point me to the right maintainer? Thanks!

the0cp:
The only failing jobs are the three Torch nightly jobs. These appear unrelated to this PR.

justinchuby:
@gramalingam

PR #2967, opened 2026-07-16
Expose the register as public register_constant_folder API (fix for Issue #2507)

### LINKED ISSUE #2507: Constant-Folding Registry
Authoring custom constant-folding for operators via the `register` method in `onnxscript.optimizer._constant_folding` feels quite convenient (see #2506 for example) - are there any plans to properly expose this to the user to make constant folding customizable? I mean, it is already possible to simply `from onnxscript.optimizer._constant_folding import register`, but this is importing from a protected memeber (at least the linter complains), so I guess it is not meant to be used this way, at least for now?

RohanMali2003:
## Summary

Exposes the internal constant-folding registration decorator as a public API: `onnxscript.optimizer.register_constant_folder`.

Previously, registering a custom constant-folding rule required importing the protected `register` function from `onnxscript.optimizer._constant_folding` directly, which linters flag as a private-member import. This re-exports it under a public, more descriptive name, following the same pattern used in #2773 for `FOLDED_FROM_KEY`.

Closes #2507

## Changes

- `onnxscript/optimizer/__init__.py`: re-export `_constant_folding.register` as `register_constant_folder`, added to `__all__`.
- `onnxscript/optimizer/_constant_folding.py`: added a docstring to `PartialEvaluatorRegistry.register` documenting the callback signature, parameters, and a usage example.
- `onnxscript/optimizer/_constant_folding_test.py`: added `test_register_constant_folder`, which registers a folder for a mock op under a dedicated test domain (`test.custom`), verifies `fold_constants` invokes it, and cleans up the registration in a `finally` block.

## Scope notes

- Only the `register` decorator is exposed; other `PartialEvaluatorRegistry` internals (e.g. `lookup_evaluators`) remain private.
- The callback signature references `OptimizerContext`/`OptimizerState`, which are currently internal types. I've documented the expected signature informally in the docstring rather than exposing those types — happy to expose them too if preferred.

## Testing

- `pytest onnxscript/optimizer/_constant_folding_test.py -v` — all passing
- `pytest --doctest-modules onnxscript/optimizer` — all passing
- `pytest tests/optimizer/test_models.py -v` — all passing (broader sanity check)
- `lintrunner` — clean

## Sidenote:

I developed and documented this with AI-assisted tooling; all changes were reviewed, and verified by running the full test suite, doctests, and lintrunner locally before submission.


RohanMali2003:

RohanMali2003:
Hi @justinchuby , gentle ping on this PR when you have a chance!

Also, I noticed the Copilot bot just flagged that the callback signature in the test and docstring still references the internal `OptimizerContext` and `OptimizerState` types. As I noted in the PR description, I held off on making those public to keep the scope tight, but I am more than happy to expose them as well if you'd prefer! Just let me know what direction you want to take. thanks!

justinchuby:
Thanks! This is a welcomed change but I may need to take some time before I can review because my plate is full with other work.

RohanMali2003:
No problem at all, and thanks for the heads-up, boss! Take your time, and I'll address any feedback whenever you get the chance to review. No rush! :)

gramalingam:
We should think through this: (a) Do we really want to expose this as a public API? (b) If so, is it in an appropriate form for freezing as part of a public API?

(I will take a look, just clarifying that this is worth thinking through)

gramalingam:
I added an AI-produced report to the issue discussion. TLDR summary: I think we should cleanup the internal implementation before making it public. The report in the issue gives some ideas how we could clean it up

PR #3006, opened 2026-08-21
Fix invalid ONNX graph from None start/end in aten_slice_scatter

### LINKED ISSUE #2372: [torchlib] slice_scatter does not support start and end to be None

gabrielfruet:
`aten::slice_scatter` produced an invalid ONNX graph whenever `start` or `end` came in as `None`. The lowering unsqueezed both bounds unconditionally, so a `None` turned into an empty-string input to `Unsqueeze`. A missing `start` now defaults to 0 and a missing `end` to `INT64_MAX`, a full slice, matching what `aten_slice` already does.

Changes:
- `torch_lib/ops/core.py`: build the `Slice` bounds conditionally, unsqueezing `start` or `end` only when they are provided.
- `tests/function_libs/torch_lib/extra_opinfo.py`: regression sample-input cases for `None` start, `None` end, both, and a step-2 variant.

Testing: the existing slice_scatter OpInfo suite now exercises the `None` bounds.

Fixes #2372

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -9130,10 +9130,8 @@ def aten_slice_scatter(
 ) -> TTensor:
     """slice_scatter(Tensor self, Tensor src, int dim=0, SymInt? start=None, SymInt? end=None, SymInt step=1) -> Tensor"""
 
-    # Although 'start' and 'end' can be None in signature, but actually 'start' must be specified
-    # Assert(start is not None)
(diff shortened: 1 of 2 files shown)


gabrielfruet:
@justinchuby Do you mind taking a look?

justinchuby (approved):
Thanks!

PR #2885, opened 2026-04-10
[torchlib] Add missing dtype parameter to aten_mean_dim

linusjuni:
Fixes #2884

 `aten_mean_dim` and `aten_mean_dim_complex` are missing the `dtype` keyword argument from their signatures, even though the ATen schema documents it (`ScalarType? dtype=None`). This causes a `TypeError` when PyTorch lowers `aten::mean.dim` with an explicit `dtype` - which happens for any model using `GlobalAveragePooling2D` (Keras/PyTorch).

- Add `dtype: int = -1` to `aten_mean_dim`, with `op.Cast` when dtype is specified
- Add `dtype: int = -1` to `aten_mean_dim_complex`, raising `NotImplementedError` for complex tensors

Follows the same pattern used by `aten_sum_dim_IntList` and `aten_sum_dim_IntList_complex`.

PR #3009, opened 2026-08-22
Honor the dtype argument of aten::mean when exporting without dim

### LINKED ISSUE #3008: aten::mean silently ignores dtype and exports the wrong output type
## Summary

While checking TorchLib operator signatures after reproducing #3002, I found that the no-dimension `aten::mean` implementation does not accept its schema's `dtype` argument. Unlike the random-operator failure, this export does not raise: it succeeds while silently producing a `FLOAT` ONNX output for `torch.mean(..., dtype=torch.float64)`.

## Environment

- ONNXScript current source: `0.7.2.dev20260822`
- PyTorch `2.10.0`
- ONNX `1.22.0`
- Python `3.14.2`
- macOS, Apple Silicon

## Steps to reproduce

```python
import onnx
import torch

class Model(torch.nn.Module):
    def forward(self, x):
        return torch.mean(x, dtype=torch.float64)

x = torch.tensor([[1e8, 1.0, -1e8]], dtype=torch.float32)
expected = Model()(x)
torch.onnx.export(Model(), (x,), "mean_dtype.onnx", dynamo=True)

model = onnx.load("mean_dtype.onnx")
actual_type = onnx.TensorProto.DataType.Name(
    model.graph.output[0].type.tensor_type.elem_type
)
print(expected.dtype, expected.item())
print(actual_type)
```

## Expected behavior

```text
torch.float64 0.3333333333333333
DOUBLE
```

The cast must happen before the reduction so accumulation uses the requested dtype.

## Actual behavior

```text
torch.float64 0.3333333333333333
FLOAT
```

The exported graph contains only `ReduceMean -> Squeeze`; no cast is emitted.

## Possible cause

`onnxscript/function_libs/torch_lib/ops/core.py` registers:

```python
@torch_op("aten::mean")
def aten_mean(self: TReal) -> TReal:
    """mean(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
```

The docstring and PyTorch schema include `dtype`, but the function signature and body omit it. The exporter therefore lowers the operation without the requested type instead of rejecting the call.

## Proposed solution

Accept `dtype: int = -1` and cast `self` before `ReduceMean` when it is supplied. I tested this temporary local change:

```python
def aten_mean(self: TReal, dtype: int = -1) -> TReal:
    if dtype != -1:
        self = op.Cast(self, to=dtype)
    result = op.ReduceMean(self)
    return op.Squeeze(result)
```

The same reproduction then exported `Cast -> ReduceMean -> Squeeze` and declared the output as `DOUBLE`.

A regression test should assert both output type and a precision-sensitive value such as `[1e8, 1, -1e8]`, so a cast-after-reduction implementation cannot pass accidentally.

Om-singhaI:
# Honor the dtype argument of aten::mean when exporting without dim

Fixes #3008

## Summary

The no dim overload of `aten::mean` declares `dtype` in its schema (`mean(Tensor self, *, ScalarType? dtype=None)`), but `aten_mean` in `onnxscript/function_libs/torch_lib/ops/core.py` did not accept the argument. Exporting `torch.mean(x, dtype=torch.float64)` for a float32 input succeeded silently, emitted only `ReduceMean -> Squeeze`, declared a FLOAT output, and accumulated in float32. For the input `[[1e8, 1.0, -1e8]]` PyTorch returns `0.3333333333333333` as float64 while the exported model returned `0.0` as float32.

## Changes

- `aten_mean` is now `trace_only=True`, takes `dtype: int = -1`, and when a dtype is given casts `self` before `ReduceMean`. The no dtype path is unchanged (`ReduceMean -> Squeeze`).
- `aten_mean_complex` takes the same argument and raises `NotImplementedError` when it is supplied, matching `aten_mean_dim_complex` and `aten_sum_complex`.
- New `ops.aten.mean.dtype` OpInfo in `tests/function_libs/torch_lib/extra_opinfo.py` (`sample_inputs_mean_dtype`) registered against `core_ops.aten_mean` in `ops_test_data.py`. It yields `make_tensor` samples of shapes `(5, 5)`, `(5,)` and `()` plus the precision sensitive tensor `[[1e8, 1.0, -1e8]]`, all with `dtype=torch.float64`.

## Why the cast happens before the reduction

`aten_mean_dim` (#2885) and `aten_sum` cast the reduced result after the reduction. That is not sufficient here: PyTorch accumulates in the requested dtype, so for `[1e8, 1.0, -1e8]` the float32 mean is `0.0` (the `1.0` is lost when added to `1e8`) while the float64 mean is `1/3`. Casting after `ReduceMean` would produce a DOUBLE tensor holding `0.0`, which still mismatches PyTorch. Casting the input first reproduces PyTorch semantics. The new test includes this sample specifically so that a cast after reduction implementation cannot pass by accident.

## Verification

Reporter's script (`torch.onnx.export(..., dynamo=True)` of `torch.mean(x, dtype=torch.float64)` with float32 input) on pristine main at a39c0a5:

```
torch eager result: 0.3333333333333333 dtype: torch.float64
ONNX output elem_type: 1 (FLOAT), expected 11 (DOUBLE)
ORT result: 0.0 dtype: float32
```

With this change:

```
torch eager result: 0.3333333333333333 dtype: torch.float64
ONNX output elem_type: 11 (DOUBLE)
ORT result: 0.3333333333333333 dtype: float64
```

`pytest tests/function_libs/torch_lib/ops_test.py -k mean`:

| State | Result |
| --- | --- |
| Pristine main, without the new test | 8 passed, 48 skipped, 8 xfailed, 60 subtests passed |
| New test with the source change stashed | 4 failed, 10 passed, 48 skipped, 8 xfailed, 60 subtests passed. All four `ops_aten_mean_dtype` samples fail with `TypeInferenceError: Inferred elem type differs from existing elem type: (1) vs (11)` |
| New test with a cast placed after the reduction | 1 failed, 8 passed, 50 skipped, 8 xfailed, 63 subtests passed. Only the `[[1e8, 1.0, -1e8]]` sample fails: `Expected 0.3333333333333333 but got 0.0` |
| New test with this change | 8 passed, 50 skipped, 8 xfailed, 64 subtests passed |

The two additional skips in the fixed state are the function proto validity checks, which skip for traced functions.

`ruff check` and `ruff format --check` (ruff 0.15.1, the lintrunner pinned version) pass on the three changed files.

Environment: Python 3.10, torch 2.13.0 (CPU), onnx 1.22.0, onnxruntime 1.23.2, macOS arm64.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/core.py
+++ b/onnxscript/function_libs/torch_lib/ops/core.py
@@ -6350,21 +6350,33 @@ def aten_maximum(self: TTensor, other: TTensor) -> TTensor:
     return op.Max(self, other)
 
 
-@torch_op("aten::mean")
-def aten_mean(self: TReal) -> TReal:
+@torch_op("aten::mean", trace_only=True)
+def aten_mean(self: TReal, dtype: int = -1) -> TReal:
     """mean(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
+
+    if dtype != -1 and dtype is not None:
+        # Cast before reducing so that the accumulation happens in the requested
+        # dtype, matching PyTorch. Casting the result afterwards would keep the
+        # precision loss of the input dtype.
+        self = op.Cast(self, to=dtype)
 
     result = op.ReduceMean(self)
     return op.Squeeze(result)
 
 
 @torch_op("aten::mean", complex=True, trace_only=True)
-def aten_mean_complex(self: TReal) -> TReal:
+def aten_mean_complex(self: TReal, dtype: int = -1) -> TReal:
     """mean(Tensor self, *, ScalarType? dtype=None) -> Tensor"""
 
     rank = len(self.shape) - 1
     dim = op.Constant(value_ints=list(range(rank)))
     result = op.ReduceMean(self, dim, keepdims=False)
+
+    if dtype != -1 and dtype is not None:
+        raise NotImplementedError(
+            "support for the dtype argument is not implemented for complex tensors"
+        )
+
     return result
 
 
--- a/tests/function_libs/torch_lib/extra_opinfo.py
+++ b/tests/function_libs/torch_lib/extra_opinfo.py
@@ -1206,6 +1206,27 @@ def sample_inputs_max_pool3d_with_indices(op_info, device, dtype, requires_grad,
         yield opinfo_core.SampleInput(arg, kwargs=kwargs)
 
 
+def sample_inputs_mean_dtype(op_info, device, dtype, requires_grad, **kwargs):
+    del op_info  # Unused
+    del kwargs  # Unused
+
+    make_arg = functools.partial(
+        torch_testing.make_tensor, device=device, dtype=dtype, requires_grad=requires_grad
+    )
+    for shape in ((S, S), (S,), ()):
+        yield opinfo_core.SampleInput(make_arg(shape), kwargs={"dtype": torch.float64})
+
+    # Precision sensitive values: accumulating in float32 gives 0.0 while accumulating
+    # in float64 gives 1/3, so an implementation that casts only the reduced result
+    # cannot pass this sample.
+    yield opinfo_core.SampleInput(
+        torch.tensor(
+            [[1e8, 1.0, -1e8]], dtype=dtype, device=device, requires_grad=requires_grad
+        ),
+        kwargs={"dtype": torch.float64},
+    )
+
+
 def sample_inputs_native_group_norm(op_info, device, dtype, requires_grad, **kwargs):
     del op_info
     make_arg = functools.partial(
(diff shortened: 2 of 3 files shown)


promiseeuler (approved):
While comparing this with the duplicate implementation I had opened, I checked out commit c48d820 locally and verified the surviving approach directly. Ruff check and format check pass on all three changed files. Exporting torch.mean([1e8, 1, -1e8], dtype=torch.float64) produces Cast -> ReduceMean -> Squeeze, returns float64, and matches PyTorch at 1/3. The OpInfo coverage is stronger than a single e2e regression, and accepting dtype on the complex overload while explicitly rejecting unsupported conversion keeps its schema behavior consistent. This looks correct to me.

justinchuby (approved):

justinchuby:
Thank you for the bug fix!

Om-singhaI:
Thanks for merging.

@promiseeuler, sorry for the duplicate effort there, and thanks for checking mine out against your own rather than just letting it go. Verifying a competing implementation and then backing it is more than anyone would expect.

If there are more torchlib lowerings worth going through, I am glad to keep picking them up.

PR #3015, opened 2026-08-27
[Optimizer] Collapse Reshape-Transpose-Reshape to Transpose

### LINKED ISSUE #1775: [Pattern] Reshape-Transpose-Reshape
image

SilasDarko:
## Summary

Implements #1775 by adding a conservative rewrite for:

`Reshape -> Transpose -> Reshape`

When the intermediate reshape only splits original axes into contiguous groups, the transpose moves those groups without reordering them internally, and the final reshape restores those groups, the sequence is replaced with a single `Transpose`.

The rule fails closed when those conditions cannot be proven from static shape information.

## Scope

This initial implementation supports fully static shapes only.

It intentionally skips chains containing size-0 or size-1 dimensions because those dimensions make axis-boundary provenance ambiguous in the current grouping logic. Symbolic or unknown shapes are also left unchanged.

The rule also validates both the input transpose permutation and the derived permutation before rewriting.

## Testing

Added coverage for:

- simple and multi-axis grouped rewrites
- non-trivial and identity permutations
- numerical equivalence
- axis-boundary crossing
- interleaved or internally reordered axis groups
- output shapes that appear valid but have different element ordering
- unknown and symbolic shapes
- size-1 dimensions
- malformed permutations
- integration with `MaterializeReshapeShape` through `optimizer.optimize()`

Local validation:

- 22 focused tests passed
- 443 relevant rewriter/optimizer tests passed, with 1 pre-existing skip
- Ruff checks passed
- ONNX checker and numerical-equivalence smoke tests passed

Fixes #1775

### DIFF
--- a/onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
+++ b/onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
@@ -1,0 +1,411 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+from __future__ import annotations
+
+import unittest
+
+import numpy as np
+import onnx
+import onnx.checker
+import onnx.helper as oh
+import onnxruntime as ort
+
+from onnxscript import ir, optimizer
+from onnxscript.rewriter import testing
+from onnxscript.rewriter.rules.common import _reshape_transpose_reshape as rtr
+
+
+def _dims_to_str(dims: list[int]) -> str:
+    return ", ".join(str(d) for d in dims)
+
+
+def _build_chain(
+    input_dims: list[int],
+    r1_dims: list[int],
+    perm: list[int],
+    output_dims: list[int],
+    *,
+    r1_shape_known: bool = True,
+    output_shape_known: bool = True,
+) -> ir.Model:
+    """Builds ``Reshape -> Transpose -> Reshape`` with non-constant shape
+    inputs (Shape() of a dummy tensor), then manually stamps the
+    shape-inference results this rule relies on -- mirroring
+    ``_materialize_reshape_shape_test.py``'s convention of simulating shape
+    inference having already run.
+    """
+    n1 = len(r1_dims)
+    n2 = len(output_dims)
+    output_sig = (
+        f"float[{_dims_to_str(output_dims)}] output" if output_shape_known else "float output"
+    )
+    text = f"""
+        <ir_version: 7, opset_import: [ "" : 17]>
+        agraph (float[{_dims_to_str(input_dims)}] data, int64[{n1}] shape1, int64[{n2}] shape2)
+            => ({output_sig})
+        {{
+            r1 = Reshape(data, shape1)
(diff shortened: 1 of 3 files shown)


SilasDarko:

SilasDarko:
Hi! Just following up on this PR when you have a chance. Most of the CI matrix is green, with the remaining failures isolated to the PyTorch-nightly jobs. Happy to address any feedback on the implementation or tests. Thanks!

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
```diff
@@ -0,0 +1,411 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+from __future__ import annotations
+
+import unittest
+
+import numpy as np
+import onnx
+import onnx.checker
+import onnx.helper as oh
```
justinchuby, line 10: Please avoid using onnx.helper. Use onnx ir apis to build any graphs.

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
```diff
@@ -0,0 +1,411 @@
...
+import onnxruntime as ort
+
+from onnxscript import ir, optimizer
+from onnxscript.rewriter import testing
+from onnxscript.rewriter.rules.common import _reshape_transpose_reshape as rtr
```
justinchuby, line 15:
```suggestion
from onnxscript.rewriter.rules.common import _reshape_transpose_reshape
```

avoid aliasing modules

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
```diff
@@ -0,0 +1,411 @@
...
+
+import unittest
+
+import numpy as np
+import onnx
+import onnx.checker
+import onnx.helper as oh
+import onnxruntime as ort
+
+from onnxscript import ir, optimizer
```
justinchuby, line 13:
```suggestion
import onnx_ir as ir
from onnxscript import optimizer
```

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
SilasDarko: Updated to use the ONNX IR APIs for graph construction. Thanks!

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
SilasDarko: Updated - removed the module alias and adjusted the references. Thanks!

onnxscript/rewriter/rules/common/_reshape_transpose_reshape_test.py
SilasDarko: Updated to import onnx_ir directly. Thanks!

PR #3023, opened 2026-08-29
Materialize the DFT axis default when converting opset 19 to 20

MohammedAlkindi:
Opset 19 defines DFT `axis` as an attribute defaulting to 1. Opset 20 moved it to an input defaulting to -2. `dft_19_20` read the attribute with a `None` default and returned early when it was absent, so the node was left untouched while the model opset was bumped. For any input of rank greater than 3 the converted model silently computes a different transform.

Measured on a rank-4 input with no `axis` attribute, numpy as ground truth, at the parent commit:

```
converted DFT inputs: ['input_x']
opset-19 result == numpy fft(axis=1): True
opset-20 result == numpy fft(axis=2): True
max abs diff: 4.825716018676758
```

onnx 1.22.0's own converter emits an axis Constant of 1 for this model, so the fix makes onnxscript agree with it.

The existing DFT tests all set `axis` explicitly, so the default path was never exercised. The added test omits it and fails on the unpatched source with `1 != 3`.

`version_converter`, `optimizer` and `rewriter`: 573 passed, 2 skipped, 6 xfailed.

The `None` guard stays: `_get_int_attribute` also returns `None` for a wrongly-typed attribute. Open PR #2941 changes only the `@register` decorator on this function, so the two do not overlap.

### DIFF
--- a/onnxscript/version_converter/_version_converter.py
+++ b/onnxscript/version_converter/_version_converter.py
@@ -160,11 +160,14 @@ def dft_19_20(node: ir.Node, op):
     dft_length = node.inputs[1] if len(node.inputs) > 1 else None
     inverse = _get_int_attribute(node, "inverse", 0)
     onesided = _get_int_attribute(node, "onesided", 0)
-    axis = _get_int_attribute(node, "axis", None)
-    if axis is not None:
-        axis_value = op.Constant(value_int=axis)
-        return op.DFT(input, dft_length, axis_value, inverse=inverse, onesided=onesided)
-    return None
+    # In opset 19 `axis` is an attribute defaulting to 1; in opset 20 it became an
+    # input defaulting to -2. An omitted attribute must therefore be materialized
+    # explicitly, or the converted node silently transforms a different axis.
+    axis = _get_int_attribute(node, "axis", 1)
+    if axis is None:
+        return None
+    axis_value = op.Constant(value_int=axis)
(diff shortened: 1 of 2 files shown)


justinchuby (approved):
Thanks!

justinchuby:
@MohammedAlkindi could you please fix lint

MohammedAlkindi:
Lint is fixed and pushed as a33e8bf. It was one duplicate blank line in `_version_converter_test.py`. With the repo's pinned ruff 0.15.1, `ruff format --check .` now reports 378 files already formatted and `ruff check onnxscript/version_converter/` passes.

On the red that remains: the three `py311-torch-nightly` jobs are failing on `main` too. Run 33344827505 on `main` fails exactly those three (ubuntu, macos, windows), so they look unrelated to this change. Happy to dig into them if you want.

MohammedAlkindi:
The `test (windows-latest, py311)` red looks like a flake rather than the change.

It passed on `00f94a2` and fails on `a33e8bf`, and the only difference between those two commits is one removed blank line in `_version_converter_test.py`. The failures are `conv2d_cpu_float16` tolerance assertions in `tests/function_libs/torch_lib/ops_test.py`, which whitespace in a version-converter test cannot reach.

I do not have rights to re-run the job, so it needs a nudge from your side. The three `py311-torch-nightly` failures are the pre-existing ones on `main` from earlier.

justinchuby (approved):

PR #3028, opened 2026-09-03
Improve error for misused positional argument (#1984)

### LINKED ISSUE #1984: informing about bad positional argument in node constructor

Alphaxiaoteng:
## Description

Calling an op with a non-Value positional argument that was meant to be an attribute/keyword — e.g. \`op.LeakyRelu(x, 1.0)\` where \`alpha\` should be a keyword — used to surface a cryptic error raised deep inside onnx_ir:

\`\`\`
AttributeError: 'float' object has no attribute '_add_usage'
\`\`\`

\`BuilderBase.call_op\` (the single point where every op dispatch builds its node) now validates positional inputs before constructing the node and raises a clear, actionable \`TypeError\` that names the op and suggests passing the value as a keyword argument:

\`\`\`
TypeError: LeakyRelu() got a non-Value positional argument 1.0 at position 2. If this was meant to be an attribute, pass it as a keyword argument (e.g. op.LeakyRelu(x, alpha=1.0)).
\`\`\`

When a schema is available, the message additionally suggests the specific attribute name.

## Why this location

The \`Node\`/\`Value\` IR was moved into the separate \`onnx_ir\` package, so the reporter's original patch (against \`ir/_core.py\`) no longer applies. \`call_op\` is the onnxscript-side equivalent of "the node constructor" and the right place to give users a readable error without touching the dependency.

## Notes / non-regression

- Legitimate scalar constants are **unaffected**: under the FULL feature set used by \`@script\` (which enables \`CAST_INPUTS\`), scalars are promoted to \`ir.Value\` constants *before* this validation runs, so \`op.Add(x, 1.0)\` still works.
- \`None\` inputs (optional inputs) and lists/tuples of \`ir.Value\` (grouped variadic inputs) are explicitly allowed.

## Tests

Added \`test_non_value_positional_argument_raises_clear_type_error\` to \`onnxscript/_internal/builder_test.py\`. The full \`builder_test.py\` suite (91 tests) passes.

Fixes #1984

### DIFF
--- a/onnxscript/_internal/builder_test.py
+++ b/onnxscript/_internal/builder_test.py
@@ -12,6 +12,7 @@ import onnx_ir as ir
 import onnxscript._internal.builder as builder
 import onnxscript.testing
 from onnxscript import script
+from onnxscript._internal.tape_builder import BuilderFeature, TapeBuilder
 from onnxscript.onnx_types import DOUBLE, FLOAT, INT64
 
 _default_opset_version = 23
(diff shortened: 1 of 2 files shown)


onnxscript/_internal/builder_test.py
```diff
@@ -1040,6 +1041,35 @@ def test_none_input_with_custom_domain(self):
...
+            TapeBuilder().LeakyRelu(x, 1.0)
+        message = str(cm.exception)
+        self.assertIn("LeakyRelu", message)
+        self.assertIn("keyword", message)
+        # Make sure we did not regress to the old cryptic AttributeError.
+        self.assertNotIn("_add_usage", message)
+
+        # The same clear error is produced for other ops (e.g. Add).
+        with self.assertRaises(TypeError):
+            TapeBuilder().Add(x, 1.0)
```
justinchuby, line 1065: You may use assertRaisesRegex to pin down the error message

onnxscript/_internal/builder_test.py
```diff
@@ -1040,6 +1041,35 @@ def test_none_input_with_custom_domain(self):
         self.assertIs(node.inputs[2], y)
         self.assertIsNotNone(result)
 
+    def test_non_value_positional_argument_raises_clear_type_error(self):
+        """GitHub issue #1984: passing a constant where an attribute/keyword was
+        expected (e.g. ``op.LeakyRelu(x, 1.0)``) used to raise a cryptic
+        ``AttributeError`` from onnx_ir (``'float' object has no attribute
+        '_add_usage'``). It should now raise a clear ``TypeError`` that tells
+        the user to pass the value as a keyword argument.
+        """
```
justinchuby, line 1050: Could you simplify the doctring to make it self-contained (no need to reference the issue)?

PR #3036, opened 2026-09-07
Fix aten_pow_scalar type promotion when the exponent is not floating point

Om-singhaI:
`aten_pow_scalar` casts the scalar base down to the exponent's dtype:

```python
return op.Pow(op.Cast(self, to=exponent.dtype), exponent)
```

When the base is a Python float and the exponent tensor is an integer, that casts the float down to int64 and builds an integer `Pow`. torch promotes the other way. A float scalar outranks an integral tensor, so `2.0 ** torch.tensor([1, 2, 3])` is float32.

Repro on main:

```python
import torch

class M(torch.nn.Module):
    def forward(self, x):
        return 2.0**x

x = torch.tensor([1, 2, 3])
onnx_program = torch.onnx.export(M(), (x,), dynamo=True, optimize=False)
onnx_program(x)
```

```
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL :
Type Error: Type (tensor(float)) of output arg (pow_1) of node (node_pow_1)
does not match expected type (tensor(int64)).
```

The graph contradicts itself. The output value carries `FLOAT` because that's the dtype torch reports for the result, while the `Cast` we emit is `to=7` and the `Pow` really returns int64. So it isn't just a wrong dtype, the model doesn't load at all.

A boolean exponent fails even earlier, for a float base and an int base both, because `Pow` accepts no boolean inputs:

```
[ONNXRuntimeError] : 10 : INVALID_GRAPH : This is an invalid model.
Type Error: Type 'tensor(bool)' of input parameter (val_1) of operator (Pow)
in node (node_pow_1) is invalid.
```

`aten_pow_tensor_scalar` right above already handles the mirror case correctly. It refuses to narrow and casts up to `FLOAT` instead. This applies the same rule in the other direction.

I promote to float32 when a float scalar meets an integral exponent, and to int64 when an int scalar meets a boolean one. Everything that already agreed with torch falls through to the original line untouched, so an int scalar over an int tensor still keeps the exponent's dtype and emits the same nodes as before.

Five tests in `e2e_ops_tests.py` next to the existing `test_pow_tensor_scalar_*` ones. Three of them fail on main (float base over int64, float base over bool, int base over bool). The other two pin the cases that must not move (float base over float16, int base over int64), and they pass either way.


Om-singhaI:
Is the `py311-torch-nightly` breakage a known one? It looks like it's already on main: [this run](https://github.com/microsoft/onnxscript/actions/runs/34070057714) fails the same three jobs and nothing else, on `addbmm`, `baddbmm`, `logit` and `_grouped_mm`. None of those go near `pow` as far as I can tell, and everything else here is green.

`Optional Lint` looks separate too, the misspell action's Docker build died on an expired Debian security release file. Anything you'd want me to do about either?

justinchuby (approved):
Thanks!

PR #3038, opened 2026-09-09
Fix numerical stability of TorchLib erfcx

### LINKED ISSUE #1223: [torchlib] Fix special.erfcx

YusefSyed:
Fixes #1223.

The current `exp(x*x) * (1 - erf(x))` decomposition loses the small erf complement for positive inputs and can combine overflow with underflow. For example, the existing OpInfo inputs around 6.8 return zero instead of approximately 0.08. This change uses piecewise Cephes rational approximations, a reciprocal positive tail, and the negative-input reflection formula. It promotes half inputs for the calculation, preserves float64 constants, and masks inactive approximation inputs because ONNX `Where` evaluates both branches. Exported graphs need neither `Erf` nor `Loop`.

The existing erfcx xfail is removed without relaxing its tolerance. New tests cover interval boundaries, tails, special values, dtypes, shapes and empty inputs. The coefficient provenance and BSD notice are included in both wheel and source distributions.

Validation:

- Focused regression and existing OpInfo tests: 20 passed, 4 expected skips; rerun immediately before submission.
- 291,148 actual ONNX Runtime output comparisons, with graph optimizations disabled and enabled, including every float16 bit pattern and dense float32/float64 inputs.
- Independent SciPy checks and 180 high-precision mpmath reference values.
- Actual `torch.onnx.export` and ONNX Runtime execution passed for float32 and float64; lint, formatting, wheel and source-distribution checks passed.

```sh
.venv/bin/python -m pytest tests/function_libs/torch_lib/special_test.py tests/function_libs/torch_lib/ops_test.py -k erfcx -q
```

The runtime validation was on macOS arm64. BFloat16 dispatch is included but was not executed in these runtime checks; no GPU or throughput claim is made.

AI disclosure: OpenAI Codex assisted with the implementation, regression tests, validation tooling and this PR description. The numerical coefficients are adapted from the attributed SciPy/XSF source, not presented as newly derived approximations.

### DIFF
--- a/onnxscript/function_libs/torch_lib/ops/special.py
+++ b/onnxscript/function_libs/torch_lib/ops/special.py
@@ -15,12 +15,86 @@ from __future__ import annotations
 import math
 from typing import Optional, Sequence
 
+from onnxscript import ir
+from onnxscript.function_libs.torch_lib.ops import common as common_ops
 from onnxscript.function_libs.torch_lib.registration import torch_op
 from onnxscript.function_libs.torch_lib.tensor_typing import TFloat
 from onnxscript.onnx_opset import opset18 as op
 from onnxscript.onnx_types import TensorType
 
 _MATH_PI = math.pi
+
+# Coefficients adapted from SciPy XSF's Cephes ``ndtr.h`` (pinned source revision
+# 5dbdff8de0dab99b475076612ea227d3f29d6cf6). The original Cephes Math Library
+# Release 2.2 is copyright Stephen L. Moshier (1984, 1987, 1988, 1992); SciPy's
+# C++ translation is BSD-3-Clause. See THIRD_PARTY_NOTICES.md for the license.
+_ERFCX_P = (
+    2.46196981473530512524e-10,
+    5.64189564831068821977e-1,
+    7.46321056442269912687,
+    4.86371970985681366614e1,
+    1.96520832956077098242e2,
+    5.26445194995477358631e2,
+    9.34528527171957607540e2,
+    1.02755188689515710272e3,
+    5.57535335369399327526e2,
+)
+_ERFCX_Q = (
+    1.0,
+    1.32281951154744992508e1,
+    8.67072140885989742329e1,
+    3.54937778887819891062e2,
+    9.75708501743205489753e2,
+    1.82390916687909736289e3,
+    2.24633760818710981792e3,
+    1.65666309194161350182e3,
+    5.57535340817727675546e2,
+)
+_ERFCX_R = (
+    5.64189583547755073984e-1,
+    1.27536670759978104416,
+    5.01905042251180477414,
+    6.16021097993053585195,
+    7.40974269950448939160,
+    2.97886665372100240670,
+)
+_ERFCX_S = (
+    1.0,
+    2.26052863220117276590,
+    9.39603524938001434673,
+    1.20489539808096656605e1,
+    1.70814450747565897222e1,
+    9.60896809063285878198,
+    3.36907645100081516050,
+)
+_ERFCX_T = (
+    9.60497373987051638749,
+    9.00260197203842689217e1,
+    2.23200534594684319226e3,
+    7.00332514112805075473e3,
+    5.55923013010394962768e4,
+)
+_ERFCX_U = (
+    1.0,
+    3.35617141647503099647e1,
+    5.21357949780152679795e2,
+    4.59432382970980127987e3,
+    2.26290000613890934246e4,
+    4.92673942608635921086e4,
+)
+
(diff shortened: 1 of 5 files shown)


onnxscript/function_libs/torch_lib/ops/special.py
```diff
@@ -15,13 +15,87 @@
...
+from onnxscript import ir
+from onnxscript.function_libs.torch_lib.ops import common as common_ops
 from onnxscript.function_libs.torch_lib.registration import torch_op
 from onnxscript.function_libs.torch_lib.tensor_typing import TFloat
 from onnxscript.onnx_opset import opset18 as op
 from onnxscript.onnx_types import TensorType
 
 _MATH_PI = math.pi
 
+# Coefficients adapted from SciPy XSF's Cephes ``ndtr.h`` (pinned source revision
```
justinchuby, line 27: Thanks. Is there a version that comes from pytorch?

YusefSyed: Codex-assisted response: yes. PyTorch has Steven G. Johnson's [`erfcx_y100` / `calc_erfcx` implementation](https://github.com/pytorch/pytorch/blob/2e2b1371931ea2a080c423ee3bc15c4fc59787a6/aten/src/ATen/native/Math.h#L1707-L2187). It uses a 100-region polynomial table for the central positive range and a continued-fraction tail. The [CPU kernel calls it](https://github.com/pytorch/pytorch/blob/2e2b1371931ea2a080c423ee3bc15c4fc59787a6/aten/src/ATen/native/cpu/UnaryOpsKernel.cpp#L619-L624); CUDA uses either [the corresponding Jiterator implementation or the same helper](https://github.com/pytorch/pytorch/blob/2e2b1371931ea2a080c423ee3bc15c4fc59787a6/aten/src/ATen/native/cuda/UnarySpecialOpsKernel.cu#L304-L317).

Those are different coefficients from the Cephes rational approximation used here. A PyTorch-sourced ONNX translation is feasible using coefficient lookup and Horner evaluation, but it would replace this approximation with the 700-coefficient table and its tail rules. I chose the current 42-coefficient rational form to avoid that lookup table in the export graph. Its SciPy provenance is explicit in the source/notice, and the focused erfcx tests still pass against PyTorch (20 passed, four expected skips). No performance comparison between the two ONNX representations has been established.

YusefSyed:
Codex-assisted CI follow-up for the unchanged head `c679e0c`:

- The torch-nightly addbmm/baddbmm/logit/grouped-mm failures also occur in the corresponding [base CI run](https://github.com/microsoft/onnxscript/actions/runs/34295172174).
- Optional Lint fails during Docker `apt-get update` on expired Debian `bullseye-security` metadata, before running lint; the [base Optional Lint run](https://github.com/microsoft/onnxscript/actions/runs/34265697030) also fails.
- The Ubuntu py310 job passed its first pytest phase, then received a runner shutdown signal during the second phase. Its log does not show a test assertion causing that cancellation.
- The Windows onnx-ir-git conv2d failures did not reproduce in this [controlled Windows recheck](https://github.com/YusefSyed/onnxscript/actions/runs/34396242151): parent `d1c005d` and head `c679e0c` ran on the same machine, dependency environment and seed. Both passed the float16 conv2d checks. The head also passed its erfcx OpInfo checks and all 19 dedicated erfcx regressions.

The comparison artifacts include source/import identity, dependency freeze, JUnit and logs. It pins ONNX IR to `641d90447cac737fabd43d94ea8c1796bde3e3b4`; the original job reported version 1.1.0 without exposing its installed Git SHA, so this is not a claim to reproduce that entire environment exactly. No conv code or test tolerance was changed. The original Windows job still needs an upstream rerun to establish its current status; I do not have permission to rerun that upstream job.

PR #3040, opened 2026-09-10
Fix optimize.py crash when the model path has no directory component

MohammedAlkindi:
`tools/optimize.py` calls `os.chdir(os.path.dirname(path))` unconditionally. For a bare filename `os.path.dirname` returns `''`, and `os.chdir('')` raises. That is the usage the module docstring documents:

```
Usage:
    python optimize.py model.onnx optimized_model.onnx
```

Before, run from the model's own directory against a real checker-valid model:

```
  File "tools/optimize.py", line 30, in main
    os.chdir(model_dir)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: ''
```

After, the same command exits 0 and writes `optimized_model.onnx`, which `onnx.checker.check_model` reloads clean. A path that does carry a directory component still works, checked from an unrelated cwd.

No test added: `tools/` has no coverage here. There is no `tools/*_test.py`, `noxfile.py`'s test session does not reach it, and `pyproject.toml` groups `tools` with `examples,docs,utils,opgen` as supporting code. Adding a test file to a directory the project does not test looked like scope creep rather than the requested fix.

`ruff check` and `ruff format --check` on the changed file: 0 violations. `lintrunner -a --take RUFF`: no lint issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/optimize.py
+++ b/tools/optimize.py
@@ -27,7 +27,8 @@ def main(args) -> None:
     # TODO: Remove this hack by fixing the optimizer to handle external data files properly.
     pwd = os.getcwd()
     model_dir = os.path.dirname(path)
-    os.chdir(model_dir)
+    if model_dir:
+        os.chdir(model_dir)
     model = onnxscript.optimizer.optimize(model)
     model = onnx.inliner.inline_local_functions(model)
     # Optimize again in case inlining created new opportunities.


tools/optimize.py
```diff
@@ -27,7 +27,8 @@ def main(args) -> None:
     # TODO: Remove this hack by fixing the optimizer to handle external data files properly.
     pwd = os.getcwd()
     model_dir = os.path.dirname(path)
-    os.chdir(model_dir)
+    if model_dir:
+        os.chdir(model_dir)
```
justinchuby, line 31:
```suggestion
model = onnx_ir.load(path)
```

could you instead update to this? This file has been outdated.

tools/optimize.py
```diff
@@ -27,7 +27,8 @@ def main(args) -> None:
...
     model = onnxscript.optimizer.optimize(model)
     model = onnx.inliner.inline_local_functions(model)
```
justinchuby, line 33:
```suggestion
```

tools/optimize.py
justinchuby, line 35:
```suggestion
```

justinchuby, line 38: construct the external_data path according to the output model name

```suggestion
onnx_ir.save(model, output_path, external_data=...)
```

MohammedAlkindi:
Done in 657f4656. `onnx_ir.load` resolves external data against the model file's own directory, so the hack and its TODO are gone along with the crash.

Two deviations worth flagging. `optimize()` already runs `InlinePass`, so the separate `inline_local_functions` call and the second optimize pass are dropped; on a model with nested custom-domain functions and a function inside an `If` subgraph, old and new output are byte-identical.

`external_data` is passed only when the input actually carries it. Unconditionally, onnx_ir's 256-byte threshold turns a 16 KB self-contained model into a 271-byte `.onnx` plus a 16 KB `.data`, and the `.onnx` alone then fails ORT with "External data path validation failed for initializer". Say if you would rather have it unconditional.

justinchuby (approved):

PR #3045, opened 2026-09-13
Fix IndexError in cast_inputs for ops with no formal inputs

bodapatisaikrishna:
## Motivation

`autocast.cast_inputs()` is the shared implementation behind both eager-mode
execution (`dynamic_cast_inputs`) and static script-translation
(`static_cast_inputs`). When it sees more positional arguments than the op's
formal input parameters, it's supposed to raise a clear error:

```python
raise ValueError(
    f"Number of actual parameters {len(args)} "
    f"exceeds number of formal parameters {len(expected_inputs)}."
)
```

But the branch that decides whether to raise that error indexes into
`expected_inputs[-1]` unconditionally:

```python
if i < len(expected_inputs):
    expected = expected_inputs[i]
elif expected_inputs[-1].variadic:   # <-- IndexError when expected_inputs == []
    ...
else:
    raise ValueError(...)
```

For an op with **zero formal input parameters** (only attributes  e.g.
`RandomNormal`, `RandomUniform`), `expected_inputs` is `[]`, so
`expected_inputs[-1]` raises an unhandled `IndexError: list index out of
range` instead of the intended `ValueError`.

This is reachable end to end through `values.Op.__call__`, which doesn't
pre-validate argument counts before dispatching to
`evaluator.eval_op` -> `autocast.dynamic_cast_inputs` -> `cast_inputs`. This
path is used for any op accessed dynamically (e.g. custom domain ops via
`Opset.__getattr__`, or an `Op` constructed directly), as opposed to the
code-generated opset functions (e.g. `opset18.RandomNormal`), which have
fixed Python signatures that Python itself rejects extra arguments for.

Repro:

```python
from onnxscript._internal import values
from onnxscript import opset18

values.Op(opset18, "RandomNormal")(1.0, 2.0, 3.0)
# IndexError: list index out of range
```

## Fix

Guard the variadic check with `expected_inputs and ...` so an empty input
list falls through to the existing `else` branch and raises the intended,
descriptive `ValueError` instead of crashing with an unrelated `IndexError`.
One-line change; no other behavior affected   the non-empty variadic and
non-variadic paths are untouched.

## Testing

Added `onnxscript/_internal/autocast_test.py` (no test file existed for this
module) covering:
- the empty-formal-inputs overflow case (was `IndexError`, now `ValueError`)
- the empty-formal-inputs no-args case (still returns `()`, no regression)
- the pre-existing non-variadic overflow case (still raises `ValueError`)
- the homogeneous-variadic case (guards against regressing the condition
  just touched)

```
$ python -m pytest onnxscript/_internal/autocast_test.py -q
4 passed
$ python -m pytest onnxscript/_internal/ -q
188 passed, 1 skipped, 1 xfailed, 3 xpassed
```

Confirmed the new failing-case test fails with the original `IndexError`
when the fix is reverted (`git stash`), and passes with the fix applied.

`ruff check` / `ruff format --check` clean on both changed files.

Found by reading `autocast.py` and testing `cast_inputs` against its own
error-message contract, not from a reported issue.


bodapatisaikrishna:

MohammedAlkindi (commented):
Verified against current main on Windows. A signature whose only param is an attribute has `len(inputs) == 0`, and `dynamic_cast_inputs(sig, (1.0,))` raises `IndexError: list index out of range` at `autocast.py:146` instead of the `ValueError` four lines below. With the guard it raises `ValueError: Number of actual parameters 1 exceeds number of formal parameters 0.`, and zero args returns `()`.

The tests earn their place: reverting only the source line while keeping `autocast_test.py` gives `1 failed, 3 passed` on that same IndexError. All four pass with the fix.

Non-blocking: `cast_inputs` is shared by `dynamic_cast_inputs` and `static_cast_inputs` (lines 173 and 202), so the guard fixes both, but the tests only cover the dynamic path.

bodapatisaikrishna:
Thanks for the thorough verification! Added the static_cast_inputs coverage you flagged in 02609c9 — mirrors both cases (raises on extra args, returns () on zero args) through the static path. converter_ is never actually touched on this code path since the guard raises before reaching it, so a typed None stand-in is enough. All 6 tests pass, mypy and ruff clean on the file.

